PR #5, opened 2024-08-06
Major refactor for SWU

AMMayberry1:
First major refactor to support Star Wars Unlimited rules! Too many changes to detail here. Some files were deliberately withheld for the next PR to have a smaller diff on some critical pieces.


.gitignore
AMMayberry1, line 1: testtest

PR #19, opened 2024-08-19
Implementing Surgical Droid

adbronico:


test/server/cardImplementations/01_SOR/2-1BSurgicalDroid.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+describe('2-1B Surgical Droid', function() {
+    integration(function() {
+        describe('2-1B Surgical Droid\'s healing', function() {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: [
+                            { card: '21b-surgical-droid', exhausted: false },
```
AMMayberry1, line 11: The default state is ready so the `exhausted: false` is unnecessary, unless you were doing it just to be explicit

test/server/cardImplementations/01_SOR/2-1BSurgicalDroid.spec.js
```diff
@@ -0,0 +1,54 @@
...
+                this.wampa = this.player2.findCardByName('wampa');
+
+                this.p1Base = this.player1.base;
+                this.p2Base = this.player2.base;
+
+                this.noMoreActions();
+            });
+
+            it('surgical droid should heal r2d2 to full', function () {
+                expect(this.r2d2.damage).toBe(1);
```
AMMayberry1, line 30: My weak preference is that we don't add expectations which are basically just checking that the test infrastructure is working (if I've left code like that in other tests please lmk and I'll remove it).

If you would prefer to keep some tests like this in for redundancy though I'm open to it

test/server/cardImplementations/01_SOR/2-1BSurgicalDroid.spec.js
```diff
@@ -0,0 +1,54 @@
...
+                expect(this.surgicalDroid.exhausted).toBe(false);
+                this.player1.clickCard(this.surgicalDroid);
+                expect(this.surgicalDroid.location).toBe('ground arena');
+                expect(this.player1).toBeAbleToSelectAllOf([this.p2Base, this.wampa]);
+                this.player1.clickCard(this.p2Base);
+                expect(this.player1).toBeAbleToSelectAllOf([this.r2d2]);
```
AMMayberry1, line 38: For single units there's also a `toBeAbleToSelect(this.r2d2)` form of the method

test/server/cardImplementations/01_SOR/2-1BSurgicalDroid.spec.js
```diff
@@ -0,0 +1,54 @@
...
+                expect(this.player1).toBeAbleToSelectNoneOf([this.surgicalDroid]);
+                this.player1.clickCard(this.r2d2);
```
AMMayberry1, line 40: Do you mind adding some whitespace in these to improve readability?

test/server/cardImplementations/01_SOR/2-1BSurgicalDroid.spec.js
```diff
@@ -0,0 +1,54 @@
...
+            it('surgical droid ability can be passed', function () {
+                expect(this.r2d2.damage).toBe(1);
+                this.player1.clickCard(this.surgicalDroid);
+                expect(this.player1).toBeAbleToSelectAllOf([this.p2Base]);
+                this.player1.clickCard(this.p2Base);
+                this.player1.clickPrompt('Pass ability');
+                expect(this.surgicalDroid.exhausted).toBe(true);
+                expect(this.r2d2.damage).toBe(1);
+            });
+        });
```
AMMayberry1, line 56: Can we also add a test where the unit's damage is > 2 so we can confirm that a max of 2 is healed

AMMayberry1 (approved):

PR #22, opened 2024-08-21
Mostly implementing Avenger

adbronico:
Need to add one more test once Leader Units are implemented

### DIFF
--- a/server/game/cardImplementations/01_SOR/AvengerHuntingStarDestroyer.ts
+++ b/server/game/cardImplementations/01_SOR/AvengerHuntingStarDestroyer.ts
@@ -1,0 +1,39 @@
+import AbilityHelper from '../../AbilityHelper';
+import Card from '../../core/card/Card';
+import { CardType, RelativePlayer, TargetMode } from '../../core/Constants';
+import Player from '../../core/Player';
+
+export default class AvengerHuntingStarDestroyer extends Card {
+    protected override getImplementationId() {
+        return {
+            id: '8240629990',
+            internalName: 'avenger#hunting-star-destroyer'
+        };
+    }
+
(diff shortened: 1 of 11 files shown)


server/game/cardImplementations/01_SOR/AvengerHuntingStarDestroyer.ts
```diff
@@ -0,0 +1,38 @@
...
+                player: RelativePlayer.Opponent,
+                controller: RelativePlayer.Opponent,
+                cardTypeFilter: CardType.NonLeaderUnit,
+                immediateEffect: AbilityHelper.immediateEffects.defeat()
+            }
+        });
+    }
+}
+
+AvengerHuntingStarDestroyer.implemented = false;
```
AMMayberry1, line 38: Can mark this true

test/server/cardImplementations/01_SOR/AvengerHuntingStarDestroyer.spec.js
```diff
@@ -0,0 +1,91 @@
...
+                // Choose the defender and check it was destroyed
+                this.player1.clickCard(this.interceptor);
+                expect(this.interceptor.location).toBe('discard');
+
+                // Ensure no damage happened
+                expect(this.p2Avenger.damage).toBe(0);
+                expect(this.p1Base.damage).toBe(0);
+            });
+
+            // it('Deployed leader units may not be chosen for Avenger ability', function () {
```
AMMayberry1, line 85: Can you add a comment here that just says `TODO LEADERS` so we know to come back here and fix it when we do that work

AMMayberry1 (approved):

PR #24, opened 2024-08-23
Implement DeckSearch, Move, and Reveal Systems, and a few example cards

adbronico:


server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    private shouldShuffle(shuffle: Derivable<boolean>, context: AbilityContext): boolean {
+        return typeof shuffle === 'function' ? shuffle(context) : shuffle;
+    }
+
+    private getDeck(player: Player): Card[] {
+        return player.drawDeck;
+    }
+
+    private selectCard(event: DeckSearchEvent, additionalProperties: unknown, cards: Card[], selectedCards: Set<Card>): void {
```
AMMayberry1, line 135: Is there a specific reason for `additionalProperties` to be `unknown` instead of `any`? Not sure when `unknown` is typically used

server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+            title = 'Select a card' + (properties.reveal ? ' to reveal' : '');
+            if (selectAmount < 0 || selectAmount > 1) {
+                title =
+                    `Select ${selectAmount < 0 ? 'all' : 'up to ' + selectAmount} cards` +
+                    (properties.reveal ? ' to reveal' : '');
+            }
+        }
+
+        if (properties.shuffle) {
+            cards.sort((a, b) => a.name.localeCompare(b.name));
```
AMMayberry1, line 166: Two comments:
1. What's the shuffling method here? It looks like it's sorting by name
2. I think we probably need to write the `ShuffleDeckSystem` to handle this unfortunately (ditto for line 214 and `MoveCardSystem` line 61)

server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+    targetMode?: TargetMode;
+    activePromptTitle?: string;
+    cardsToSearch?: number | ((context: AbilityContext) => number); // Number of cards to search through
+    selectCardCount?: number | ((context: AbilityContext) => number);
+    reveal?: boolean;
+    shuffle?: boolean | ((context: AbilityContext) => boolean);
+    title?: string;
+    immediateEffect?: GameSystem<IGameSystemProperties>; // What should be done with the selected cards?
+    message?: string;
+    uniqueNames?: boolean;
```
AMMayberry1, line 23: I think this parameter is probably worth keeping but I think it needs a clearer name

server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+                    '{0} puts {1} card{2} on the bottom of their conflict deck',
+                    event.player,
+                    cardsToMove.length,
+                    cardsToMove.length > 1 ? 's' : ''
+                );
+            }
+        }
+    }
+
+    private defaultDoneHandle(properties: IDeckSearchProperties, context: AbilityContext, event: DeckSearchEvent, selectedCards: Set<Card>): void {
```
AMMayberry1, line 234: nit, typo: `defaultDoneHandle`

server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+        this.handleDoneMessage(properties, context, event, selectedCards);
+
+        const gameAction = this.generatePropertiesFromContext(event.context).immediateEffect;
```
AMMayberry1, line 237: nit: `gameAction` is now `gameSystem`

server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+        if (gameAction) {
+            const selectedArray = Array.from(selectedCards);
+            event.context.targets = selectedArray;
+            gameAction.setDefaultTargetFn(() => selectedArray);
+            context.game.queueSimpleStep(() => {
+                if (gameAction.hasLegalTarget(context)) {
+                    gameAction.resolve(null, context);
+                }
+            }, 'resolve deck search');
```
AMMayberry1, line 246: small edit for clarity: "resolve effect on searched cards", since search has already resolved if I'm understanding right

server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+
+        if (properties.placeOnBottomInRandomOrder) {
+            const cardsToMove = allCards.filter((card) => !selectedCards.has(card));
+            if (cardsToMove.length > 0) {
+                shuffleArray(cardsToMove);
+                for (const card of cardsToMove) {
+                    event.player.moveCard(card, Location.Deck, { bottom: true });
+                }
+                context.game.addMessage(
+                    '{0} puts {1} card{2} on the bottom of their conflict deck',
```
AMMayberry1, line 225: nit: just "deck" instead of "conflict deck"

test/server/cardImplementations/01_SOR/MonMothmaVoiceoftheRebellion.spec.js
```diff
@@ -0,0 +1,120 @@
...
+                this.pykeSentinel = this.player1.findCardByName('pyke-sentinel');
+                this.volunteerSoldier = this.player1.findCardByName('volunteer-soldier');
+
+                this.noMoreActions();
+            });
+
+            it('should prompt to choose a Rebel from the top 5 cards', function () {
+                this.player1.clickCard(this.monMothma);
+                expect(this.player1).toHavePrompt('Select a card to reveal');
+                expect(this.player1).toHaveDisabledPromptButton(this.cellBlockGuard.title);
```
AMMayberry1, line 30: Might be worth adding support for this method to accept a list

test/server/cardImplementations/01_SOR/MonMothmaVoiceoftheRebellion.spec.js
```diff
@@ -0,0 +1,120 @@
...
+                this.monMothma = this.player1.findCardByName('mon-mothma#voice-of-the-rebellion');
+                this.battlefieldMarine = this.player1.findCardByName('battlefield-marine');
+
+                this.cartelSpacer = this.player1.findCardByName('cartel-spacer');
+                this.cellBlockGuard = this.player1.findCardByName('cell-block-guard');
+                this.pykeSentinel = this.player1.findCardByName('pyke-sentinel');
+                this.volunteerSoldier = this.player1.findCardByName('volunteer-soldier');
+
+                this.noMoreActions();
+            });
```
AMMayberry1, line 25: Tests look great overall, very thorough. Couple edge cases that might be worth adding:
- Deck has less than five cards remaining
- Deck has no cards remaining

test/server/cardImplementations/02_SHD/GreefKargaAffableCommissioner.spec.js
```diff
@@ -0,0 +1,40 @@
...
+                expect(this.player1).toHaveDisabledPromptButton(this.atst.title);
+                expect(this.player1).toHaveDisabledPromptButton(this.pykeSentinel.title);
+                expect(this.player1).toHaveDisabledPromptButton(this.cartelSpacer.title);
+                expect(this.player1).toHaveDisabledPromptButton(this.battlefieldMarine.title);
+                this.player1.clickPrompt(this.foundling.title);
+                expect(this.foundling.location).toBe('hand');
+                expect(this.getChatLogs(2)).toContain('player1 takes Foundling');
+            });
+
+            //TODO: do we need more tests?
```
AMMayberry1, line 37: I think the Mon Mothma tests give pretty thorough coverage of the base system mechanics, for Greef this basic test is enough. Can remove this comment

test/server/cardImplementations/01_SOR/MonMothmaVoiceoftheRebellion.spec.js
```diff
@@ -0,0 +1,120 @@
...
+                expect(this.player1).toHavePrompt('Select a card to reveal');
+                expect(this.player1).toHaveDisabledPromptButton(this.cellBlockGuard.title);
+                expect(this.player1).toHaveDisabledPromptButton(this.volunteerSoldier.title);
+                expect(this.player1).toHaveDisabledPromptButton(this.pykeSentinel.title);
+                expect(this.player1).toHaveDisabledPromptButton(this.cartelSpacer.title);
+                expect(this.player1).toHavePromptButton(this.battlefieldMarine.title);
+                expect(this.player1).toHavePromptButton('Take nothing');
+            });
+
+            it('should reveal the chosen Rebel', function() {
```
AMMayberry1, line 38: I'm not sure if it's the best test practice but it might be okay to stitch the first 4 test cases here into one big test. The reason being that (a) each individual test case has to go through a lot of setup and the time starts to add up across tests, and (b) these are really testing different aspects of the same code action (play Mon Mothma and select Marine). So it feels okay to me to combine them, but I'm fine if you want to keep them as-is for cleanliness. Ditto for Tarkin's tests.

test/helpers/PlayerInteractionWrapper.js
```diff
@@ -142,7 +142,14 @@ class PlayerInteractionWrapper {
     }
 
     get deck() {
-        return this.player.deck;
+        return this.player.drawDeck;
+    }
+
+    setDeck(newContents = []) {
+        newContents.reverse().forEach((name) => {
```
AMMayberry1, line 149: Do we need to empty the deck before doing this, or is the intent that we're just moving a subset of cards to the top?

server/game/cardImplementations/01_SOR/GrandMoffTarkinDeathStarOverseer.ts
```diff
@@ -0,0 +1,29 @@
...
+        return {
+            id: '9266336818',
+            internalName: 'grand-moff-tarkin#death-star-overseer'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Search the top 5 cards of your deck for up to 2 Imperial cards, then reveal and draw it.',
+            optional: false,
```
AMMayberry1, line 16: I'm fine omitting `optional: false` from these since it's the default

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -3,20 +3,22 @@ import type { Card } from '../core/card/Card';
...
+    // TODO: remove completely if faceup logic is not needed
+    // faceup?: boolean;
     bottom?: boolean;
     changePlayer?: boolean;
     discardDestinationCards?: boolean;
 }
 
-/** @deprecated This system was imported from L5R but has not been tested */
+/**
+ * This system has been partially tested.
```
AMMayberry1, line 20: I would call it about as tested as any other system at this point :P probably can remove this comment

server/game/cardImplementations/01_SOR/GrandMoffTarkinDeathStarOverseer.ts
```diff
@@ -0,0 +1,29 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.deckSearch({
+                selectCardCount: 2,
+                cardsToSearch: 5,
+                cardCondition: (card) => card.hasSomeTrait(Trait.Imperial),
+                immediateEffect: AbilityHelper.immediateEffects.moveCard({
```
AMMayberry1, line 21: As discussed, we most likely need `DrawCardSystem` for this

server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+    selectCardCount?: number | ((context: AbilityContext) => number);
+    reveal?: boolean;
+    shuffle?: boolean | ((context: AbilityContext) => boolean);
+    title?: string;
+    immediateEffect?: GameSystem<IGameSystemProperties>; // What should be done with the selected cards?
+    message?: string;
+    uniqueNames?: boolean;
+    player?: Player;
+    choosingPlayer?: Player;
+    placeOnBottomInRandomOrder?: boolean;
```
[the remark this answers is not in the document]
adbronico: It looks like if it's true, it returns them to the bottom of the deck. If it's false, it's not clear to me what happens. I suspect they remain on top?

I was thinking we could eliminate this by making the remainingCardsHandler be set up to return the cards to the bottom by default. Thoughts?

server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+        }
+        properties.cardCondition = properties.cardCondition || (() => true);
+        return properties;
+    }
+
+    public override getEffectMessage(context: AbilityContext): [string, any[]] {
+        const properties = this.generatePropertiesFromContext(context);
+        const amount = this.getAmount(properties.cardsToSearch, context);
+        const message =
+            amount > 0
```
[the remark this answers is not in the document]
adbronico: -1 is the entire deck. I'm not sure that 0 is ever a valid case, I'll check some jigoku cards

server/game/gameSystems/DeckSearchSystem.ts
```diff
@@ -0,0 +1,284 @@
...
+        const properties = this.generatePropertiesFromContext(context, additionalProperties) as IDeckSearchProperties;
+        if (this.getAmount(properties.cardsToSearch, context) === 0) {
+            return false;
+        }
+        const player = properties.player || context.player;
+        return this.getDeck(player).length > 0 && super.canAffect(player, context);
+    }
+
+    public override generatePropertiesFromContext(context: AbilityContext, additionalProperties = {}): IDeckSearchProperties {
+        const properties = super.generatePropertiesFromContext(context, additionalProperties) as IDeckSearchProperties;
```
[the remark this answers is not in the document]
adbronico: Since cardCondition's default is a function that returns true, I'm not sure if we can check that super well. Does something like this work in TS?
if (Contract.assertTrue(properties.cardCondition !== (() => true))) {
            return;
        }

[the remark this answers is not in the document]
adbronico: Also, the method needs to return something unlike an event handler. Should I return the properties? Or should I maybe be handling the Contract checks elsewhere?

server/game/gameSystems/DeckSearchSystem.ts
adbronico: I'm not sure why it name sorts here, it looks like it does the shuffling separately later on.

How would I call the shuffle system from here?

server/game/gameSystems/DeckSearchSystem.ts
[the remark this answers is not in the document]
AMMayberry1: Yeah, looks like they remain on top. I think that is the best solution since afaik the only standard handling for the searched cards in SWU is randomized on bottom, otherwise it's a custom card action.

server/game/gameSystems/DeckSearchSystem.ts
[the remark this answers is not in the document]
AMMayberry1: Oh yeah fair, I guess if cardCondition doesn't default to null then we can't do much there. Nbd.

The method is supposed to return a properties object of whatever the `TProperties` for the class is - in this case, `IDeckSearchProperties`.

server/game/gameSystems/DeckSearchSystem.ts
AMMayberry1: Discussed in chat - see `executeHandler` in `PlayUnitAction` and `InitiateAttackAction` for examples of triggering a system directly from code using an event window.

server/game/gameSystems/DeckSearchSystem.ts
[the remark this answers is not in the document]
AMMayberry1: Yeah if that's the case then I'd still prefer to have `TargetMode` be the primary condition and then we check `amount` iff the target mode uses it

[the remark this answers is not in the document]
adbronico: I think that is already more or less happening in selectCard. The code we're commenting on is for the number of cards to search. I've updated the names of things for clarity

test/server/cardImplementations/01_SOR/MonMothmaVoiceoftheRebellion.spec.js
```diff
@@ -0,0 +1,126 @@
...
+        describe('Mon Mothma\'s Ability', function() {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['mon-mothma#voice-of-the-rebellion'],
+                        deck: ['cell-block-guard', 'pyke-sentinel', 'volunteer-soldier', 'cartel-spacer', 'academy-defense-walker', 'wampa', 'viper-probe-droid', 'snowtrooper-lieutenant'],
+                        deckSize: 8,
+                        leader: ['leia-organa#alliance-general'],
+                        resources: ['atst', 'atst', 'atst', 'atst', 'atst', 'atst']
```
AMMayberry1, line 93: Can omit the resources line now

test/server/cardImplementations/01_SOR/MonMothmaVoiceoftheRebellion.spec.js
```diff
@@ -0,0 +1,126 @@
...
+
+        describe('Mon Mothma\'s Ability', function() {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['mon-mothma#voice-of-the-rebellion'],
+                        deck: ['cell-block-guard', 'pyke-sentinel', 'volunteer-soldier', 'cartel-spacer', 'academy-defense-walker', 'wampa', 'viper-probe-droid', 'snowtrooper-lieutenant'],
+                        deckSize: 8,
+                        leader: ['leia-organa#alliance-general'],
```
AMMayberry1, line 92: I changed things so that leader and base can no longer be a list, must be a single value. But the default resource count is now 20 so probably don't need to worry about aspect cost, can just remove the leader completely in almost all cases and use the default.

server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -0,0 +1,285 @@
...
+        if (event.amount === -1) {
+            cards = cards.filter((card) => properties.cardCondition(card, context));
+        }
+        events.push(event);
+        this.selectCard(event, additionalProperties, cards, new Set());
+
+        return events;
+    }
+
+    private getNumCards(numCards: Derivable<number>, context: AbilityContext): number {
```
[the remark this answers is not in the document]
adbronico: Some of the properties can be either a number or a function to determine the number. Derivable is the type the L5R code used to join those together, then it gets used in the getNumCards and other functions

server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -0,0 +1,285 @@
...
+        //     cards.sort((a, b) => a.name.localeCompare(b.name));
+        // }
+
+        context.game.promptWithHandlerMenu(choosingPlayer, {
+            activePromptTitle: title,
+            context: context,
+            cards: cards,
+            cardCondition: (card: Card, context: AbilityContext) =>
+                properties.cardCondition(card, context) &&
+                (!properties.chosenCardsMustHaveUniqueNames || !Array.from(selectedCards).some((sel) => sel.name === card.name)) &&
```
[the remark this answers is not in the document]
adbronico: There's no harm in leaving it I think. It is an effect I can see being introduced someday.

server/game/gameSystems/RevealSystem.ts
```diff
@@ -0,0 +1,52 @@
...
+        const properties = this.generatePropertiesFromContext(event.context, additionalProperties) as IRevealProperties;
+        if (properties.chatMessage) {
+            event.context.game.addMessage(
+                '{0} reveals {1} due to {2}',
+                properties.player || event.context.player,
+                event.card,
+                event.context.source
+            );
+        }
+        event.card.facedown = false;
```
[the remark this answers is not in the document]
adbronico: I think we can just comment this out with a TODO for now?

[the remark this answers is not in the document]
AMMayberry1: Yeah that or just remove it completely, either is fine for me

server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    public override generateEventsForAllTargets(context: AbilityContext, additionalProperties = {}): GameEvent[] {
+        const events: GameEvent[] = [];
+
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        const player = properties.player || context.player;
+        const event = this.generateEvent(player, context, additionalProperties) as any;
+        const deckLength = this.getDeck(player).length;
+        const amount = event.amount === -1 ? deckLength : event.amount > deckLength ? deckLength : event.amount;
```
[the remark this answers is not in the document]
AMMayberry1: Sounds good

server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -144,17 +143,20 @@ export class SearchDeckSystem extends PlayerTargetSystem<ISearchDeckProperties>
...
+            case TargetMode.UpToVariable:
+            case TargetMode.Exactly:
+            case TargetMode.ExactlyVariable:
+                selectAmount = this.getNumCards(properties.selectCount, context);
+                break;
+            case TargetMode.Unlimited:
+                selectAmount = -1;
+                break;
+            case TargetMode.Single:
+            default:
```
AMMayberry1, line 157: One very small favor to ask - if the above switch cases grab all of the expected values, can we repeat the pattern I've used elsewhere which is throwing an assertion failure in the default case:

```
default:
    Contract.fail(`Invalid targetMode: ${properties.targetMode}`);
    selectAmount = -1; // or some other useful default value that will let things try to move forward
    break;
```

This type of contract check probably catches the most errors of any of them in my work on the repo.

server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -248,15 +249,15 @@ export class SearchDeckSystem extends PlayerTargetSystem<ISearchDeckProperties>
         }
     }
 
-    private handleDoneMessage(properties: ISearchDeckProperties, context: AbilityContext, event: any, selectedCards: Set<Card>): void {
+    private searchCompleteMessageHandler(properties: ISearchDeckProperties, context: AbilityContext, event: any, selectedCards: Set<Card>): void {
```
AMMayberry1, line 252: Sorry, just now catching this - maybe this should be named `searchCompleteHandler` or something since the name implies it's just doing a message, but it is also potentially activating other handlers. Or just fold the functionality directly into `onSearchComplete`.

test/helpers/IntegrationHelper.js
```diff
@@ -387,6 +387,39 @@ var customMatchers = {
...
+            compare: function (cards, player, numCards) {
+                var result = {};
+                const deck = player.deck;
+                const L = deck.length;
+                result.pass = L >= numCards;
+                if (result.pass) {
+                    for (let card of cards) {
+                        result.pass = card.location === 'deck';
+                        if (!result.pass) {
+                            result.message = `Expected ${card.title} to be in the deck.`;
```
AMMayberry1, line 402: Should we be appending these strings together to have a message with all of the missing cards? Ditto for line 413.

test/helpers/IntegrationHelper.js
```diff
@@ -387,6 +387,39 @@ var customMatchers = {
...
+    toAllBeInBottomOfDeck: function () {
+        return {
+            compare: function (cards, player, numCards) {
+                var result = {};
+                const deck = player.deck;
+                const L = deck.length;
+                result.pass = L >= numCards;
+                if (result.pass) {
+                    for (let card of cards) {
+                        result.pass = card.location === 'deck';
```
AMMayberry1, line 400: I think as-is this means that the final test pass result will depend on what happens with the last card - shouldn't it be `result.pass = result.pass && card.location === 'deck'`? Ditto for line 411.

AMMayberry1 (approved):

PR #25, opened 2024-08-24
Added "gain ability" support and Vambrace Grappleshot as example

AMMayberry1:
Re-enabled the `GainAbility` stuff from L5R to support dynamically gained abilities. This is particularly important for upgrades as most of them grant a triggered ability to the attached card.

Added Vambrace Grappleshot as an example. The same tools will work for units that grant dynamic abilities.

Also did some cleanup of the upgrade test infrastructure to fix a bug that was causing sporadic failures.

### DIFF
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -80,7 +80,7 @@ export class Card extends OngoingEffectSource {
         public readonly owner: Player,
         private readonly cardData: any
     ) {
-        super(owner.game);
+        super(owner.game, cardData.title);
 
         this.validateCardData(cardData);
         this.validateImplementationId(cardData);
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -556,7 +556,7 @@ export class Card extends OngoingEffectSource {
 
     /**
      * Deals with the engine effects of leaving play, making sure all statuses are removed. Anything which changes
-     * the state of the card should be here. This is also called in some strange corner cases e.g. for attachments
+     * the state of the card should be here. This is also called in some strange corner cases e.g. for upgrades
      * which aren't actually in play themselves when their parent (which is in play) leaves play.
      */
     public leavesPlay() {
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -346,7 +346,7 @@ export class Card extends OngoingEffectSource {
 
     // *************************************** EFFECT HELPERS ***************************************
     public isBlank(): boolean {
-        return this.anyEffect(EffectName.Blank);
+        return this.hasEffect(EffectName.Blank);
     }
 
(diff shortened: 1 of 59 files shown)


server/game/gameSystems/TriggerAbilitySystem.ts
```diff
@@ -7,6 +7,7 @@ import type TriggeredAbility from '../core/ability/TriggeredAbility';
 import type { TriggeredAbilityContext } from '../core/ability/TriggeredAbilityContext';
 import { CardTargetSystem, ICardTargetSystemProperties } from '../core/gameSystem/CardTargetSystem';
 
+/** @deprecated Need to rename this so I stop confusing it with {@link ITriggeredAbilityProps} lol */
```
adbronico, line 10: +1

test/server/cardImplementations/02_SHD/VambraceGrappleshot.spec.js
```diff
@@ -0,0 +1,69 @@
...
+                this.player1.clickCard(this.snowspeeder);
+
+                expect(this.snowspeeder.damage).toBe(5);
+                expect(this.marine.damage).toBe(3);
+                expect(this.snowspeeder.exhausted).toBe(true);
+            });
+
+            it('should not have any effect after being removed', function () {
+                this.marine.removeUpgrade(this.vambraceGrappleshot);
+                this.vambraceGrappleshot.moveTo('discard');
```
adbronico, line 33: Are both of these necessary? What is the location of the grappleshot after removeUpgrade?

server/game/core/card/Card.ts
```diff
@@ -361,6 +362,10 @@ export class Card extends OngoingEffectSource {
         return !this.facedown && !this.hasRestriction(AbilityRestriction.InitiateKeywords, context);
     }
 
+    protected getGainedAbilityEffects<TAbility>(abilityType: AbilityType): TAbility[] {
+        return this.getEffectValues(EffectName.GainAbility).filter((ability) => ability.abilityType === abilityType);
```
adbronico, line 366: Does this cover all possible gained abilities? i.e. triggered, reactions, constants?

adbronico (commented):
Couple questions but no major thoughts

server/game/core/card/Card.ts
AMMayberry1: Yeah you can retrieve any type using the combination of generic constraint and ability type, though you need to make sure that the two match.

test/server/cardImplementations/02_SHD/VambraceGrappleshot.spec.js
AMMayberry1: I decided that `attachUpgrade` / `removeUpgrade` will just connect the two cards together, the zone move should happen separately since that has more rules and it's not consistent where the card is coming from / going to.

PR #26, opened 2024-08-25
Implement events, add Daring Raid and Repair

AMMayberry1:
Implemented the code for Event card types. Added Daring Raid and Repair to test the system. Need to add Restock to test that the event sequencing happens right, but waiting on other changes to merge first.

Also fixed a bug with upgrade aspect costs and added a test for it, as well as some general cleanup.

### DIFF
--- a/server/game/cardImplementations/02_SHD/DaringRaid.ts
+++ b/server/game/cardImplementations/02_SHD/DaringRaid.ts
@@ -11,7 +11,7 @@ export default class DaringRaid extends EventCard {
 
     public override setupCardAbilities() {
         this.setEventAbility({
-            title: 'Deal 2 damage to a friendly ground unit and an enemy ground unit',
+            title: 'Deal 2 damage to a unit or base',
             targetResolver: {
                 immediateEffect: AbilityHelper.immediateEffects.damage({ amount: 2 })
             },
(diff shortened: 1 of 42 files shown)


server/game/core/ability/CardAbility.js
```diff
@@ -6,11 +6,11 @@ const EnumHelpers = require('../utils/EnumHelpers.js');
 const { addInitiateAttackProperties } = require('../attack/AttackHelper.js');
 
 class CardAbility extends CardAbilityStep {
-    constructor(game, card, properties, abilityType = AbilityType.Action) {
+    constructor(game, card, properties, type = AbilityType.Action) {
```
adbronico, line 9: Why the rename here? The enum is still called AbilityType

server/game/cardImplementations/02_SHD/DaringRaid.ts
```diff
@@ -0,0 +1,24 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '7826408293',
+            internalName: 'daring-raid',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Deal 2 damage to a friendly ground unit and an enemy ground unit',
```
adbronico, line 14: This title should be updated to "Deal 2 damage to a unit or a base"

adbronico (changes requested):
Daring Raid title needs fixing, and a question about type vs. abilityType

adbronico (approved):
to me, looks good

PR #28, opened 2024-08-26
Finish Base implementation, add Tarkintown

AMMayberry1:
Finalized the infrastructure for adding Base cards. Added Tarkintown as an example. Wanted to test that the epic action limit is preserved across turns, so did a lot of cleanup + debugging of the non-action phases to enable the option to move to a second action phase during the test.

This ended up meaning a lot of cleanup of multiple parts of the code, but now the regroup and setup phases are known to be working correctly (though claim initiative is not currently tested).


server/game/gameSystems/ReadySystem.ts
```diff
@@ -0,0 +1,45 @@
...
+    protected override readonly targetTypeFilter = [WildcardCardType.Unit, CardType.Leader, CardType.Event];
+    protected override readonly defaultProperties: IReadySystemProperties = {
+        isRegroupPhaseReadyStep: false
+    };
+
+    public eventHandler(event): void {
+        event.card.ready();
+    }
+
+    public override canAffect(card: Card, context: AbilityContext): boolean {
```
adbronico, line 27: I think we need a check here eventually for effects that prevent readying like Frozen in Carbonite and No Good to Me Dead. Something like checking checking for a cardCannot be readied ongoing effect?

If we implement that when we add those cards, I'd recommend leaving a TODO here for it

server/game/core/gameSteps/phases/RegroupPhase.ts
```diff
@@ -22,4 +25,23 @@ export class RegroupPhase extends Phase {
             player.drawCardsToHand(2);
         }
     }
+
+    private readyAllCards() {
```
adbronico, line 29: Likewise to in the other comment, do we need to do any sort of cardCannot(BeReadied) restrictions here?

AMMayberry1: Isn't needed here since this calls the ReadySystem, and we'll have the check there

PR #31, opened 2024-08-31
LookAtSystem & Viper Probe Droid

adbronico:


server/game/cardImplementations/01_SOR/ViperProbeDroid.ts
```diff
@@ -0,0 +1,23 @@
...
+            id: '8986035098',
+            internalName: 'viper-probe-droid'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Look at an opponent\'s hand.',
+            immediateEffect: AbilityHelper.immediateEffects.lookAt((context) => ({
+                target: context.player.opponent.hand.sort((a, b) => a.name.localeCompare(b.name)),
```
AMMayberry1, line 16: Any specific reason we want them sorted?

server/game/gameSystems/LookAtSystem.ts
```diff
@@ -0,0 +1,80 @@
...
+        const obj = { a: cards, b: context };
+        event.stateBeforeResolution = cards.map((a) => {
+            return { card: a, location: a.location };
+        });
+        event.context = context;
+    }
+
+    public override eventHandler(event, additionalProperties = {}): void {
+        const context = event.context;
+        const properties = this.generatePropertiesFromContext(context, additionalProperties) as ILookAtProperties;
```
AMMayberry1, line 61: The cast here isn't needed, the parent class method returns `TProperties` which in this case is `ILookAtProperties`

server/game/gameSystems/LookAtSystem.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    public override addPropertiesToEvent(event, cards, context: AbilityContext, additionalProperties): void {
+        if (!cards) {
+            cards = this.generatePropertiesFromContext(context, additionalProperties).target;
+        }
+        if (!Array.isArray(cards)) {
+            cards = [cards];
+        }
+        event.cards = cards;
+        const obj = { a: cards, b: context };
```
AMMayberry1, line 52: Looks like this isn't used?

server/game/gameSystems/LookAtSystem.ts
```diff
@@ -0,0 +1,80 @@
...
+        event.stateBeforeResolution = cards.map((a) => {
```
AMMayberry1, line 53: Looks like this is only here because a couple of L5R cards had a specific kind of reaction to `LookAt.` Fine to leave it here or remove it, but if we leave it here we should add a note as to what it's intended for

server/game/gameSystems/LookAtSystem.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public getMessage(message, context): string {
+        if (typeof message === 'function') {
+            return message(context);
+        }
+        return message;
+    }
+
+    public override isEventFullyResolved(event): boolean {
```
AMMayberry1, line 73: Looks like this is exactly the same as the parent class method, so we can remove it. Ditto for `checkEventCondition` below

AMMayberry1: I _think_ this is also true for `generateEventsForAllTargets`, it's coded slightly differently but I'm pretty sure the behavior is exactly the same.

test/server/cardImplementations/01_SOR/ViperProbeDroid.spec.js
```diff
@@ -0,0 +1,36 @@
...
+    integration(function() {
+        describe('Viper Probe Droid\'s Ability', function() {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['viper-probe-droid']
+                    },
+                    player2: {
+                        hand: ['wampa', 'battlefield-marine', 'pyke-sentinel']
```
AMMayberry1, line 11: Would be good to add some extra units in the opponent's arena(s) so we have some negative results to be sure the ability is restricted to cards from hand

server/game/gameSystems/LookAtSystem.ts
```diff
@@ -0,0 +1,80 @@
...
+        }
+        event.cards = cards;
+        const obj = { a: cards, b: context };
+        event.stateBeforeResolution = cards.map((a) => {
+            return { card: a, location: a.location };
+        });
+        event.context = context;
+    }
+
+    public override eventHandler(event, additionalProperties = {}): void {
```
AMMayberry1, line 59: Small nit, but I'm trying to have a convention of having the handler method be the first method in the class since they tend to be the most important (still working on updating the existing classes)

server/game/cardImplementations/01_SOR/ViperProbeDroid.ts
adbronico: @AMMayberry1 It was like that in the L5R code. I don't think it's necessary, but I suspect the reason is so that the cards are not in draw order. This would give someone information about how long someone has held a card etc.

server/game/gameSystems/LookAtSystem.ts
adbronico: CardTargetSystem's version of generateEventsForAllTargets and checkEventCondition will cause the tests to fail. These seem to match the ones in GameTargetSystem but not CardTargetSystem

server/game/core/Constants.ts
```diff
@@ -306,4 +306,11 @@ export enum AbilityRestriction {
     Ready = 'ready',
     ReceiveDamage = 'receiveDamage',
     TriggerAbilities = 'triggerAbilities',
+}
+
+export enum ViewCardType {
```
AMMayberry1, line 311: Two small comments:

1. Since this is only relevant for one situation we can probably move it to `ViewCardSystem.ts` to avoid cluttering this file too much
2. How would you feel about renaming it to `ViewCardMode` since (a) 'mode' is typically how a parameter like this is named elsewhere in the repo (e.g., `TargetMode`) and (b) 'CardType' here sounds like `CardType` as in unit, upgrade, event etc.

server/game/gameSystems/ViewCardSystem.ts
```diff
@@ -0,0 +1,76 @@
+import { AbilityContext } from '../core/ability/AbilityContext';
+import { BaseCard } from '../core/card/BaseCard';
+import { ViewCardType } from '../core/Constants';
+import { GameEvent } from '../core/event/GameEvent';
+import { CardTargetSystem, ICardTargetSystemProperties } from '../core/gameSystem/CardTargetSystem';
+import Player from '../core/Player';
+
+export interface IViewCardProperties extends ICardTargetSystemProperties {
+    viewType: ViewCardType;
+    sendChatMessage?: boolean;
```
AMMayberry1, line 10: Is there any situation where we wouldn't send a chat message currently, or are you putting this here for when we add a UI for this later and the message is then optional?

server/game/gameSystems/ViewCardSystem.ts
```diff
@@ -0,0 +1,76 @@
...
+    message?: string | ((context) => string);
+    messageArgs?: (cards: any) => any[];
+
+    /** The player who is viewing or revealing the card. */
```
AMMayberry1, line 14: Good docstr, this is helpful

server/game/gameSystems/ViewCardSystem.ts
```diff
@@ -0,0 +1,76 @@
...
+        const context = event.context;
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        if (properties.sendChatMessage) {
+            const messageArgs = this.getMessageArgs(event, context, additionalProperties);
+            context.game.addMessage(this.getMessage(properties.message, context), ...messageArgs);
+        }
+    }
+
+    public override canAffect(card: BaseCard, context: AbilityContext) {
+        // TODO: What situations would mean that a card cannot be looked at?
```
AMMayberry1, line 29: This is interesting - I finally looked up the rules for these. "Look At" has no restrictions at all, but "Reveal" can only be cards in deck, hand, or resource zone. Neither one has any specific restrictions on faceup / facedown.

Probably worth adding a check here for the "Reveal" case on the off chance it ever becomes relevent - like for example if something triggers off a reveal, it would be important that the card could be legally revealed. Unlikely to be relevant but it's easy to add and hard to remember so we might as well do it now, imo.

Could be handled either as a check of the mode property here or as an override in the Reveal class (which would still call `super.canAffect()`

server/game/gameSystems/LookAtSystem.ts
```diff
@@ -1,77 +1,39 @@
...
-export interface ILookAtProperties extends ICardTargetSystemProperties {
-    message?: string | ((context) => string);
-    messageArgs?: (cards: any) => any[];
-}
+export type ILookAtProperties = Omit<IViewCardProperties, 'viewType'>;
 
-export class LookAtSystem extends CardTargetSystem<ILookAtProperties> {
+export class LookAtSystem extends ViewCardSystem {
     public override readonly name = 'lookAt';
     public override readonly eventName = EventName.OnLookAtCards;
```
AMMayberry1, line 9: nit: if each event is only going to look at a single card at a time, then I think a more consistent naming for this is `OnLookAtCard` (singular)

server/game/gameSystems/ViewCardSystem.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public getMessage(message, context: AbilityContext): string {
+        if (typeof message === 'function') {
+            return message(context);
+        }
+        return message;
+    }
+
+    public getMessageArgs(event: any, context: AbilityContext, additionalProperties) {
```
AMMayberry1, line 71: Is this needed if both of the derived classes replace it? Can just mark it abstract if we want to require that they each have some implementation of it.

server/game/gameSystems/ViewCardSystem.ts
AMMayberry1: Actually I just thought of the answer to my own question - in the case when the opponent is looking at top card of a deck (e.g. Thrawn), that shouldn't be broadcast. For now we can leave this as-is since we have nothing in place for the UI but we should probably leave something like TODO LOOKAT here to remember

adbronico: @AMMayberry1 Yeah, it looks like all lookAt/reveal logic is going through the chat log atm, which is not going to work for Thrawn or for Twin Suns in the future if we support that. We need some sort of popup that shows info to only certain players

server/game/gameSystems/RevealSystem.ts
```diff
@@ -14,22 +15,33 @@ export class RevealSystem extends ViewCardSystem {
...
         } else {
-            propertyWithViewType = Object.assign(propertiesOrPropertyFactory, { viewType: ViewCardType.Reveal });
+            propertyWithViewType = Object.assign(propertiesOrPropertyFactory, { viewType: ViewCardMode.Reveal });
         }
 
         super(propertyWithViewType);
     }
 
+    public override canAffect(card: BaseCard, context: AbilityContext): boolean {
+        // TODO: What situations would mean that a card cannot be looked at?
```
AMMayberry1, line 35: Can we remove the comment now that we've figured this out?

AMMayberry1 (approved):

PR #32, opened 2024-09-01
Update docs, migrate cards from `cardImplementations/` to `cards/`

AMMayberry1:
Reworked the contributor's guide to reflect our changes for SWU. Moved all card implementations from `cardImplementations/` folders to `cards/` folders.

Also did some various cleanup of interfaces and card implementations since a lot of this stuff will be the examples that people refer to.


docs/implementing-cards.md
```diff
@@ -1,388 +1,570 @@
...
 
-### Keywords
+There are several ability types in SWU, each with its own initialization method. As shown above in the example of an action ability, each method accepts a property object defining the ability's behavior. Use the `AbilityHelper` import to get access to tools to help with implementations. Additionally, see [Interfaces.ts](../server/game/Interfaces.ts) for a list of available parameters for each ability type.
 
-Most Keywords are automatically parsed from the card text. It isn't necessary to explicitly implement them unless they are provided by a conditional persistent effect.
+The ability types and methods are:
 
-### Static bonuses from attachments
+| Ability Type | Method | Definition |
+| --- | --- | --- |
```
adbronico, line 94: We should add an example column here with a couple varied examples

docs/implementing-cards.md
```diff
@@ -1,388 +1,570 @@
...
-    match: card => card.isDefending(),
-    targetController: 'opponent',
-    effect: AbilityDsl.effects.modifyMilitarySkill(-1)
+// While this card is in the ground arena, all of the opponent's units in the space arena get -1/-1
+this.constantAbility({
+    sourceLocationFilter: Location.GroundArena,
+    targetLocationFilter: Location.SpaceArena,
+    targetCardType: WildcardCardType.Unit,
+    targetController: RelativePlayer.Opponent,
+    ongoingEffect: AbilityHelper.ongoingEffects.modifyStats({ power: -1, hp: -1 })
```
adbronico, line 168: modifyStats should be 1, 1 not -1, -1

docs/implementing-cards.md
```diff
@@ -1,388 +1,570 @@
...
-    ]
+// Entrenched ability
+this.addConstantAbilityTargetingAttached({
+    title: 'Attached unit cannot attack bases',
+    ongoingEffect: AbilityHelper.ongoingEffects.cannotAttackBase(),
 });
 ```
 
-#### Applying effects to cards which aren't in play
+#### **IGNORE THIS SECTION, STILL WIP**: ~~Applying effects to cards which aren't in play~~
```
adbronico, line 229: Can we update all Ignore this, WIP and TBD type notes in here to all be TODO for future finding and referencing?

docs/implementing-cards.md
```diff
@@ -1,388 +1,570 @@
...
 ```javascript
-this.action({
-    title: 'Sacrifice to discard an attachment'
-    target: {
-        cardType: 'attachment',
-        gameAction: AbilityDsl.actions.discardFromPlay()
-    },
-    // ...
+class AbilityContext {
+    constructor(properties) {
```
adbronico, line 478: Add a TODO to make this look more SWU-ish (this could apply in a few places)

adbronico (changes requested):
This looks really good. There's a few small things but overall this is a great start.

docs/implementing-cards.md
AMMayberry1: Do you mean examples of abilities or examples of code implementations?

docs/implementing-cards.md
AMMayberry1: Why is that? It's just a random example, and the comment says -1/-1

docs/implementing-cards.md
AMMayberry1: Can you elaborate? What would make the context class more SWUish?

docs/implementing-cards.md
adbronico: @AMMayberry1 just like example cards for each one so people would know what to look

docs/implementing-cards.md
adbronico: I must have misread the comment as 1/1. My b

docs/implementing-cards.md
adbronico: @AMMayberry1 Stuff like rings and stage being removed for example. It isn't high priority, but someone might assume this is an accurate breakdown of the context object when they're new.

AMMayberry1: Oh lol right I don't even notice ring stuff anymore, good catch I'll remove it.

I think Stage might actually still be relevant (I think it means "ability resolution stage") but we're not using it in many places. I'll add a TODO in the code for later investigation to see if we still need it.

adbronico (approved):
Merge or merge not, there is no try

PR #34, opened 2024-09-03
Add support for leaders

AMMayberry1:
Added the full functionality of leaders. Simplified and cleaned up the exhaust and pay resource systems a little bit. Updated the documentation to reflect the process for leaders and some potential pitfalls.

Added Krennic and Tarkin leaders to proof the system. Added a general test file for leader behaviors and some new test infra to support leader testing.

Some other general improvement:
- Some bug fixes and improvements to the test infra
- Added type guards on the `isX()` methods on `Card.ts` so that they can be used to remove casts in certain situations


server/game/abilities/keyword/RestoreAbility.ts
```diff
@@ -13,6 +13,7 @@ export class RestoreAbility extends TriggeredAbility {
         return {
             title: `Restore ${restoreAmount}`,
             when: { onAttackDeclared: (event, context) => event.attack.attacker === context.source },
+            locationFilter: WildcardLocation.AnyArena,
```
adbronico, line 16: What does adding the AnyArena accomplish here?

server/game/cards/01_SOR/DirectorKrennicAspiringToAuthority.ts
```diff
@@ -0,0 +1,29 @@
...
+        return {
+            id: '8560666697',
+            internalName: 'director-krennic#aspiring-to-authority',
+        };
+    }
+
+    private buildKrennicAbilityProperties() {
+        return {
+            title: 'Give each friendly damaged unit +1/+0',
+            matchTarget: (card) => card.isUnit() && card.damage !== 0,
```
adbronico, line 15: Is this set up by default to only affect friendly units?

server/game/cards/01_SOR/Tarkintown.ts
```diff
@@ -16,8 +16,7 @@ export default class Tarkintown extends BaseCard {
         this.setEpicActionAbility({
             title: 'Deal 3 damage to a damaged non-leader unit',
             targetResolver: {
-                cardTypeFilter: WildcardCardType.NonLeaderUnit,
-                cardCondition: (card) => (card as UnitCard).damage !== 0,
+                cardCondition: (card) => card.isUnit() && card.damage !== 0,
```
adbronico, line 19: Does the isUnit() method only return true for non-leader units?

server/game/cards/02_SHD/GroguIrresistible.ts
```diff
@@ -15,6 +15,8 @@ export default class GroguIrresistible extends NonLeaderUnitCard {
             title: 'Exhaust an enemy unit',
             cost: AbilityHelper.costs.exhaustSelf(),
             targetResolver: {
+                locationFilter: WildcardLocation.AnyArena,
+                cardCondition: (card) => card.controller !== this.controller,
```
adbronico, line 19: Is the AnyArena needed? And Is this not redundant with the controller check below?

server/game/core/card/Card.ts
```diff
@@ -205,77 +210,42 @@ export class Card extends OngoingEffectSource {
...
     public createActionAbility(properties: IActionAbilityProps): ActionAbility {
         properties.cardName = this.title;
         return new ActionAbility(this.game, this, properties);
     }
 
 
     // ******************************************* CARD TYPE HELPERS *******************************************
-    // TODO: convert these to use ts type narrowing for simpler conversions to derived types (see https://www.typescriptlang.org/docs/handbook/2/classes.html#this-based-type-guards)
-    public isEvent(): boolean {
+    public isEvent(): this is EventCard {
```
adbronico, line 220: I love/hate the syntax here lol

server/game/core/card/LeaderUnitCard.ts
```diff
@@ -1,39 +1,104 @@
...
 export class LeaderUnitCard extends LeaderUnitCardParent {
-    private _isDeployed = false;
-
-    public get isDeployed() {
-        return this._isDeployed;
-    }
-
     public override get type() {
-        return this._isDeployed ? CardType.LeaderUnit : CardType.Leader;
+        return this._deployed ? CardType.LeaderUnit : CardType.Leader;
```
adbronico, line 18: Not a must-have, but I think I prefer the "isDeployed" naming convention. It makes it clear without needing code hints that this is a boolean field

server/game/core/card/UpgradeCard.ts
```diff
@@ -40,20 +39,6 @@ export class UpgradeCard extends UpgradeCardParent {
         return this._parentCard;
     }
 
-    /** The card that this card is underneath */
-    public set parentCard(card: Card | null) {
```
adbronico, line 44: Did this get moved elsewhere or is it being handled in the mixins?

test/server/cards/01_SOR/DirectorKrennicAspiringToAuthority.spec.js
```diff
@@ -0,0 +1,80 @@
...
+                expect(this.wampa.hp).toBe(5);
+
+                expect(this.battlefieldMarine.power).toBe(3);
+                expect(this.battlefieldMarine.hp).toBe(3);
+
+                expect(this.cartelSpacer.power).toBe(3);
+                expect(this.cartelSpacer.hp).toBe(3);
+
+                expect(this.consularSecurityForce.power).toBe(3);
+                expect(this.consularSecurityForce.hp).toBe(7);
```
adbronico, line 29: shouldnt this be 6? theres 1 damage on it in setup

adbronico (changes requested):

server/game/core/card/LeaderUnitCard.ts
danbastin: agreed here

server/game/abilities/keyword/RestoreAbility.ts
AMMayberry1: It's because of new requirements around leader abilities. To reduce the possibility of mixing up leader and leader unit side abilities, there's now a requirement that any ability which is attached to a leader explicitly declare its active locations instead of using the default.

This gets handled automatically for abilities that are written in the .ts file. The side effect is that keyword abilities bypass that. So we either need to refactor to try and make an exception for keyword abilities or just make sure to explicitly apply a `locationFilter` property for triggered keyword abilities.

server/game/core/card/LeaderCard.ts
```diff
@@ -4,14 +4,33 @@ import Contract from '../utils/Contract';
...
-    public override isLeader() {
+    public override isLeader(): this is LeaderCard {
         return true;
     }
-}
\ No newline at end of file
+
+    // this is overriden in the LeaderUnit derived class
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public deploy() {}
```
danbastin, line 29: Do we need to have the deploy function here if only the Unit side cares about it? If so I believe we can label this function `public abstract deploy() : void` in order to avoid having to use the linter override

server/game/cards/01_SOR/DirectorKrennicAspiringToAuthority.ts
AMMayberry1: Yes, and the tests confirm that

server/game/cards/01_SOR/Tarkintown.ts
AMMayberry1: The word `Unit` is used in the code the same way it is in the game rules, so it can refer to leader units, token units, and units that are neither of the above

server/game/core/Constants.ts
```diff
@@ -7,7 +7,6 @@ export enum Location {
     Discard = 'discard',
     GroundArena = 'ground arena',
     Hand = 'hand',
-    Leader = 'leader',
```
danbastin, line 10: What is the reason for removing a dedicated location for Leaders?

server/game/cards/02_SHD/GroguIrresistible.ts
AMMayberry1: It's needed to prevent the ability from being able to target undeployed leaders

server/game/core/card/Card.ts
AMMayberry1: Yeah it's a little clunky but it solves a real problem that's kind of unique to TS

server/game/core/card/UpgradeCard.ts
AMMayberry1: It got accidentally left in, I meant to remove it in the upgrade PR. The intent is that outside code cannot change the attached card directly, it must use the `attach()` and `unattach()` methods which keep state consistent

test/server/cards/01_SOR/DirectorKrennicAspiringToAuthority.spec.js
AMMayberry1: I followed the game rules pretty literally so HP always refers to max HP value, damage is not being subtracted out. The unit is defeated when damage == HP

server/game/core/Constants.ts
AMMayberry1: Yeah so this was actually a mistake on my part that I was slow to correct. Per the swu rules, leader and base are actually both in the same "base zone" and there is no leader zone.

server/game/core/card/LeaderCard.ts
AMMayberry1: Oh yeah, normally I would prefer that but `abstract` doesn't work with mixins unfortunately :/ and if I have to choose between the two I'd much rather have mixins

server/game/cards/02_SHD/GroguIrresistible.ts
AMMayberry1: Oh I missed the controller comment though. Yeah that was a mistake, good catch

server/game/core/card/LeaderUnitCard.ts
danbastin: As discussed in chat I'm good with this.

server/game/cards/01_SOR/Tarkintown.ts
adbronico: Then wouldn't this allow Tarkintown to target a leader unit?

test/server/cards/01_SOR/DirectorKrennicAspiringToAuthority.spec.js
adbronico: I think this is a little unintuitive so I might suggest we rename the hp to maxHp if that's the case

server/game/cards/01_SOR/Tarkintown.ts
AMMayberry1: Good question. That isn’t an issue because GameSystems can declare a set of card types that are legally allowed to target, and the DamageSystem can’t target undeployed leaders

test/server/cards/01_SOR/DirectorKrennicAspiringToAuthority.spec.js
AMMayberry1: That’s reasonable, I could go either way. I’ll poll in the chat to see what people find clearest.

server/game/cards/01_SOR/Tarkintown.ts
AMMayberry1: Ope you’re right that’s a bug 😅

adbronico (approved):
Begun, the merge wars have

PR #37, opened 2024-09-08
Refactor `generateEventsForAllTargets` to `queueGenerateEventGameSteps`

AMMayberry1:
Realized that some of the "meta" game systems from the legacy L5R code override the event generation method to queue game steps which generate the events. Refactored the existing `generateEventsForAllTargets` method back to the way it was to be able to support that usage, but renamed it to `queueGenerateEventGameSteps` to make the purpose and usage clearer.


adbronico (approved):
Always two there are, no more, no less. A merger and an approver.

PR #39, opened 2024-09-11
Overhaul attack ability system, add Leia leader

AMMayberry1:
Revamped the way attacks are declared in abilities to unblock Ambush and some other card abilities. Added the Leia leader as an example.

Needed to make a lot of small engine updates:
- Added a check in the ability resolver to skip abilities that have no cost and no effect in order to reduce UX noise
- Refactored the flow for playing events so that they move to the discard earlier in the process (this facilitates skipping resolution of events with no effect)
- Some light refactoring of ability properties so that things end up in the right place
- Added the `InitiateAttackWithUnitSystem` to facilitate triggering attacks on units from ability code
- Majorly stripped down the AttackHelper code since most of the logic is in the new system


server/game/Interfaces.ts
```diff
@@ -67,19 +68,26 @@ export interface IAbilityProps<Context> {
...
+    optional?: boolean;
+
+    /**
+     * Indicates that an attack should be triggered from a friendly unit.
+     * Shorthand for `AbilityHelper.immediateEffects.attack(AttackSelectionMode.SelectAttackerAndTarget)`.
+     * Can either be an {@link IInitiateUnitAttackProperties} property object or a function that creates one from
      * an {@link AbilityContext}.
      */
-    initiateAttack?: IInitiateAttack | ((context: AbilityContext) => IInitiateAttack);
+    initiateAttack?: IInitiateUnitAttackProperties | ((context: AbilityContext) => IInitiateUnitAttackProperties);
```
adbronico, line 82: I think this could just be IInitiateAttackProperties. Unit seems implied

server/game/core/ability/CardAbility.js
```diff
@@ -140,7 +135,7 @@ class CardAbility extends CardAbilityStep {
     displayMessage(context, messageVerb = context.source.isEvent() ? 'plays' : 'uses') {
         if (
             context.source.isEvent() &&
-            context.source.location !== Location.Hand
+            context.source.location !== Location.Discard
```
adbronico, line 138: What's the cause and/or impact of this change?

server/game/core/ability/EventAbility.ts
```diff
@@ -17,7 +17,7 @@ export class EventAbility extends CardAbility {
     }
 
     public override meetsRequirements(context: AbilityContext = this.createContext(), ignoredRequirements = []) {
-        if (!ignoredRequirements.includes('location') && !this.isInValidLocation(context)) {
+        if (!ignoredRequirements.includes('location') && this.card.location !== Location.Discard) {
```
adbronico, line 20: Will this run into problems with the Set 3 card that lets you play events from discard?

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
```diff
@@ -0,0 +1,129 @@
...
+                expect(this.battlefieldMarine.exhausted).toBe(true);
+                expect(this.regionalGovernor.damage).toBe(3);
+                expect(this.battlefieldMarine.damage).toBe(1);
+
+                // second attack
+                expect(this.player1).toBeActivePlayer();
+                expect(this.player1).toBeAbleToSelectExactly([this.fleetLieutenant, this.allianceXwing]);
+                expect(this.player1).toHaveEnabledPromptButton('Pass ability');
+                this.player1.clickCard(this.allianceXwing);
+                expect(this.player1).toHaveEnabledPromptButton('Pass ability');
```
adbronico, line 36: Should the pass ability be present once the card is selected here?

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
```diff
@@ -0,0 +1,129 @@
...
+        });
+
+        describe('Leia\'s deployed ability', function() {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['atst', 'fleet-lieutenant', { card: 'battlefield-marine', exhausted: true }],
+                        spaceArena: ['alliance-xwing'],
+                        leader: { card: 'leia-organa#alliance-general', deployed: true }
```
adbronico, line 75: We should add something to the testing readme to show this is how to have a deployed leader. My first instinct would have been to include Leia in the ground arena

server/game/gameSystems/InitiateAttackWithUnitSystem.ts
```diff
@@ -0,0 +1,76 @@
...
+import type { TriggeredAbilityContext } from '../core/ability/TriggeredAbilityContext';
+import { CardTargetSystem, ICardTargetSystemProperties } from '../core/gameSystem/CardTargetSystem';
+import { UnitCard } from '../core/card/CardTypes';
+import { InitiateAttackAction } from '../actions/InitiateAttackAction';
+import { AbilityContext } from '../core/ability/AbilityContext';
+import Contract from '../core/utils/Contract';
+import { IAttackProperties } from './AttackStepsSystem';
+import * as GameSystemLibrary from './GameSystemLibrary';
+
+export interface IInitiateUnitAttackProperties extends IAttackProperties {
```
adbronico, line 12: The "unit" and "WithUnit" parts of this seem redundant. Or is this to signify that it will be attacking with a specifically chosen unit as opposed to just declaring an attack?

adbronico (changes requested):
A couple questions/comments

server/game/core/ability/CardAbility.js
AMMayberry1: It has to do with the fact that we now bypass abilities that have no effect on game state to reduce UX noise. What happened is that events which are no-ops were getting bypassed, but the effect that moved the event card to discard was actually happening as part of the event ability resolution (not the play event card resolution). So it was getting skipped if we skipped the ability resolution.

So I moved the event card -> discard move a little earlier in the pipeline and now it works. It's just that some checks were expecting event cards to be in hand or in a special play zone since that's how it worked in l5r.

server/game/core/ability/EventAbility.ts
AMMayberry1: That's what `ignoredRequirements` is for, in that case we'd just include 'location' in the ignored requirements. We haven't done much play-from-out-of-play stuff yet so we still need to iron out the details but I know there's support for it in l5r so it's just a matter of learning the details

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
AMMayberry1: I would argue yes, since the cost (exhausting the attacker) hasn't been paid yet so there's no issue if the user changes their mind and decides to pass at the second step. We could talk to Ross, but it would actually require extra work to keep it from being present at that stage - there's a check in the target resolver to see if passing is allowed and if so it just presents the button, it doesn't have awareness of whether the button was presented in an earlier stage already.

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
AMMayberry1: I agree - are you still thinking of doing that readme?

server/game/gameSystems/InitiateAttackWithUnitSystem.ts
AMMayberry1: Yeah I really wrestled with the naming here. Basically there's two closely related but distinctly different systems:

- The one that walks through the attack steps and manages all of the stages from beginning to end. This accepts the attacker as a property (so it has to be already decided) and the target is the attack target, so a target resolver would be presented to decide the attack target if needed. This is currently called `AttackStepsSystem` iirc.
- An outer system (this one) which allows you to choose both the attacker and the defender in two different steps. It's functionally allowing you to choose a unit and then trigger an attack with that unit, which ends up calling the `AttackStepsSystem` above. I am currently calling it `InitiateAttackWithUnit` to indicate that you are choosing the attacking unit, but I don't love that name and am very open to suggestions

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
adbronico: I wanted to but I'm not really sure I will be able to get to it soon

server/game/gameSystems/InitiateAttackWithUnitSystem.ts
adbronico: What about something like TriggeredAttackSystem for this one?

AMMayberry1: I prefer using `Initiate` since it's calling the `InitiateAttackAbility` class (which is what's called when a user clicks on the card and attacks). Or else change both to `Trigger`.

Bigger picture though, I'm not sure if that clear disambiguates from the other attack system since they're both triggering / initiating an attack

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
AMMayberry1: Kk, we can figure out if we want to deal with it later or I can carve out some time. Seems like people are onboarding to tests without too much issue so it may not be urgent.

server/game/Interfaces.ts
AMMayberry1: Actually I guess this needs to match the system name as discussed below, since they're paired

server/game/gameSystems/InitiateAttackWithUnitSystem.ts
AMMayberry1: Eh, actually maybe it's clear enough, since `AttackSteps` is not really ever going to be used anyway except in some very special cases

adbronico (approved):
I am a reviewer, like my father before me.

PR #40, opened 2024-09-11
Implemented I Am Your Father

Ennis0:
Proof of concept for abilities that give players arbitrary choices, using targetResolvers with TargetMode.Select


server/game/cards/01_SOR/IAmYourFather.ts
```diff
@@ -0,0 +1,38 @@
...
+                targetUnit: {
+                    controller: RelativePlayer.Opponent,
+                    cardTypeFilter: WildcardCardType.Unit
+                },
+                opponentsChoice: {
+                    mode: TargetMode.Select,
+                    dependsOn: 'targetUnit',
+                    choosingPlayer: RelativePlayer.Opponent,
+                    choices: (context) => ({
+                        [`Let ${context.targets.targetUnit.title} take 7 damage`]: AbilityHelper.immediateEffects.damage({
```
AMMayberry1, line 26: It's not a huge deal but I don't love the "Let" language on these, I would prefer it being framed as a choice of which effect to resolve:
- "<unit> takes 7 damage"
- "Opponent draws 3 cards"

That said if you like it as-is I won't block on it, just a suggestion

test/server/cards/01_SOR/IAmYourFather.spec.js
```diff
@@ -0,0 +1,50 @@
...
+                    }
+                });
+            });
+
+            it('can only select opponent\'s units, including leaders', function () {
+                this.player1.clickCard(this.iAmYourFather);
+                expect(this.player1).toBeAbleToSelectExactly([this.viperProbeDroid, this.darthVader]);
+            });
+
+            it('gives the opponent a choice after selecting the target unit', function () {
```
AMMayberry1, line 24: Since there is a lot of setup for each test case (each case has to do the entire game setup), we try to consolidate tests as much as possible. So in this case instead of having a test which confirms each stage of the process, I would prefer consolidating down to 2 cases (one for each button) in which one or both tests have expectations regarding the right targets being available

AMMayberry1 (changes requested):

test/server/cards/01_SOR/IAmYourFather.spec.js
Ennis0: Got it. I figured each test should be testing exactly one thing as much as possible(the thing you describe in the first argument to it()), but I can condense the tests down.

AMMayberry1: Yeah, I think ideally that would be the case but our test infra kind of sucks and we have a finite amount of github actions compute time available so we need to condense a bit

AMMayberry1 (approved):

PR #41, opened 2024-09-11
Fixed a couple of mistakes in the state watchers update.

Ennis0:


AMMayberry1 (approved):
Thanks for catching all that!

PR #42, opened 2024-09-11
adding in Imperial Interceptor

redwulfe:
This adds in the imperial Interceptor and its unit test.


server/game/cards/01_SOR/ImperialInterceptor.ts
```diff
@@ -0,0 +1,31 @@
...
+            id: '9002021213',
+            internalName: 'imperial-interceptor'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'You may deal 3 damage to a space unit.',
+            optional: true,
+            targetResolvers: {
```
AMMayberry1, line 17: Since there's only one target resolver, you can simplify it to use `targetResolver` (singular) and to have no name:

```
targetResolver: {
      controller: RelativePlayer.Any,
      locationFilter: Location.SpaceArena,
      immediateEffect: AbilityHelper.immediateEffects.damage({ amount: 3 })
}
```

Then the `effectArgs` on line 26 would reference `context.target` instead of `context.targets.spaceUnit`

server/game/cards/01_SOR/ImperialInterceptor.ts
```diff
@@ -0,0 +1,31 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '9002021213',
+            internalName: 'imperial-interceptor'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'You may deal 3 damage to a space unit.',
```
AMMayberry1, line 15: nit, consistency: we typically aren't using a period at the end of titles / messages like this

test/server/cards/01_SOR/ImperialInterceptor.spec.js
```diff
@@ -0,0 +1,60 @@
...
+                        spaceArena: ['system-patrol-craft']
+                    },
+                    player2: {
+                        groundArena: ['wampa', 'superlaser-technician'],
+                        spaceArena: ['gladiator-star-destroyer']
+                    }
+                });
+            });
+
+            it('can only select space units & can select itself', function () {
```
AMMayberry1, line 19: Because our tests require a lot of setup (each test case has to start an entire game instance), we try to consolidate test cases down as much as we can. This is not normal practice since typically you want test cases to be small, but since our test framework needs some improvement we need to handle things a little differently.

In addition, we typically assume that if a unit is selectable at the target phase that the effect would work correctly for any selectable card. Obviously there are exceptions if some card is very different than the others - in this case it may be worth having a separate test which damages the interceptor to confirm that the timing is working correctly.

To summarize, these are the test cases I would recommend for this card:
1. A test that resolves the ability, where we have one `toBeAbleToSelectExactly` check, check that the 'Pass ability' button is enabled, and then choose a unit which is not the interceptor and check that the damage is applied
2. Same test as 1 (we could skip the `toBeAbleToSelectExactly`) but we select the interceptor for the damage and check that it is in discard at the end
3. A test where we pass the ability (you already have this one)

server/game/cards/01_SOR/ImperialInterceptor.ts
redwulfe: Changed it

server/game/cards/01_SOR/ImperialInterceptor.ts
redwulfe: no problem

test/server/cards/01_SOR/ImperialInterceptor.spec.js
redwulfe: test updated

server/game/cards/01_SOR/ImperialInterceptor.ts
```diff
@@ -0,0 +1,29 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '9002021213',
+            internalName: 'imperial-interceptor'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'You may deal 3 damage to a space unit',
```
AMMayberry1, line 15: Hey sorry, missed this on the first pass - another style thing is we don't use "You" language for the ability titles like this. So I think in this case it would just be "Deal 3 damage to a space unit". We also don't include "may" since that is implied by the presence of the "Pass ability button"

redwulfe: No worries all fixed.

AMMayberry1 (approved):

PR #44, opened 2024-09-12
Fix "when defeated" trigger issues, add Vanguard Infantry

AMMayberry1:
We discovered a bug causing when-defeated triggers to completely fail. The reason was that triggers were being activated after events were resolved, so if a card left the arena during the event resolution window (e.g. due to defeat resolving) then its abilities would be unregistered and the defeat ability would never fire.

As a (possibly temporary) mitigation, we changed the resolution to work as follows:

1. Triggers are activated before event resolution, all triggered abilities are saved
2. Resolve event(s)
3. Activate triggers _again_ to catch any triggers such as when-played that would've been missed at step 1. Duplicate triggers are ignored.
4. Check if any triggering events were cancelled; if so, remove the relevant triggered abilities from the list to resolve
5. Resolve all remaining triggered abilities

If the ability is a "then" ability (i.e. triggers are being held until the parent ability finishes resolving), we don't active any triggers but instead add them to the parent triggered ability window.

Added Vangard Infantry as a proof card to confirm that everything is working.

TODO: we need to investigate whether it would be better for cards to have their abilities online while not in the arena, to simplify cases like this. It's also needed for exploit and possibly ambush.

### DIFF
--- a/server/game/cards/01_SOR/VanguardInfantry.ts
+++ b/server/game/cards/01_SOR/VanguardInfantry.ts
@@ -1,0 +1,23 @@
+import AbilityHelper from '../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../core/card/NonLeaderUnitCard';
+
+export default class VanguardInfantry extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '5575681343',
+            internalName: 'vanguard-infantry',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenDefeatedAbility({
+            title: 'Give an experience token to a unit',
+            optional: true,
+            targetResolver: {
+                immediateEffect: AbilityHelper.immediateEffects.giveExperience()
+            }
+        });
+    }
+}
+
+VanguardInfantry.implemented = true;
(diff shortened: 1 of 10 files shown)


server/game/cards/01_SOR/VanguardInfantry.ts
```diff
@@ -0,0 +1,23 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '5575681343',
+            internalName: 'vanguard-infantry',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenDefeatedAbility({
+            title: 'Give an experience token to a unit',
```
adbronico, line 14: Capitalize Experience

server/game/core/event/EventWindow.js
```diff
@@ -60,14 +63,21 @@ class EventWindow extends BaseStepWithPipeline {
         this.events.forEach((event) => event.checkCondition());
     }
 
-    openWindow(abilityType) {
+    /** open a triggering window for any ability type other than {@link AbilityType.Triggered} */
+    openNonTriggeredWindow(abilityType) {
```
adbronico, line 67: So this is used for things like constant effects (like snoke or Dodonna?)

adbronico (changes requested):
Just one nitpick change

server/game/core/event/EventWindow.js
AMMayberry1: It’s for replacement effects since those have different timing than regular triggered abilities. Constant abilities do have a window where they might need to be ordered but it’s not in response to a game event like triggered abilities or replacement effects are.

I meant to rename this method to be more specific, will do that.

PR #45, opened 2024-09-13
Fix triggering requirements for triggered and action abilities

AMMayberry1:
L5R has different rules than SWU for allowing triggered abilities and actions to active. Made some updates to correct those requirements so our cards can function correctly.

**Triggered abilities:** this was pretty straightforward, there was already an "optional" flag that we could use in cases like Death Trooper for allowing a targeted to fizzle. It just wasn't working as expected for some reason, needed some additional implementation.

**Action abilities:** went through and changed how `meetsRequirements` works so that if the ability is an action and has a cost that impacts game state, it's automatically allowed to activate.

Needed to add some changes to the way card selectors work to satisfy the above. Added tests for some units that need the above and confirmed that they pass.

Also did some cleanup to prevent the use of multiple types of system specifiers (initiateAttack, immediateEffect, targetResolver(s)) on the same ability since the interactions between these are non-obvious and error-prone.


server/game/Interfaces.ts
```diff
@@ -19,7 +19,7 @@ export type ITriggeredAbilityProps = ITriggeredAbilityWhenProps | ITriggeredAbil
 export type IReplacementEffectAbilityProps = IReplacementEffectAbilityWhenProps | IReplacementEffectAbilityAggregateWhenProps;
 
 /** Interface definition for addActionAbility */
-export interface IActionAbilityProps<Source = any> extends Omit<IAbilityProps<AbilityContext<Source>>, 'optional'> {
+export type IActionAbilityProps<Source = any> = Exclude<IAbilityPropsWithSystems<AbilityContext<Source>>, 'optional'> & {
```
adbronico, line 22: I'm not super familiar with Typescript syntax. What is the purpose of the change from an interface to a type here, and is the '&' symbol meaningful in some way?

server/game/cards/01_SOR/DeathTrooper.ts
```diff
@@ -18,13 +18,15 @@ export default class DeathTrooper extends NonLeaderUnitCard {
                     cardTypeFilter: WildcardCardType.Unit,
                     controller: RelativePlayer.Self,
                     locationFilter: Location.GroundArena,
-                    immediateEffect: AbilityHelper.immediateEffects.damage({ amount: 2 })
+                    immediateEffect: AbilityHelper.immediateEffects.damage({ amount: 2 }),
+                    optional: true
```
adbronico, line 22: This ability isn't optional. I saw in the PR description that there was some special handling for this, but I feel like this is really unclear/unintuitive to be setting to true here. Is there more to it?

adbronico (changes requested):
See comments

server/game/Interfaces.ts
AMMayberry1: Yeah there's some nuance, but at the end of the day the behavior is the same.

The goal here is to make it so that the same properties object can't have both `targetResolver` and `immediateEffect` (or one of a couple other properties but we'll narrow the discussion to those two).

Previously, it was legal to do that because the ability properties interface was just a big union:

```typescript
interface IAbilityProps {
    targetResolver: IActionTargetResolver;
    immediateEffect: GameSystem | GameSystem[];
    title: string;
    locationFilter: LocationFilter | LocationFilter[];
    // ... some other stuff ...
}

interface ITriggeredAbilityProps extends IAbilityProps {
    // add some triggered stuff here
}

interface IActionAbilityProps extends IAbilityProps {
    // add some action stuff here
}
```

However, to split things into two groups ("types that use `immediateEffect`" and "types that use `targetResolver`"), I had to break those out into sub-interfaces and then join them back with a type union like this:

```typescript
 interface IAbilityProps {
    title: string;
    locationFilter: LocationFilter | LocationFilter[];
    // ... some other stuff ...
}

interface IImmediateEffectAbilityProps extends IAbilityProps {
    immediateEffect: GameSystem | GameSystem[];
}

interface ITargetResolverAbilityProps<Context> extends IAbilityProps<Context> {
    targetResolver: IActionTargetResolver;
}

// this forces the user to select one version or the other but not both
type IAbilityPropsWithSystems = IImmediateEffectAbilityProps | ITargetResolverAbilityProps;
```

However, since the derived interfaces `ITriggeredAbilityProps` and `IActionAbilityProps` are now extending a _type_ instead of an interface, they themselves have to become types. The & in this context is functionally the same as `extends`, it's used for appending new properties to an existing type:

```typescript
type ITriggeredAbilityProps = IAbilityPropsWithSystems & {
    // add some triggered stuff here
}

type IActionAbilityProps = IAbilityPropsWithSystems & {
    // add some action stuff here
}
```

server/game/Interfaces.ts
```diff
@@ -133,14 +108,70 @@ export type WhenType = {
...
+
+interface TargetResolversAbilityProps<Context> extends IAbilityProps<Context> {
+    targetResolvers: IActionTargetsResolver;
+}
+
+interface ImmediateEffectAbilityProps<Context> extends IAbilityProps<Context> {
+    immediateEffect: GameSystem | GameSystem[];
+}
+
+interface HandlerAbilityProps<Context> extends IAbilityProps<Context> {
```
Ennis0, line 154: I'd prefer "AbilityPropsWithX" over "XAbilityProps"

server/game/cards/01_SOR/DeathTrooper.ts
Ennis0: The target that must be fulfilled if possible but the ability still happens if not possible. Naming that property is tricky; the best I can think of is canResolveWithout.

server/game/core/ability/abilityTargets/CardTargetResolver.js
```diff
@@ -78,15 +78,27 @@ class CardTargetResolver {
...
-            if (legalTargets.length === 1) {
-                context.targets[this.name] = legalTargets[0];
-                if (this.name === 'target') {
-                    context.target = legalTargets[0];
-                }
+
+        const legalTargets = this.selector.getAllLegalTargets(context, player);
+        if (legalTargets.length === 0) {
+            if (context.stage === Stage.PreTarget) {
+                // if there are no targets at the pretarget stage, delay targeting until after costs are paid
```
Ennis0, line 85: By my reading of the official rules, target selection shouldn't happen until after costs anyways(example in 1.8.9). Could this system be brought in line with that? Would it potentially cause issues where we have to undo paying costs and we don't have the infrastructure for that or something?

AMMayberry1: Discussed in chat

server/game/cards/01_SOR/DeathTrooper.ts
AMMayberry1: Discussed in chat

adbronico (approved):
Good Veldakin good... merge him. Do it!

PR #46, opened 2024-09-13
Added Headhunting and the SequentialSystem

AMMayberry1:
Added a new meta-system back from jigoku, the SequentialSystem. This is used to chain a sequence of effects together with a shared triggering window at the end (effectively the same as a set of "then" effects).

Demonstrated the use of it with the Headhunters card, along with the SelectCardSystem for making card selections in a context where target resolvers aren't allowed - in this case, inside the chain of systems within SequentialSystem. Did some cleanup in there as well.


server/game/cards/02_SHD/Headhunting.ts
```diff
@@ -0,0 +1,39 @@
...
+            ])
+        });
+    }
+
+    private buildBountyHunterAttackEffect() {
+        return AbilityHelper.immediateEffects.selectCard({
+            innerSystem: AbilityHelper.immediateEffects.attack({
+                targetCondition: (card) => !card.isBase(),
+                effects: AbilityHelper.ongoingEffects.conditionalAttackStatBonus(
+                    (attacker: UnitCard) => attacker.hasSomeTrait(Trait.BountyHunter),
```
adbronico, line 30: Can we de-anonymize these properties? I.e. something like
cardCondition: (attacker: UnitCard) => attacker.hasSomeTrait(Trait.BountyHunter),
bonus: { power: 2, hp: 0}

If not possible, not a dealbreaker

server/game/gameSystems/SelectCardSystem.ts
```diff
@@ -32,16 +31,15 @@ export interface ISelectCardProperties extends ICardTargetSystemProperties {
 
 /**
  * A wrapper system for adding a target selection prompt around the execution the wrapped system.
- * Only used for adding a selection effect to a system that is part of a cost.
+ * Functions the same as a targetResolver and used in situations where one can't be created (e.g., costs).
  */
-// TODO: why is this class needed for evaluating costs when systems already have target evaluation and selection built in?
 export class SelectCardSystem extends CardTargetSystem {
```
adbronico, line 36: I think we should add to the Readme something that says what each system is used for. It's not clear to me what situations call for SelectCard vs. not

server/game/gameSystems/SequentialSystem.ts
```diff
@@ -0,0 +1,73 @@
...
+    public constructor(gameSystems: GameSystem[]) {
+        super({ gameSystems: gameSystems });
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public override eventHandler() {}
+
+    public override queueGenerateEventGameSteps(events: GameEvent[], context: AbilityContext, additionalProperties = {}): void {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        for (const gameSystem of properties.gameSystems) {
```
adbronico, line 20: Is the order of the gamesystems going to be consistent in all cases?

test/server/cards/02_SHD/Headhunting.spec.js
```diff
@@ -0,0 +1,132 @@
...
+                expect(this.theMandalorian.damage).toBe(3);
+
+                expect(this.player2).toBeActivePlayer();
+            });
+
+            it('should be able to be passed', function () {
+                this.player1.clickCard(this.headhunting);
+
+                // first attack - skip target resolution, go straight to next attack
+                expect(this.player1).toBeAbleToSelectExactly([this.atst, this.reputableHunter, this.theMandalorian, this.wampa]);
```
adbronico, line 64: I think we need a way to differentiate the passes here. If someone plays this card and then passes, why is it passing the first attack and not the whole effect? This can be a TODO, but I would imagine we might need options like "Pass attack" vs. "Pass all attacks" if we don't want to make them choose the amount of attacks up front.

This brings up an interesting case Im not sure of. I know there are windows for trigger resolution within an attack, but if a unit is defeated by the first attack of Headhunting, does its When Defeated resolve before the second attack? I assume all of them are resolved at the end of the whole event resolution.

adbronico (changes requested):
See comments

server/game/gameSystems/SelectCardSystem.ts
AMMayberry1: Agreed. There are some missing pieces that will come in the Ambush PR, at that point I'll have enough to fill out a readme section on how this is all used.

server/game/gameSystems/SequentialSystem.ts
AMMayberry1: Oh yeah I forgot to add a docstr on this system. The game systems will always execute sequentially. It's basically an implementation of the text "do [X], then do [Y], then do...".

test/server/cards/02_SHD/Headhunting.spec.js
AMMayberry1: Will discuss the first question in the UI chat. For the second question, the on defeat will happen in the next available trigger window (so immediately after the unit is defeated). Per this rule:

IMG_6590

AMMayberry1: Though right now the implementation is buggy, but I'm planning a future PR focused on windows and timing so I'll address it then.

AMMayberry1: Actually maybe I will just put a TODO for the first one and then trigger the discussion, since this is blocking some stuff.

adbronico (approved):
Your Jedi merge tricks won't work on me

PR #47, opened 2024-09-15
Implemented set 1 Chewbacca unit("Loyal Companion").

Ennis0:


adbronico (approved):
But what about the droid attack on the Wookiees?

test/server/cards/01_SOR/ChewbaccaLoyalCompanion.spec.js
```diff
@@ -0,0 +1,23 @@
...
+                        groundArena: ['wampa']
+                    },
+                    player2: {
+                        groundArena: [{ card: 'chewbacca#loyal-companion', exhausted: true }]
+                    }
+                });
+            });
+
+            it('readies Chewbacca when he is attacked', function () {
+                this.player1.clickCard(this.wampa);
```
AMMayberry1, line 17: Can we set it up so that it readies twice in the same phase just to confirm the trigger works multiple times. Can either exhaust it manually and pass player 2 or have it do an attack, either's fine

Ennis0: Sure.

test/server/core/abilities/SimultaneousTriggers.spec.js
```diff
@@ -0,0 +1,47 @@
...
+                });
+            });
+
+            it('should prompt the active player(controller of Sabine) which player\'s triggers to resolve first', function () {
+                this.player1.clickCard(this.sabineWren);
+                //Don't need to click Chewbacca due to sentinel
+                expect(this.player1).toHavePrompt('Both players have triggered abilities in response. Choose a player to resolve all of their abilities first:');
+                expect(this.player2).toHavePrompt('Waiting for opponent to choose a player to resolve their triggers first');
+                this.player1.clickPrompt('You');
+                expect(this.chewbacca.exhausted).toBe(true);
```
AMMayberry1, line 22: Can you add some whitespace just to make these more readable

test/server/core/abilities/SimultaneousTriggers.spec.js
```diff
@@ -0,0 +1,47 @@
+describe('Simultaneous triggers', function() {
```
AMMayberry1, line 1: Would prefer to match the test file names as closely to the real files as possible - could we move this to be core/gameSteps/abilityWindow/TriggeredAbilityWindow.spec.js. Then we can add in more tests here as we fill out the functionality.

test/server/core/abilities/SimultaneousTriggers.spec.js
Ennis0: Done.

test/server/core/abilities/SimultaneousTriggers.spec.js
Ennis0: Done.

AMMayberry1 (approved):

PR #48, opened 2024-09-15
Implemented Yoda, Old Master

Ennis0:


server/game/cards/01_SOR/YodaOldMaster.ts
```diff
@@ -0,0 +1,30 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenDefeatedAbility({
+            title: 'Choose any number of players to draw 1 card',
+            targetResolver: {
+                mode: TargetMode.Select,
+                choices: (context) => ({
+                    //TODO: Make this selection prompt less awkward: create a way to select any number of options from an arbitrary list.
```
AMMayberry1, line 19: Agreed. It looks like this is the only ability that allows you to select multiple players so there's not a strong need to create a special tool for this situation, but maybe more generally something that would do like checkboxes could be helpful in this and a few other situations. Good note to discuss with the FE folks at some point.

test/server/cards/01_SOR/YodaOldMaster.spec.js
```diff
@@ -0,0 +1,68 @@
...
+                expect(this.player1.hand.length).toBe(2);
+                expect(this.player2.hand.length).toBe(0);
+                expect(this.player2).toBeActivePlayer();
+            });
+
+            it('should draw a card for each selected player', function () {
+                this.player1.clickCard(this.yoda);
+                this.player1.clickCard(this.vigilantHonorGuards);
+                expect(this.yoda).toBeInLocation('discard');
+                expect(this.player1).toHaveEnabledPromptButtons(['Me', 'Opponent', 'Me and Opponent', 'No one']);
```
AMMayberry1, line 35: No need to change, just a note for the future - in a situation where there are multiple tests that set up the exact same scenario, I'm okay if we don't repeat the exact same expectation every time. So e.g. in this case, since the these two expectations are covered in the first test, we could also remove them in subsequent tests just to make the overall code a little shorter.

It's totally personal preference, I'm very fine to keep this as-is if you prefer. Just an fyi in case you want to have to put in less code in the future.

AMMayberry1 (approved):

PR #49, opened 2024-09-15
Add some polish to lasting effects

AMMayberry1:
There was some basic infra in place for lasting effects but it badly needed some polish to be a little more comprehensible. Added helper methods that focus on the two common cases ("for this phase" and "for this attack"), and made the process for creating them much more streamlined.

Added a few example cards with the new setup and updated the documentation.

Added a helper method provided by Matchstick for simplifying situations where a derived GameSystem needs to strip out hidden properties and re-add them for passing back up the chain.

Added a guard that will catch if two implemented cards have the same class name (since I made that mistake recently...)


docs/implementing-cards.md
```diff
@@ -844,61 +843,49 @@ this.action({
...
+            })
+        }
+    });
+}
 ```
 
-To apply an effect to last until the end of the round, use `untilEndOfRound`:
+#### "For this attack" effects
+
+Any lasting effects applied to the attacker or the defender for the duration of the attack can be added via the `attackerLastingEffects` and `defenderLastingEffects` properties of an attack. There is also a condition property which can be used to control whether the effect is applied. See Fleet Lieutenant for an example:
```
adbronico, line 876: We should add a TODO somewhere here for multiple target attacks (i.e. Maul) or multiple attackers on one target (not yet a thing but possible)

server/game/cards/01_SOR/SurpriseStrike.ts
```diff
@@ -0,0 +1,25 @@
...
+        return {
+            id: '3809048641',
+            internalName: 'surprise-strike',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Attack with a unit. It gets +3/+0 for this attack.',
+            targetResolver: {
```
adbronico, line 16: What limits this to only choosing friendly units? Is it the attack itself?

adbronico (dismissed):
Merge you will. Good relations with the Wookiees you have.

PR #51, opened 2024-09-18
Add system for "if" conditions and some cards to test triggered and constant abilities

AMMayberry1:
Rounded out the ConditionalSystem so that it could be used to support most cases of the word "if" in card abilities (the primary exception being "if you do", which will require special logic).

Implemented a few cards to demonstrate how to implement the word "if" when used in:
- Normal triggered abilities
- Triggered abilities on attachments
- Constant abilities on attachments

Examples for other cases (particularly actions) will come in future PRs but the core functionality should be the same for all.

Also did some improvements to engine components to resolve issues found during testing, mainly in the TriggeredAbilityWindow. The fact that it would error out in a situation where there are multiple triggers on the same unit became a blocker, so for now we are just presenting every ability by text name regardless of any additional context. This will need to be improved eventually, but for the simple cases that come up in tests it's fine for now.

### DIFF
--- a/test/server/cards/01_SOR/JedhaAgitator.spec.js
+++ b/test/server/cards/01_SOR/JedhaAgitator.spec.js
@@ -1,27 +1,27 @@
 describe('Jedha Agitator', function() {
     integration(function() {
-        // describe('Jedha Agitator\'s on attack ability', function() {
-        //     beforeEach(function () {
-        //         this.setupTest({
-        //             phase: 'action',
-        //             player1: {
-        //                 groundArena: ['jedha-agitator'],
-        //             },
-        //             player2: {
-        //                 groundArena: ['wampa'],
-        //                 spaceArena: ['cartel-spacer']
-        //             }
-        //         });
-        //     });
+        describe('Jedha Agitator\'s on attack ability', function() {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['jedha-agitator'],
+                    },
+                    player2: {
+                        groundArena: ['wampa'],
+                        spaceArena: ['cartel-spacer']
+                    }
+                });
+            });
 
-        //     it('should do nothing if no leader is deployed', function () {
-        //         this.player1.clickCard(this.jedhaAgitator);
-        //         this.player1.clickCard(this.p2Base);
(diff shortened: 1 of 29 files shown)


test/server/cards/01_SOR/JedhaAgitator.spec.js
```diff
@@ -0,0 +1,84 @@
...
+
+                expect(this.player1).toHaveEnabledPromptButton('If you control a leader unit, deal 2 damage to a ground unit or base');
+                expect(this.player1).toHaveEnabledPromptButton('Saboteur: defeat all shields');
+                expect(this.jedhaAgitator.exhausted).toBe(true);
+
+                this.player1.clickPrompt('If you control a leader unit, deal 2 damage to a ground unit or base');
+                expect(this.player1).toBeAbleToSelectExactly([this.wampa, this.jedhaAgitator, this.battlefieldMarine, this.p1Base, this.p2Base, this.bobaFett]);
+                this.player1.clickCard(this.wampa);
+                expect(this.wampa.damage).toBe(2);
+                expect(this.p2Base.damage).toBe(2);
```
Ennis0, line 55: Jedha Agitator has saboteur, so it will do 3 damage here once Boba Fett, Daimyo is actually implemented. Probably swap Boba for a leader with an irrelevant ability.

test/server/cards/01_SOR/VadersLightsaber.spec.js
```diff
@@ -0,0 +1,74 @@
...
+            });
+
+            it('should deal 4 damage to a ground unit when attached to the Darth Vader unit', function () {
+                this.player1.clickCard(this.vadersLightsaber);
+                expect(this.player1).toBeAbleToSelectExactly([this.darthVader, this.wampa]);    // cannot attach to vehicles
+                this.player1.clickCard(this.darthVader);
+
+                expect(this.darthVader).toHaveExactUpgradeNames(['vaders-lightsaber']);
+                expect(this.player1).toBeAbleToSelectExactly([this.wampa, this.darthVader]);
+                expect(this.player1).not.toHavePassAbilityPrompt();
```
Ennis0, line 26: The when played for Vader's Lightsaber is a "may" ability. Am I misreading or does this line contradict that?

PR #52, opened 2024-09-19
Added Iden to test conditional action

AMMayberry1:
Added the Iden leader to test that an "if" condition on an action ability works correctly. To facilitate this, added the `UnitsDefeatedThisPhaseWatcher` and some small improvements to the test infra.


server/game/core/Constants.ts
```diff
@@ -318,8 +318,9 @@ export enum AbilityRestriction {
 }
 
 export enum StateWatcherName {
+    AttacksThisPhase = 'attacksThisPhase',
```
adbronico, line 321: Does this record the attack as a whole, units that were attacked, or units that did the attacking?

server/game/gameSystems/DeployLeaderSystem.ts
```diff
@@ -1,6 +1,7 @@
 import type { AbilityContext } from '../core/ability/AbilityContext';
 import type { Card } from '../core/card/Card';
 import { Arena, CardType, EventName } from '../core/Constants';
+import { GameEvent } from '../core/event/GameEvent';
```
adbronico, line 4: Is this import used?

server/game/cards/01_SOR/IdenVersioInfernoSquadCommander.ts
```diff
@@ -0,0 +1,51 @@
...
+    protected override setupLeaderSideAbilities() {
+        this.addActionAbility({
+            title: 'Heal 1 from base if an opponent\'s unit was defeated this phase',
+            cost: AbilityHelper.costs.exhaustSelf(),
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => {
+                    const opponentUnitsDefeatedThisPhase = this.unitsDefeatedThisPhaseWatcher.getDefeatedUnitsControlledByPlayer(context.source.controller.opponent);
+                    return opponentUnitsDefeatedThisPhase.length > 0;
+                },
+                trueImmediateEffect: AbilityHelper.immediateEffects.heal((context) => {
```
adbronico, line 30: This seems a bit wordy but I can't think of a better name for it atm

server/game/cards/01_SOR/IdenVersioInfernoSquadCommander.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addActionAbility({
+            title: 'Heal 1 from base if an opponent\'s unit was defeated this phase',
+            cost: AbilityHelper.costs.exhaustSelf(),
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => {
+                    const opponentUnitsDefeatedThisPhase = this.unitsDefeatedThisPhaseWatcher.getDefeatedUnitsControlledByPlayer(context.source.controller.opponent);
+                    return opponentUnitsDefeatedThisPhase.length > 0;
```
adbronico, line 28: Will opponentUnitsDefeatedThisPhase always be an array or could it ever be undefined/null?

test/server/cards/01_SOR/IdenVersioInfernoSquadCommander.spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+                this.player1.clickCard(this.idenVersio);
+                this.player1.clickPrompt('Heal 1 from base if an opponent\'s unit was defeated this phase');
+                expect(this.idenVersio.exhausted).toBe(true);
+                expect(this.p1Base.damage).toBe(5);
+
+                this.idenVersio.exhausted = false;
+                this.player2.passAction();
+
+                // case 3: enemy unit defeated
```
adbronico, line 41: I feel like each of these cases should be its own unit test. I know we've been consolidating them sometimes but I think we might want to start making our tests more granular

test/server/cards/01_SOR/IdenVersioInfernoSquadCommander.spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+                this.player2.passAction();
+
+                // case 2: enemy unit defeated
+                this.player1.clickCard(this.wampa);
+                this.player1.clickCard(this.specforceSoldier);
+                expect(this.p1Base.damage).toBe(4);
+            });
+
+            // TODO: once leader shields and defeat timing is fixed, add a test for Iden's ability to heal base when she is defeated
```
adbronico, line 89: Can we add another TODO to make sure she properly interacts with simultaneous defeats (i.e. if Superlaser kills her and also enemy units)

adbronico (changes requested):
Broadly looks good. See comments

server/game/core/Constants.ts
AMMayberry1: image

test/server/cards/01_SOR/IdenVersioInfernoSquadCommander.spec.js
AMMayberry1: The problem is that our tests are very heavy and take a lot of time since to set up, so there would be a concrete impact if we suddenly were adding a lot more tests. Eventually we should look at reducing that but it probably requires state backup tbh (or it might just be that jasmine is slow in general).

I'm particularly aware of it because we have finite Github Actions time, otherwise I'd be more 50/50 about it.

AMMayberry1: E.g., when I tried to run the L5R tests the first time with all of its cards, I think it took 5 minutes.

adbronico (approved):
I have a bad feeling about this...

PR #54, opened 2024-09-19
Changes to trigger resolution to more closely match the rules.

Ennis0:
Constructing an EventWindow now specifies whether or not triggers triggered during that window should resolve right after those events. If this is set to false, the EventWindow will add the triggers triggered by its events to the TriggeredAbilityWindow belonging to its parent EventWindow. This structure implements the game rules concept of nested actions and triggers.


server/game/core/gameSteps/AbilityResolver.js
```diff
@@ -75,6 +76,7 @@ class AbilityResolver extends BaseStepWithPipeline {
                 context: this.context
             };
             if (this.context.ability.isCardPlayed()) {
+                resolveTriggersAfter = true;
```
AMMayberry1, line 79: As noted in the comment on PlayUnitAction, I don't think this is the spot to control this since whether or not the play card action resolves immediately is contextual (e.g. u-wing play cards are nested, alliance dispatcher is not). If we need to keep it here for now for stuff to work right that's fine, but we should a TODO.

That said, I think from other comments / discussion this might not be needed here.

server/game/core/gameSteps/abilityWindow/TriggeredAbilityWindow.ts
```diff
@@ -338,6 +331,6 @@ export class TriggeredAbilityWindow extends BaseStep {
     }
 
     public override toString() {
-        return this.toStringName;
+        return `'TriggeredAbilityWindow: ${this.triggeringEvents.map((event) => event.name).join(', ')}'`;
```
AMMayberry1, line 334: Oh yeah good catch

server/game/core/event/EventWindow.js
```diff
@@ -57,6 +59,12 @@ class EventWindow extends BaseStepWithPipeline {
     setCurrentEventWindow() {
         this.previousEventWindow = this.game.currentEventWindow;
         this.game.currentEventWindow = this;
+        if (this.resolveTriggersAfter) {
+            this.triggeredAbilityWindow = new TriggeredAbilityWindow(this.game, this, AbilityType.Triggered);
```
AMMayberry1, line 63: Since we're initializing `this.triggeredAbilityWindow` here, can remove the initialize-if-null logic in `triggerEventsForWindow` (and probably just omit that method completely tbh)

server/game/core/gameSteps/abilityWindow/TriggeredAbilityWindow.ts
```diff
@@ -45,31 +44,26 @@ export class TriggeredAbilityWindow extends BaseStep {
...
 
-    public emitEvents() {
-        this.eventsEmitted = true;
+    public addTriggeringEvents(events:GameEvent[]) {
+        this.triggeringEvents.push(...events);
+    }
 
-        const events = this.eventWindow.events.filter((event) => !this.eventsToExclude.includes(event));
+    public emitEvents() {
+        const events = this.triggeringEvents.filter((event) => !this.eventsToExclude.includes(event));
```
AMMayberry1, line 55: I missed this previously - can we add a check here that also filters out any events that are already cancelled before we emit them

server/game/core/event/EventWindow.js
```diff
@@ -57,6 +59,12 @@ class EventWindow extends BaseStepWithPipeline {
...
+        } else if (this.previousEventWindow) {
+            this.triggeredAbilityWindow = this.previousEventWindow.triggeredAbilityWindow;
+            this.triggeredAbilityWindow.addTriggeringEvents(this.events);
```
AMMayberry1, line 66: We actually can't set the triggering events here because it's possible that the set of events may change before we reach `executeHandler`. For example, `createContingentEvents` could do it - it is used for generating the events that defeat upgrades if the owning unit is defeated, as well as potentially some additional cases down the road. Replacement effects can also add new events.

Also, I suspect that once we do a deep dive on unit defeat timing we will find that we need to insert defeat events into the window in the middle of resolution (see my comment on the triggered ability window tests). So all that to say, better if we keep the set of events being triggered as flexible as possible until `executeHandler` is totally done.

server/game/core/gameSteps/AbilityResolver.js
```diff
@@ -88,6 +90,7 @@ class AbilityResolver extends BaseStepWithPipeline {
                 }));
             }
             if (this.context.ability.isActivatedAbility()) {
+                resolveTriggersAfter = [AbilityType.Action, AbilityType.Event, AbilityType.Triggered].includes(this.context.ability.type);
```
AMMayberry1, line 93: Same comment as above, I think this is maybe not the right place to do this determination because there may be additional context. For example, I think this logic doesn't work if the ability currently being resolved is a "then" ability, since `context.ability` would be pointing at the original card ability (I _think_).

I'm assuming the goal here is something like "make sure that a card ability gets a triggering window at the end", in which case maybe this should be something that gets controlled by CardAbility.ts? Basically the way it works is that every resolvable ability is a CardAbilityStep. The whole printed ability (potentially multiple steps) is a CardAbility.

So the way it works in practice if there are "then" abilities is that you trigger a CardAbility and it will generate additional CardAbilitySteps to handle the "then" part(s) before it finishes resolving. So it feels like CardAbility is where we need to enforce that trigger resolution happens. Maybe something like a `resolveTriggersAfter = false` constructor parameter on CardAbilityStep, and then CardAbility just overrides it with true when it calls `super()`.

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
```diff
@@ -45,48 +45,60 @@ describe('Leia Organa, Alliance General', function() {
...
-                        leader: 'leia-organa#alliance-general'
-                    },
-                    player2: {
-                        groundArena: ['sundari-peacekeeper'],
-                        spaceArena: ['tie-advanced']
-                    }
-                });
-            });
 
-            it('can be activated with no target', function () {
```
AMMayberry1, line 84: Why remove this test?

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
```diff
@@ -45,48 +45,60 @@ describe('Leia Organa, Alliance General', function() {
...
+            it('should allow appropriate attack triggers to happen when either attack is declared', function () {
```
AMMayberry1, line 62: Good addition, awesome to see that it works!!

test/server/core/gameSteps/abilityWindow/TriggeredAbilityWindow.spec.js
```diff
@@ -47,5 +47,61 @@ describe('Simultaneous triggers', function() {
                 expect(this.chewbacca.exhausted).toBe(false);
             });
         });
+
+        describe('Two units with a when defeated ability killing each other', function () {
```
AMMayberry1, line 51: Also very cool to see that these work!! One thing to note though (might be worth a TODO for the benefit of anyone reading the tests) is that this case only works because the "when defeat" abilities are triggered by the defeat of the card that owns the ability.

For example, Iden's ability may not trigger correctly in this situation because if her defeat event resolves first, she will have left the field before the other unit defeat resolves.

I think defeat logic is definitely something for a different PR though just b/c the timing is so weird.

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
```diff
@@ -45,48 +45,60 @@ describe('Leia Organa, Alliance General', function() {
...
+                // unit with trigger first
                 this.player1.clickCard(this.leiaOrgana);
                 this.player1.clickPrompt('Attack with a Rebel unit');
+                this.player1.clickCard(this.sabineWren);
+                this.player1.clickCard(this.sundariPeacekeeper);
+
+                expect(this.sabineWren.damage).toBe(0);
+                expect(this.sundariPeacekeeper.damage).toBe(0);
+                expect(this.p2Base.damage).toBe(0);
+                expect(this.player1).toBeAbleToSelectExactly([this.sundariPeacekeeper, this.p1Base, this.p2Base]);
```
AMMayberry1, line 72: Can you just add a comment explaining that this block is triggering the Sabine ability, since this test setup is kind of complex (and ditto below)

server/game/core/event/ThenEventWindow.js
```diff
@@ -20,9 +20,6 @@ class ThenEventWindow extends EventWindow {
         for (let event of this.events) {
             this.previousEventWindow.addEvent(event);
         }
```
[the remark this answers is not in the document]
Ennis0: Looks that way. I will do so.

server/game/core/event/EventWindow.js
```diff
@@ -57,6 +59,12 @@ class EventWindow extends BaseStepWithPipeline {
...
+        }
```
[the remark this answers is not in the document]
Ennis0: An EventWindow with no parent that isn't set to resolveTriggersAfter = true will skip both blocks. I checked, and it looks like this is true of onBeginRound, onPhaseCreated, and onRoundEnded. None of the tests fail, so I don't think anything's going wrong with the gamestate.

server/game/core/event/EventWindow.js
```diff
@@ -21,6 +21,8 @@ class EventWindow extends BaseStepWithPipeline {
 
         this.triggeredAbilityWindow = null;
 
+        this.resolveTriggersAfter = resolveTriggersAfter;
```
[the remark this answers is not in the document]
Ennis0: I can't think of anything great. Does "resolveTriggersAfterEvents" disambiguate at all?

server/game/actions/PlayUnitAction.ts
```diff
@@ -35,7 +35,7 @@ export class PlayUnitAction extends PlayCardAction {
                 controller: player
             }).generateEvent(context.source, context),
             cardPlayedEvent
-        ]);
+        ], true);
```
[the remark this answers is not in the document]
Ennis0: That is an important point. Where do you think including the parameter should be included? PlayCardAction? All the way up in PlayerOrCardAbility? A property of PlayCardContext or AbilityContext? I'm not sure exactly the boundaries of what all these things are used for so I don't know quite where this would fit.

server/game/core/gameSteps/abilityWindow/TriggeredAbilityWindow.ts
Ennis0: Done.

test/server/cards/01_SOR/LeiaOrganaAllianceGeneral.spec.js
Ennis0: I think I just overlooked it when merging. I'll go look for it and paste it back in.

server/game/core/event/EventWindow.js
Ennis0: Done. Some EventWindows still won't have a TriggeredAbiltyWindow at all(see the other thread about setCurrentEventWindow) so I use this.triggeredAbilityWindow?. in a few spots.

server/game/core/event/EventWindow.js
Ennis0: So, do this(line 66) in executeHandler instead? I'm not sure if the TriggeredEventWindow would need to know this EventWindow's events in the meantime.

server/game/core/event/EventWindow.js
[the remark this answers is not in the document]
AMMayberry1: I don't think so - maybe a different direction, like "hasTriggeringWindow" since in some cases it technically doesn't have a triggering window? Or like "parentWindowResolvesTrigger" or something

server/game/actions/PlayUnitAction.ts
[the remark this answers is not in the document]
AMMayberry1: I think PlayCardAction would work since it's only relevant for those classes.

server/game/core/event/EventWindow.js
AMMayberry1: Yeah I think so, iirc that's the way it was beforehand so that should still work

server/game/core/event/EventWindow.js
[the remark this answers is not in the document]
Ennis0: How about ownsATriggerWindow?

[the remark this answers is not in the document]
AMMayberry1: Maybe hasTriggerWindow or hasTriggeredAbilityWindow (since "trigger" could mean a lot of things)? "Owns" feels a little weird in this context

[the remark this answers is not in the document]
Ennis0: Well all EventWindows will have a TriggeredAbilityWindow. It's just a question of whether that window is shared with other EventWindows and which EventWindow will cause it to resolve.

server/game/core/ability/PlayerOrCardAbility.js
```diff
@@ -54,6 +55,8 @@ class PlayerOrCardAbility {
         this.keyword = null;
         this.type = type;
         this.optional = !!properties.optional;
+        //Nested abilities rule(Comp Rules 2.0 7.6.11): triggers should always resolve triggers triggered during their effects right after their effects
+        this.resolveTriggersAfter = this.type === AbilityType.Triggered || !!properties.resolveTriggersAfter;
```
AMMayberry1, line 59: I don't think we can really do a proper test of nesting until on-defeat timing is working correctly, sadly. Could you just leave a TODO here though that we need to test this?

server/game/core/event/EventWindow.js
```diff
@@ -6,7 +6,7 @@ const { default: Contract } = require('../utils/Contract.js');
 // const KeywordAbilityWindow = require('../gamesteps/keywordabilitywindow.js');
 
 class EventWindow extends BaseStepWithPipeline {
-    constructor(game, events) {
+    constructor(game, events, ownsATriggerWindow = false) {
```
AMMayberry1, line 9: Need jsdoc comment explaining what the trigger parameter does

server/game/core/gameSteps/abilityWindow/InitiateAbilityEventWindow.js
```diff
@@ -65,7 +65,7 @@ class InitiateAbilityEventWindow extends EventWindow {
         this.eventsToExecute = this.events.sort((event) => event.order);
 
         // we emit triggered abilities here to ensure that they get triggered in case e.g. a card is defeated during event resolution
-        this.triggerEventsForWindow();
+        this.triggeredAbilityWindow?.emitEvents();
```
AMMayberry1, line 68: Shouldn't need the `?` anymore I think?

server/game/core/event/EventWindow.js
```diff
@@ -93,10 +102,11 @@ class EventWindow extends BaseStepWithPipeline {
     }
 
     executeHandler() {
+        this.triggeredAbilityWindow.addTriggeringEvents(this.events);
```
AMMayberry1, line 105: Could we save a step and just pass the events in at `emitEvents()`?

server/game/core/ongoingEffect/OngoingEffectEngine.ts
```diff
@@ -75,7 +75,7 @@ export class OngoingEffectEngine {
                     }
                     const actionEvents = [];
                     properties.gameAction.queueGenerateEventGameSteps(actionEvents, context);
-                    this.game.queueSimpleStep(() => this.game.openThenEventWindow(actionEvents), 'openAdditionThenEventWindow');  // TODO: why is it using this window type?
+                    this.game.queueSimpleStep(() => this.game.openEventWindow(actionEvents), 'openAdditionThenEventWindow');  // TODO: why is it using this window type?
```
AMMayberry1, line 78: Can you do me a favor and change the step name to be `'openDelayedActionsWindow'` and also remove the comment, I know what this is doing now

server/game/core/event/EventWindow.js
Ennis0: Yes.

server/game/core/Game.js
```diff
@@ -842,13 +842,14 @@ class Game extends EventEmitter {
      * Creates a game GameEvent, and opens a window for it.
      * @param {String} eventName
      * @param {Object} params - parameters for this event
+     * @param {boolean} resolveTriggersAfter - whether the game's GlobalTriggeredAbilityWindow should be resolved after the event
      * @param {(GameEvent) => void} handler - (GameEvent + params) => undefined
      * returns {GameEvent} - this allows the caller to track GameEvent.resolved and
      * tell whether or not the handler resolved successfully
      */
-    createEventAndOpenWindow(eventName, params = {}, handler = () => true) {
+    createEventAndOpenWindow(eventName, params = {}, resolveTriggersAfter = false, handler = () => true) {
```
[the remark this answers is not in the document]
Ennis0: Good catch.
PlayerOrCardAbility and derived classes now also have a resolveTriggersAfter which doesn't make sense as ownsTriggerWindow. Do you think that one's okay as resolveTriggersAfter or is there a better name in that case?

AMMayberry1 (approved):

PR #57, opened 2024-09-19
Implemented Vigilant Honor Guards

Ennis0:


test/server/cards/01_SOR/VigilantHonorGuards.spec.js
```diff
@@ -0,0 +1,45 @@
...
+                this.player1.clickCard(this.consularSecurityForce);
+                expect(this.player1).toBeAbleToSelectExactly([this.vigilantHonorGuards, this.p2Base]); //no sentinel
+                this.player1.clickCard(this.p2Base);
+
+                //checking if they regain sentinel when fully healed
+                this.player2.clickCard(this.repair);
+                this.player2.clickCard(this.vigilantHonorGuards);
+                expect(this.vigilantHonorGuards.damage).toBe(0);
+
+                this.moveToNextActionPhase();
```
AMMayberry1, line 35: Prefer to just un-exhaust the unit when that's an option since this will move through the entire regroup phase, which might be kind of slow. Probably not noticeable for a single test but it'll add up

AMMayberry1 (approved):

PR #58, opened 2024-09-20
Add `simultaneous` meta-system and The Force is With Me

AMMayberry1:
Add the second meta-system, `SimultaneousSystem`. This is for the (more common) case where multiple effects resolve in the same triggering window. Add The Force is With Me as a test card.

Since L5R had multiple ways of handling abilities with multiple effects which did not clearly state the differences between them, we're consolidating down to "sequential" and "simultaneous" and we'll expand as needed. Thus we're removing the option to provide an array of GameSystems as an immediate effect, instead forcing people to use `AbilityHelper.immediateEffects.simultaneous()` to make the intent clear.

Updated the documentation to include how to use the new systems, as well as backfilling some documentation for other recent PRs.

Fixed a bug in the `.toHaveExactUpgradeNames()` test expectation.


server/game/cards/01_SOR/TheForceIsWithMe.ts
```diff
@@ -0,0 +1,32 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '7861932582',
+            internalName: 'the-force-is-with-me',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Give 2 Experience, a Shield if there is a Force unit, and optionally attack',
```
Ennis0, line 15: "...Shield if you control a Force unit..." is more accurate.

server/game/cards/01_SOR/TheForceIsWithMe.ts
```diff
@@ -0,0 +1,32 @@
...
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                immediateEffect: AbilityHelper.immediateEffects.simultaneous([
```
adbronico, line 18: I love everything about how this code works. It makes so much sense.

adbronico (dismissed):
The Force Is With Merge

PR #59, opened 2024-09-20
Migrate set cards into subfolders by type

AMMayberry1:
Organizing the card tests / impls into subfolders within each set by card type to make things a little easier to navigate.


adbronico (approved):
Boring conversation anyway

PR #60, opened 2024-09-21
Update the get-cards script

AMMayberry1:


adbronico (approved):
It's an older code, but it checks out.

PR #63, opened 2024-09-22
Implement Smuggle

adbronico:
TODO for future PRs:
1. Tech / gain Smuggle
2. First Light / additional cost logic
3. Armed to the Teeth implementation


server/game/core/card/Card.ts
```diff
@@ -38,6 +38,9 @@ export class Card extends OngoingEffectSource {
     public static implemented = false;
 
     public readonly aspects: Aspect[] = [];
+    public readonly smuggleAspects?: Aspect[];
+    public readonly smuggleCost?: number;
+    public readonly additionalSmuggleCosts?: boolean;
```
AMMayberry1, line 43: Just checking, is this temporary? I was expecting that the smuggle details would be wrapped up in the keyword instance, since that's how it's done for raid and restore and there might be multiple instance of smuggle present on one card so it can't necessarily be represented like this.

server/game/core/ability/PlayCardFromHandAction.ts
```diff
@@ -0,0 +1,22 @@
+import * as CostLibrary from '../../costs/CostLibrary';
+import { IActionTargetResolver } from '../../TargetInterfaces';
+import { Card } from '../card/Card';
+import { PlayType } from '../Constants';
+import { ICost } from '../cost/ICost';
+import { PlayCardAction } from './PlayCardAction';
+
+export class PlayCardFromHandAction extends PlayCardAction {
```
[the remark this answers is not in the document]
adbronico: @AMMayberry1 I'm drawing a blank on what we could do here besides renaming this. I assume we would need a new play type for playing from the Discard (TIE Silencer) and for Deck (such as Ezra Bridger). Would there be some sort of way to make this like templated like PlayCardAction<Location> (I don't know Typescript well enough to have the syntaxes down or know what it can do)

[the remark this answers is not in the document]
AMMayberry1: Yeah sorry, thinking on it more I think the changes will be small enough that we don't need to account for them. I was thinking we could just use this class with some additional input parameters to control where it's legal to be played from but on reflection that might not end up being the best solution so we can just ignore that for now.

test/server/core/abilities/keyword/Smuggle.spec.js
```diff
@@ -0,0 +1,72 @@
...
+                    },
+                    player2: {
+                    }
+                });
+            });
+
+            it('a unit can be played for its smuggle cost', function () {
+                expect(this.player1.countSpendableResources()).toBe(18);//Sanity check before we Smuggle
+                this.player1.clickCard(this.collectionsStarhopper);
+                expect(this.collectionsStarhopper).toBeInLocation('space arena');
```
AMMayberry1, line 29: Check that it comes in exhausted

test/server/core/abilities/keyword/Smuggle.spec.js
```diff
@@ -0,0 +1,72 @@
...
+                this.player1.clickCard(this.smugglersAid);
+                expect(this.p1Base.damage).toBe(0);
+                expect(this.player1.countExhaustedResources()).toBe(3);
+                expect(this.player1.countSpendableResources()).toBe(15);
+                expect(this.mercenaryGunship).toBeInLocation('resource');
+            });
+
+            it('a card without Smuggle cannot be played from resources', function () {
+                expect(this.battlefieldMarine).not.toHaveAvailableActionWhenClickedBy(this.player1);
+                expect(this.player1).toBeActivePlayer();
```
AMMayberry1, line 60: This check is redundant

test/server/core/abilities/keyword/Smuggle.spec.js
```diff
@@ -0,0 +1,72 @@
...
+            });
+
+            it('Cards with different smuggle aspects than play aspects only care about the smuggle aspects', function () {
```
AMMayberry1, line 63: Oh good test, didn't think of this right away

server/game/actions/PlayUpgradeAction.ts
```diff
@@ -29,12 +29,18 @@ export class PlayUpgradeAction extends PlayCardAction {
...
-                .attachUpgrade({ upgrade: context.source, takeControl: context.source.controller !== context.player })
-                .generateEvent(context.target, context),
-            cardPlayedEvent
-        ], this.resolveTriggersAfter);
+        const events = [context.game.actions
+            .attachUpgrade({ upgrade: context.source, takeControl: context.source.controller !== context.player })
+            .generateEvent(context.target, context),
+        cardPlayedEvent];
+
+        if (this.playType === PlayType.Smuggle) {
```
AMMayberry1, line 37: Will discuss in chat, but this code should be moved into some kind of base class instead of being repeated in three places

server/game/core/ability/KeywordHelpers.ts
```diff
@@ -102,9 +105,39 @@ function parseNumericKeywordValueIfEnabled(keyword: KeywordName, cardText: strin
     return Number(match.value[1]);
 }
 
+/**
+ * Checks if the Smuggle keyword is enabled and returns
+ *
+ * @returns null if the keyword is not enabled, or the numeric value if enabled
+ */
+function parseSmuggleIfEnabled(keyword: KeywordName, cardText: string, cardName: string): KeywordWithCostValues {
+    Contract.assertTrue(KeywordName.Smuggle === keyword);
```
Ennis0, line 114: Why have a function argument if it must always have the same specific value?

server/game/core/Player.js
```diff
@@ -875,9 +898,20 @@ class Player extends GameObject {
...
      */
-    exhaustResources(count) {
+    exhaustResources(count, priorityResources = []) {
+        let priorityResourcesExhausted = 0;
         let readyResources = this.resources.filter((card) => !card.exhausted);
-        for (let i = 0; i < Math.min(count, readyResources.length); i++) {
+        if (priorityResources.length > 0) {
+
+        }
+        priorityResources.forEach((priority) => {
```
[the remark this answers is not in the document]
AMMayberry1: Had this in the back of my mind because I had a nagging feeling there was something even simpler - realized you could just make an array of all ready resources with the priority ones first, then just do `.slice(0, count).forEach((resource) => resource.exhaust())`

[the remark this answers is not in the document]
adbronico: @AMMayberry1 How would I get the priority ones into the beginning of the array?

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -159,6 +160,13 @@ export function putIntoPlay<TContext extends AbilityContext = AbilityContext>(pr
 export function ready<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IReadySystemProperties, TContext> = {}): GameSystem<TContext> {
     return new ReadySystem<TContext>(propertyFactory);
 }
+
+/**
+ * default changePlayers = false
+ */
+export function resourceCard<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IResourceCardProperties, TContext> = {}): CardTargetSystem<TContext> {
```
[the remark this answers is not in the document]
adbronico: I think I'm OK with it given that there is nothing called a "resource card" in the game and that it's listed as a game system

server/game/core/Player.js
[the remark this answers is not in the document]
AMMayberry1: Just use filters to generate a list of ready priority resources and the concatenate the list of ready non-priority resources onto it. Or, you could use a filter to generate the list of all ready resources of either type and then sort by priority.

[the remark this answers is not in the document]
adbronico: Right, but to get the non-priority list, wouldn't I need to filter them or loop anyway?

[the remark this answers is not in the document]
AMMayberry1: Ah right I keep forgetting it's a list and not a property of the cards themselves. It would still be doable but not worth the effort imo, probably just easiest to go with something like the code example above

server/game/gameSystems/ResourceCardSystem.ts
```diff
@@ -0,0 +1,73 @@
...
+        changePlayer: false,
+        readyResource: false
+    };
+
+    public eventHandler(event: any, additionalProperties = {}): void {
+        const context = event.context;
+        // TODO: remove this completely if determinmed we don't need card snapshots
+        // event.cardStateWhenMoved = card.createSnapshot();
+        const properties = this.generatePropertiesFromContext(context, additionalProperties) as IResourceCardProperties;
+        //TODO: Is there a better/cleaner way to handle one or multiple cards here?
```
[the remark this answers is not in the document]
adbronico: I don't think so

server/game/gameSystems/ResourceCardSystem.ts
```diff
@@ -0,0 +1,73 @@
...
+        const destinationController = Array.isArray(properties.target)
+            ? properties.changePlayer
+                ? properties.target[0].controller.opponent
+                : properties.target[0].controller
+            : properties.changePlayer
+                ? properties.target.controller.opponent
+                : properties.target.controller;
+        return [
+            'move {0} to {1}\'s resources',
+            [properties.target, destinationController]
```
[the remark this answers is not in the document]
adbronico: I'm thinking for now we make this only support one card, until it needs to support more?

server/game/core/ability/PlayCardAction.ts
```diff
@@ -62,4 +77,14 @@ export abstract class PlayCardAction extends PlayerAction {
         const resourceCost = this.cost.find((cost) => cost.getAdjustedCost);
         return resourceCost ? resourceCost.getAdjustedCost(context) : 0;
     }
+
+    public handleSmuggle(context: PlayCardContext) {
+        if (this.playType === PlayType.Smuggle) {
+            context.game.openEventWindow([
```
AMMayberry1, line 83: Sorry, I don't know how I missed this before - since the new resource is considered to enter play simultaneously with the old one leaving per the rules, this event needs to go into the same window as the event that moves the smuggle card into play (or discard for events).

That should be pretty straightforward, just convert this to be like `generateSmuggleResourceEvent` or something and have it return the event, which you then insert into the window with the event that moves the card. Similar to how the putIntoPlay and cardPlayedEvent are emitted together for units, just insert the smuggle event into that same window with them.

For units and upgrades it's very obvious where the window that moves the card is being created since it's in the ctor, for event cards it happens here: https://github.com/SWU-Karabast/forceteki/blob/9c391fe88a42113edcc91a31af2bb15fbc0ffb4d/server/game/core/gameSteps/AbilityResolver.js#L259

Lmk if that needs additonal explanation, I think the overall code change should be relatively small.

server/game/core/Player.js
```diff
@@ -636,16 +637,23 @@ class Player extends GameObject {
...
+                cost = card.cost;
+                break;
+            case PlayType.Smuggle:
+                const smuggleInstance = card.getKeywordWithCostValues(KeywordName.Smuggle);
                 aspects = smuggleInstance.aspects;
                 cost = smuggleInstance.cost;
-            }
+                break;
+            default:
+                Contract.fail('Invalid Play Type');
```
AMMayberry1, line 654: Ideally would be \`Invalid Play Type '${playingType}'\` so we get additional info in the log if we ever need to debug this

server/game/core/Player.js
```diff
@@ -919,21 +927,29 @@ class Player extends GameObject {
      */
     // TODO: Create an ExhaustOrReadyResourcesSystem
     exhaustResources(count, priorityResources = []) {
-        let priorityResourcesExhausted = 0;
-        let readyResources = this.resources.filter((card) => !card.exhausted);
-        if (priorityResources.length > 0) {
+        Contract.assertTrue(this.countSpendableResources() >= count,
```
AMMayberry1, line 930: @Ennis0 was also working on this method and convinced me that this isn't necessary, we should be allowed to try and exhaust / ready more resources than available and it will just do as many as it can. What would your vote be?

server/game/gameSystems/ResourceCardSystem.ts
```diff
@@ -28,15 +29,21 @@ export class ResourceCardSystem<TContext extends AbilityContext = AbilityContext
...
-        const cards = [].concat(properties.target);
-        cards.forEach((card) => {
-            const player = properties.changePlayer && card.controller.opponent ? card.controller.opponent : card.controller;
-            player.moveCard(card, Location.Resource);
-            if (properties.readyResource) {
-                context.game.openEventWindow(ready({ target: card }).generateEvent(context.source, context));
-            }
-        });
+        // TODO: If we ever need to resource multiple cards at once, this will need an update
+        const card = Array.isArray(properties.target) ? properties.target[0] as Card : properties.target as Card;
```
AMMayberry1, line 33: Can we override `generatePropertiesFromContext` and add a contract check there that `properties.target.length === 1` if it is an array, that way if someone tries it without realizing they'll get notified and also we don't have to put that check everywhere in here this gets used.

server/game/gameSystems/ResourceCardSystem.ts
```diff
@@ -28,15 +29,21 @@ export class ResourceCardSystem<TContext extends AbilityContext = AbilityContext
...
+
+        const player = properties.targetPlayer === RelativePlayer.Opponent ? card.controller.opponent : card.controller;
```
AMMayberry1, line 35: Not for this PR, but this kind of makes me wonder if there needs to a be `WildcardRelativePlayer` type that just has `WildcardRelativePlayer.Any`. Because `RelativePlayer.Any` is technically a legal value here but it makes no sense for it to be that.

For now would be great to add a contract check in the overridden `generatePropertiesFromContext` for that and a TODO to investigate further.

server/game/core/ability/PlayCardAction.ts
```diff
@@ -1,16 +1,36 @@
...
+    /**
+     * Returns the appropriate title for the action based on the PlayType
+     * @param title The play title
+     * @param playType The play type
+     * @returns The play title for the action
+     */
+    private static getTitle(title: string, playType: PlayType): string {
+        switch (playType) {
+            case PlayType.Smuggle:
+                return title + 'with Smuggle';
```
AMMayberry1, line 30: Missing a space I think, right now it would be `Play this eventwith Smuggle`

server/game/actions/PlayUpgradeAction.ts
```diff
@@ -2,18 +2,18 @@ import { AbilityContext } from '../core/ability/AbilityContext';
...
 export class PlayUpgradeAction extends PlayCardAction {
     // we pass in a targetResolver holding the attachUpgrade system so that the action will be blocked if there are no valid targets
-    public constructor(card: Card) {
-        super(card, 'Play this upgrade', [], { immediateEffect: attachUpgrade<AbilityContext<UpgradeCard>>((context) => ({
-            upgrade: context.source
-        })) });
+    public constructor(card: Card, playType: PlayType = PlayType.PlayFromHand) {
+        super(card, 'Play this upgrade', playType, [],
+            { immediateEffect: attachUpgrade<AbilityContext<UpgradeCard>>((context) => ({
+                upgrade: context.source
```
AMMayberry1, line 15: nit: I would vote just have the immediateEffect all be one line, but nbd if you prefer it this way

server/game/cards/01_SOR/units/SuperlaserTechnician.ts
```diff
@@ -14,11 +13,7 @@ export default class SuperlaserTechnician extends NonLeaderUnitCard {
         this.addWhenDefeatedAbility({
             title: 'Put Superlaser Technician into play as a resource and ready it',
             optional: true,
-            // TODO: figure out how to change this from sequential to simultaneous while keeing the ready effect
-            immediateEffect: AbilityHelper.immediateEffects.sequential([
-                // TODO: create a MoveCardToResourceSystem and accompanying function in immediateEffects
-                AbilityHelper.immediateEffects.moveCard((context) => ({ target: context.source, destination: Location.Resource })),
-                AbilityHelper.immediateEffects.ready((context) => ({ target: context.source }))])
+            immediateEffect: AbilityHelper.immediateEffects.resourceCard({ target: this, readyResource: true })
```
AMMayberry1, line 16: `AbilityHelper.immediateEffects.resourceCard((context) => ({ target: context.source, readyResource: true })))`

server/game/cards/02_SHD/upgrades/HotshotDL44Blaster.ts
```diff
@@ -0,0 +1,38 @@
...
+export default class HotshotDL44Blaster extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '5874342508',
+            internalName: 'hotshot-dl44-blaster',
+        };
+    }
+
+    public override canAttach(targetCard: Card, controller: Player = this.controller): boolean {
+        if (targetCard.hasSomeTrait(Trait.Vehicle)) {
```
AMMayberry1, line 16: I added a one-liner way of doing this since it comes up a lot, take a look at Jedi Lightsaber impl

server/game/core/Constants.ts
```diff
@@ -222,7 +222,6 @@ export enum KeywordName {
     Saboteur = 'saboteur',
     Sentinel = 'sentinel',
     Shielded = 'shielded',
-    /** @deprecated Not implemented yet */
```
AMMayberry1, line 225: 🎉🎉🎉

server/game/core/ability/KeywordHelpers.ts
```diff
@@ -102,9 +105,37 @@ function parseNumericKeywordValueIfEnabled(keyword: KeywordName, cardText: strin
...
+        return null;
+    }
+
+    if (matchIter.next().done !== true) {
+        throw new Error(`Expected to match at most one instance of enabled keyword ${KeywordName.Smuggle} in card ${cardName}, but found multiple`);
+    }
+
+    const smuggleCost = Number(match.value[1]);
+    const aspectString = match.value[2];
+    const smuggleAspects = EnumHelpers.checkConvertToEnum(aspectString.toLowerCase().split(' '), Aspect);
```
AMMayberry1, line 128: I keep forgetting to ask - can you send me two examples of cards from the swu site that have mismatched smuggle text so I can report it to them, we've had a couple glitches like this come up now and it's probably worth at least bringing it up

server/game/core/ability/PlayCardAction.ts
```diff
@@ -1,16 +1,36 @@
...
+        super(card, PlayCardAction.getTitle(title, playType), additionalCosts.concat(CostLibrary.payPlayCardResourceCost(playType)), targetResolver);
+
+        this.playType = playType;
+    }
+
+    /**
+     * Returns the appropriate title for the action based on the PlayType
+     * @param title The play title
+     * @param playType The play type
+     * @returns The play title for the action
```
AMMayberry1, line 25: Not a change request, just a note that for really simple methods like this I'm fine omitting the detailed param / return descriptors. Totally fine if you prefer having them though, just a heads-up to potentially save time in the future

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -41,7 +40,9 @@ export class PlayableOrDeployableCard extends Card {
     }
 
     public override getActions(): PlayerOrCardAbility[] {
-        return this.defaultActions.concat(super.getActions());
+        const actions = this.defaultActions.concat(super.getActions());
+
+        return actions;
```
AMMayberry1, line 45: Why split into two lines?

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -33,34 +33,37 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
         // event.cardStateWhenMoved = card.createSnapshot();
         const properties = this.generatePropertiesFromContext(context, additionalProperties) as IMoveCardProperties;
-        if (properties.switch && properties.switchTarget) {
-            const otherCard = properties.switchTarget;
-            card.owner.moveCard(otherCard, card.location);
-        }
-        const player = properties.changePlayer && card.controller.opponent ? card.controller.opponent : card.controller;
-        player.moveCard(card, properties.destination, { bottom: !!properties.bottom });
+        // TODO: Is there a better/cleaner way to handle one or multiple cards here?
+        const cards = [].concat(properties.target);
```
AMMayberry1, line 40: This is a clever one-liner for handling this situation regardless of whether target is an array or not, I like it

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -33,34 +33,37 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
+        cards.forEach((card) => {
```
AMMayberry1, line 41: Do you mind changing this into a proper `for (const card of cards)` loop? Only reason is that when debugging it will completely step over this entire code block if it's in an `<array>.forEach()` call, which is super annoying

server/game/gameSystems/ResourceCardSystem.ts
```diff
@@ -0,0 +1,76 @@
...
+        // TODO: remove this completely if determinmed we don't need card snapshots
+        // event.cardStateWhenMoved = card.createSnapshot();
+        const properties = this.generatePropertiesFromContext(context, additionalProperties) as IResourceCardProperties;
+        // TODO: If we ever need to resource multiple cards at once, this will need an update
+        const card = Array.isArray(properties.target) ? properties.target[0] as Card : properties.target as Card;
+
+        const player = properties.targetPlayer === RelativePlayer.Opponent ? card.controller.opponent : card.controller;
+        player.moveCard(card, Location.Resource);
+        if (properties.readyResource) {
+            context.game.openEventWindow(ready({ target: card }).generateEvent(context.source, context));
```
AMMayberry1, line 38: This line it shouldn't be here - what this line currently does is interrupt the flow of resourcing the card to ready the card _in its pre-move location_, since the move hasn't resolved yet. The contingent event code below should be correct

server/game/gameSystems/ResourceCardSystem.ts
```diff
@@ -0,0 +1,76 @@
...
+
+            event.setContingentEventsGenerator((event) => {
+                const readyResourceEvent = ready({ target: card }).generateEvent(context.source, context);
+                const contingentEvents = [readyResourceEvent];
+
+                return contingentEvents;
```
AMMayberry1, line 44: Could just do `return [readyResourceEvent];`

test/server/core/abilities/keyword/Smuggle.spec.js
```diff
@@ -0,0 +1,72 @@
...
+                expect(this.armedToTheTeeth).toBeInLocation('ground arena');
+                expect(this.wampa.upgrades).toContain(this.armedToTheTeeth);
+
+                // This costs 6 due to the lack of a red aspect on base or leader
+                expect(this.player1.countExhaustedResources()).toBe(6);
+                expect(this.player1.countSpendableResources()).toBe(12);
+                expect(this.mercenaryGunship).toBeInLocation('resource');
+            });
+
+            it('an event can be played for its smuggle cost', function () {
```
AMMayberry1, line 48: Would you mind changing this test to be covert strength instead of smuggler's aid? Reason being that it would be nice to do a check at the targeting stage that the swap has already resolved (event in discard, top card in resource line)

server/game/core/Player.js
adbronico: @Ennis0 @AMMayberry1 Would an example be like if you attack with Boba but can only ready 1 resource?

server/game/core/ability/PlayCardAction.ts
```diff
@@ -1,16 +1,36 @@
...
+     */
+    private static getTitle(title: string, playType: PlayType): string {
+        switch (playType) {
+            case PlayType.Smuggle:
+                return title + 'with Smuggle';
+            default:
+                return title;
```
AMMayberry1, line 32: Can we explicitly cover both play types and have a Contract.fail as the default

server/game/core/Player.js
AMMayberry1: Yeah correct. It's just a design philosophy question, it's more flexible if we let the caller ask for more resources than are available but could allow bugs to sneak through. Forcing the caller to check how many are available first is more work but might catch some issues occasionally. It's a small enough thing that I'm good either way.

server/game/cards/01_SOR/units/SuperlaserTechnician.ts
adbronico: This one pains me. Saying this there is so much clearer. Otherwise we're setting target to context.source, which is super nebulous

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
adbronico: There was something in the middle originally. I've made it more concise

server/game/gameSystems/MoveCardSystem.ts
adbronico: I have no memory of doing this

server/game/gameSystems/ResourceCardSystem.ts
adbronico: I noticed that this doesn't seem to actually be working

server/game/core/Player.js
adbronico: @AMMayberry1 I think it depends to me whether the ability is something we can choose. Like if we are not giving them the option of when to choose Boba's leader-side ability, then we should make sure it uses as much, but if we let them choose, we could probably get rid of this check? Does that make sense?

AMMayberry1 (approved):

PR #65, opened 2024-09-24
Refactor to support better type inference

AMMayberry1:
Major refactor to add much better support for type inference in common scenarios, as well as an improvement to error handling.

**Error handling and type assertions**
We decided that it is okay for `Contract` methods to throw an error even in prod, so we are able to refactor to use those as one-liners instead of needing to have extra logic that checks the result and returns something. In addition, we can now use assertion annotations for type narrowing. The end result is that what used to look like this

```typescript
  public override executeHandler(context: PlayCardContext): void {
      if (!Contract.assertTrue(context.source.isEvent())) {
          return;
      }

      context.game.addMessage(
          '{0} plays {1}',
          context.player,
          context.source,
      );
      context.game.resolveAbility((context.source as EventCard).getEventAbility().createContext());
  }
```

Now looks like this:

```typescript
  public override executeHandler(context: PlayCardContext): void {
      // after this check, context.source is inferred to be of type EventCard
      Contract.assertTrue(context.source.isEvent());

      context.game.addMessage(
          '{0} plays {1}',
          context.player,
          context.source,
      );

      // cast to EventCard no longer needed
      context.game.resolveAbility(context.source.getEventAbility().createContext());
  }
```

**Type inference for `context.source`**
A frequent frustration is that `context.source` appears as the `any` type or as a generic `Card`, which creates issues when trying to use properties that are only available in a subclass (e.g., damage only exists on `UnitCard`).

Added a generic parameter to GameSystem so that the AbilityContext it accepts can be specified. The end result is that `context.source` will always be of type `this`, so that the properties and methods of the current card can be accessed without a cast. See the Shield ability as an example:

```typescript
// shield ability before
this.addReplacementEffectAbility({
    title: 'Defeat shield to prevent attached unit from taking damage',
    when: {
        onDamageDealt: (event, context) => event.card === (context.source as UpgradeCard).parentCard
    },
    replaceWith: {
        target: this,
        replacementImmediateEffect: AbilityHelper.immediateEffects.defeat()
    },
    effect: 'shield prevents {1} from taking damage',
    effectArgs: (context) => [(context.source as UpgradeCard).parentCard],
});
```

```typescript
// shield ability after - no casts needed
this.addReplacementEffectAbility({
    title: 'Defeat shield to prevent attached unit from taking damage',
    when: {
        onDamageDealt: (event, context) => event.card === context.source.parentCard
    },
    replaceWith: {
        target: this,
        replacementImmediateEffect: AbilityHelper.immediateEffects.defeat()
    },
    effect: 'shield prevents {1} from taking damage',
    effectArgs: (context) => [context.source.parentCard],
});
```

Also the Vambrace Grappleshot ability, which can directly reference the `activeAttack` property now:

```typescript
  this.addGainTriggeredAbilityTargetingAttached({
      title: 'Exhaust the defender on attack',
      when: { onAttackDeclared: (event, context) => event.attack.attacker === context.source },
      immediateEffect: AbilityHelper.immediateEffects.exhaust((context) => {
          return { target: context.source.activeAttack?.target };    // no cast needed
      })
  });
```

Also, because `context.source` now has type data, there will be auto-complete for its properties.

### DIFF
--- a/server/game/Interfaces.ts
+++ b/server/game/Interfaces.ts
@@ -15,12 +15,12 @@ import { IInitiateAttackProperties } from './gameSystems/InitiateAttackSystem';
 // ********************************************** EXPORTED TYPES **********************************************
 
 /** Interface definition for addTriggeredAbility */
-export type ITriggeredAbilityProps = ITriggeredAbilityWhenProps | ITriggeredAbilityAggregateWhenProps;
-export type IReplacementEffectAbilityProps = IReplacementEffectAbilityWhenProps | IReplacementEffectAbilityAggregateWhenProps;
+export type ITriggeredAbilityProps<TSource extends Card = Card> = ITriggeredAbilityWhenProps<TSource> | ITriggeredAbilityAggregateWhenProps<TSource>;
+export type IReplacementEffectAbilityProps<TSource extends Card = Card> = IReplacementEffectAbilityWhenProps<TSource> | IReplacementEffectAbilityAggregateWhenProps<TSource>;
 
 /** Interface definition for addActionAbility */
-export type IActionAbilityProps<Source = any> = Exclude<IAbilityPropsWithSystems<AbilityContext<Source>>, 'optional'> & {
-    condition?: (context?: AbilityContext<Source>) => boolean;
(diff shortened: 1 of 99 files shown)


adbronico (approved):
Would it help if I got out and pushed?

PR #66, opened 2024-09-25
Test improvements

AMMayberry1:
Added a section to the test cheat sheet on different test actions that are available. Changed unit property name generation so that a test property is generated for both title and title + subtitle. Cleaned up some expectations a bit.


adbronico (approved):
Had a slight weapons malfunction but, uh, everything's perfectly alright now.

PR #67, opened 2024-09-25
Re-enable parallel test run

AMMayberry1:
The parallel test setting was having issue in the past so we disabled it in github actions and recommended people not use it. It currently seems that the problem is gone, so re-enabling everything for now and will monitor to see if any issues occur.


adbronico (approved):
Somehow... Palpatine returned

PR #68, opened 2024-09-25
Improve some card properties and fix ActiveAttack issue

AMMayberry1:
Improved some of the common card properties to make them safer / easier to use:

- Added a `remainingHp` property that returns `max(0, hp - damage)`
- Renamed the `hp` and `power` properties to `getHp()` and `getPower()` to make it clearer that they involve some complex checking logic
- Added a guard on the `.upgrades` property so that it can't be used when the card is not in an arena (same as for damage)
- Added a similar guard on the `.activeAttack` property

As a result of the last change above, found an issue where units that were defeated in combat would still have an "active" attack after moving to discard. This does not currently cause any problems but is not in line with the rules as the unit is no longer involved in an attack once defeated. It could also cause confusion in the future. Therefore, added additional logic to clean up the property immediately when the unit leaves the play area for any reason.


adbronico (approved):
We're fine. We're all fine here now. Thank you. How are you?

PR #69, opened 2024-09-27
Add Ambush and Overwhelm

AMMayberry1:
Implemented Ambush and Overwhelm as well as a few test units.

"Gain Overwhelm" works and is tested. "Gain Ambush" requires some tweaking to the play unit sequencing and will be enabled in a future PR.


server/game/abilities/keyword/AmbushAbility.ts
```diff
@@ -1,36 +1,54 @@
...
+    public static buildAmbushAbilityProperties<TSource extends Card = Card>(): ITriggeredAbilityProps<TSource> {
+        return {
+            title: 'Ambush',
+            optional: true,
+            when: { onUnitEntersPlay: (event, context) => event.card === context.source },
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: AmbushAbility.unitWouldHaveAmbushTarget<TSource>,
+                onTrue: AbilityHelper.immediateEffects.sequential([
+                    AbilityHelper.immediateEffects.ready(),
+                    AbilityHelper.immediateEffects.attack((context) => {
```
Ennis0, line 24: Is there a way this attack could be tagged as being due to ambush? Would be useful for set 2 Enfys Nest.

test/server/core/abilities/keyword/Ambush.spec.js
```diff
@@ -1,34 +1,82 @@
...
+                        hand: ['syndicate-lackeys'],
+                        groundArena: ['wampa']
+                    },
+                    player2: {
+                        spaceArena: ['cartel-spacer']
+                    }
+                });
+            });
+
+            it('enters play and there is only no target, the Ambush prompt will not happen', function () {
```
Ennis0, line 76: "only" is ungrammatical here

test/server/core/abilities/keyword/Ambush.spec.js
```diff
@@ -1,34 +1,82 @@
...
+                this.player1.clickCard(this.syndicateLackeys);
+                expect(this.player2).toBeActivePlayer();
```
Ennis0, line 78: Could make sure the ambush unit isn't readying.

server/game/abilities/keyword/AmbushAbility.ts
```diff
@@ -1,36 +1,54 @@
...
+                onFalse: AbilityHelper.immediateEffects.noAction()
+            })
+        };
+    }
 
-//     public constructor(game: Game, card: Card) {
-//         if (!Contract.assertTrue(card.isUnit())) {
-//             return;
-//         }
+    private static unitWouldHaveAmbushTarget<TSource extends Card = Card>(context: TriggeredAbilityContext<TSource>): boolean {
```
adbronico, line 36: Is this the best way to do this? I haven't seen us set up an ability like this before

server/game/cards/02_SHD/units/ClanChallengers.ts
```diff
@@ -0,0 +1,22 @@
...
+        return {
+            id: '3487311898',
+            internalName: 'clan-challengers',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'While this unit is upgraded, it gains Overwhelm',
+            condition: (context) => context.source.upgraded,
```
adbronico, line 16: Can we update "upgraded" to be a function (i.e. isUpgraded() or hasUpgrades() or something)?

server/game/core/utils/Contract.ts
```diff
@@ -55,6 +55,12 @@ export function assertNotNullLike<T>(val: T, message?: string): asserts val is N
     }
 }
 
+export function assertIsNullLike<T>(val: T, message?: string): asserts val is null | undefined {
+    if (val != null) {
```
adbronico, line 59: This check catches both null and undefined?

test/server/core/abilities/keyword/Overwhelm.spec.js
```diff
@@ -0,0 +1,91 @@
...
+            });
+
+            it('and the unit is defeated before damage is resolved, all damage goes to base', function () {
+                this.player1.clickCard(this.emperorPalpatine);
+                this.player1.clickCard(this.deathStarStormtrooper);
+                expect(this.p2Base.damage).toBe(9);
+            });
+        });
+    });
+});
```
adbronico, line 91: Where do we want to handle the interaction between Overwhelm and the Blizzard Assault AT-AT?

adbronico (commented):
No blocking changes but a few notes

server/game/core/utils/Contract.ts
AMMayberry1: Yeah, this is like the one good use case for == instead of ===

test/server/core/abilities/keyword/Overwhelm.spec.js
AMMayberry1: Unfortunately we never got a final answer from the judge chat on how that works. My plan is to just hold off on implementing it until the set 3 rules update comes out and hope it clarifies things, then go from there

server/game/abilities/keyword/AmbushAbility.ts
AMMayberry1: To do what specifically? Can you elaborate on what the concern is about

AMMayberry1: Two things:

1. It's not just checking for the presence of an enemy unit, per the specific rule it's checking if there is a unit that this unit can attack. So that means checking for cases like e.g. if there's a Sabine unit alone in the ground arena but her ability is live b/c of space arena units, the opponent can't do a ground ambush b/c there's no attackable ground unit.

Those rules get pretty involved so if we tried to rewrite them here we'd just be rewriting or separating out all of the checks in the attack system into a helper. The latter would be the more okay of the two but even then, the helper would need to have passed in all of the same data that goes into making an AttackSystem so it kind of gets to the point of asking what would be better about it. The biggest advantage would be we're not creating a transient object, but idk if that's worth the trade.

2. We do quietly use this pattern in a places already. If you look at the PlayUpgrade action, you can see it's including an `attachUpgrade` system call that it passes to super() as a target resolver so the ability knows which targets are legal to play the upgrade on: https://github.com/AMMayberry1/forceteki/blob/f34c6f076e4d3bb09285aacab91b88ff9cfaa2ea/server/game/actions/PlayUpgradeAction.ts#L15

So overall there's probably room to optimize a little to make this cleaner but I think doing that in a way that doesn't just shove the complexity somewhere else would take a lot of work, which I have trouble justifying given how infrequently we need to do this.

server/game/abilities/keyword/AmbushAbility.ts
AMMayberry1: I think that's a good idea but for a simple feature like that I like waiting until that card that needs it is being implemented so we can write and test it in the same PR. I'll add a TODO here in the meantime

PR #71, opened 2024-09-28
Implemented Relentless, Konstantine's Folly

Ennis0:


server/game/ongoingEffects/OngoingEffectLibrary.ts
```diff
@@ -46,7 +46,7 @@ export = {
     //     OngoingEffectBuilder.card.static(EffectName.AttachmentRestrictTraitAmount, object),
     // attachmentTraitRestriction: (traits) => OngoingEffectBuilder.card.static(EffectName.AttachmentTraitRestriction, traits),
     // attachmentUniqueRestriction: () => OngoingEffectBuilder.card.static(EffectName.AttachmentUniqueRestriction),
-    // blank: (blankTraits = false) => OngoingEffectBuilder.card.static(EffectName.Blank, blankTraits),
+    blank: (blankTraits = false) => OngoingEffectBuilder.card.static(EffectName.Blank, blankTraits),
```
AMMayberry1, line 49: 1. Can you remove "blankTraits" for now, there's nothing in SWU that erases traits currently so it just adds confusion imo. Will be easy to add back in later if we need it.

2. Blanking for units will need to have some different behaviors to handle the possibility of being blanked and then gaining new abilities, so we need to separate it out from blanking events. Could we name this "blankEventCard" or something

server/game/cards/01_SOR/units/RelentlessKonstantinesFolly.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar): void {
+        this.cardsPlayedThisPhaseWatcher = AbilityHelper.stateWatchers.cardsPlayedThisPhase(registrar, this);
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'The first event played by each opponent each round loses all abilities',
+            ongoingEffect: AbilityHelper.ongoingEffects.blank(false),
+            //An event card goes to the discard before its effects resolve, so filtering for being in the discard is appropriate to allow playing an event but blanking its effects.
```
AMMayberry1, line 25: I was confused why this is the case, since the "play event" ability should not be getting blanked by the change you made. I investigated and there's a bug related to some old blanking code ported from l5r. Take a look at this commit with the bug fix, it passes your tests: https://github.com/AMMayberry1/forceteki/commit/fbcad5f10ea8a4043d17a36d8755c5fe1dddb05a

I would prefer we use that version since it's clearer and simpler (and it fixes the bug).

server/game/cards/01_SOR/units/RelentlessKonstantinesFolly.ts
```diff
@@ -0,0 +1,42 @@
...
+            //Filtering for anywhere(WildcardLocation.Any), causes the event card to lose the ability to even play it in the first place.
```
AMMayberry1, line 26: nit: can you add a space after the `//` for these lines for readability. again, the linter should be catching this, I need to investigate

test/server/cards/01_SOR/units/RelentlessKonstantinesFolly.spec.js
```diff
@@ -0,0 +1,72 @@
...
+                //next round, it should nullify the first event played again
+                this.moveToNextActionPhase();
+                this.player1.pass();
+                exhaustedResourcesBeforeCardPlay = this.player2.countExhaustedResources();
+                this.player2.clickCard(this.momentOfPeace);
+                expect(this.player2.countExhaustedResources()).toBe(exhaustedResourcesBeforeCardPlay + 1);
+                expect(this.relentless).toHaveExactUpgradeNames([]);
+            });
+
+            it('should not nullify a second or later event even if Relentless was played after the first event', function () {
```
AMMayberry1, line 45: Great work on the tests! This covers every case I could think of.

server/game/cards/01_SOR/units/RelentlessKonstantinesFolly.ts
Ennis0: Sure. I've switched it over.

AMMayberry1 (approved):

PR #73, opened 2024-09-28
Add Snoke and logic for defeat due to HP effects

AMMayberry1:
Added the engine infrastructure to handle defeat due to HP modifying effects.

Add the concept of "subwindows" to EventWindow. A subwindow is a new window that resolves immediately after the current window's events resolve, before triggered abilities resolve. This is needed to account for the timing of defeat due to combat, damage, etc.

Added Snoke and several tests that his ability works and that relevant triggered abilities happen in the right time window. There are more tests that need to be added for more advanced interactions involving multiple layers of resolving constant abilities / nested triggered abilities, but these will be reserved for a dedicated future PR.

Also added Dodonna and some basic tests that his ability interacts correctly with defeat. More tests are needed but are blocked by implementing units that would be needed for the test.

Added a few additional units for testing purposes, keeping it to the simplest possible ones. More complicated cases reserved for the future timing windows test PR.


server/game/cards/01_SOR/upgrades/JediLightsaber.ts
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        return super.canAttach(targetCard, controller);
+    }
+
+    public override setupCardAbilities() {
+        this.addGainTriggeredAbilityTargetingAttached({ title: 'Give the defender -2/-2 for this phase',
+            when: { onAttackDeclared: (event, context) => event.attack.attacker === context.source },
+
+            // need to check if the target is a base - if so, don't apply the stat modifier effect
```
Ennis0, line 28: Can we mess with canAffect() or maybe StaticOngoingEffect.canBeApplied() so that the effect just isn't applied if it's invalid for the target? I wanted basically that(except for targeting) for an ability that gave sentinel, but couldn't figure it out and made it a TODO. What do you think?

server/game/core/Game.js
```diff
@@ -868,6 +868,17 @@ class Game extends EventEmitter {
         return this.queueStep(new EventWindow(this, events, ownsTriggerWindow));
     }
 
+    /**
+     * Creates a "sub-window" for events which will have priority resolution and
```
Ennis0, line 872: Do you want this to coexist with contingent events? What is the difference?

server/game/core/event/EventWindow.js
```diff
@@ -63,8 +68,18 @@ class EventWindow extends BaseStepWithPipeline {
...
+    addSubwindowEvents(events) {
+        this.subwindowEvents = this.subwindowEvents.concat(events);
+    }
+
     setCurrentEventWindow() {
         this.previousEventWindow = this.game.currentEventWindow;
+        this.windowDepth = this.previousEventWindow ? this.previousEventWindow.windowDepth + 1 : 0;
+
+        if (this.windowDepth >= 50) {
+            throw new Error('Event window depth has reached 50, likely caught in an infinite loop');
```
Ennis0, line 80: Smart!

AMMayberry1: Thanks 😄

server/game/cards/01_SOR/upgrades/JediLightsaber.ts
AMMayberry1: Yeah, generally speaking it would be useful if an effect could declare default target conditions the way a GameSystem does.

It would be a pretty big change to the way effects are set up though, and the whole effect system is badly in need of an overhaul as is. So I'm kind of inclined to postpone any big improvements on the system until after launch, unless it starts to become a really big issue

AMMayberry1: Actually, I'll poke around at it and evaluate how bad it would actually be. There's definitely stuff that would be easier if it was changed

adbronico (approved):

server/game/cards/01_SOR/upgrades/JediLightsaber.ts
```diff
@@ -0,0 +1,41 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => context.event.attack.target.isBase(),
+                onTrue: AbilityHelper.immediateEffects.noAction(),
+                onFalse: AbilityHelper.immediateEffects.forThisPhaseCardEffect((context) => ({
+                    effect: AbilityHelper.ongoingEffects.modifyStats({ power: -2, hp: -2 }),
+                    target: (context.event.attack as Attack).target
+                }))
+            }) },
+        (context) => context.source.parentCard?.hasSomeTrait(Trait.Force));
```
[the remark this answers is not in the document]
AMMayberry1: Good callout, I can refactor so that this is also a part of the properties object. There's a handful of similar updates needed to the upgrade system so I added a TODO and I'll include it in a future PR that hits all of that.

PR #75, opened 2024-10-01
Overhaul gain ability system for Krell and Heroic Resolve

AMMayberry1:
Encountered some bugs trying to implement Krell and realized that the whole gain ability system needed an overhaul to clean up some l5r behaviors and add the support for some gain ability edge cases - Krell and Heroic Resolve.

Had to make tweaks in a lot of places to get things to line up correctly, but both cards are now working as intended and gained abilities are in much better shape overall.


test/server/cards/02_SHD/upgrades/HeroicResolve.spec.js
```diff
@@ -0,0 +1,170 @@
...
+                expect(this.player1).not.toHavePassAbilityButton();
+                expect(this.player1).toBeAbleToSelectExactly([this.p2Base, this.specforceSoldier]);
+
+                this.player1.clickCard(this.specforceSoldier);
+                expect(this.frontierAtrt.isUpgraded()).toBe(false);
+                expect(this.p2HeroicResolve).toBeInLocation('discard');
+                expect(this.specforceSoldier).toBeInLocation('discard');
+                expect(this.p2Base.damage).toBe(8);
+                expect(this.player1.countExhaustedResources()).toBe(4);
+            });
```
Ennis0, line 167: Do you think we might want to make sure that the ability can't defeat a Heroic Resolve on another unit?

server/game/cards/01_SOR/units/GeneralKrellHeartlessTactician.ts
```diff
@@ -0,0 +1,28 @@
...
+            internalName: 'general-krell#heartless-tactician',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'Other friendly units gain when defeated ability',
+            targetController: RelativePlayer.Self,
+            matchTarget: (card, context) => card !== context.source,
+            ongoingEffect: AbilityHelper.ongoingEffects.gainAbility({
```
Ennis0, line 18: The draw should be optional.

PR #79, opened 2024-10-03
implement protector and infiltrator's skills

VJubert:


test/server/cards/02_SHD/upgrades/InfiltratorsSkill.spec.js
```diff
@@ -0,0 +1,22 @@
...
+                    player1: {
+                        groundArena: [{ card: 'battlefield-marine', upgrades: ['infiltrators-skill'] }],
+                    },
+                    player2: {
+                        groundArena: ['pyke-sentinel']
+                    }
+                });
+            });
+
+            it('should ignore sentinel', function () {
```
AMMayberry1, line 16: Can you also add a test for the defeat shield effect

test/server/cards/02_SHD/upgrades/InfiltratorsSkill.spec.js
```diff
@@ -0,0 +1,22 @@
...
+                this.player1.clickCard(this.battlefieldMarine);
+                expect(this.player1).toBeAbleToSelectExactly([this.pykeSentinel, this.p2Base]);
```
AMMayberry1, line 18: Can we confirm that it can actually resolve the attack against base (do the click and confirm that the damage happens)

test/server/cards/02_SHD/upgrades/Protector.spec.js
```diff
@@ -0,0 +1,23 @@
...
+                    },
+                    player2: {
+                        groundArena: [{ card: 'snowspeeder', upgrades: ['protector'] }]
+                    }
+                });
+            });
+
+            it('should be sentinel', function () {
+                this.player1.clickCard(this.battlefieldMarine);
+                //Snowspeeder automatically selected due to sentinel
```
AMMayberry1, line 18: We need to confirm that the attack actually resolves (confirm that the damage happened and / or cards are moved to discard), there are a lot of ways that the attack resolution could silently fail and move directly to player 2 being active

test/server/cards/02_SHD/upgrades/InfiltratorsSkill.spec.js
```diff
@@ -0,0 +1,34 @@
...
+                // ignore sentinel
+                this.player1.clickCard(this.p2Base);
+                expect(this.p2Base.damage).toBe(4);
+            });
```
AMMayberry1, line 22: nit: whitespace between test cases. I'll update the linter to catch this

test/server/cards/02_SHD/upgrades/InfiltratorsSkill.spec.js
```diff
@@ -0,0 +1,34 @@
...
+            it('should ignore shield', function () {
+                this.player1.clickCard(this.battlefieldMarine);
+                expect(this.player1).toBeAbleToSelectExactly([this.pykeSentinel, this.p2Base]);
+
+                // defeat shielded pyke sentinel
+                this.player1.clickCard(this.pykeSentinel);
+                expect(this.battlefieldMarine.damage).toBe(2);
+                expect(this.pykeSentinel).toBeInLocation('discard');
```
AMMayberry1, line 30: Can we change this test to use a unit that will survive the attack? We want the most precise test possible, since some effect could defeat the unit and shield together (though unlikely).

So the best test would be to have a bigger unit and see that (a) its shield is defeated and (b) it has taken damage from the attack, meaning that the shield was defeated before the attack damage happened.

test/server/cards/02_SHD/upgrades/InfiltratorsSkill.spec.js
```diff
@@ -0,0 +1,36 @@
...
+                // ignore sentinel
+                this.player1.clickCard(this.p2Base);
+                expect(this.p2Base.damage).toBe(4);
+            });
+
+            it('should ignore shield', function () {
+                this.player1.clickCard(this.battlefieldMarine);
+                expect(this.player1).toBeAbleToSelectExactly([this.niimaOutpostConstables, this.p2Base]);
+
+                // defeat shielded pyke sentinel
```
AMMayberry1, line 28: nit: comment is stale

AMMayberry1, line 28: nit: comment is stale

AMMayberry1 (approved):

PR #81, opened 2024-10-03
feat(card): Implement Recruit

joshfullmer:


test/server/cards/01_SOR/events/Recruit.spec.js
```diff
@@ -0,0 +1,114 @@
...
+                    expect(this.player1).toHaveEnabledPromptButtons([this.viperProbeDroid.title, 'Take nothing']);
+                    expect(this.player1).toHaveDisabledPromptButtons([this.confiscate.title, this.iAmYourFather.title]);
+                    this.player1.clickPrompt(this.viperProbeDroid.title);
+
+                    expect(this.player1.deck.length).toBe(2);
+                    expect([this.confiscate, this.iAmYourFather]).toAllBeInBottomOfDeck(this.player1, 2);
+                });
+
+
+                it('when the deck is empty', function() {
```
AMMayberry1, line 50: Small note on the naming - jasmine constructs tests by concatenating the hierarchical names together. So this test would end up being named `Recruit - integration - when there is one valid option, when the deck is empty`.

I would recommend just making this case its own describe block with its own setup just for clarity. Each individual case has to go through the whole game initialization process anyway, so there's no impact on performance or functionality.

AMMayberry1: Looking further, probably a lot of these names could be slightly adjusted to make sense when concatenated together

joshfullmer: Done in https://github.com/SWU-Karabast/forceteki/pull/81/commits/213827b45e2721ddcf2442a0abfd0774405a6ea6

AMMayberry1 (approved):

PR #83, opened 2024-10-03
Implementing Open Fire

wfowlks:


test/server/cards/01_SOR/events/OpenFire.spec.js
```diff
@@ -0,0 +1,56 @@
...
+
+            it('can damage a unit without a shield, dealing damage to the unit with health to spare', function () {
+                this.player1.clickCard(this.openFire);
+                expect(this.player1).toBeAbleToSelectExactly([this.wampa, this.cartelSpacer, this.pykeSentinel, this.fleetLieutenant]);
+
+                this.player1.clickCard(this.wampa);
+                expect(this.wampa.damage).toBe(4);
+            });
+
+            it('can damage a damaged unit and destroy it', function () {
```
AMMayberry1, line 37: In most contexts I'm all for super-thorough tests, in this case unfortunately we have a finite amount of Github Actions runner time and the jasmine tests are pretty heavy. So, we try to consolidate tests down as much as possible.

I think the first two tests have unique value and are good - we actually haven't written a test that checks shield blocking against direct damage effects yet, so the first one covers that case now.

The latter two tests I would prefer to cut, as both defeat from damage and stat bonuses due to upgrades are thoroughly tested elsewhere.

AMMayberry1:
Looks like the implementation file is in the wrong place, it should go under `SOR_01/events/`

AMMayberry1:
Ah apologies, I was going through and auto-updating PRs after the last merge and hit yours by mistake, looks like you're still doing changes.

Also please do reach out in the discord server so we can add you to the contributor channels (if you're not in there already)

test/server/cards/01_SOR/events/OpenFire.spec.js
wfowlks: That sounds reasonable :)

wfowlks:
Should be good now. Sorry the card was in the wrong place because I copied it over to the forked repo in the wrong location. Now i'm all set up to work off the forked branch

AMMayberry1 (approved):

PR #87, opened 2024-10-03
Foundling, Protector Of The Throne, Mandalorian Warrior

VJubert:


test/server/cards/02_SHD/upgrades/Foundling.spec.js
```diff
@@ -0,0 +1,32 @@
...
+                // check battlefield marine trait
+                expect(this.battlefieldMarine.hasSomeTrait('mandalorian')).toBeFalse();
+
+                // add battlefield marine foundling
+                this.player1.clickCard(this.foundling);
+                expect(this.player1).toBeAbleToSelectExactly([this.battlefieldMarine, this.snowspeeder, this.pykeSentinel]);
+                this.player1.clickCard(this.battlefieldMarine);
+
+                // check players units traits
+                expect(this.battlefieldMarine.hasSomeTrait('mandalorian')).toBeTrue();
```
AMMayberry1, line 27: This is a good start but we always want tests to confirm actual card behavior, which is sometimes tricky because it may involve multiple cards.

In this case the way we would want to confirm that Foundling is working correctly is to use another card that targets Mandalorian cards. Unfortunately, none are implemented yet. If you would be willing to add a simple one like Mandalorian Warrior or Mandalorian Armor in this PR and then use it in this test case, that'd be amazing

test/server/cards/02_SHD/upgrades/Foundling.spec.js
```diff
@@ -0,0 +1,32 @@
...
+                        hand: ['foundling'],
+                        groundArena: ['battlefield-marine', 'pyke-sentinel'],
+                    },
+                    player2: {
+                        groundArena: ['snowspeeder']
+                    }
+                });
+            });
+
+            it('should gain mandalorian\'s trait', function () {
```
AMMayberry1, line 17: nit: right now the name of this test will be `Foundling - integration - Foundling should gain mandalorian's trait`. Could we change here and line 3 so that this part becomes something like "Foundling's ability should give the attached card the Mandalorian trait"

test/server/cards/02_SHD/units/MandalorianWarrior.spec.js
```diff
@@ -0,0 +1,25 @@
...
+                        groundArena: ['protector-of-the-throne', 'battlefield-marine'],
+                    },
+                    player2: {
+                        groundArena: ['snowspeeder']
+                    }
+                });
+            });
+
+            it('should give an experience to mandalorian unit', function () {
+                // add experience to a mandalorian
```
AMMayberry1, line 18: Don't necessarily need to add comments for really simple tests like this, where the test name pretty much explains the whole test. Not necessarily a problem, if you find it helpful then definitely keep doing it. Just not a requirement unless the test is more complex

test/server/cards/02_SHD/units/MandalorianWarrior.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('Mandalorian Warrior', function () {
+    integration(function () {
+        describe('Mandalorian Warrior\'s ability', function () {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['mandalorian-warrior'],
+                        groundArena: ['protector-of-the-throne', 'battlefield-marine'],
```
AMMayberry1, line 9: We typically try to add units to make sure all of the filters are working correctly. In this case, could you please add:
- A deployed Mandalorian leader to confirm that leader units are targetable
- An enemy Mandalorian unit to confirm that enemy units are targetable

test/server/cards/02_SHD/units/ProtectorOfTheThrone.spec.js
```diff
@@ -0,0 +1,36 @@
...
+                    phase: 'action',
+                    player1: {
+                        groundArena: [{ card: 'protector-of-the-throne', upgrades: ['shield'] }],
+                    },
+                    player2: {
+                        groundArena: ['wampa', 'jedha-agitator'],
+                    }
+                });
+            });
+            it('should give it sentinel only as long as it is upgraded', function () {
```
AMMayberry1, line 15: nit: newline before test case

AMMayberry1: Great test otherwise! This covers everything we would want.

test/server/cards/02_SHD/upgrades/Foundling.spec.js
```diff
@@ -0,0 +1,37 @@
...
+                        hand: ['foundling', 'mandalorian-warrior'],
+                        groundArena: ['battlefield-marine', 'pyke-sentinel', 'protector-of-the-throne'],
+                    },
+                    player2: {
+                        groundArena: ['snowspeeder']
+                    }
+                });
+            });
+
+            it('\'s ability should give the attached card the Mandalorian trait', function () {
```
AMMayberry1, line 17: nit: jasmine adds a space in between the "describe" string and the "it" string. So in this case it would come out as `Foundling 's ability should...` instead of `Foundling's ability should...`. If you move the `\'s` into `describe` it will look correct

test/server/cards/02_SHD/upgrades/Foundling.spec.js
```diff
@@ -0,0 +1,37 @@
...
+                this.player1.clickCard(this.battlefieldMarine);
+
+                // check players units traits
+                expect(this.battlefieldMarine.hasSomeTrait('mandalorian')).toBeTrue();
+                expect(this.pykeSentinel.hasSomeTrait('mandalorian')).toBeFalse();
+
+                this.player2.pass();
+                // mandalorian warrior should be able to add experience to battlefield marine
+                this.player1.clickCard(this.mandalorianWarrior);
+                expect(this.player1).toBeAbleToSelectExactly([this.battlefieldMarine, this.protectorOfTheThrone]);
```
AMMayberry1, line 33: Since this is kind of a unique case, let's go ahead and add the actual click and confirm that battlefield marine gains the experience correctly. The expectation would be `.hasExactUpgradeNames(['experience'])`

test/server/cards/02_SHD/units/MandalorianWarrior.spec.js
```diff
@@ -0,0 +1,25 @@
...
+                    player2: {
+                        groundArena: ['snowspeeder']
+                    }
+                });
+            });
+
+            it('should give an experience to mandalorian unit', function () {
+                // add experience to a mandalorian
+                this.player1.clickCard(this.mandalorianWarrior);
+                this.player1.clickPrompt('Give 1 experience to another Mandalorian unit');
```
AMMayberry1, line 20: Could you add `expect(this.player1).toHavePassAbilityButton();`. I forgot about that one

test/server/cards/02_SHD/units/ProtectorOfTheThrone.spec.js
VJubert: pasted from supercommando squad 😆

server/game/cards/02_SHD/upgrades/Foundling.ts
```diff
@@ -0,0 +1,21 @@
...
+        return {
+            id: '7687006104',
+            internalName: 'foundling',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addConstantAbilityTargetingAttached({
+            title: 'Give the Mandalorian trait to the attached card',
+            ongoingEffect: OngoingEffectBuilder.card.static(EffectName.AddTrait, 'mandalorian'),
```
AMMayberry1, line 16: Sorry, missed this before - should use the enum `Trait.Mandalorian` and not the string

test/server/cards/02_SHD/upgrades/Foundling.spec.js
```diff
@@ -0,0 +1,39 @@
...
+                // check battlefield marine trait
+                expect(this.battlefieldMarine.hasSomeTrait('mandalorian')).toBeFalse();
+
+                // add battlefield marine foundling
+                this.player1.clickCard(this.foundling);
+                expect(this.player1).toBeAbleToSelectExactly([this.battlefieldMarine, this.snowspeeder, this.pykeSentinel, this.protectorOfTheThrone]);
+                this.player1.clickCard(this.battlefieldMarine);
+
+                // check players units traits
+                expect(this.battlefieldMarine.hasSomeTrait('mandalorian')).toBeTrue();
```
AMMayberry1, line 27: For consistency, let's also use `Trait.Mandalorian` throughout the tests. It gets really messy if we use strings instead of the enum name.

AMMayberry1 (approved):

PR #88, opened 2024-10-03
feat(card): Implement Asteroid Sanctuary

joshfullmer:


test/server/cards/01_SOR/events/AsteroidSanctuary.spec.js
```diff
@@ -0,0 +1,68 @@
...
+                this.player1.clickCard(this.asteroidSanctuary);
+                expect(this.player1).toBeAbleToSelectExactly([this.deathTrooper, this.superlaserTechnician]);
+
+                this.player1.clickCard(this.deathTrooper);
+                expect(this.player1).toBeAbleToSelectExactly([this.viperProbeDroid, this.deathStarStormtrooper]);
+
+                this.player1.clickCard(this.viperProbeDroid);
+                expect(this.deathTrooper.exhausted).toBe(true);
+                expect(this.viperProbeDroid).toHaveExactUpgradeNames(['shield']);
+                expect(this.asteroidSanctuary).toBeInLocation('discard');
```
AMMayberry1, line 27: Generally we don't need to test this, we can assume that event resolution is correctly putting the card in discard. If the goal is to check that the card is done resolving, could do `expect(this.player2).toBeActivePlayer()`

server/game/cards/01_SOR/events/AsteroidSanctuary.ts
```diff
@@ -0,0 +1,33 @@
...
+            title: 'Exhaust an enemy unit and give a shield to a friendly unit that costs 3 or less',
+            targetResolvers: {
+                exhaust: {
+                    locationFilter: WildcardLocation.AnyArena,
+                    controller: RelativePlayer.Opponent,
+                    immediateEffect: AbilityHelper.immediateEffects.exhaust()
+                },
+                shield: {
+                    controller: RelativePlayer.Self,
+                    cardCondition: (card) => (card as UnitCard).cost <= 3,
```
AMMayberry1, line 25: We have a typescript typing trick available that makes this a little simpler (and the type conversion safer):

```typescript
cardCondition: (card) => card.isUnitCard() && card.cost <= 3,
```

The `card.isUnitCard()` will cause `card` to be identified as type `UnitCard` if the check succeeds. Won't go into the technical details unless you're interested, but that's our preferred way of handling this situation now (might be some older code that still does it the other way).

joshfullmer: Ahh, this is a nice use of type predicates! Done in https://github.com/SWU-Karabast/forceteki/pull/88/commits/2b1010e504f7710ed6260379ea9f7e47893d0901 

Also, fixed the doc where I found the original usage.

test/server/cards/01_SOR/events/AsteroidSanctuary.spec.js
joshfullmer: I opted to include the `toBeActivePlayer` checks to confirm there are no additional actions need from the player. Done in https://github.com/SWU-Karabast/forceteki/pull/88/commits/a726380c1f2d950046f1404d41725e69d45ac4a1

AMMayberry1 (approved):

PR #89, opened 2024-10-03
Events

VJubert:
- Takedown
- Fell the Dragon
- Make an Opening
- Rival's Fall
- Covert Strength
- Moment of the Glory


server/game/cards/01_SOR/events/MakeAnOpening.ts
```diff
@@ -0,0 +1,32 @@
...
+        this.setEventAbility({
+            title: 'Give a unit -2/-2 for this phase. Heal 2 damage from your base',
+            targetResolvers: {
+                downgradeUnit: {
+                    cardTypeFilter: WildcardCardType.Unit,
+                    immediateEffect: AbilityHelper.immediateEffects.forThisPhaseCardEffect({
+                        effect: AbilityHelper.ongoingEffects.modifyStats({ power: -2, hp: -2 })
+                    })
+                },
+                heal: {
```
AMMayberry1, line 23: So this case might get a little complicated. The first thing is that I think you may have misread the card, it specifically says "heal your base" not "heal a base."

That said, since there is only one target for the effect, we need to change this. We only want to use target resolvers if there are potentially multiple targets for the player to choose. For the heal case there are not, the target is known in advance. This code works but it is kind of going about it slightly the wrong way.

In that case normally we use the `immediateEffect` field instead of `targetResolver`. I recommend you first take a look at the comment I left on Smuggler's Aid since that will kind of show the difference.

This case is more complex than Smuggler's Aid for two reasons:

1. There are multiple effects
2. We have one effect that needs a target decided by the player (-2/-2), and one that does not (heal)

To get some initial context, you should the read the wiki about multiple effects: https://github.com/SWU-Karabast/forceteki/wiki/Advanced-Card-Behaviors#abilities-with-multiple-effects

So what we need to do is use the `simultaneous()` effect to create two effects that resolve together, and then use `selectCard` to let the player choose the target for the -2/-2. It will look kind of like this:

```typescript
  this.setEventAbility({
      title: 'Give a unit -2/-2 for this phase. Heal 2 damage from your base',
          AbilityHelper.immediateEffects.selectCard({
              innerSystem: AbilityHelper.immediateEffects.forThisPhaseCardEffect({
                  effect: AbilityHelper.ongoingEffects.modifyStats({ power: -2, hp: -2 }),
              }),
          }),
          AbilityHelper.immediateEffects.heal((context) => ({ amount: 2, target: context.source.controller.base }))
      ]),
  });
```

Unfortunately there are not many cards yet that use `selectCard` so we don't have great examples for it, you could look at [Headhunting](https://github.com/SWU-Karabast/forceteki/blob/944e671f65218a4220b75afb3ef32495627f3c2d/server/game/cards/02_SHD/events/Headhunting.ts#L27) as an example but that's a complex one.

The tl;dr is basically it functions just like `targetResolver`, and it's what we have to use in situations where we are mixing things that need the player to choose a target (the -2/-2) and things that don't (the heal effect).

server/game/cards/01_SOR/events/Takedown.ts
```diff
@@ -0,0 +1,25 @@
...
+    protected override getImplementationId () {
+        return {
+            id: '4849184191',
+            internalName: 'takedown',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setEventAbility({
+            title: 'Defeat an unit with 5 or less remaining HP',
```
AMMayberry1, line 15: nit, grammar: should be "Defeat _**a**_ unit". Ditto for Rival's Fall

server/game/cards/02_SHD/events/FellTheDragon.ts
```diff
@@ -0,0 +1,25 @@
...
+            internalName: 'fell-the-dragon',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setEventAbility({
+            title: 'Defeat a non-leader unit with 5 or more power',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                cardCondition: (card, _) => card.isNonLeaderUnit() && card.getPower() >= 5,
```
AMMayberry1, line 18: Since we have to do the `isNonLeaderUnit()` check here anyway you might as well remove the `cardTypeFilter` since they're redundant

server/game/cards/02_SHD/events/SmugglersAid.ts
```diff
@@ -0,0 +1,25 @@
...
+        return {
+            id: '0866321455',
+            internalName: 'smugglers-aid',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setEventAbility({
+            title: 'Heal 3 damage from you base',
+            targetResolver: {
```
AMMayberry1, line 16: Similar situation here as for Make an Opening, except that this case is much simpler since there is only one effect. Because we know the target in advance (controller's base), we do not need a target resolve. Can just do the `immediateEffect` directly:

```typescript
    this.setEventAbility({
        title: 'Heal 3 damage from your base',
        immediateEffect: AbilityHelper.immediateEffects.heal((context) => ({ target: context.source.controller.base, amount: 3 }))
    });
}
```

It might be a little confusing that we pass in the `context` to the `heal()`, but everything that is under `immediateEffects` has the option to accept either (a) a properties object or (b) a function that accepts `context` and generates a properties object. In this case we do option (b) because we need to get the source card so that we can find out who is the controlling player whose base will be healed.

test/server/cards/01_SOR/events/MakeAnOpening.spec.js
```diff
@@ -0,0 +1,37 @@
...
+                    }
+                });
+            });
+
+            it('should downgrade an unit and heal base', function () {
+                this.player1.clickCard(this.makeAnOpening);
+                expect(this.player1).toBeAbleToSelectExactly([this.pykeSentinel, this.gideonHask, this.atst, this.isbAgent, this.cartelSpacer, this.sabineWren]);
+
+                this.player1.clickCard(this.isbAgent);
+                // VBL : maybe a bug somewhere, if I keep this test here he fails, but at the end of event it works
```
AMMayberry1, line 24: This is intended behavior, none of the card effects resolve until all targets are selected. However, as I mentioned in the comment on the implementation, the player should not be required to choose a target for the other effect since it is known in advance ("your base")

test/server/cards/01_SOR/events/MakeAnOpening.spec.js
```diff
@@ -0,0 +1,37 @@
...
+                // expect(this.isbAgent.getPower()).toBe(0);
+                // expect(this.isbAgent.remainingHp).toBe(1);
+
+                this.p1Base.damage = 5;
```
AMMayberry1, line 28: Prefer if we do setup stuff like this at the beginning of the test just so it's easy to see

test/server/cards/01_SOR/events/MakeAnOpening.spec.js
```diff
@@ -0,0 +1,37 @@
+describe('Make an Opening', function () {
+    integration(function () {
+        describe('Make an Opening\'s ability', function () {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['make-an-opening'],
+                        groundArena: ['pyke-sentinel', { card: 'gideon-hask#ruthless-loyalist', upgrades: ['entrenched'] }],
```
AMMayberry1, line 9: Is there a reason to have so many units + upgrades in this test?

test/server/cards/01_SOR/events/Takedown.spec.js
```diff
@@ -0,0 +1,56 @@
...
+
+            it('should defeat an ally', function () {
+                this.player1.clickCard(this.takedown);
+                expect(this.player1).toBeAbleToSelectExactly([this.pykeSentinel, this.isbAgent, this.cartelSpacer, this.bobaFett, this.sabineWren]);
+
+                this.player1.clickCard(this.pykeSentinel);
+                expect(this.pykeSentinel).toBeInLocation('discard');
+            });
+
+            it('should defeat an leader', function () {
```
AMMayberry1, line 36: nit: "a leader", a couple other places as well

test/server/cards/01_SOR/events/Takedown.spec.js
```diff
@@ -0,0 +1,56 @@
...
+                this.player1.clickCard(this.takedown);
+                expect(this.player1).toBeAbleToSelectExactly([this.pykeSentinel, this.isbAgent, this.cartelSpacer, this.bobaFett, this.sabineWren]);
+
+                this.player1.clickCard(this.sabineWren);
+                expect(this.sabineWren.deployed).toBeFalse();
+            });
+
+            it('should defeat an downgrade unit', function () {
```
AMMayberry1, line 44: "Downgrade" isn't an official game term so it might be confusing. Maybe "a unit with an hp reducing effect"?

test/server/cards/02_SHD/events/FellTheDragon.spec.js
```diff
@@ -0,0 +1,36 @@
...
+                    },
+                    player2: {
+                        groundArena: ['atst', 'isb-agent'],
+                        spaceArena: ['cartel-spacer'],
+                        leader: { card: 'darth-vader#dark-lord-of-the-sith', deployed: true }
+                    }
+                });
+            });
+
+            it('should defeat an ennemy', function () {
```
AMMayberry1, line 19: nit, typo: ennemy. Ditto Rival's Fall

test/server/cards/01_SOR/events/Takedown.spec.js
AMMayberry1: Love this test though! Great idea, I don't think we have a similar test that covers this case with hp reducing effects yet.

test/server/cards/02_SHD/events/FellTheDragon.spec.js
```diff
@@ -0,0 +1,36 @@
+describe('Fell the Dragon', function() {
+    integration(function() {
+        describe('Fell the Dragon\'s ability', function() {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['fell-the-dragon'],
+                        groundArena: ['pyke-sentinel', { card: 'scout-bike-pursuer', upgrades: ['academy-training'], damage: 4 }, 'avenger#hunting-star-destroyer'],
```
AMMayberry1, line 9: Nice, like the setup on scout bike pursuer. Good test of those effects.

test/server/cards/02_SHD/events/MomentOfGlory.spec.js
```diff
@@ -0,0 +1,30 @@
...
+                        spaceArena: ['imperial-interceptor']
+                    }
+                });
+            });
+
+            it('can buff a unit', function () {
+                this.player1.clickCard(this.momentOfGlory);
+                expect(this.player1).toBeAbleToSelectExactly([this.pykeSentinel, this.sabineWren, this.wampa, this.imperialInterceptor]);
+
+                this.player1.clickCard(this.pykeSentinel);
```
AMMayberry1, line 23: This is another case where we don't want to just check a property, we prefer to check a real game effect. For buffs like this I will typically pick an enemy unit that would normally kill the attacker but since it has the buff it does not die.

So if Pyke Sentinel is normally 2/3 and is now 6/7 I'll pick something with (3 < power < 7) and hp > 6. So like AT-ST would be a nice simple one.

test/server/cards/02_SHD/events/CovertStrength.spec.js
```diff
@@ -0,0 +1,48 @@
...
+                        leader: { card: 'sabine-wren#galvanized-revolutionary', deployed: true, damage: 4 }
+                    },
+                    player2: {
+                        groundArena: ['wampa'],
+                        spaceArena: ['imperial-interceptor']
+                    }
+                });
+            });
+
+            it('can heal a unit', function () {
```
AMMayberry1, line 20: Awesome tests! No feedback, looks great.

test/server/cards/01_SOR/events/Takedown.spec.js
VJubert: experience from old engine ^^

test/server/cards/01_SOR/events/MakeAnOpening.spec.js
```diff
@@ -0,0 +1,32 @@
...
+                });
+            });
+
+            it('should reduce hp to a unit and heal your base', function () {
+                this.p1Base.damage = 5;
+                this.player1.clickCard(this.makeAnOpening);
+                expect(this.player1).toBeAbleToSelectExactly([this.pykeSentinel, this.isbAgent, this.cartelSpacer, this.sabineWren]);
+
+                this.player1.clickCard(this.isbAgent);
+                expect(this.isbAgent.getPower()).toBe(0);
```
AMMayberry1, line 25: Sorry, missed this - can we select a unit that has power > 2 so we can confirm that the effect is reducing by exactly 2

test/server/cards/01_SOR/events/Takedown.spec.js
```diff
@@ -0,0 +1,56 @@
...
+                    },
+                    player2: {
+                        groundArena: ['atst', 'isb-agent'],
+                        spaceArena: ['cartel-spacer'],
+                        leader: { card: 'sabine-wren#galvanized-revolutionary', deployed: true }
+                    }
+                });
+            });
+
+            it('should defeat a ennemy', function () {
```
AMMayberry1, line 20: nit, typo: ennemy (a few other places as well)

test/server/cards/01_SOR/events/MakeAnOpening.spec.js
```diff
@@ -16,17 +16,29 @@ describe('Make an Opening', function () {
                 });
             });
 
-            it('should reduce hp to a unit and heal your base', function () {
+            it('should fully-reduce hp to a unit and heal your base', function () {
```
AMMayberry1, line 19: Does the test with isb agent cover anything that the test with at-st does not? If not, let's remove the isb agent test

VJubert: isb agent test is here to see if -2/-2 can go behind 0

AMMayberry1: Oh gotcha. That's not needed here, we have other tests for power modifying effects which confirm this. So we can remove this test.

VJubert: He cover enemy unit while atst cover ally unit 😁

AMMayberry1 (approved):

PR #90, opened 2024-10-03
Implementing Waylay

wfowlks:
Implemented Waylay, also fixed the test in OpenFire and cleaned up some other upgrades that my IDE picked up as unused


test/server/cards/01_SOR/events/Confiscate.spec.js
```diff
@@ -48,7 +48,7 @@ describe('Confiscate', function() {
                 expect(this.wampa.isUpgraded()).toBe(false);
 
                 // this expectation will automatically check that entrenched is in the owning player's discard
-                expect(this.entrenched).toBeInLocation('discard');
+                expect(this.entrenched).toBeInLocation('discard', this.player1);
```
AMMayberry1, line 51: Oh yeah good catch, thanks

test/server/cards/01_SOR/events/OpenFire.spec.js
```diff
@@ -24,6 +24,8 @@ describe('Open Fire', function() {
 
                 this.player1.clickCard(this.cartelSpacer);
                 expect(this.cartelSpacer).toHaveExactUpgradeNames([]);
```
AMMayberry1, line 26: Thanks for fixing all this - I also realized that this line is maybe more clear as `expect(this.cartelSpacer.isUpgraded()).toBe(false)`

test/server/cards/01_SOR/events/Waylay.spec.js
```diff
@@ -0,0 +1,54 @@
...
+                        groundArena: [{ card: 'pyke-sentinel', upgrades: ['entrenched'] }],
+                    },
+                    player2: {
+                        groundArena: ['wampa', 'superlaser-technician'],
+                        spaceArena: [{ card: 'imperial-interceptor', upgrades: ['academy-training'] }]
+                    }
+                });
+            });
+
+            it('can return a friendly or enemy unit to its owner\'s hand', function () {
```
AMMayberry1, line 18: Waylay is kind of unique and we haven't tested the `returnToHand()` stuff much here yet so this is one where I'd actually prefer to add a more robust set of tests (apologies for the lack of clarity on that).

If you don't mind, could you add these cases:
- Add a deployed leader and confirm it doesn't appear as targetable for waylay
- A case where we select a friendly unit to return to hand
- Confirm that the waylay-ed unit can be immediately played again

If you think of any additional ones feel free to add them. One exception is that I do eventually want to use Waylay also for testing things like the Snoke effect, or triggers like "when played," but we will do those in a separate PR since that's more for testing the engine stuff than Waylay itself.

test/server/cards/01_SOR/events/Waylay.spec.js
```diff
@@ -0,0 +1,54 @@
...
+                        hand: ['entrenched'],
+                    },
+                    player2: {
+                        hand: ['waylay'],
+                        groundArena: ['wampa'],
+                    }
+                });
+            });
+
+            it('when played on a friendly upgraded unit, it will cause the upgrade to be in the owner\'s discard', function () {
```
AMMayberry1, line 41: Love this test but since it's more testing the behavior of the upgrade system than Waylay, so do you mind moving it to `Upgrade.spec.js`?

test/server/cards/01_SOR/events/Waylay.spec.js
```diff
@@ -0,0 +1,54 @@
...
+                this.player1.clickCard(this.entrenched);
```
AMMayberry1, line 42: Can you do me a small favor and add a comment here that says this:

`TODO: add an option to set an upgrade's controller in test setup`

We should make this simpler but it's not an urgent task

test/server/cards/01_SOR/events/Waylay.spec.js
wfowlks: Will do

test/server/cards/01_SOR/events/Waylay.spec.js
wfowlks: I was between `returnToHand()` and `returnToHandFromPlay()` and it was a toss up. I only did a shallow dive but my concerns were mostly around if we play this again the state of the object wont maintain damage etc. I will be honest i didnt deep dive into each of the helper methods. So I'm open to using which ever you prefer, and dont mind doinig more extensive tests

test/server/cards/01_SOR/events/Waylay.spec.js
```diff
@@ -0,0 +1,53 @@
...
+                this.player1.clickCard('waylay');
+                expect(this.player1).toBeAbleToSelectExactly([this.wampa, this.pykeSentinel, this.imperialInterceptor, this.superlaserTechnician]);
+
+                this.player1.clickCard(this.superlaserTechnician);
+                expect(this.superlaserTechnician).toBeInLocation('hand', this.player2);
+            });
+
+            it('should allow the player to select a friendly unit to return to hand and should remove damage and be playable', function () {
+                this.pykeSentinel.damage = 2;
+                this.pykeSentinel.exhausted = false; // Making sure it doesn't retain state of unexhausted
```
AMMayberry1, line 30: Does it work without this? Since other people tend to base their tests on existing ones, I want to avoid adding unnecessary code so that people don't think it's required for things to work right

test/server/cards/01_SOR/events/Waylay.spec.js
```diff
@@ -0,0 +1,53 @@
...
+                this.player1.clickCard('waylay');
+                expect(this.player1).toBeAbleToSelectExactly([this.wampa, this.pykeSentinel, this.superlaserTechnician, this.imperialInterceptor]);
+
+                this.player1.clickCard(this.pykeSentinel);
+                expect(this.pykeSentinel).toBeInLocation('hand', this.player1);
+                expect(this.entrenched).toBeInLocation('discard', this.player2);
+
+                this.player2.passAction();
+                this.player1.clickCard(this.pykeSentinel);
+                expect(this.pykeSentinel.damage).toBe(0); // Just making sure that the damage is not added back
```
AMMayberry1, line 45: Yeah good call

test/server/core/card/Upgrade.spec.js
```diff
@@ -105,5 +105,34 @@ describe('Upgrade cards', function() {
...
+                        groundArena: ['pyke-sentinel'],
+                    },
+                    player2: {
+                        hand: ['entrenched'],
+                        groundArena: ['wampa'],
+                    }
+                });
+            });
+
+            it('and a unit is returned to its owner\'s hand, the upgrade should be in the upgrade\'s owner\'s hand', function () {
```
AMMayberry1, line 124: I think this is a typo, shouldn't it be "the upgrade should be in the upgrade's owner's **discard**"?

server/game/cards/01_SOR/events/Waylay.ts
```diff
@@ -0,0 +1,25 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Return a non-leader unit to its owner\'s hand',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                locationFilter: WildcardLocation.AnyArena,
+                immediateEffect: AbilityHelper.immediateEffects.returnToHand()
```
AMMayberry1, line 19: Apologies, missed this earlier - could you try `returnToHandFromPlay()`? That's typically the one we want to use in this specific case. If it gives you any problems ping me in the discord

AMMayberry1: Then you can remove the `locationFilter` because `returnToHandFromPlay` will set it automatically

test/server/cards/01_SOR/events/Waylay.spec.js
wfowlks: I can add do not copy. To the comment. But I just wanted to be sure testing this mechanic, if you think it's unnecessary I'm happy to remove it but I'm more inclined to add a THIS IS NOT NECESSARY FOR OTHER TESTS to the comment

test/server/core/card/Upgrade.spec.js
wfowlks: You are correct sir

test/server/cards/01_SOR/events/Waylay.spec.js
AMMayberry1: Hmm I might be misunderstanding. I'm trying to say that the unit should start the test ready and this line is redundant. If that's not true there's a problem.

test/server/core/card/Upgrade.spec.js
AMMayberry1: Still says hand

wfowlks: yeah sorry, had to merge the changes from main, then i stash applied and lost the change... it worked on my machine :P 
I'll have this fixed pronto

PR #91, opened 2024-10-04
Implemented Leia Organa, Defiant Princess

Ennis0:


server/game/gameSystems/ReadyResourcesSystem.ts
```diff
@@ -0,0 +1,41 @@
+import { AbilityContext } from '../core/ability/AbilityContext';
+import type { Card } from '../core/card/Card';
+import { CardType, Location, WildcardCardType } from '../core/Constants';
+import Player from '../core/Player.js';
+import * as EnumHelpers from '../core/utils/EnumHelpers';
+import { IPlayerTargetSystemProperties, PlayerTargetSystem } from '../core/gameSystem/PlayerTargetSystem.js';
+
+
+export interface IReadyResourcesSystemProperties extends IPlayerTargetSystemProperties {
+    amount?: number
```
AMMayberry1, line 10: My thought would be to make the amount non-optional so that the card writer is required to provide a number. The reason just being that when I see:

`immediateEffect: AbilityHelper.immediateEffects.readyResources(),`

with no amount, it's not obvious to me that its default behavior is to ready 1 resource

server/game/core/Player.js
```diff
@@ -882,6 +882,16 @@ class Player extends GameObject {
         }
     }
 
+    /**
+     * Ready the specified number of resources
+     */
+    readyResources(count) {
```
AMMayberry1, line 888: Need logic to handle the case when count > number of available resources. My thought is to just add a contract check for that and make it illegal to do instead of making it allowed, I think allowing it as an option would create too many potential errors.

server/game/gameSystems/ReadyResourcesSystem.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public override getEffectMessage(context: TContext): [string, any[]] {
+        const properties = this.generatePropertiesFromContext(context);
+        const message = 'ready {0} resources';
+        return [message, [properties.amount]];
+    }
+
+    public override canAffect(player: Player, context: TContext, additionalProperties = {}): boolean {
+        return player.countExhaustedResources() > 0;
```
AMMayberry1, line 29: This check is slightly off, because an effect technically can target a player with no exhausted resources, it just won't do anything. However, if the effect is a cost, this must be true.

Take a look at how the `ReadySystem` handles `isCost` in this check as an example

server/game/gameSystems/ReadyResourcesSystem.ts
```diff
@@ -0,0 +1,41 @@
...
+import Player from '../core/Player.js';
+import * as EnumHelpers from '../core/utils/EnumHelpers';
+import { IPlayerTargetSystemProperties, PlayerTargetSystem } from '../core/gameSystem/PlayerTargetSystem.js';
+
+
+export interface IReadyResourcesSystemProperties extends IPlayerTargetSystemProperties {
+    amount?: number
+}
+
+export class ReadyResourcesSystem<TContext extends AbilityContext = AbilityContext, TProperties extends IReadyResourcesSystemProperties = IReadyResourcesSystemProperties> extends PlayerTargetSystem<TContext, TProperties> {
```
AMMayberry1, line 13: Since readying and exhausting resources are so similar, we probably want to have a base class that implements the core logic for both and then extend it with subclasses for each. Take a look at the ReadySystem and ExhaustSystem for an example of how to do this with the parent class ExhaustOrReadySystem.

server/game/gameSystems/ReadyResourcesSystem.ts
```diff
@@ -0,0 +1,41 @@
...
+        const message = 'ready {0} resources';
+        return [message, [properties.amount]];
+    }
+
+    public override canAffect(player: Player, context: TContext, additionalProperties = {}): boolean {
+        return player.countExhaustedResources() > 0;
+    }
+
+    public override defaultTargets(context: TContext): Player[] {
+        return [context.player];
```
AMMayberry1, line 33: Looks like the parent class does this behavior already, so we can remove this override

test/server/cards/01_SOR/units/LeiaOrganaDefiantPrincess.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+            it('should be able to exhaust an enemy unit', function () {
+                this.player1.clickCard(this.leiaOrgana);
+                this.player1.clickPrompt('Exhaust a unit');
+                expect(this.player1).toBeAbleToSelectExactly([this.leiaOrgana, this.atst, this.cartelSpacer, this.jynErso, this.wampa, this.allianceXwing]);
+                this.player1.clickCard(this.wampa);
+                expect(this.wampa.exhausted).toBe(true);
+            });
+
+            it('should be able to exhaust a friendly unit', function () {
```
AMMayberry1, line 36: I think Chaqa wrote the note in the excel that we need to test exhausting both a friendly and an enemy unit, this test feels unnecessary to me with all the other tests we have in place so I'd prefer to cut it.

I would like though to test what happens if there is no other unit on the field and the player selects "exhaust a unit," can we confirm that nothing happens? Since basically it would auto-target Leia as the only unit and she's exhausted already. But the button should still appear correctly.

server/game/gameSystems/ReadyResourcesSystem.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export interface IReadyResourcesSystemProperties extends IPlayerTargetSystemProperties {
+    amount?: number
+}
+
+export class ReadyResourcesSystem<TContext extends AbilityContext = AbilityContext, TProperties extends IReadyResourcesSystemProperties = IReadyResourcesSystemProperties> extends PlayerTargetSystem<TContext, TProperties> {
+    protected override defaultProperties: IReadyResourcesSystemProperties = {
+        amount: 1
+    };
+
```
AMMayberry1, line 17: Need to set an eventName property so that this system's event could be responded to with a triggered ability.

We have onSpendResources but adding onReadyResources and onExhaustResources would probably be clearer

server/game/gameSystems/ReadyResourcesSystem.ts
Ennis0: The parent class PlayerTargetSystem returns the player's opponent as the default target.

server/game/core/Player.js
Ennis0: What kind of errors do you think this could cause? This function is just a slight alteration of exhaustResources() right above it, which has been fine. count being higher than the number of exhausted resources is handled by the Math.min() in in for loop declaration.

server/game/gameSystems/ReadyResourcesSystem.ts
Ennis0: I based a lot of this class off of SearchDeckSystem(because it also targets a player), and the canAffect() in that class is only true if it could have an effect. Should that be changed as well?

test/server/cards/01_SOR/units/LeiaOrganaDefiantPrincess.spec.js
```diff
@@ -0,0 +1,50 @@
+describe('Leia Organa, Defiant Princess', function() {
+    integration(function() {
+        describe('Leia Organa, Defiant Princess\'s when played ability', function() {
+            beforeEach(function () {
+                this.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['leia-organa#defiant-princess'],
+                        groundArena: [{ card: 'atst', exhausted: true }],
+                        spaceArena: [{ card: 'cartel-spacer', exhausted: false }],
```
AMMayberry1, line 10: Don't need to explicitly mark `exhausted: false` on any of the units, it's the default

test/server/cards/01_SOR/units/LeiaOrganaDefiantPrincess.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+            it('should be able to exhaust any unit', function () {
+                this.player1.clickCard(this.leiaOrgana);
+                this.player1.clickPrompt('Exhaust a unit');
+                expect(this.player1).toBeAbleToSelectExactly([this.leiaOrgana, this.atst, this.cartelSpacer, this.jynErso, this.wampa, this.allianceXwing]);
+                this.player1.clickCard(this.wampa);
+                expect(this.wampa.exhausted).toBe(true);
+            });
+
+            it('should do nothing if the controller selects "Exhaust a unit" when no other unit is on the field', function () {
```
AMMayberry1, line 36: Can we create a new describe block and test setup for this case since it's so different

server/game/gameSystems/ReadyResourcesSystem.ts
AMMayberry1: Yeah that's incorrect as well then, probably it's just that search deck isn't currently a cost for anything so it hasn't come up. We can change it in this PR or just add a TODO comment there.

server/game/gameSystems/ReadyResourcesSystem.ts
Ennis0: There's really not much shared between them(as it stands, looking at PayResourceCostSystem.ts)(just having an "amount" property). It doesn't seem worth the extra file to me.

test/server/cards/01_SOR/units/LeiaOrganaDefiantPrincess.spec.js
Ennis0: I thought we were trying to limit the number of describe blocks for some reason. I was probably thinking of it() blocks/actually test count.

server/game/gameSystems/ReadyResourcesSystem.ts
AMMayberry1: Oh yeah that's fair actually, I guess since this system is targeting players there are far fewer checks to do so the system is much simpler. Seems good to me.

AMMayberry1:
Could you do me a favor and add a comment in ExhaustOrReadySystem: `TODO: change this to no longer be usable on resources and convert everything over to using ReadyResourceSystem or ExhaustResourceSystem`

test/helpers/PlayerInteractionWrapper.js
```diff
@@ -115,6 +115,9 @@ class PlayerInteractionWrapper {
                 });
             }
         } else {
+            if (leaderOptions.deployed === false) {
+                leaderCard.undeploy();
```
AMMayberry1, line 119: Can we make this idempotent? Currently it will break if the user sets `deployed: false` when the leader is already undeployed

test/server/cards/01_SOR/units/LeiaOrganaDefiantPrincess.spec.js
AMMayberry1: Following up on this - is it intentional to leave the explicit `exhausted: false` here? Since tests serve as documentation for writing other tests I would prefer to omit properties that aren't needed so that other people don't see them and think that they are needed.

If you prefer to keep them here to be explicit that's fine, in that case though can we get a comment explaining that's the reason and that the parameter is already false by default

test/server/cards/01_SOR/units/LeiaOrganaDefiantPrincess.spec.js
```diff
@@ -0,0 +1,56 @@
...
+                        leader: { card: 'jyn-erso#resisting-oppression', deployed: true, exhausted: false }
+                    },
+                    player2: {
+                        groundArena: [{ card: 'wampa', exhausted: false }],
+                        spaceArena: ['alliance-xwing']
+                    }
+                });
+            });
+
+            it('should be able to ready a friendly resource', function () {
```
AMMayberry1, line 20: Sorry, I somehow completely missed this until now - technically per RAW Leia should be able to ready a resource for the opponent.

That said, it may be the case we decide there's no reason to ever do that currently so we just shortcut and always do a friendly resource. If you prefer that option lmk and we can confirm with the rest of the team that they think it's a good idea.

test/server/cards/01_SOR/units/LeiaOrganaDefiantPrincess.spec.js
Ennis0: I changed it so that only the exhausted units explicitly have the exhausted property.

test/server/cards/01_SOR/units/LeiaOrganaDefiantPrincess.spec.js
Ennis0: I think I recall someone telling me the opponent's resource case wasn't actually needed. I'd agree with that too: I can't think of even an edge case where someone would want to ready an opponent's resource(at least in a standard 2 player game).

test/helpers/PlayerInteractionWrapper.js
Ennis0: Changed. Let me know if I did it right.

test/server/cards/01_SOR/units/LeiaOrganaDefiantPrincess.spec.js
AMMayberry1: Yeah I vaguely remembered some of the TWI cards keying off of ready resources but none of them would change that. I'm fine to leave as is until some real situation arises where it would be useful

test/helpers/PlayerInteractionWrapper.js
AMMayberry1: Looks perfect! Thanks

AMMayberry1 (approved):

PR #94, opened 2024-10-04
Feature/FirstLegionSnowTrooper

CheBato:
Added First Legion Snowtrooper card + Tests


server/game/cards/01_SOR/units/FirstLegionSnowtrooper.ts
```diff
@@ -0,0 +1,22 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '4619930426',
+            internalName: 'first-legion-snowtrooper',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'While attacking a damaged unit, this unit gets +2/+0 and gains Overwhelm. (Deal excess damage to the opponent\'s base.)',
```
AMMayberry1, line 15: We typically don't copy the text verbatim, just enough to communicate to the player what's about to happen (often times this text will be displayed on a button or something). In this case you could just remove the overwhelm explainer text

server/game/cards/01_SOR/units/FirstLegionSnowtrooper.ts
```diff
@@ -0,0 +1,22 @@
...
+            condition: (context) => context.source.isAttacking() && context.source.activeAttack?.target.isUnit() && (context.source.activeAttack?.target.location !== 'discard' && context.source.activeAttack?.target.damage > 0),
```
AMMayberry1, line 16: There's almost nowhere we need to use raw strings, we have enums that cover pretty much everything. In this case should use `Location.Discard` instead of `'discard'`

AMMayberry1: Although actually I think this check might not be fully correct, because it's possible that the card ended up somewhere other than discard (e.g., some hypothetical effect could bounce it back to hand). So it should be like this:

```typescript
import * as EnumHelpers from '../../../core/utils/EnumHelpers';

// inside card impl

condition: (context) => context.source.isAttacking() && context.source.activeAttack?.target.isUnit() && (EnumHelpers.isArena(context.source.activeAttack?.target.location) && context.source.activeAttack?.target.damage > 0),
```

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
```diff
@@ -0,0 +1,86 @@
...
+                    player1: {
+                        groundArena: ['first-legion-snowtrooper'],
+                    },
+                    player2: {
+                        groundArena: ['yoda#old-master'],
+                    }
+                });
+            });
+
+            it(' Ability should do nothing', function () {
```
AMMayberry1, line 16: So the jasmine test framework concatenates the test strings together hierarchically to make the test name. With what you have right now the test name would be `First Legion Snow Trooper - integration - First Legion Snow Trooper while attacking non-damaged unit.  Ability should do nothing`

Typically we structure the names to be more like `First Legion Snow Trooper - integration - Snow Trooper's ability should have no effect when attacking a non-damaged unit`. Ditto for all test names throughout

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
```diff
@@ -0,0 +1,86 @@
...
+                this.player1.clickCard(this.firstLegionSnowtrooper);
+                this.player1.clickCard(this.yodaOldMaster);
+                expect(this.firstLegionSnowtrooper.exhausted).toBe(true);
+                expect(this.firstLegionSnowtrooper.damage).toBe(2);
+                expect(this.yodaOldMaster.damage).toBe(2);
+                expect(this.player2).toBeActivePlayer();
+            });
+        });
+        describe('First Legion Snow Trooper with a damaged enemy unit.', function() {
```
AMMayberry1, line 25: Do you mind adding a newline before this? For some reason the linter doesn't catch this case, I need to investigate

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
```diff
@@ -0,0 +1,86 @@
...
+                });
+            });
+
+            it('First legion Snowtrooper should receive overwhelm and +2/+0, defeating yoda and dealing 1 damage to opponents base.', function () {
+                this.player1.clickCard(this.firstLegionSnowtrooper);
+                this.player1.clickCard(this.yoda);
+                expect(this.firstLegionSnowtrooper.exhausted).toBe(true);
+                expect(this.firstLegionSnowtrooper.damage).toBe(2);
+                expect(this.p2Base.damage).toBe(6);
+                expect(this.player2).toHaveEnabledPromptButtons(['You', 'Opponent', 'You and Opponent', 'No one']);
```
AMMayberry1, line 45: Do you mind switching to a unit that doesn't have an on-defeat just to make the test as simple as possible

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
```diff
@@ -0,0 +1,86 @@
...
+                        groundArena: [{ card: 'yoda#old-master', damage: 1 }],
+                        base: { card: 'dagobah-swamp', damage: 5 }
+                    }
+                });
+            });
+
+            it('First legion Snowtrooper should receive overwhelm and +2/+0, defeating yoda and dealing 1 damage to opponents base.', function () {
+                this.player1.clickCard(this.firstLegionSnowtrooper);
+                this.player1.clickCard(this.yoda);
+                expect(this.firstLegionSnowtrooper.exhausted).toBe(true);
```
AMMayberry1, line 42: Please break up code blocks with newlines in longer tests for readability

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
```diff
@@ -0,0 +1,86 @@
...
+                expect(this.firstLegionSnowtrooper.damage).toBe(2);
+                expect(this.p2Base.damage).toBe(6);
+                expect(this.player2).toHaveEnabledPromptButtons(['You', 'Opponent', 'You and Opponent', 'No one']);
+                this.player2.clickPrompt('You');
+                expect(this.player2).toBeActivePlayer();
+            });
+
+            it('First legion Snowtrooper attacking base and should not receive +2/+0 and overwhelm.', function () {
```
AMMayberry1, line 50: We have this unfortunate situation where the jasmine tests are very slow and we have a finite amount of free github actions runner time, so even though it's bad practice we try to consolidate tests if we can. In this case, could you merge this test into the previous one by setting `snowtrooper.exhausted = false` in between attacks? Please look at Iden Versio's tests for an example of how to do this and how to document it with comments

The test cases themselves are great!

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
```diff
@@ -0,0 +1,86 @@
...
+                        groundArena: [{ card: 'first-legion-snowtrooper', upgrades: ['experience'] }],
+                    },
+                    player2: {
+                        groundArena: [{ card: 'yoda#old-master', damage: 1 }],
+                        base: { card: 'dagobah-swamp', damage: 5 }
+                    }
+                });
+            });
+
+            it('First legion Snowtrooper attacking yoda and should receive +2/+0, overwhelm +1/+1 from experience upgrade. Defeating Yoda and dealing 2 to base', function () {
```
AMMayberry1, line 73: This is a good catch, we actually don't yet have any tests that cover overwhelm + upgrades. Since this is more a test of those systems though and not really the card implementation, can you do me a favor and move this test to `Upgrades.spec.js`?

server/game/cards/01_SOR/units/FirstLegionSnowtrooper.ts
CheBato: thank you for the comment, i'll be more vigilant in searching for a more appropriate enumerator.

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: When it came to this test i wanted to see how the conditional would react when Yoda was removed from play and the snowtrooper still had the isAttacking() True. If its unnecessary I'll change the unit.

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: i've moved it to Upgrades.spec.js.

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: I merged both test cases into one.

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: I added the comments and newlines for better readability.

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: Added the newline.

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: I fixed the texts to adhere to the standard. Thank you for the explanation.

server/game/cards/01_SOR/units/FirstLegionSnowtrooper.ts
CheBato: I fixed the text to adhere to the standard. Thank you for the explanation.

server/game/cards/01_SOR/units/FirstLegionSnowtrooper.ts
```diff
@@ -0,0 +1,22 @@
...
+        return {
+            id: '4619930426',
+            internalName: 'first-legion-snowtrooper',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'While attacking a damaged unit, this unit gets +2/+0 and gains Overwhelm.',
+            condition: (context) => context.source.isAttacking() && context.source.activeAttack?.target.isUnit() && (EnumHelpers.isArena(context.source.activeAttack?.target.location) && context.source.activeAttack?.target.damage > 0),
```
AMMayberry1, line 16: Sorry, one more request to make here - this line is super long, do you mind breaking it up into multiple lines

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
```diff
@@ -0,0 +1,73 @@
...
+                    }
+                });
+            });
+
+            it('should have no effect when attacking a non-damaged-unit', function () {
+                // actions
+                this.player1.clickCard(this.firstLegionSnowtrooper);
+                this.player1.clickCard(this.yodaOldMaster);
+
+                // check board state
```
AMMayberry1, line 21: This test is simple enough that the explainer comments are optional (feel free to keep them if you find them helpful, just letting you know they aren't strictly necessary)

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
```diff
@@ -0,0 +1,73 @@
+describe('First Legion Snow Trooper', function() {
+    integration(function() {
+        describe('First Legion Snow Troopers ability', function() {
```
AMMayberry1, line 3: nit, grammar: should be `'First Legion Snow Trooper\'s ability'`

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
AMMayberry1: It's good to test the snow trooper isAttacking case, we don't have great test coverage for that yet. But we have really good tests for when-defeated ability timing, so we can omit that and switch to a unit that doesn't have one for simplicity

test/server/core/card/Upgrade.spec.js
```diff
@@ -105,5 +105,33 @@ describe('Upgrade cards', function() {
...
+                        groundArena: [{ card: 'first-legion-snowtrooper', upgrades: ['experience'] }],
+                    },
+                    player2: {
+                        groundArena: [{ card: 'death-trooper', damage: 1 }],
+                        base: { card: 'dagobah-swamp', damage: 5 }
+                    }
+                });
+            });
+
+            it('First legion Snowtrooper attacking Death trooper should receive +2/+0, overwhelm and +1/+1 from experience upgrade. Defeating Death trooper and dealing 3 to base', function () {
```
AMMayberry1, line 123: Missed this test case name, still needs to be updated. Thanks!

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: Alrighty i'll swap to another unit.

server/game/cards/01_SOR/units/FirstLegionSnowtrooper.ts
CheBato: sure no problem!

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: if its all the same to you I'd like to keep them to have some form of structure.

AMMayberry1: Yep fine by me!

test/server/core/card/Upgrade.spec.js
CheBato: updated the case name.

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: fixed the text

server/game/cards/01_SOR/units/FirstLegionSnowtrooper.ts
CheBato: broke into three lines.

test/server/cards/01_SOR/units/FirstLegionSnowTrooper.spec.js
CheBato: swapped to  SnowTrooper lieutenant

AMMayberry1 (approved):

PR #95, opened 2024-10-04
More complexe cards

VJubert:


server/game/cards/02_SHD/leaders/BoKatanKryzePrincessInExile.ts
```diff
@@ -0,0 +1,61 @@
...
+                    onTrue: AbilityHelper.immediateEffects.damage({ amount: 1 }),
+                    onFalse: AbilityHelper.immediateEffects.noAction(),
+                })
+            }
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities () {
+        this.addOnAttackAbility({
+            title: 'You may deal 1 damage to a unit. If you attacked with another Mandalorian unit this phase, you may deal 1 damage to a unit',
```
AMMayberry1, line 38: I messaged you in discord about what the behavior is if we target a unit with both damage but it is defeated by the first one. Depending on how that is handled in the engine, we might want to make some changes here.

I think ideally the behavior I would prefer is that if the unit would be defeated by the first damage we don't allow the user to select it for the second damage, that seems like it would be misleading. Instead they could just pass the second damage if they want.

That would require using `sequential()` and `selectCard()` for these, but first let's see what the current behavior is

AMMayberry1: Sorry to keep waffling on this, it's kind of a complicated situation - but I think we do actually have to use `sequential()` and `selectCard()` here. The reason is not so much targeting as the fact that if the unit dies to the first instance of damage, the second instance of damage should not resolve.

This matters if there was, for example, a new triggered ability that said "**When a friendly unit receives damage: draw a card**. If the target unit dies to the first Bo-Katan damage, then the target player should only draw 1 card, not 2. But the way it is currently, the player would draw 2 cards. So we need to use `sequential()` to force the first damage to resolve before the second one does.

server/game/cards/01_SOR/units/BountyHunterCrew.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public override setupCardAbilities () {
+        this.addWhenPlayedAbility({
+            title: 'Return an event from a discard pile',
+            targetResolver: {
+                optional: true,
+                cardTypeFilter: CardType.Event,
+                locationFilter: Location.Discard,
+                immediateEffect: AbilityHelper.immediateEffects.returnToHand({ locationFilter: Location.Discard })
```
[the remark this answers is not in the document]
VJubert: both are needed, the first is to get targets, the second is used to filter targets
I agree it's redundant but I did not find a way to remove  one

test/server/cards/02_SHD/leaders/ReyMoreThanAScavenger.js
```diff
@@ -0,0 +1,123 @@
...
+                        spaceArena: ['grey-squadron-ywing']
+                    }
+                });
+            });
+
+            it('should give an experience to a unit with 2 or less power (be careful Raid)', function () {
+                this.player1.clickCard(this.rey);
+                // need to order triggers between restore & on attack
+                this.player1.clickPrompt('Give an Experience token to a unit with 2 or less power');
+                // rey should not be selectable because Red Three give her Raid 1
```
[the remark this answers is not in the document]
VJubert: experience from an misplay 😆

test/server/cards/02_SHD/leaders/BoKatanKryzePrincessInExile.spec.js
```diff
@@ -0,0 +1,109 @@
...
+                this.player1.clickCard(this.bokatanKryze);
+                this.player1.clickCard(this.p2Base);
+                expect(this.player1).toBeAbleToSelectExactly([this.mandalorianWarrior, this.battlefieldMarine, this.bokatanKryze, this.protectorOfTheThrone, this.allianceXwing]);
+                this.player1.clickCard(this.protectorOfTheThrone);
+                expect(this.p2Base.damage).toBe(4);
+                expect(this.protectorOfTheThrone.damage).toBe(1);
+                expect(this.player2).toBeActivePlayer();
+                this.player2.pass();
+
+                // attack with a mandalorian
```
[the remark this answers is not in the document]
VJubert: that can be a test, you have to attack with another mandalorian, maybe if you find a way to unexhausted her, she should not deal 2 damages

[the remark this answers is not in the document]
AMMayberry1: Ohh good catch, I missed that it said "another"

server/game/cards/02_SHD/leaders/BoKatanKryzePrincessInExile.ts
AMMayberry1: Hey so I think we had another miscommunication, apologies for making the explanation so confusing. What you did is not quite what I meant 😅 but I think it's okay, I have a plan to make some changes to the way the engine works so that we can get around this problem.

My final answer: can you please revert the change you did here, it is correct each target resolver to have its own optional: true but the ability itself should not. At that point we will be good.

server/game/cards/02_SHD/leaders/BoKatanKryzePrincessInExile.ts
```diff
@@ -0,0 +1,60 @@
...
+            optional: true,
+            targetResolvers: {
+                firstDmg: {
+                    activePromptTitle: 'Deal 1 damage to a unit',
```
AMMayberry1, line 42: The damage selector already has a built-in prompt so these are not necessary - it's not really important to tell the player that it is the second instance of damage, they will know that from context. We try to keep the targeting prompts as simple as possible

AMMayberry1: Ahh, okay. That is in the test harness. Basically, every time `clickCard()` is called it checks to make sure that something about the prompt has changed to make sure that we did not click on a card that does nothing. In this case it is working right, it's just that the prompt for part 1 and part 2 is exactly the same so it gets confused.

You can use `clickCardNonChecking()` instead of `clickCard()` in cases where it is expected that the prompt will not change, please leave a comment explaining the reason

server/game/cards/01_SOR/units/BountyHunterCrew.ts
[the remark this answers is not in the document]
AMMayberry1: This is a bug, I will take a look when I have time (probably tonight)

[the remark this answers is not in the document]
AMMayberry1: Or if you want to investigate using the debug steps from the wiki that would also be awesome

server/game/cards/02_SHD/leaders/BoKatanKryzePrincessInExile.ts
```diff
@@ -0,0 +1,59 @@
...
+                })
+            }
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities () {
+        this.addOnAttackAbility({
+            title: 'You may deal 1 damage to a unit. If you attacked with another Mandalorian unit this phase, you may deal 1 damage to a unit',
+            optional: true,
+            targetResolver: {
```
AMMayberry1, line 40: Couple of things - first, ignore my earlier comment, we should be using `simultaneous()` not `sequential()`, apologies for the confusion.

This is kind of confusing having a targetResolver inside of which is a sequential with another selectCard, it's very hard to read and understand. I would find it much more intuitive to just have the meta-system be the root and then use two selectCard calls:

```typescript
    AbilityHelper.immediateEffects.selectCard({ innerSystem: AbilityHelper.immediateEffect.damage(...) }),
    AbilityHelper.immediateEffect.conditional({ onTrue: AbilityHelper.immediateEffects.selectCard(...), ... ),
])
```

test/server/cards/02_SHD/leaders/BoKatanKryzePrincessInExile.spec.js
```diff
@@ -0,0 +1,154 @@
...
+
+                // 2 triggers as we attack with another mandalorian (2 damage to 1 unit who die on first damage)
+                this.bokatanKryze.exhausted = false;
+                this.player2.passAction();
+                this.player1.clickCard(this.bokatanKryze);
+                this.player1.clickCard(this.p2Base);
+                expect(this.player1).toBeAbleToSelectExactly([this.mandalorianWarrior, this.battlefieldMarine, this.bokatanKryze, this.protectorOfTheThrone, this.allianceXwing, this.jedhaAgitator]);
+                expect(this.player1).toHavePassAbilityButton();
+                // prompt does not change between 2 effects of bo katan ability
+                this.player1.clickCardNonChecking(this.jedhaAgitator);
```
AMMayberry1, line 145: In this situation we should use clickCard() because the prompt does change (Jedha Agitator is removed from the list of options). However, if we change to `simultaneous()` as I suggested above, then the prompt will not change and we should keep using `clickCardNonChecking()`.

server/game/cards/02_SHD/leaders/BoKatanKryzePrincessInExile.ts
```diff
@@ -0,0 +1,59 @@
...
+                    onFalse: AbilityHelper.immediateEffects.noAction(),
+                })
+            }
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities () {
+        this.addOnAttackAbility({
+            title: 'You may deal 1 damage to a unit. If you attacked with another Mandalorian unit this phase, you may deal 1 damage to a unit',
+            optional: true,
```
AMMayberry1, line 39: This should not be optional because that would mean passing the entire ability, including both instances of damage. Instead this should be non-optional and each target should be optional.

test/server/cards/02_SHD/leaders/BoKatanKryzePrincessInExile.spec.js
```diff
@@ -0,0 +1,154 @@
...
+                this.player1.clickCard(this.battlefieldMarine);
+                expect(this.battlefieldMarine.damage).toBe(2);
+
+                // 2 triggers as we attack with another mandalorian (2 damage to 1 unit who die on first damage)
+                this.bokatanKryze.exhausted = false;
+                this.player2.passAction();
+                this.player1.clickCard(this.bokatanKryze);
+                this.player1.clickCard(this.p2Base);
+                expect(this.player1).toBeAbleToSelectExactly([this.mandalorianWarrior, this.battlefieldMarine, this.bokatanKryze, this.protectorOfTheThrone, this.allianceXwing, this.jedhaAgitator]);
+                expect(this.player1).toHavePassAbilityButton();
```
AMMayberry1, line 143: Once we update the use of `optional` in the ability as I mentioned above, can we add a tests to confirm that it works if we pass either the first or second damage (i.e., if we pass first damage, second damage should still be allowed).

AMMayberry1 (approved):

PR #100, opened 2024-10-07
Units

VJubert:


server/game/cards/02_SHD/leaders/QiraIAloneSurvived.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    protected override setupLeaderUnitSideAbilities () {
+        this.addTriggeredAbility({
+            title: ' Heal all damage from each unit. Then, deal damage to each unit equal to half its remaining HP, rounded down.',
+            when: {
+                [EventName.OnLeaderDeployed]: (event, context) => event.card === context.source
+            },
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.heal((context) => ({ target: context.source.controller.getUnitsInPlay().concat(context.source.controller.opponent.getUnitsInPlay()), amount: (c) => c.damage })),
```
AMMayberry1, line 35: For readability, can we use `card` instead of `c` here and split the `heal()` and `damage()` into multiple lines like this:

```typescript
AbilityHelper.immediateEffects.heal((context) => ({ 
    target: context.source.controller.getUnitsInPlay().concat(context.source.controller.opponent.getUnitsInPlay()), 
    amount: (c) => c.damage 
}))
```

 Otherwise, these look exactly as they should! Great job 😄

server/game/cards/02_SHD/units/BountyGuildInitiate.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public override setupCardAbilities () {
+        this.addWhenPlayedAbility({
+            title: 'Deal 2 damage to a ground unit if you control another Bounty Hunter unit',
+            targetResolver: {
+                locationFilter: Location.GroundArena,
+                optional: true,
+                immediateEffect: AbilityHelper.immediateEffects.conditional({
+                    condition: (context) => context.source.controller.getOtherUnitsInPlay(context.source, WildcardLocation.AnyArena, (card) => card.hasSomeTrait(Trait.BountyHunter)).length > 0,
```
AMMayberry1, line 20: We have a helper for this - can just do `context.source.controller.isTraitInPlay(Trait.BountyHunter)`

server/game/cards/02_SHD/units/GreySquadronYWing.ts
```diff
@@ -0,0 +1,25 @@
...
+            id: '9472541076',
+            internalName: 'grey-squadron-ywing'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addOnAttackAbility({
+            title: 'An opponent chooses a unit or base they control. You may deal 2 damage to it',
+            targetResolver: {
+                choosingPlayer: RelativePlayer.Opponent,
```
AMMayberry1, line 17: Just leaving a placeholder comment here from our discussion in chat

server/game/cards/02_SHD/units/HuntingNexu.ts
```diff
@@ -0,0 +1,23 @@
...
+        return {
+            id: '8991513192',
+            internalName: 'hunting-nexu'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addConstantAbility({
+            title: 'While you control another Aggression unit, this unit gains Raid 2',
+            condition: (context) => context.source.controller.getOtherUnitsInPlayWithAspect(context.source, Aspect.Aggression).length > 0,
```
AMMayberry1, line 16: Do you mind adding a helper similar to the one mentioned above called `Player.isAspectInPlay()`? Probably will be useful for other cases like this in the future

server/game/cards/02_SHD/units/HuntingNexu.ts
```diff
@@ -0,0 +1,23 @@
...
+            matchTarget: (card, context) => card === context.source,
```
AMMayberry1, line 17: Can remove this, by default a constant ability will target only the source card

server/game/gameSystems/DamageSystem.ts
```diff
@@ -23,7 +22,11 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext> exte
     protected override readonly targetTypeFilter = [WildcardCardType.Unit, CardType.Base];
 
     public eventHandler(event): void {
-        event.card.addDamage(event.damage);
+        if (typeof event.damage === 'number') {
```
AMMayberry1, line 25: To prevent any weird fallthrough situations due to bugs, let's make this a switch statement that checks for `case 'number'` and `case 'function'` and in the `default` case has Contract.fail(\`Unexpected type: ${typeof event.damage}\`).

Ditto for HealSystem.

server/game/gameSystems/DeployLeaderSystem.ts
```diff
@@ -17,7 +18,15 @@ export class DeployLeaderSystem<TContext extends AbilityContext = AbilityContext
...
 
+        const deployedEvent = new GameEvent(this.eventName, {
+            player: event.context.source.controller,
+            card: event.card,
+            context: event.context
+        });
+
+        event.context.game.addMessage('{0} plays {1}', event.context.player, event.context.source);
         event.card.deploy();
+        event.context.game.openEventWindow(deployedEvent, true);
```
AMMayberry1, line 29: This is currently resolving the `deployedEvent` in a separate window than the `moveCard` action which happens inside of `deploy()`. It's effectively resolving them sequentially, but we need them to resolve in the same window simultaneously.

The way to do that is to have this event create a "contingent event", which means basically that when this event resolves it will create a second event that also resolves in the same window.

An example of how to set that up is in DefeatSystem below. You can see that the `updateEvent()` method calls a method in `CardTargetSystem`, which in turn calls `event.setContingentEventsGenerator`. You need to do the same but do it all inside of `DeployLeaderSystem.updateEvent()` and just have `event.setContingentEventsGenerator()` create the `deployedEvent`.

https://github.com/SWU-Karabast/forceteki/blob/1ce928369b827de16f5ff381ec531939f116c181/server/game/gameSystems/DefeatCardSystem.ts#L66

It's a little complex sorry, lmk if it would be helpful for me to provide more detail there. If it's helpful, you can see here where the event window generates the contingent events and adds them to the window to be resolved: https://github.com/SWU-Karabast/forceteki/blob/1ce928369b827de16f5ff381ec531939f116c181/server/game/core/event/EventWindow.js#L45

test/server/cards/02_SHD/leaders/QiraIAloneSurvived.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+                expect(context.hylobonEnforcer.damage).toBe(2);
+                expect(context.qira.exhausted).toBeTrue();
+                expect(context.player1.countExhaustedResources()).toBe(1);
+                expect(context.hylobonEnforcer).toHaveExactUpgradeNames(['shield']);
+
+                // try to give a shield to a unit with less than 2 hp
+                context.player2.passAction();
+                context.player1.clickCard(context.hylobonEnforcer);
+                expect(context.p2Base.damage).toBe(3);
```
AMMayberry1, line 33: Why the attack?

test/server/cards/02_SHD/leaders/QiraIAloneSurvived.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+                context.player1.clickPrompt('Deploy Qi’ra');
+
+                expect(context.player2).toBeActivePlayer();
+                expect(context.qira.damage).toBe(4); // 8/2 = 4
+                expect(context.deathStarStormtrooper.damage).toBe(0); // 1/2 = 0
+                expect(context.greenSquadronAwing.damage).toBe(0); // shield => 0
+                expect(context.greenSquadronAwing.isUpgraded()).toBeFalse(); // lost his shield
+                expect(context.ruggedSurvivors.damage).toBe(3); // (5+2=7)/2 = 3
+                expect(context.greySquadronYwing.damage).toBe(1); // 3/2 = 1
+                expect(context.hylobonEnforcer.damage).toBe(2);// 4/2 = 1
```
AMMayberry1, line 73: This is fantastic! Covers all the basic cases. Only extra thing I can think of to check is maybe have one more test that has either Snoke or Dodonna and confirm that the values are calculated correctly with hp modifiers.

test/server/cards/02_SHD/units/Crosshair.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+                context.crosshair.exhausted = false;
+                context.player2.passAction();
+                context.player1.clickCard(context.crosshair);
+                context.player1.clickPrompt('Attack');
+                context.player1.clickCard(context.p2Base);
+                expect(context.p2Base.damage).toBe(4);
+
+                // after the end of phase, boost are reverted back
+                context.player2.passAction();
+                context.player1.passAction();
```
AMMayberry1, line 59: Can we use `context.moveToNextActionPhase()` to have less code and for clarity of what's going on

test/server/cards/02_SHD/units/HuntingNexu.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+            it('should have Raid 2 because we control an Aggression unit.', function () {
+                const { context } = contextRef;
+                context.player1.clickCard(context.huntingNexu);
+                // autoselect base
+                expect(context.player2).toBeActivePlayer();
+                expect(context.p2Base.damage).toBe(6);
+            });
+        });
+
+        describe('Bounty Guild Initiate\'s ability', function() {
```
AMMayberry1, line 24: nit: wrong unit name

test/server/cards/02_SHD/units/OutlandTieVanguard.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+                        hand: ['outland-tie-vanguard'],
+                        groundArena: ['greedo#slow-on-the-draw', 'wampa'],
+                    },
+                    player2: {
+                        spaceArena: ['green-squadron-awing']
+                    }
+                });
+            });
+
+            it('should deal 2 damage to unit because we control another Bounty Hunter unit.', function () {
```
AMMayberry1, line 17: nit: wrong test name

test/server/cards/02_SHD/units/RhokaiGunship.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+                    }
+                });
+            });
+
+            it('should deal 1 damage to a unit or a base.', function () {
+                const { context } = contextRef;
+                context.player1.passAction();
+                context.player2.clickCard(context.greenSquadronAwing);
+                context.player2.clickCard(context.rhokaiGunship);
+                expect(context.player1).toBeAbleToSelectExactly([context.greenSquadronAwing, context.battlefieldMarine, context.p1Base, context.p2Base]);
```
AMMayberry1, line 22: nit: could use one or two lines of whitespace to break up the long code block

test/server/cards/02_SHD/units/RuthlessAssassin.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+            });
+
+            it('should deal 2 damage to a friendly unit.', function () {
+                const { context } = contextRef;
+                context.player1.clickCard(context.ruthlessAssassin);
+                // must select a friendly unit
+                expect(context.player1).toBeAbleToSelectExactly([context.ruthlessAssassin, context.greedo, context.greenSquadronAwing]);
+                expect(context.player1).not.toHaveChooseNoTargetButton();
+                context.player1.clickCard(context.ruthlessAssassin);
+                expect(context.ruthlessAssassin.damage).toBe(2);
```
AMMayberry1, line 25: nit again: also could use some whitespace

test/server/cards/02_SHD/units/WildRancor.spec.ts
```diff
@@ -0,0 +1,29 @@
+describe('Wild Rancor', function() {
+    integration(function(contextRef) {
+        describe('Wild Rancor\'s ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['wild-rancor'],
+                        groundArena: ['battlefield-marine'],
```
AMMayberry1, line 9: Let's add a leader for this check just to be sure that the method for getting all units in an arena is correctly getting leaders, I don't know if we've tested that yet

server/game/cards/02_SHD/units/GreySquadronYWing.ts
VJubert: y wing was deleted

test/server/cards/02_SHD/leaders/QiraIAloneSurvived.spec.ts
VJubert: to check if damage + grit is working

server/game/core/Player.js
```diff
@@ -288,16 +288,33 @@ class Player extends GameObject {
     }
 
     /**
-     * Returns if a card is in play (unots, upgrades, provinces, holdings) that has the passed trait
+     * Returns if a card is in play (units, upgrades, provinces, holdings) that has the passed trait
```
AMMayberry1, line 291: This is a mistake in the docs that I missed - since these methods currently only return units, can we update the docstr for this the surrounding methods to say that

server/game/core/Player.js
```diff
@@ -288,16 +288,33 @@ class Player extends GameObject {
...
      * @param {string} trait
-     * @returns {boolean} true/false if the trait is in pay
+     * @param {any} ignoreUnit
+     * @returns {boolean} true/false if the trait is in play
      */
-    isTraitInPlay(trait) {
+    isTraitInPlay(trait, ignoreUnit = undefined) {
```
AMMayberry1, line 296: nit: I think `null` and `undefined` are mostly interchangeable in javascript but I think in this situation it's more conventional to use `null`. Same for `isAspectInPlay`

server/game/core/Player.js
```diff
@@ -288,16 +288,33 @@ class Player extends GameObject {
...
+        if (ignoreUnit !== undefined) {
```
AMMayberry1, line 297: Could do this all in one statement using a ternary operator, i.e.

```javascript
return ignoreUnit != undefined
    ? this.getOtherUnitsInPlay(ignoreUnit).some((card) => card.hasSomeTrait(trait))
    : this.getUnitsInPlay().some((card) => card.hasSomeTrait(trait));
```

Also, this is the one situation it's better to use `!=` instead of `!==` because `!=` will match both `null` and `undefined`

Ditto on both of the above for `isAspectInPlay`

server/game/cards/01_SOR/units/TieAdvanced.ts
```diff
@@ -15,11 +15,11 @@ export default class TieAdvanced extends NonLeaderUnitCard {
             title: 'Give 2 Experience tokens to a friendly Imperial unit',
             targetResolver: {
                 controller: RelativePlayer.Self,
-                cardCondition: (card) => card.hasSomeTrait(Trait.Imperial) && card !== this,
+                cardCondition: (card, context) => card.hasSomeTrait(Trait.Imperial) && card !== context.source,
```
AMMayberry1, line 18: Thank you for catching all of these!!

test/server/cards/02_SHD/units/Crosshair.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+                    player1: {
+                        groundArena: ['crosshair#following-orders']
+                    },
+                    player2: {
+                        groundArena: ['wampa'],
+                    }
+                });
+            });
+
+            it('should make your opponent choose a unit to deal 2 damage', function () {
```
AMMayberry1, line 16: Just caught this - looks like wrong test name

test/server/cards/02_SHD/units/KoskaReeves.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+                        groundArena: ['koska-reeves#loyal-nite-owl'],
+                        spaceArena: ['green-squadron-awing']
+                    },
+                    player2: {
+                        groundArena: ['wampa'],
+                    }
+                });
+            });
+
+            it('should deal 2 damage to a friendly unit.', function () {
```
AMMayberry1, line 18: nit: might be better as `should deal 2 damage to a friendly unit when upgraded`

test/server/cards/02_SHD/units/RuthlessAssassin.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+                        groundArena: ['greedo#slow-on-the-draw'],
+                        spaceArena: ['green-squadron-awing']
+                    },
+                    player2: {
+                        groundArena: ['wampa'],
+                    }
+                });
+            });
+
+            it('should deal 2 damage to a friendly unit.', function () {
```
AMMayberry1, line 18: nit: friendly _ground_ unit

VJubert: that can be a space unit

AMMayberry1: Oh yeah right I'm thinking of death trooper

server/game/gameSystems/DamageSystem.ts
```diff
@@ -23,7 +22,16 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext> exte
     protected override readonly targetTypeFilter = [WildcardCardType.Unit, CardType.Base];
 
     public eventHandler(event): void {
-        event.card.addDamage(event.damage);
+        switch (typeof event.damage) {
```
AMMayberry1, line 25: I just realized there's an issue here - we need it to be the case that the `damage` value is set when the event is created, not when it is resolved.

There are two problems that could happen currently:

1. It is possible that attributes of the card could change during resolution, so that the `event.damage` function gives different values depending on when it is called
2. There are some cards such as Boba Fett's Armor that modify the damage value before it resolves, so the damage must be determined already

Maybe the simplest way to fix this is to do the calculation in `addPropertiesToEvent()` instead of here. That way it will be set once and not changed.

AMMayberry1 (approved):

PR #104, opened 2024-10-08
feat(card): Implement Boba Fett leader and Cantina Bouncer

joshfullmer:


server/game/cards/01_SOR/leaders/BobaFettCollectingTheBounty.ts
```diff
@@ -0,0 +1,35 @@
...
+            internalName: 'boba-fett#collecting-the-bounty',
+        };
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addTriggeredAbility({
+            title: 'Ready a resource',
+            cost: [AbilityHelper.costs.exhaustSelf()],
+            when: {
+                onCardReturnedToHand: (event, context) => event.card.isUnit() && event.card.controller !== context.source.controller && context.source.controller.resources.some((resource) => resource.exhausted),
```
AMMayberry1, line 17: I just realized we don't have it yet, but we definitely need an "onLeavesPlay" event that happens in parallel with any card being defeated / returned to hand from play.

If you're willing to tackle it that'd be awesome, it should be pretty straightforward but the place to add it is not super obvious. If you're up for it, instructions are below, otherwise just ping me and I can do a patch in the next day or two.

We have a concept of "contingent events" which are events that get emitted if-and-only-if a prior event actually resolves. Current examples are (1) overwhelm damage, which only happens if the original damage instance isn't prevented by a shield, and (2) defeating upgrades on a unit leaving play.

The code location for (2) above is where the "onLeavesPlay" event should also be emitted (note you will need add a new entry for it in Constants.ts). 

Go to `CardTargetSystem.addLeavesPlayPropertiesToEvent()` and update it to emit an anonymous event of type 'onLeavesPlay.` For an example of how to create such an event, see here: https://github.com/SWU-Karabast/forceteki/blob/22a2a2a5cb1565efd74d000fb29bcd8dc747b37d/server/game/core/gameSteps/AbilityResolver.js#L78

server/game/cards/01_SOR/leaders/BobaFettCollectingTheBounty.ts
```diff
@@ -0,0 +1,35 @@
...
+                onCardDefeated: (event, context) => event.card.isUnit() && event.card.controller !== context.source.controller && context.source.controller.resources.some((resource) => resource.exhausted),
```
AMMayberry1, line 18: We have to be intentional about how we set up the exhausted condition here, since it is possible that state will change in between the trigger time and the actual resolution time of the ability. See for example how Jedha Agitator is coded, with a conditional in the effect instead of in the trigger.

The difference is between these two ways of expressing the ability (pretending that the exhaust condition was an actual part of the text):

**When an enemy unit leaves play and you control an exhausted resource**: You may exhaust this leader. If you do, ready a resource.

**When an enemy unit leaves play**: If you control an exhausted resource, you may exhaust this leader. If you do, ready a resource.

In the latter case, the condition isn't checked until the time that the ability resolves. So if there were, for example, a triggered ability in the same window that caused a resource to become exhausted, we could activate Boba1's ability without missing the opportunity.

That said, the downside of doing it that was is that it means if there are multiple triggered abilities in the same window, the player will be prompted to resolve all of them. That includes the Boba1 ability, which may or may not actually have an effect (though the player should be able to realize that).

Lmk in chat if that's not clear. I'm not 100% sure what the best option is, I'm leaning towards thinking the Jedha Agitator way is better but curious for your thoughts

server/game/cards/01_SOR/leaders/BobaFettCollectingTheBounty.ts
```diff
@@ -0,0 +1,35 @@
...
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities() {
+        this.addTriggeredAbility({
+            title: 'Ready 2 resources',
+            when: {
+                onAttackCompleted: (event, context) => event.attack.attacker === context.source,
+            },
+            immediateEffect: AbilityHelper.immediateEffects.readyResources({ amount: 2 })
```
AMMayberry1, line 30: This is missing the check that an enemy left play this phase. This will also be kind of difficult to set up, let me know if you want to tackle it or for me to help.

Take a look at Iden Versio's implementation, specifically the state watcher piece. This is how we record game history. There is currently a state watcher for _defeat_ (the one Iden uses), you would need to add a parallel one for leaves play.

It's not hard if you follow the template, but it's a bit of work just fyi.

server/game/cards/01_SOR/units/CantinaBouncer.ts
```diff
@@ -0,0 +1,24 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Return a non-leader unit to its owner\'s hand.',
+            optional: true,
+            targetResolver: {
+                cardCondition: (card) => card.isNonLeaderUnit(),
+                immediateEffect: AbilityHelper.immediateEffects.returnToHand(),
```
AMMayberry1, line 18: Can you switch to `returnToHandFromPlay()`, I have a todo to convert this to refactor those to be more separate.

test/server/cards/01_SOR/leaders/BobaFettCollectingTheBounty.spec.js
```diff
@@ -0,0 +1,94 @@
...
+                    },
+                    player2: {
+                        groundArena: ['cell-block-guard'],
+                        leader: { card: 'luke-skywalker#faithful-friend', deployed: true },
+                    },
+                });
+            });
+
+            it('should ready a resource when an enemy unit leaves play - defeated', function() {
+                this.player1.clickCard(this.rivalsFall);
```
AMMayberry1, line 22: For tests like this that are primarily attack-based we tend to consolidate them down to fewer tests as much as possible by un-exhausting units in between, bc Jasmine tests are extremely heavy and we have finite github actions time.

It's bad test practice but kind of a practical limitation we're trying to observe for now. Check out the Iden tests and see how that's set up as a reference.

AMMayberry1: I don't have time to do a thorough review of tests on this pass but I think there's plenty in this review already, I'll take a detailed look at tests once the impl stuff is updated.

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -0,0 +1,42 @@
+import { Card } from '../core/card/Card';
+import { UnitCard } from '../core/card/CardTypes';
+import { UpgradeCard } from '../core/card/UpgradeCard';
+import { StateWatcherName } from '../core/Constants';
+import Player from '../core/Player';
+import { StateWatcher } from '../core/stateWatcher/StateWatcher';
+import { StateWatcherRegistrar } from '../core/stateWatcher/StateWatcherRegistrar';
+
+export interface CardLeftPlayEntry {
+    card: UnitCard | UpgradeCard;
```
AMMayberry1, line 10: We have the type union `InPlayCard` that covers this, just needs to be updated so it is exported

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -0,0 +1,42 @@
...
+    controlledBy: Player;
+}
+
+export type ICardsLeftPlayThisPhase = CardLeftPlayEntry[]
```
AMMayberry1, line 14: nit: missing semicolon, iirc there's a linter bug that prevents us from just adding that as a thing it catches automatically

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -0,0 +1,42 @@
...
+export class CardsLeftPlayThisPhaseWatcher extends StateWatcher<CardLeftPlayEntry[]> {
+    public constructor(registrar: StateWatcherRegistrar, card: Card) {
+        super(StateWatcherName.CardsLeftPlayThisPhase, registrar, card);
+    }
+
+    public override getCurrentValue() {
+        return super.getCurrentValue();
+    }
+
+    public getCardsLeftPlayControlledByPlayer(controller: Player): (UnitCard | UpgradeCard)[] {
```
AMMayberry1, line 25: The attack and play watchers both have a `filter` parameter on their corresponding methods to make retrieving results a little more succinct, would be good to have that here as well I think since it would be used in your Boba impl

server/game/cards/01_SOR/units/CantinaBouncer.ts
```diff
@@ -0,0 +1,24 @@
...
+            internalName: 'cantina-bouncer',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Return a non-leader unit to its owner\'s hand.',
+            optional: true,
+            targetResolver: {
+                cardCondition: (card) => card.isNonLeaderUnit(),
```
AMMayberry1, line 17: If we're just checking card type we can use the helper `cardTypeFilter: WildcardCardType.NonLeaderUnit`

test/server/cards/01_SOR/leaders/BobaFettCollectingTheBounty.spec.js
```diff
@@ -0,0 +1,108 @@
...
+                    },
+                    player2: {
+                        groundArena: ['viper-probe-droid', 'cell-block-guard', 'wampa'],
+                        leader: { card: 'luke-skywalker#faithful-friend', deployed: true },
+                    },
+                });
+            });
+
+            it('should ready a resource when an enemy unit leaves play', function() {
+                const reset = () => {
```
AMMayberry1, line 22: Ohh nice, love this pattern for keeping it cleaner.

test/server/cards/01_SOR/leaders/BobaFettCollectingTheBounty.spec.js
```diff
@@ -0,0 +1,108 @@
...
+
+            it('should ready 2 resources when Boba Fett completes an attack if an enemy unit left play this turn', function() {
+                this.player1.clickCard(this.wampa);
+                this.player2.passAction();
+                this.player1.clickCard(this.bobaFett);
+
+                expect(this.player1.countSpendableResources()).toBe(3);
+            });
+
+            it('should not ready resources if Boba Fett dies while attacking, even if an enemy unit left play this turn', function() {
```
AMMayberry1, line 98: Love this test! Pretty sure this is our first instance of "when completes an attack" so this is good test coverage for that feature. Glad to see it works out of the box

test/server/cards/01_SOR/leaders/BobaFettCollectingTheBounty.spec.js
```diff
@@ -0,0 +1,108 @@
...
+                this.player1.clickCard(this.rivalsFall);
+                this.player1.clickCard(this.viperProbeDroid);
+
+                expect(this.player1.countSpendableResources()).toBe(1);
+                expect(this.bobaFett.exhausted).toBeTrue();
+
+                // Case 2 - when returning an enemy unit to hand
+                reset();
+
+                this.player1.clickCard(this.cantinaBouncer);
```
AMMayberry1, line 38: Thanks for adding cantina bouncer for this test, gives me a lot more confidence that everything is working right under the hood

AMMayberry1 (commented):
Just leaving a note - please resolve previous comments when you submit changes if you don't mind, reduces the work on my side. Thanks!

AMMayberry1 (approved):

PR #108, opened 2024-10-09
Feature/cards

CheBato:
Added Boba Fett Disintegrator ability + tests.


server/game/cards/01_SOR/units/BobaFettDisintegrator.ts
```diff
@@ -0,0 +1,39 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '4156799805',
+            internalName: 'boba-fett#disintegrator'
+        };
+    }
+
+    // setup watcher
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar) {
+        this.cardsPlayedThisPhaseWatcher = AbilityHelper.stateWatchers.cardsPlayedThisPhase(registrar, this);
```
AMMayberry1, line 19: So as we discussed in the chat, there is a difference between "played" and "enters play", which is mostly important for leaders (leaders enter play but are not played).

The Boba effect is supposed to work the same for both leaders and units - i.e., if they entered play this phase (either by being deployed or being played from hand / smuggled), the Boba effect does not work on them. If they entered play the previous round, it does work.

So the first thing is that the `CardsPlayedThisPhaseWatcher` will not work for this card because it is only looking for played cards, not cards entering play. You will need to copy CardsPlayedThisPhaseWatcher.ts and create a new, slightly modified CardsEnteredPlayThisPhaseWatcher.ts that looks for `onUnitEntersPlay` instead of `onCardPlayed` in its `setupWatcher()` method.

The other issue is that currently there is a bug where the `onUnitEntersPlay` event is not firing correctly for leaders at all. Val is working on this in his PR. If he gets it merged soon enough then we can just merge it into yours, otherwise we can just comment out the leader tests for now (with a TODO comment explaining why) and do a new PR to update it once Val's PR is in.

server/game/cards/01_SOR/units/BobaFettDisintegrator.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'If this unit is attacking an exhausted unit that didn\'t enter play this round, deal 3 damage to the defender.',
+            immediateEffect: AbilityHelper.immediateEffects.conditional((attackContext) => ({
+                condition: () => {
+                    // we find the cards that the opponent played this phase and check if it's the card targeted
+                    const playedCardsByOpponentThisPhaseWithCriteria = this.cardsPlayedThisPhaseWatcher.getCardsPlayed((playedCardEntry) =>
```
AMMayberry1, line 28: This name is a little confusing (what is the criteria?), might be easier to understand if you just directly compute the boolean here i.e.: `const targetCardPlayedThisPhase = this.cardsPlayedThisPhaseWatcher.getCardsPlayed(...).length !== 0`

server/game/cards/01_SOR/units/BobaFettDisintegrator.ts
```diff
@@ -0,0 +1,39 @@
...
+                        playedCardEntry.playedBy === attackContext.source.activeAttack.target.owner && attackContext.source.activeAttack.target === playedCardEntry.card);
+                    return playedCardsByOpponentThisPhaseWithCriteria.length === 0 && attackContext.source.activeAttack.target.isNonLeaderUnit() && attackContext.source.activeAttack.target.exhausted;
```
AMMayberry1, line 30: As discussed in the comment above, should not be filtering out leader units

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
```diff
@@ -0,0 +1,119 @@
...
+                this.player1.clickCard(this.bobaFett);
+                this.player1.clickCard(this.lomPyke);
+
+                // check board state
+                expect(this.player2).toBeActivePlayer();
+                expect(this.lomPyke.damage).toBe(3);
+                expect(this.bobaFett.damage).toBe(4);
+            });
+
+            it('Should activate when attacking a exhausted unit not played in this phase.', function() {
```
AMMayberry1, line 45: Two things:

1. We can stitch all of these test cases together into one I think
2. For the test(s) where the Boba effect is expected to do damage, instead of targeting a unit that was put in place by the test (since that works a little weird) let's play unit from hand, then use `this.moveToNextActionPhase()`

server/game/cards/01_SOR/units/BobaFettDisintegrator.ts
CheBato: I added the CardsEnteredPlayedThisPhaseWatcher and updated the BobaFett implementation with it.

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
CheBato: When it comes to this comment i'm not quite sure what you mean. Do you mean to create 1 it() function with all the cases or multiple it() functions with one describe?. I'll fix number 2 when nr 1 is clarified.

AMMayberry1: Oh yeah sorry - it's the firs tthing you said, create 1 `it()` with multiple cases. It's because each `it()` takes a very long time to run (compared to other test frameworks I've used). Take a look at the first test case for General Krell to see what I mean: https://github.com/SWU-Karabast/forceteki/blob/main/test/server/cards/01_SOR/units/GeneralKrellHeartlessTactician.spec.js

CheBato: No worries, i'll rewrite the test cases. Thanks for the explanation.

server/game/stateWatchers/CardsEnteredPlayThisPhaseWatcher.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Card } from '../core/card/Card';
+
+export interface EnteredCardEntry {
+    card: EnteredPlayCard,
+    playedBy: Player
+}
+
+export type ICardsEnteredPlayThisPhase = EnteredCardEntry[];
+
+export class CardsEnteredPlayThisPhaseWatcher extends StateWatcher<EnteredCardEntry[]> {
```
AMMayberry1, line 16: Can you add a comment here noting that there's a known issue where this currently doesn't work with leaders

server/game/cards/01_SOR/units/BobaFettDisintegrator.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'If this unit is attacking an exhausted unit that didn\'t enter play this round, deal 3 damage to the defender.',
+            immediateEffect: AbilityHelper.immediateEffects.conditional((attackContext) => ({
+                condition: () => {
+                    // we find the cards that the opponent played this phase and check if it's the card targeted
+                    return this.cardsEnteredPlayThisPhaseWatcher.getCardsPlayed((playedCardEntry) =>
+                        playedCardEntry.playedBy === attackContext.source.activeAttack.target.owner && attackContext.source.activeAttack.target === playedCardEntry.card).length === 0 &&
```
AMMayberry1, line 29: Missed this before - since `activeAttack` is nullable we should probably do `attackContext.source.activeAttack?.target.owner` with the `?` here (same for other uses of the property in this statement).

It would conceivably get set to null if the attacker were defeated before this ability triggered, for example, b/c it gets forcibly set to null if the card is moved out of the arena.

server/game/cards/01_SOR/units/BobaFettDisintegrator.ts
```diff
@@ -0,0 +1,39 @@
...
+        this.cardsEnteredPlayThisPhaseWatcher = AbilityHelper.stateWatchers.cardsEnteredPlayThisPhase(registrar, this);
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'If this unit is attacking an exhausted unit that didn\'t enter play this round, deal 3 damage to the defender.',
+            immediateEffect: AbilityHelper.immediateEffects.conditional((attackContext) => ({
+                condition: () => {
+                    // we find the cards that the opponent played this phase and check if it's the card targeted
+                    return this.cardsEnteredPlayThisPhaseWatcher.getCardsPlayed((playedCardEntry) =>
```
AMMayberry1, line 28: If this is all one statement you can simplify the `{ return ...; }` format and just do `condition: () => this.cardsEnteredPlayThisWatcher...`

server/game/core/card/CardTypes.ts
```diff
@@ -57,6 +57,10 @@ export type PlayableCard =
     UpgradeCard |
     NonLeaderUnitCard;
 
+/** Type union for any card type that can enter play (with deploy or played) */
+export type EnteredPlayCard =
```
AMMayberry1, line 61: This concept already exists, it's `InPlayCard` which covers upgrades, units, and leaders.

server/game/stateWatchers/CardsEnteredPlayThisPhaseWatcher.ts
```diff
@@ -0,0 +1,53 @@
+import { StateWatcher } from '../core/stateWatcher/StateWatcher';
+import { StateWatcherName } from '../core/Constants';
+import { StateWatcherRegistrar } from '../core/stateWatcher/StateWatcherRegistrar';
+import Player from '../core/Player';
+import { EnteredPlayCard, PlayableCard } from '../core/card/CardTypes';
+import Game from '../core/Game';
+import { Card } from '../core/card/Card';
+
+export interface EnteredCardEntry {
+    card: EnteredPlayCard,
```
AMMayberry1, line 10: Should be `InPlayCard` (see above comment in CardTypes.ts)

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
```diff
@@ -0,0 +1,115 @@
...
+                        groundArena: ['boba-fett#disintegrator'],
+                    },
+                    player2: {
+                        leader: 'luke-skywalker#faithful-friend',
+                        hand: ['lom-pyke#dealer-in-truths', 'consular-security-force', 'entrenched'],
+                    }
+                });
+            });
+
+            it('while attacking an exhausted unit that didn\'t enter play this round, deal 3 damage to the defender.', function () {
```
AMMayberry1, line 18: nit, grammar: this currently reads as `Boba Fett Disintegrator's ability while attacking an exhausted unit that didn't enter play this round, deal 3 damage to the defender.`

Ideally would be something like `Boba Fett Disintegrator's ability, while attacking an exhausted unit that didn't enter play this round, should deal 3 damage to the defender`

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
```diff
@@ -0,0 +1,115 @@
+describe('Boba Fett Disintegrator', function() {
```
AMMayberry1, line 1: nit, consistency: comma between title and subtitle - `Boba Fett, Disintegrator`

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
```diff
@@ -0,0 +1,115 @@
...
+
+                this.player1.clickCard(this.bobaFett);
+                this.player1.clickCard(this.consularSecurityForce);
+
+                // check board state
+                expect(this.consularSecurityForce.damage).toBe(3);
+                expect(this.bobaFett.damage).toBe(3);
+
+                // reset board state
+                this.bobaFett.damage = 0;
```
AMMayberry1, line 36: fellarmir had a good idea for making this a little simpler by creating a reset method so we can have less code being repeated: https://github.com/SWU-Karabast/forceteki/blob/965dae3d78844c0cee7dfff102b6f76c623ad6c3/test/server/cards/01_SOR/leaders/BobaFettCollectingTheBounty.spec.js#L22

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
```diff
@@ -0,0 +1,115 @@
...
+                expect(this.consularSecurityForce.exhausted).toBe(true);
+
+                // reset state
+                this.consularSecurityForce.damage = 0;
+                this.bobaFett.exhausted = false;
+                this.bobaFett.damage = 0;
+                this.consularSecurityForce.exhausted = false;
+
+                // Case 6 Ability should activate when selecting vambrace-grappleshots ability before Boba Fetts.
+                this.player2.clickCard(this.entrenched);
```
AMMayberry1, line 103: Why is entrenched needed here? Could we just reset consular security force's damage to 0?

server/game/cards/01_SOR/units/BobaFettDisintegrator.ts
CheBato: added optional ? to activeAttack since it can be nullable in BobaFettDisintegrator.ts.

server/game/stateWatchers/CardsEnteredPlayThisPhaseWatcher.ts
CheBato: Changed EnteredPlayCard into InPlayCard.

server/game/cards/01_SOR/units/BobaFettDisintegrator.ts
CheBato: made it all into a one statement.

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
CheBato: changed grammar of test text.

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
CheBato: Added reset function to the tests to make it simpler.

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
CheBato: The reason why I used entrenched here is to see how much damage consularSecurityForce will receive. I wanted to see if the ability worked with vambrace-grappleshot, and since vambrace gives +2 power, I wouldn't be able to see if the damage is correct or not accurately.

test/server/cards/01_SOR/units/BobaFettDisintegrator.spec.js
CheBato: fixed title and subtitle.

AMMayberry1 (approved):

PR #110, opened 2024-10-09
Added the functionality of the game ending in a draw.

Ennis0:
Also implemented Sabine leader

### DIFF
--- a/server/game/cards/01_SOR/leaders/SabineWrenGalvanizedRevolutionary.ts
+++ b/server/game/cards/01_SOR/leaders/SabineWrenGalvanizedRevolutionary.ts
@@ -1,0 +1,36 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { LeaderUnitCard } from '../../../core/card/LeaderUnitCard';
+
+export default class SabineWrenGalvanizedRevolutionary extends LeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '4841169874',
+            internalName: 'sabine-wren#galvanized-revolutionary',
+        };
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addActionAbility({
+            title: 'Deal 1 damage to each base',
+            cost: AbilityHelper.costs.exhaustSelf(),
+            immediateEffect: AbilityHelper.immediateEffects.damage((context) => ({
+                amount: 1,
+                target: context.game.getPlayers().map((player) => player.base)
+            }))
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities() {
+        this.addOnAttackAbility({
+            title: 'Deal 1 damage to each enemy base',
+            immediateEffect: AbilityHelper.immediateEffects.damage((context) => ({
+                amount: 1,
+                target: context.game.getPlayers()
+                    .filter((player) => player !== context.player)
+                    .map((player) => player.base)
+            }))
+        });
+    }
+}
+
+SabineWrenGalvanizedRevolutionary.implemented = true;
--- a/server/game/cards/01_SOR/leaders/SabineWrenGalvanizedRevolutionary.ts
+++ b/server/game/cards/01_SOR/leaders/SabineWrenGalvanizedRevolutionary.ts
@@ -15,7 +15,7 @@ export default class SabineWrenGalvanizedRevolutionary extends LeaderUnitCard {
             cost: AbilityHelper.costs.exhaustSelf(),
             immediateEffect: AbilityHelper.immediateEffects.damage((context) => ({
                 amount: 1,
-                target: context.game.getPlayers().map((player) => player.base)
+                target: [context.player.base, context.player.opponent.base]
             }))
         });
     }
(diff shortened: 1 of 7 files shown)


server/game/cards/01_SOR/leaders/SabineWrenGalvanizedRevolutionary.ts
```diff
@@ -0,0 +1,36 @@
...
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities() {
+        this.addOnAttackAbility({
+            title: 'Deal 1 damage to each enemy base',
+            immediateEffect: AbilityHelper.immediateEffects.damage((context) => ({
+                amount: 1,
+                target: context.game.getPlayers()
+                    .filter((player) => player !== context.player)
```
AMMayberry1, line 29: For simplicity, maybe just `target: context.player.opponent.base`? We've been making the assumption that it's only two players pretty much everywhere so might as well leverage it here

server/game/core/Game.js
```diff
@@ -443,40 +443,47 @@ class Game extends EventEmitter {
...
-    //         if (player.honor >= honorRequiredToWin) {
-    //             this.recordWinner(player, 'honor');
-    //         } else if (player.opponent && player.opponent.honor <= 0) {
-    //             this.recordWinner(player, 'dishonor');
-    //         }
-    //     }
-    // }
+    checkWinCondition() {
+        let losingPlayers = [];
+        for (const player of this.getPlayers()) {
```
AMMayberry1, line 450: Bit simpler as `losingPlayers = this.getPlayers().filter((player) => player.base.damage >= player.base.getHp())`

server/game/core/Game.js
```diff
@@ -443,40 +443,47 @@ class Game extends EventEmitter {
...
+            if (player.base.damage >= player.base.getHp()) {
+                losingPlayers.push(player);
+            }
+        }
+        if (losingPlayers.length === 1) {
+            this.endGame(losingPlayers[0].opponent, 'base destroyed');
+        } else if (losingPlayers.length === 2) { // draw game
+            this.endGame(null, 'both bases destroyed');
+        }
```
AMMayberry1, line 459: I would recommend that we add a final `else` with `Contract.fail` just to prevent any chance of a weird fallthrough case due to a bug or something

test/server/cards/01_SOR/leaders/SabineWrenGalvanizedRevolutionary.spec.js
```diff
@@ -0,0 +1,44 @@
...
+                    player1: {
+                        leader: { card: 'sabine-wren#galvanized-revolutionary', deployed: true }
+                    },
+                    player2: {
+                        groundArena: ['rebel-pathfinder']
+                    }
+                });
+            });
+
+            it('should deal 1 damage to the opponent\'s base on attack', function () {
```
AMMayberry1, line 35: These are great! Only additional test I can think of would be this for a timing test: enemy base at 29 damage, Sabine attacks something, then confirm that when the end game prompt happens she has done 0 damage to the defender (i.e., game ended before her attack resolved).

If we wanted to be extra about it, could also add an additional "on attack" ability with like vambrace grappleshot or something to confirm that game end happens when the player chooses to resolve the Sabine ability. But I'd be fine without that as well, it's not strictly covering anything new.

test/server/core/Game.spec.js
```diff
@@ -0,0 +1,45 @@
...
+                    },
+                    player2: {
+                        base: { card: 'administrators-tower', damage: 29 }
+                    }
+                });
+            });
+
+            it('should cause that player to lose the game', function () {
+                this.player1.clickCard(this.rebelPathfinder);
+                expect(this.player1).toHavePrompt('Game Won');
```
AMMayberry1, line 40: Can we find a way to check which player won, and do that check for both players' prompts?

server/game/cards/01_SOR/leaders/SabineWrenGalvanizedRevolutionary.ts
Ennis0: I figured it wouldn't hurt to make it multiplayer compatible ahead of time, but if you say so I'll change it.

server/game/core/Game.js
Ennis0: A final else block would be executed every resolveGameState() in which no one wins.

test/server/core/Game.spec.js
Ennis0: Right now the winner is printed in the prompt's menuTitle. I'm not really sure what menuTitle and promptTitle will correspond to in the actual UI, so: should I also add the winner's name to the promptTitle or should I check the menuTitle in the test somehow?

AMMayberry1: I think `toHavePrompt` will pass if it matches either promptTitle or menuTitle

server/game/core/Game.js
AMMayberry1: Ah right 😅

AMMayberry1 (approved):

PR #111, opened 2024-10-10
Claim Initiative, Test Typings, and Baze Malbus Implementation

CDMayberry:
## Initiative claiming functional
claiming initiative now properly flags the player as "passed", simply had it call pass after it finishes, and the button to claim initiative is removed for the rest of the phase. Added an event window on claiming initiative, but nothing specifically tests it yet.

## server and test build changes
I reorganized the build process somewhat. The server build is now output to ./build/server by default, and test builds to ./build/test. server and test build independently of each other but test is still aware of server. JS/TS files under ./tests can now reference '../server' instead of '../build' as the folders are in the same relative position, which provides better typing support and more intuitive require(...) paths.

There is a now a npm script `npm run test-fast` that can be used the same as `npm test`. This will only rebuild the test folder rather than both the server and test folders, unless it detects that the server hasn't been built yet, at which point it just does a full rebuild. Might make re-running a single test file a little faster but shouldn't be used if you're also changing server code.

`npm test` can build both the server and test folders in parallel now.

Added a sub-folder for custom npm scripts, ./scripts. These just are for doing npm commands with a little more complexity. Not absolutely required, but felt it was handy to have.

the new eslint configs messed with building TS files in the test folder, so I kinda just hit it with a hammer until it stopped complaining. Will probably need a refactor from someone who has a better understanding of it.

## Test changes
hasInitiative can now be flagged in the setupTest options player object. This will properly setup the given player has having the initiative at the start of the game.

While digging into that, I also changed a few arrays calls, some functions were sorting the array's in-place which was throwing off some testing. Changed all calls to use .concat() first to create an array copy to then sort, and separately fixed a check or two to properly sort.

### General typing improvements
I've added an IntegrationsHelper.d.ts file. Trying to convert the JS file to TS wasn't worth the hassle due to heavy usage of `this` and other side-effects caused by integration. Custom Matchers have been added so they should now show up in intellisense for tests when using `exact()`. There are typings for the options object and the some of the game objects, but require using a new test context object to get typing/intellisense for.

### limited TS support and test context
With the test folder structure changes, we can now create TS specs. However, it's not feasible to make `this` type safe as-is (better to just rewrite in vitest). As a workaround it's now possible to pass a contextRef in with the integration function call ala `integration(function(contextRef) { ... })`. This has a `context` object which contains the exact same fields that are accessible via `this` with some additional typings, using reference sorcery. 

Instead of calling `this.setupTest({ ... })`, you can call `contextRef.setupTest({ ... })`. In a given `it(...)` block, you can destructure the contextRef to get the context specific to that test, `const { context } = contextRef`, and use context as if you were using `this`. This gives much stronger typing and intellisense support. 

contextRef is entirely optional, and `this` inside of tests functions exactly the same as before.

NOTE: This doesn't take into account any cards defined under the player objects, so the context has a general purpose "any" accessor. No type safety but at the very least it's no worse than before.

See the specs for examples.

## Oh right, Implemented Baze Malbus, Temple Guardian
Dude's implemented.


test/helpers/GameFlowWrapper.js
```diff
@@ -32,13 +32,11 @@ class GameFlowWrapper {
     }
 
     get firstPlayer() {
```
AMMayberry1, line 34: I wonder if we can just remove this, there really isn't a similar concept in swu that's relevant and afaik the method isn't used. There's "acting player" and "player who currently controls initiative," but since initiative can change hands during the round it doesn't carry quite same meaning as "first player."

Iirc the term is actually used once in the rules but it's to determine who gets to choose initiative at the very beginning of the game

test/helpers/GameFlowWrapper.js
```diff
@@ -50,7 +48,7 @@ class GameFlowWrapper {
             return player.hasPrompt('Waiting for opponent to take an action or pass') ? 1 : 0;
         };
 
-        var playersInPromptedOrder = this.allPlayers.sort((playerA, playerB) =>
+        var playersInPromptedOrder = this.allPlayers.concat().sort((playerA, playerB) =>
```
AMMayberry1, line 51: Good catch on needing the copies. Are there any alternative shorthands for doing it besides `concat()` that might be a little clearer? The `concat()` by itself is pretty unintuitive for folks who don't know about the behavior, if I saw this in the code previously I probably would've assumed it was a bug. Not a hard block, just curious if there are easy alternatives

test/helpers/GameFlowWrapper.js
```diff
@@ -74,15 +72,16 @@ class GameFlowWrapper {
...
+        this.allPlayersInInitiativeOrder().forEach((player) => player.clickPrompt('No'));
     }
 
 
     /**
      * Skips setup phase with defaults
      */
     skipSetupPhase() {
-        this.selectInitiativePlayer(this.player1);
+        //console.log('Setting Initiative to ' + this.firstPlayer.player.id + ', default: ' + this.game.initiativePlayer?.id)
```
AMMayberry1, line 83: Left in by accident?

test/helpers/IntegrationHelper.d.ts
```diff
@@ -0,0 +1,87 @@
+type BaseCard = import('../../server/game/core/card/BaseCard').BaseCard;
```
AMMayberry1, line 1: Just curious, what's the `.d.ts` mean?

test/server/cards/01_SOR/units/BazeMalbusTempleGuardian.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                    const { context } = contextRef;
+                    
+                    expect(context.player2).toBeActivePlayer();
+                    context.player2.clickCard(context.wampa);
+                    context.player2.clickCard(context.p1Base);
+                    expect(context.p1Base.damage).toBe(4);
+                    expect(context.player1).toBeActivePlayer();
+                });
+
+                it('should give it sentinel if the player claims initiative', function () {
```
AMMayberry1, line 28: So because the jasmine tests are so heavy and slow and we have finite github actions time, I'm asking people to consolidate test cases as much as they can to reduce usage. Not good test practice but I don't want to run things down before we (hopefully) migrate to vitest and things aren't hella slow.

So in this case I would probably combine this test and the above into one case, with comments marking each one

test/server/cards/01_SOR/units/BazeMalbusTempleGuardian.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                });
+
+                it('should give it sentinel', function () {
+                    const { context } = contextRef;
+
+                    expect(context.player1).toBeActivePlayer();
+                    context.player1.passAction();
+                    context.player2.clickCard(context.battlefieldMarine);
+                    
+                    // Baze Malbus automatically selected due to sentinel
```
AMMayberry1, line 68: Can we extend this case to confirm that he loses sentinel if the other player claims initiative

test/helpers/GameFlowWrapper.js
CDMayberry: alternative is [...this.allPlayers] to destructure the array into another array.

test/helpers/GameFlowWrapper.js
CDMayberry: Yeah can remove

test/helpers/IntegrationHelper.d.ts
CDMayberry: definition file, no code, only typings. It's primarily used to pair with js files rather than turn them into TS files.

test/server/cards/01_SOR/units/BazeMalbusTempleGuardian.spec.ts
CDMayberry: Fair, I can consolidate

test/helpers/GameFlowWrapper.js
CDMayberry: Makes sense, it's basically a shortcut for initiative player anyway.

server/game/cards/01_SOR/units/BazeMalbusTempleGuardian.ts
```diff
@@ -0,0 +1,23 @@
...
+            id: '5879557998',
+            internalName: 'baze-malbus#temple-guardian'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addConstantAbility({
+            title: 'While you have the initiative, this unit gains Sentinel',
+            condition: (context) => context.source.controller.hasInitiative(),
+            matchTarget: (card, context) => card === context.source,
```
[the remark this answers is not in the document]
CDMayberry: Gotcha

test/helpers/DeckBuilder.js
```diff
@@ -88,13 +88,13 @@ class DeckBuilder {
 
     getNamedCardsInPlayerEntry(playerEntry) {
         let namedCards = [];
-        if (typeof playerEntry === 'number' || typeof playerEntry == null) {
+        if (typeof playerEntry === 'number' || typeof playerEntry === 'boolean' || playerEntry === null) {
```
AMMayberry1, line 91: Now that I'm looking at it, these should probably just throw TestSetupError

test/helpers/IntegrationHelper.js
```diff
@@ -278,6 +277,37 @@ var customMatchers = {
             }
         };
     },
+    toHaveClaimInitiativeAbilityButton: function (util, customEqualityMatchers) {
```
AMMayberry1, line 280: We need to add this to the wiki once merged, lmk if you're down to do it or else I'm happy to. Ditto for toHaveInitiative and player.claimInitiative().

test/server/cards/01_SOR/units/BazeMalbusTempleGuardian.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+                    context.player2.clickCard(context.battlefieldMarine);     
+                                   
+                    // Baze Malbus automatically selected due to sentinel
+                    expect(context.battlefieldMarine.damage).toBe(2);
+                    expect(context.bazeMalbus.damage).toBe(3);
+
+                    context.player1.clickCard(context.bazeMalbus);
+                    context.player1.clickCard(context.p2Base);
+                    
+                    expect(context.player2).toHaveClaimInitiativeAbilityButton();
```
AMMayberry1, line 68: Just for readability it would be helpful to add comment descriptors by case, since we have to make these so long - see GeneralKrellHeartlessTactician.spec.js for an example (doesn't have to look exactly like that, but just convey similar information would be great for folks trying to read the test for the first time in case someone breaks it)

package.json
```diff
@@ -1,45 +1,50 @@
...
-    "typescript-eslint": "^8.0.1",
-    "underscore": "^1.13.6",
-    "uuid": "^10.0.0",
-    "zod": "^3.23.8"
-  },
-  "devDependencies": {
-    "@types/node": "^20.14.10"
-  }
-}
+    "type": "commonjs",
```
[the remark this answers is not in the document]
CDMayberry: Not sure, I did add a package for copying files around so it perhaps shifted something. ESLint also started auto-formatting some of the files when I open/save them, so it might be the culprit.

test/helpers/DeckBuilder.js
CDMayberry: I almost started to do that but the function calling getNamedCardsInPlayerEntry is looping through the entire Player Options object and catching any setting, this was to filter out things w/o breaking. 

I can work on getting that calling function to use only specific fields rather than the entire object, then we can make this throw an error.

test/helpers/IntegrationHelper.js
CDMayberry: Ah right, I can add it once it's merged.

test/server/cards/01_SOR/units/BazeMalbusTempleGuardian.spec.ts
CDMayberry: Added comments for each case within a test.

test/helpers/DeckBuilder.js
CDMayberry: Ah right, so number specifically is because some fields can pass it to "pad" the particular area. Changed it around slightly but should handle these better now.

CDMayberry:
@AMMayberry1 Smacked the linter hard enough, seems to be working now. Got the requested changes in, lmk if anything else crops up.

AMMayberry1 (approved):

test/helpers/DeckBuilder.js
AMMayberry1: Ahhh forgot there was a legit case for that, my bad. Much easier to read now either way, thanks for handling that!

PR #112, opened 2024-10-11
Add "distribute healing / damage among targets" prompt and test cards

AMMayberry1:
Added logic to set up a prompt for abilities that ask the player to distribute X healing or damage across targets. Created new systems for distributing that will create the prompt and then create individual instances of healing and damage for each target.

Implemented Palpatine unit, Overwhelming Barrage, and Midnight Repairs as test cases.


adbronico (dismissed):
Never tell me the odds!

test/server/cards/01_SOR/events/OverwhelmingBarrage.spec.js
```diff
@@ -0,0 +1,102 @@
...
+                expect(this.battlefieldMarine.damage).toBe(2);
+                expect(this.tielnFighter).toBeInLocation('discard');
+                expect(this.bobaFett.damage).toBe(1);
+
+                // attack into wampa to confirm stats buff
+                this.atst.damage = 0;
+                this.player2.clickCard(this.atst);
+                this.player2.clickCard(this.wampa);
+                expect(this.wampa).toBeInLocation('ground arena');
+                expect(this.wampa.damage).toBe(6);
```
Ennis0, line 45: AT-ST will do 7 damage here when Boba Fett, Daimyo is fully implemented.

test/server/cards/01_SOR/events/OverwhelmingBarrage.spec.js
```diff
@@ -0,0 +1,102 @@
...
+                this.player1.clickCard(this.wampa);
+                expect(this.player1).toBeAbleToSelectExactly([this.battlefieldMarine, this.jynErso, this.atst, this.tielnFighter, this.bobaFett]);
+                this.player1.clickPrompt('Choose no targets');
+
+                expect(this.jynErso.damage).toBe(0);
+                expect(this.wampa.damage).toBe(0);
+                expect(this.atst.damage).toBe(0);
+                expect(this.wampa.damage).toBe(0);
+                expect(this.tielnFighter.damage).toBe(0);
+                expect(this.bobaFett.damage).toBe(0);
```
Ennis0, line 77: Is there any value in checking that the wampa still gets the buff even if the second part of the ability is skipped?

test/server/cards/01_SOR/events/OverwhelmingBarrage.spec.js
```diff
@@ -0,0 +1,102 @@
...
+                expect(this.player2).toBeActivePlayer();
+            });
```
Ennis0, line 79: Should probably make sure the player can't assign less than the full damage in total(comp rules 8.5.5).

test/server/cards/01_SOR/events/OverwhelmingBarrage.spec.js
AMMayberry1: Oh lol right I keep forgetting he has an ability

test/server/cards/01_SOR/events/OverwhelmingBarrage.spec.js
AMMayberry1: Right now I'm expecting that it will be the responsibility of the FE to enforce that the amount falls in the legal range. I do have checks in the prompt that confirm the amount is in the valid range, they throw an exception if so.

PR #115, opened 2024-10-13
Damage when drawing from empty deck

Ennis0:
Converted TargetResolver classes to typescript and made a base class for them with their shared code.
Implemented a PlayerTargetResolver to automate letting players choose a player for effects.
Implemented the game mechanic of players taking damage to their base when they would draw from an empty deck.


server/game/TargetInterfaces.ts
```diff
@@ -2,27 +2,37 @@ import type { AbilityContext } from './core/ability/AbilityContext';
...
+import { RelativePlayer, TargetMode, LocationFilter, CardTypeFilter } from './core/Constants';
+import { PlayerTargetSystem } from './core/gameSystem/PlayerTargetSystem';
 
 // allow block comments without spaces so we can have compact jsdoc descriptions in this file
 /* eslint @stylistic/lines-around-comment: off */
 
 // ********************************************** EXPORTED TYPES **********************************************
 export type ITriggeredAbilityTargetResolver<TContext extends TriggeredAbilityContext = TriggeredAbilityContext> =
-    | (ICardTargetResolver<TContext> & ITriggeredAbilityCardTargetResolver<TContext>)
+    | (ICardTargetResolver<TContext>)
```
AMMayberry1, line 13: Can remove the parens

server/game/core/ability/PlayerOrCardAbility.js
```diff
@@ -103,12 +103,11 @@ class PlayerOrCardAbility {
     }
 
     buildTargetResolver(name, properties) {
-        if (properties.mode === TargetMode.Select) {
-            return new SelectTargetResolver(name, properties, this);
-        } else if (properties.mode === TargetMode.Ability) {
-            return new AbilityTargetResolver(name, properties, this);
+        switch (properties.mode) {
+            case TargetMode.Select: return new SelectTargetResolver(name, properties, this);
```
AMMayberry1, line 107: I've never seen one-line case statements like this before - did the linter force it? If not I'd recommend splitting them into two lines just for style consistency, I find it kind of hard to read like this

server/game/core/ability/abilityTargets/TargetResolverBaseClass.ts
```diff
@@ -0,0 +1,55 @@
...
+import { AbilityContext } from '../AbilityContext';
+import * as Contract from '../../utils/Contract';
+import { GameSystem } from '../../gameSystem/GameSystem';
+import PlayerOrCardAbility from '../PlayerOrCardAbility';
+import { RelativePlayer } from '../../Constants';
+
+/**
+ * Base class for all target resolvers.
+ */
+export abstract class TargetResolverBaseClass<TProps extends ITargetResolverBase<AbilityContext>> {
```
AMMayberry1, line 11: Typically we don't include "Class" in the name of a class, but in SWU it's always weird because "Base" is a type of card so it's ambiguous. Maybe the cleanest way is just to put "TargetResolver" and let the `abstract` make it clear that it's a base class.

server/game/core/ability/abilityTargets/TargetResolverBaseClass.ts
```diff
@@ -0,0 +1,55 @@
...
+    protected dependentTarget = null;
+    protected dependentCost = null;
+    public constructor(protected name: string, protected properties: TProps, ability: PlayerOrCardAbility) {
```
AMMayberry1, line 14: nit: space between properties and constructor

server/game/core/ability/abilityTargets/TargetResolverBaseClass.ts
```diff
@@ -0,0 +1,55 @@
...
+        if (this.properties.dependsOn) {
+            const dependsOnTarget = ability.targetResolvers.find((target) => target.name === this.properties.dependsOn);
+
+            // assert that the target we depend on actually exists
+            Contract.assertNotNullLike(dependsOnTarget);
+
+            dependsOnTarget.dependentTarget = this;
```
AMMayberry1, line 21: Hmm, this might be a bug in the engine - since a target can have multiple dependent targets this probably ought to be an array. That would involve a number of other changes though, so I would put a guard here that doesn't use `Contract` but maybe instead just uses an `if` statement that throws an error saying "Multiple dependent targets not implemented yet."

I'm _pretty_ sure no card we have implemented yet would hit that error, but if so it would be good to know now and we can figure out where to go from there.

server/game/core/ability/abilityTargets/TargetResolverBaseClass.ts
```diff
@@ -0,0 +1,55 @@
...
+        return !!this.properties.dependsOn || this.hasLegalTarget(context);
+    }
+
+    protected abstract hasLegalTarget(context): boolean;
+
+    protected getGameSystems(context: AbilityContext): GameSystem | GameSystem[] {
+        return this.properties.immediateEffect ? [this.properties.immediateEffect] : [];
+    }
+
+
```
AMMayberry1, line 36: nit: inconsistent number of newlines between methods

server/game/core/ability/abilityTargets/TargetResolverBaseClass.ts
```diff
@@ -0,0 +1,55 @@
...
+    protected abstract getAllLegalTargets(context: AbilityContext): any[];
```
AMMayberry1, line 37: This is a personal thing but typically I put the abstract methods at the top right under the constructor so that they're easy to find for anyone implementing a subclass. Won't block on it, just a suggestion.

server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -45,38 +40,29 @@ class CardTargetResolver {
         return CardSelector.for(Object.assign({}, properties, { cardCondition: cardCondition, targets: true }));
     }
 
-    getContextCopy(card, context) {
-        let contextCopy = context.copy();
+    private getContextCopy(card: Card, context: AbilityContext) {
+        const contextCopy = context.copy({});
```
AMMayberry1, line 44: Let's just add a default value to the `copy()` method so we don't have to do this

server/game/core/ability/abilityTargets/PlayerTargetResolver.ts
```diff
@@ -0,0 +1,83 @@
...
+    public constructor(name: string, properties: IPlayerTargetResolver<AbilityContext>, ability: PlayerOrCardAbility) {
+        super(name, properties, ability);
+
+        if (this.properties.immediateEffect) {
+            this.properties.immediateEffect.setDefaultTargetFn((context) => context.targets[name]);
+        }
+    }
+
+    protected override hasLegalTarget(context: any): boolean {
+        // I don't think there are cases of player-targeting abilities that are sometimes invalid, but I could be wrong.
```
AMMayberry1, line 18: We need to check here because it depends on what gamesystems are involved. E.g., if there was some card that said "chosen player cannot draw cards for this phase" and this targetResolver was wrapping the "DrawCardSystem", this check could fail (assuming that player is the only legal target due to some filter). Same for `getAllLegalTargets`.

One thing to consider would be making a stripped-down version of `BaseCardSelector` that manages player targets. That would also be useful because we could eventually extend it with additional options like "canChooseMultiple" so that maybe Yoda would have checkboxes instead of an awkward set of buttons like it does now.

That could be a TODO though, but definitely we need some checks here either way.

server/game/core/ability/abilityTargets/PlayerTargetResolver.ts
```diff
@@ -0,0 +1,83 @@
...
+            return;
+        }
+
+        const player = context.choosingPlayerOverride || this.getChoosingPlayer(context);
+        if (player === context.player.opponent && context.stage === Stage.PreTarget) {
+            targetResults.delayTargeting = this;
+            return;
+        }
+        const promptTitle = this.properties.activePromptTitle || 'Choose a player';
+        const choices = ['You', 'Opponent'];
```
AMMayberry1, line 45: Similar comment here - I think based on the way that the CardTargetResolver behaves, people would expect to have a little more control over this using filters / properties and more automation determining if a given player is a legal target or not. We need the behavior between the two to be as consistent as possible.

This is a probably large and definitely non-urgent task so feel free to freeze this part and just do the DrawSystem updates if you prefer. That said, definitely it would improve some cards in the long run. Up to you.

server/game/core/ability/abilityTargets/PlayerTargetResolver.ts
```diff
@@ -0,0 +1,83 @@
...
+            (chosenPlayer) => {
+                return () => {
+                    context.targets[this.name] = chosenPlayer;
+                    if (this.name === 'target') {
+                        context.target = chosenPlayer;
+                    }
+                };
+            }
+        );
+        if (player !== context.player.opponent && context.stage === Stage.PreTarget) {
```
AMMayberry1, line 56: If the logic for this method ends up being so similar to the corresponding method in CardTargetResolver after making other changes, then definitely we should refactor the common logic into a `resolve()` method in the base class and have specific abstract methods it calls that handles the specifics that are relevant to player vs card targeting.

If we end up making a `PlayerSelector` that functions similar enough to `BaseCardSelector` then that would make it even more straightforward

test/server/gameSystems/DrawSystem.spec.js
```diff
@@ -0,0 +1,50 @@
...
+            it('over the amount in their deck, they should draw the remainder and take damage equal to thrice the difference', function () {
+                this.player1.clickCard(this.missionBriefing);
+                this.player1.clickPrompt('You');
+                expect(this.player1.base.damage).toBe(3);
+                expect(this.player1.hand.length).toBe(2);
+                expect(this.wampa).toBeInLocation('hand');
+                expect(this.player1.deck.length).toBe(0);
+            });
+
+            it('over the amount in their deck, they should draw the remainder and take damage equal to thrice the difference', function () {
```
AMMayberry1, line 26: Since the test case name is what gets emitted when the test fails, need to make sure every case has a distinct name so that people can reliably find it in case of failure. Probably just putting "controller" and "opponent" in the names here would be enough

server/game/gameSystems/DrawSystem.ts
```diff
@@ -38,5 +39,23 @@ export class DrawSystem<TContext extends AbilityContext = AbilityContext> extend
...
         event.amount = amount;
+        // If there are not enough cards left in deck to draw the specified amount, instead draw the remainder and take damage to base equal to thrice the difference.
+        event.checkCondition = () => {
+            if (event.cancelled || event.resolved || event.name === EventName.Unnamed) {
+                return;
+            }
+            if (!event.condition(this)) {
+                event.cancel();
+            }
+            if (event.amount > event.player.drawDeck.length) {
```
AMMayberry1, line 50: Will discuss in chat, but I think this logic needs to be in `eventHandler`

test/server/gameSystems/DrawSystem.spec.js
```diff
@@ -0,0 +1,50 @@
+describe('Drawing cards', function() {
+    integration(function() {
+        describe('When a player draws cards', function() {
```
AMMayberry1, line 3: I want to start moving tests towards using the new scheme that Matchstick added since it allows autocomplete. I'll be doing a big PR soon to refactor the existing tests as well. Please take a look at the Baze Malbus test to see the changes, but it boils down to:

1. Switch from spec.js to spec.ts
2. Change to `integration(function(contextRef)` at the top
3. Add `const { context } = contextRef;` in each test case
4. Use `contextRef` instead of `this` everywhere

server/game/core/ability/abilityTargets/TargetResolverBaseClass.ts
Ennis0: Why should we not use a Contract? I put one in becuase I'm not sure how to throw an error otherwise and nothing tripped it.

server/game/gameSystems/DrawSystem.ts
Ennis0: So how exactly should the damage be done? I'd think it needs to be a damage event, rather than just directly adding damage in DrawSystem.eventHandler, for consistency and futureproofing(potential triggers/replacement effects for damage on base) right? Should it be a subevent? A new event window?

server/game/core/ability/abilityTargets/TargetResolverBaseClass.ts
AMMayberry1: In my mind they mean slightly different things. Contract failure means "you have put the game in an illegal state and have a bug to fix." Whereas this would be saying "there is a missing feature we need that needs to be added, and you just happened to run into it."

For a while I had a similar kind of thing in TriggeredAbilityWindow because we didn't have the full logic for resolving multiple triggers. In this case it would look like:

```typescript
if (dependsOnTarget.dependentTarget != null) {
    throw new Error("Having multiple dependent targets for the same source target has not yet been implemented");
}

dependsOnTarget.dependentTarget = this;
```

AMMayberry1: To elaborate on why the library is called "Contract" - in my work we refer to those kinds of checks as "contract checks," meaning that it is checking that the software contracts - e.g. between, two objects that are passing data back and forth - have not been violated.

server/game/gameSystems/DrawSystem.ts
AMMayberry1: So, looking at it, I think it might actually be a replacement effect? It uses the word "instead" which is one of the words indicating replacement.

My first thought would then be to use the replacement mechanism to check if the draw event will overshoot the number of cards left in deck, then if so, optionally create a new draw event with the right draw number (if there are any cards at all to draw) and a damage event.

Probably the easiest way to create both is to just have the `replacementImmediateEffect` parameter of `ReplacementEffectSystem` be a `simultaneous()` with the two events. I _think_ that would work but we've never tried it before so there might be some hiccups. We may need to add a new solution for how to have replacementEffect create multiple events.

AMMayberry1: Actually, I thought about this a little more - I guess replacement effect isn't the right way to do it since that kind of implies that there is an outside effect that is altering the normal resolution. But creating the damage events is part of the normal resolution of draw. So then I think it's probably a contingent event.

AMMayberry1: Apologies for all the back-and-forth on this, more complicated than I was expecting. Thanks again for tackling it.

Ennis0: No worries. I considered several different implementations myself before I got to where I submitted the PR. Just checking: does making it a contingent event imply, as far as the game engine is concerned, that the original draw event happened successfully?

server/game/core/ability/abilityTargets/PlayerTargetResolver.ts
Ennis0: I put a big TODO at the top of the class to create a PlayerSelector that can determine target legality, and marked a couple functions in the class as placeholder for now.

server/game/gameSystems/DrawSystem.ts
AMMayberry1: Sort of - it's more the reverse, the game engine won't create the contingent event until it's sure the source event will resolve.

Overwhelm is a good example - the overwhelm event is created as a contingent event off the attacker damage, but only after the replacement effect step because shield might cancel the attacker event.

Ennis0: Right. I'm concerned(probably unnecessarily) that having a false draw event(no cards actually drawn) resolve could be bad in the future if there's such a thing as triggering on draw. For now though I've coded the contingent event implementation and tests are passing.

server/game/core/ability/abilityTargets/PlayerTargetResolver.ts
```diff
@@ -0,0 +1,83 @@
+import { TargetResolver } from './TargetResolver';
+import { IPlayerTargetResolver } from '../../../TargetInterfaces';
+import { AbilityContext } from '../AbilityContext';
+import type Player from '../../Player';
+import PlayerOrCardAbility from '../PlayerOrCardAbility';
+import { Stage } from '../../Constants';
+
+// This currently assumes that every player will always be a legal target for any effect it's given.
+// TODO: Make a PlayerSelector class(see the use of property "selector" in CardTargetResolver) to help determine target legality. Use it to replace placeholder override functions below.
```
AMMayberry1, line 9: I'm not 100% opposed to checking the class in like this, since it is providing some useful functionality and some of the missing pieces (specifically the legality checks) are not relevant given the current cards in the game.

My concern however is that people will try to use this in the future thinking that it has the same level of functionality as `CardTargetSelector` (understandably), which will cause a lot of frustration / debugging effort while they figure out that's not the case.

Unless you're planning to address these in the very near future (i.e. in your next PR or two), I think there are a few extra things we should do in this PR to maximize the utility of this and make it less likely that people will get tripped up:

- Add a similar TODO comment on IPlayerTargetResolver explaining that the functionality is incomplete, and one on `AbilityRestriction` in Constants.ts explaining that player restrictions are not currently respected by PlayerTargetResolver
- I think we should add enough functionality to be able to support existing cases like Yoda - i.e., the ability to also select "both" and "none." That seems like such an obvious feature that I doubt there's any amount of documenting we could do that would cause people to not expect it to be there. It should be very easy to add flags for those via `IPlayerTargetProperties` though, and then we can add it to Yoda in this PR to document the functionality.

server/game/core/ability/abilityTargets/TargetResolver.ts
```diff
@@ -0,0 +1,72 @@
...
+    protected abstract checkTarget(context: AbilityContext): boolean;
+
+    protected abstract hasTargetsChosenByInitiatingPlayer(context: AbilityContext): boolean;
+
+    protected canResolve(context) {
+        // if this depends on another target, that will check hasLegalTarget already
+        return !!this.properties.dependsOn || this.hasLegalTarget(context);
+    }
+
+    protected resolve(context: AbilityContext, targetResults, passPrompt = null) {
```
AMMayberry1, line 43: This is a good start! One more big suggestion I have.

Typically when you provide a method like this in a base class, it's doing one of two things:

1. It provides some basic default functionality which is enough for at least some of the derived classes to use without modifying / overriding
2. It makes calls into abstract methods that the derived classes are required to provide implementations for, thus each derived class can specialize its functionality to their own case

This pattern you have here of requiring derived classes to overload but also call super() is a little tricky because it's not super clear that that's the intent and it's also not enforceable by the transpiler the same way that `abstract` is.

An easy way to fix this via (2) above is to, instead of overriding and calling `super.resolve()`, add a new abstract method named `resolveInternal` or `resolveImpl` or something and call that at the end of this method, and then the child classes override that.

I do think though you could go further with that pattern, there is still a lot of code shared between the `resolve()` methods of the derived classes that could be pulled in here. The 'pay costs first' button, the waiting prompt, etc are all  in common.

It might require making a couple more abstract methods so that different pieces of the pipeline can be implemented by subclasses - e.g., you might need like `abstract buildChoices()`, `abstract buildHandlers()`, stuff like that. But at the end of the day it would make the whole set of them much easier to understand and maintain since we could modify all common behaviors in the base class instead of needing to duplicate changes across all of them.

test/server/gameSystems/DrawSystem.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+                context.player1.clickCard(context.missionBriefing);
+                context.player1.clickPrompt('You');
+                expect(context.player1.base.damage).toBe(3);
+                expect(context.player1.hand.length).toBe(2);
+                expect(context.wampa).toBeInLocation('hand');
+                expect(context.player1.deck.length).toBe(0);
+            });
+
+            it('over the amount in their deck due to and during an opponent\'s action, they should draw the remainder and take damage equal to thrice the difference', function () {
```
AMMayberry1, line 28: Can we add a timing test where the damage ends the game, and confirm that any drawn cards are in hand at the time of game end

server/game/gameSystems/DrawSystem.ts
AMMayberry1: This is a valid concern (if I understand right) and I think I have an answer for this, messaged in discord to continue the discussion

server/game/core/cardSelector/BaseCardSelector.js
```diff
@@ -2,7 +2,7 @@ const { Location, RelativePlayer, WildcardLocation } = require('../Constants');
 const Contract = require('../utils/Contract');
 const EnumHelpers = require('../utils/EnumHelpers');
 
-// TODO: once converted to TS, make this abstract
+// TODO: once converted to TS, make this abstract. Also, move the code in CardSelector.js to here.
```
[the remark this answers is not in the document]
Ennis0: That makes sense.

server/game/core/ability/abilityTargets/TargetResolver.ts
Ennis0: Thank you for the guidance. It feels obvious now but I don't think I can remember doing a refactor like this(making a base class and pulling out shared code) in recent memory so I didn't really think of making new functions.

server/game/core/ability/abilityTargets/TargetResolver.ts
```diff
@@ -3,44 +3,60 @@ import { AbilityContext } from '../AbilityContext';
...
+
     public constructor(protected name: string, protected properties: TProps, ability: PlayerOrCardAbility) {
         if (this.properties.dependsOn) {
             const dependsOnTarget = ability.targetResolvers.find((target) => target.name === this.properties.dependsOn);
 
             // assert that the target we depend on actually exists
             Contract.assertNotNullLike(dependsOnTarget);
+            // TODO: Change the dependsOn system to allow multiple dependent targets.
+            if (dependsOnTarget.dependentTarget != null) {
+                throw new Error('Attempting to set a dependent target for source target <insert target name here> but it already has one. Having multiple dependent targets for the same source target has not yet been implemented');
```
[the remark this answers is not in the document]
Ennis0: Totally missed that. Thank you.

server/game/core/ability/PlayerOrCardAbility.js
```diff
@@ -104,9 +104,12 @@ class PlayerOrCardAbility {
...
     buildTargetResolver(name, properties) {
         switch (properties.mode) {
-            case TargetMode.Select: return new SelectTargetResolver(name, properties, this);
-            case TargetMode.Player: return new PlayerTargetResolver(name, properties, this);
-            default: return new CardTargetResolver(name, properties, this);
+            case TargetMode.Select:
+                return new SelectTargetResolver(name, properties, this);
+            case TargetMode.Player:
+                return new PlayerTargetResolver(name, properties, this);
+            default:
```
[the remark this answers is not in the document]
Ennis0: I tried this, and it looks like target selectors default to TargetMode.Single, but assignment of that default happens in the constructor for CardTargetResolver, which calls this.GetSelector, which calls CardSelector.for, which calls CardSelector.getDefaultedProperties. So in this function mode can in fact be unassigned, and there's a ton of cards that use that defaulting behavior.

server/game/core/ability/abilityTargets/TargetResolver.ts
AMMayberry1: Of course! Yeah I don't know that we've done it much in this repo since most methods are pretty small and self-contained.

server/game/core/ability/PlayerOrCardAbility.js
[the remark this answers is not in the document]
AMMayberry1: Ahh okay. Yeah looking at it now, TargetMode is weird because is implicitly mixes card targeting options with player / select targeting options. I guess it's fine to have this is as the default then, might be good to leave a comment on it though

server/game/core/ability/PlayerOrCardAbility.js
```diff
@@ -107,9 +107,20 @@ class PlayerOrCardAbility {
...
-            default:
+            case TargetMode.AutoSingle:
+            case TargetMode.Exactly:
+            case TargetMode.ExactlyVariable:
+            case TargetMode.MaxStat:
+            case TargetMode.Single:
+            case TargetMode.Unlimited:
+            case TargetMode.UpTo:
+            case TargetMode.UpToVariable:
+            case undefined: // CardTargetResolver contains behavior that defaults the mode to TargetMode.Single if it is not defined yet.
```
AMMayberry1, line 120: Need to add an explicit case here for null as well because switch uses `===` comparison (as opposed to `==`) so null and undefined are different

server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -37,7 +37,7 @@ export class CardTargetResolver extends TargetResolver<ICardTargetResolver<Abili
               (properties.immediateEffect == null || properties.immediateEffect.hasLegalTarget(contextCopy) &&
                 (!properties.cardCondition || properties.cardCondition(card, contextCopy)));
         };
-        return CardSelector.for(Object.assign({}, properties, { cardCondition: cardCondition, targets: true }));
+        return CardSelectorFactory.for(Object.assign({}, properties, { cardCondition: cardCondition, targets: true }));
```
AMMayberry1, line 40: nit: this is just a common programming convention, but if we're renaming this to "factory" then typically the corresponding method would be called "create" (it's kind of the de-facto method name used with the factory pattern)

server/game/core/ability/abilityTargets/PlayerTargetResolver.ts
```diff
@@ -26,58 +28,65 @@ export class PlayerTargetResolver extends TargetResolver<IPlayerTargetResolver<A
...
+                    return () => {
+                        context.targets[this.name] = chosenPlayer;
+                        if (this.name === 'target') {
+                            context.target = chosenPlayer;
+                        }
+                    };
+                }
+            );
+
+        /* if (player !== context.player.opponent && context.stage === Stage.PreTarget) {
```
AMMayberry1, line 81: Just leaving a reminder to remove this if we can't sort out the issue in chat discussion

server/game/core/gameSteps/prompts/HandlerMenuMultipleSelectionPrompt.js
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * The properties option object may contain the following:
+ * choices            - an array of titles for menu buttons
+ * maxSelected        - the maximum number of choices the player is allowed to submit. Leave undefined to allow the player to submit any number.
+ * handler            - a handler function that is called with the results of the prompt(which choices were selected) after the user clicks 'Done'
+ * activePromptTitle  - the title that should be used in the prompt for the choosing player.
+ * waitingPromptTitle - the title to display for opponents.
+ * source             - what is at the origin of the user prompt, usually a card; used to provide a default waitingPromptTitle, if missing
+ */
+class HandlerMenuMultipleSelectionPrompt extends HandlerMenuPrompt {
```
AMMayberry1, line 16: This is great!! Very useful, nice work putting it together so fast.

One note for long-term is that we may eventually remove some of this logic and let the FE manage it as like checkboxes or something, in which case we'll just do what we do with e.g. Overwhelming Barrage where we provide the context and then just accept a chunk of data back from the FE indicating the user selection (as opposed to providing buttons and managing click-by-click in the BE). But for first release we want to minimize complexity in the FE so this is perfect for what we need for launch.

test/server/cards/01_SOR/units/YodaOldMaster.spec.js
```diff
@@ -20,9 +20,11 @@ describe('Yoda, Old Master', function() {
                 this.player1.clickCard(this.yoda);
                 this.player1.clickCard(this.rogueSquadronSkirmisher);
                 expect(this.yoda).toBeInLocation('discard');
-                expect(this.player1).toHaveEnabledPromptButtons(['You', 'Opponent', 'You and Opponent', 'No one']);
+                expect(this.player1).toHaveEnabledPromptButtons(['You', 'Opponent', 'Done']);
```
AMMayberry1, line 23: Soooo much cleaner, thanks a ton for taking care of that!!

test/server/gameSystems/DrawSystem.spec.ts
```diff
@@ -51,6 +51,19 @@ describe('Drawing cards', function() {
...
+                const { context } = contextRef;
+
+                context.player2.setBaseStatus({ card: context.p2Base.internalName, damage: 27 });
+                context.player1.clickCard(context.missionBriefing);
+                context.player1.clickPrompt('Opponent');
+
+                expect(context.p2Base.damage).toBe(30);
+                expect(context.player1).toHavePrompt('player1 has won the game!');
+                expect(context.player2).toHavePrompt('player1 has won the game!');
+                expect(context.player2.hand.length).toBe(1);
```
AMMayberry1, line 65: Just thought of this - for this test and the other end game timing tests you did, could real quick add a check confirming that the player who took the game-ending action is still the active player? That will confirm that the action didn't finish resolving

AMMayberry1 (approved):

PR #116, opened 2024-10-14
Add cost adjusters

AMMayberry1:
Brought the jigoku concept of "cost reducers" forward as "cost adjusters." These are ongoing effects used for adjusting the play cost of cards.

Refactored constant abilities to live on the base Card class since they are used for events to reduce their own cost.

Refactored LastingCardEffect to be PlayerCardEffect, in line with the behavior the class in jigoku.

Implemented some basic cases using Del Meeko, Bendu, Force Choke.


server/game/Interfaces.ts
```diff
@@ -30,6 +34,19 @@ export type IActionAbilityProps<TSource extends Card = Card> = Exclude<IAbilityP
     phase?: PhaseName | 'any';
 };
 
+export interface IOngoingEffectProps {
+    targetLocationFilter?: Location | Location[];
+    canChangeZoneOnce?: boolean;
+    canChangeZoneNTimes?: number;
```
Ennis0, line 40: Do these need to be two properties? Also, where exactly will they be used? I can sort of imagine an effect granting a when defeated needing to persist into the discard; is that the kind of thing it's for?

server/game/core/Player.js
```diff
@@ -674,69 +657,20 @@ class Player extends GameObject {
             adjuster.canAdjust(playingType, card, target, ignoreType, penaltyAspect)
         );
         var costIncreases = matchingAdjusters
-            .filter((a) => a.getAmount(card, this) < 0)
-            .reduce((cost, adjuster) => cost - adjuster.getAmount(card, this), 0);
+            .filter((adjuster) => adjuster.direction === CostAdjustDirection.Increase)
```
Ennis0, line 660: It seems a bit silly to call out the adjustment direction with its own enum rather than just using positive and negative numbers. Do the rules treat increases and decreases differently or something?

server/game/core/Player.js
```diff
@@ -674,69 +657,20 @@ class Player extends GameObject {
...
         var costDecreases = matchingAdjusters
-            .filter((a) => a.getAmount(card, this) > 0)
+            .filter((adjuster) => adjuster.direction === CostAdjustDirection.Decrease)
             .reduce((cost, adjuster) => cost + adjuster.getAmount(card, this), 0);
 
         baseCost += costIncreases;
         var reducedCost = baseCost - costDecreases;
 
-        var costFloor = Math.min(baseCost, Math.max(...matchingAdjusters.map((a) => a.costFloor)));
+        // TODO: not 100% sure what the use case for this line is
```
Ennis0, line 669: I'm guessing it's for effects that say "reduce cost to a minimum of X". None of those in SWU, though of course there is always a minimum of 0.

server/game/core/Player.js
Ennis0, line 660: What's the advantage to using an enum to track adjustment direction rather than checking positive/negative?

server/game/core/Player.js
```diff
@@ -674,69 +657,20 @@ class Player extends GameObject {
...
-            .reduce((cost, adjuster) => cost - adjuster.getAmount(card, this), 0);
+            .filter((adjuster) => adjuster.direction === CostAdjustDirection.Increase)
+            .reduce((cost, adjuster) => cost + adjuster.getAmount(card, this), 0);
         var costDecreases = matchingAdjusters
-            .filter((a) => a.getAmount(card, this) > 0)
+            .filter((adjuster) => adjuster.direction === CostAdjustDirection.Decrease)
             .reduce((cost, adjuster) => cost + adjuster.getAmount(card, this), 0);
 
         baseCost += costIncreases;
         var reducedCost = baseCost - costDecreases;
```
Ennis0, line 667: reducedCost makes more sense as adjustedCost I believe.

server/game/core/card/EventCard.ts
```diff
@@ -63,4 +77,25 @@ export class EventCard extends EventCardParent {
         properties.cardName = this.title;
         this._eventAbility = new EventAbility(this.game, this, properties);
     }
+
+    /** Add a constant ability on the event card that decreases its cost under the given condition */
+    protected addDecreaseCostAbility(properties: IDecreaseEventCostAbilityProps<this>): void {
```
Ennis0, line 82: Jabba's Rancor has the same self cost reduction type effect as Force Choke. Should this function be on a more generic card class?

server/game/cards/01_SOR/events/ForceChoke.ts
```diff
@@ -0,0 +1,34 @@
...
+            id: '1446471743',
+            internalName: 'force-choke',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addDecreaseCostAbility({
+            title: 'If you control a Force unit, this costs 1 resource less to play',
+            amount: 1,
+            condition: (context) => context.source.controller.isTraitInPlay(Trait.Force)
```
adbronico, line 17: Does isTraitInPlay only check units, or also upgrades? I ask becasue there is at least 1 card in Set 3 (Padawan Starfighter) that looks for Force units OR upgrades, so we might need to add in a list of card types to the method.

server/game/core/Constants.ts
```diff
@@ -134,6 +134,8 @@ export enum CardType {
 export enum WildcardCardType {
     Any = 'any',
     NonLeaderUnit = 'nonLeaderUnit',
+    /** Any card type that can be played from hand */
+    Playable = 'playable',
```
adbronico, line 138: Would a card played from Smuggle or Discard also count for these?

adbronico (changes requested):
See comments

server/game/core/Constants.ts
AMMayberry1: It's really just a shorthand for the list: event, unit, upgrade (excluding tokens). So the type of play card action doesn't come to bear. I can update the comment to be clearer though

server/game/core/card/EventCard.ts
AMMayberry1: I wrote it this way because I didn't want to give event cards access to generic constant abilities, since this is the only type of constant ability an event card can have (currently). But it might make sense to make it a more generic helper that's available to more cards.

server/game/core/Player.js
AMMayberry1: Good callout, looking at it now I think I'm over-engineering and we can just let amount be positive or negative.

server/game/Interfaces.ts
AMMayberry1: Oh yeah these are legacy, I'll either add a comment or remove them

adbronico (approved):
*wookiee noises*

PR #118, opened 2024-10-16
Feature/engine - Scry Mechanic

CheBato:
- Added the Scry GameSystem.
- Added R2D2 + tests.
- Added Inferno Four + tests.


test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+                        hand: ['inferno-four#unforgetting'],
+                        deck: ['foundling', 'pyke-sentinel', 'atst', 'cartel-spacer', 'wampa'],
+                    },
+                    player2: {
+                        spaceArena: ['tie-advanced'],
+                    }
+                });
+            });
+
+            it('While playing/defeating lets you look at the top 2 cards of the deck and decide whether to put either them on the bottom or top of deck in any order.', function () {
```
AMMayberry1, line 17: nit: test names shouldn't start with a capital letter since it will end up in the middle of a sentence: `Inferno Four - Unforgetting's ability While playing/defeating...`

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+                const { context } = contextRef;
+                let preSwapDeck = context.player1.deck;
+
+                // Case 1 on play move top card to bottom
+                context.player1.clickCard(context.infernoFour);
+                context.player1.clickPrompt('pyke-sentinel-bottom');
+                context.player1.clickPrompt('foundling-bottom');
+
+
```
AMMayberry1, line 26: nit: excess whitespace in several places

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+                });
+            });
+
+            it('While playing/defeating lets you look at the top 2 cards of the deck and decide whether to put either them on the bottom or top of deck in any order.', function () {
+                const { context } = contextRef;
+                let preSwapDeck = context.player1.deck;
+
+                // Case 1 on play move top card to bottom
+                context.player1.clickCard(context.infernoFour);
+                context.player1.clickPrompt('pyke-sentinel-bottom');
```
AMMayberry1, line 23: At each stage we should use `expect().toHaveExactPromptButtons()` to confirm the set of available prompt buttons. Also, at least for this test we should confirm that cards haven't moved until all selections are done.

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+                expect(context.player1.deck[3]).toEqual(preSwapDeck[1]);
+                expect(context.player1.deck[4]).toEqual(preSwapDeck[0]);
+                expect(context.player2).toBeActivePlayer();
+
+
+                // restart state
+                preSwapDeck = context.player1.deck;
+
+                // Case 2 on attack move to top
+                context.player2.clickCard(context.tieAdvanced);
```
AMMayberry1, line 42: For this test could we instead try splitting top / bottom to make sure that works (or add a third test if you prefer)

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+                // Check board state
+                // preswap deck deck: ['atst', 'cartel-spacer', 'wampa', 'pyke-sentinel', 'foundling']
+                // expected after deck: ['cartel-spacer', 'atst', 'wampa', 'pyke-sentinel', 'foundling']
+                expect(context.player1.deck.length).toBe(5);
+                expect(context.player1.deck[0]).toEqual(preSwapDeck[1]);
+                expect(context.player1.deck[1]).toEqual(preSwapDeck[0]);
+                expect(context.player1).toBeActivePlayer();
+            });
+        });
```
AMMayberry1, line 55: Need to add a test of what happens if there is one card in deck or zero cards in deck. Thinking on it now, probably we need to have logic for the one-card case that just does the "look at" and then skips the button altogether since the result is the same whatever they push

test/server/cards/01_SOR/units/R2D2IgnoringProtocol.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+                // Case 2 on attack move to top
+                context.player1.clickCard(context.r2d2);
+                context.player1.clickCard(context.battlefieldMarine);
+                context.player1.clickPrompt('pyke-sentinel-top');
+                // Check board state
+                expect(context.player1.deck.length).toBe(5);
+                expect(context.player1.deck).toEqual(preSwapDeck);
+                expect(context.player2).toBeActivePlayer();
+            });
+        });
```
AMMayberry1, line 46: Would also be good to have the empty deck case here

test/server/cards/01_SOR/events/ShootFirst.spec.ts
```diff
@@ -1,8 +1,6 @@
 describe('Shoot First', function () {
     integration(function (contextRef) {
         describe('Shoot First Ability', function () {
-            const { context } = contextRef;
```
AMMayberry1, line 4: Thanks!

server/game/gameSystems/ScrySystem.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        // @ts-ignore
+        const cardHandler = (player, card) => {
+            decision_mapper = decision_mapper.filter((a) => a !== card + '-top' && a !== card + '-bottom');
+            if (decision_mapper.length > 0) {
+                context.game.promptWithHandlerMenu(player, {
+                    activePromptTitle: 'Select card to move to the top or bottom of the deck',
+                    context: context,
+                    choices: decision_mapper,
+                    choiceHandler: (choice) => {
```
AMMayberry1, line 42: There's another option here worth considering, which is using `handlers` instead of `choiceHandler`. The way handlers works is that you provide an array of choices (which you already have) and an array of handlers that is the same length. Then if choice index `i` is chosen, handler index `i` is called.

It might take a couple tries to get the indexing right but it would decouple the button names from the implementation, which would be really nice if we ever need to change them

server/game/gameSystems/ScrySystem.ts
```diff
@@ -0,0 +1,84 @@
...
+import { CardType, Location, WildcardCardType } from '../core/Constants';
+import { CardTargetSystem, type ICardTargetSystemProperties } from '../core/gameSystem/CardTargetSystem';
+import { GameEvent } from '../core/event/GameEvent';
+import { moveCard } from './GameSystemLibrary';
+
+export interface IScryProperties extends ICardTargetSystemProperties {
+    amount: number;
+}
+
+export class ScrySystem<TContext extends AbilityContext = AbilityContext> extends CardTargetSystem<TContext, IScryProperties> {
```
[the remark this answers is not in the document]
adbronico: On one hand I agree, on the other Scry is a pretty ubiquitous term.

server/game/cards/01_SOR/units/InfernoFourUnforgetting.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Look at the top 2 cards of your deck. Put any number of them on the bottom of your deck and the rest on top in any order.',
+            immediateEffect: AbilityHelper.immediateEffects.scry({
+                amount: 2,
+            })
+        });
+        this.addWhenDefeatedAbility({
```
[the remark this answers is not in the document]
CheBato: changed to when with addTriggeredAbility

server/game/cards/01_SOR/units/InfernoFourUnforgetting.ts
```diff
@@ -0,0 +1,29 @@
...
+            id: '9133080458',
+            internalName: 'inferno-four#unforgetting'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Look at the top 2 cards of your deck. Put any number of them on the bottom of your deck and the rest on top in any order.',
+            immediateEffect: AbilityHelper.immediateEffects.scry({
+                amount: 2,
```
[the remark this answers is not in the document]
CheBato: they have become one liners!

server/game/gameSystems/ScrySystem.ts
```diff
@@ -0,0 +1,84 @@
...
+    };
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public override eventHandler(event): void { }
+
+    public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext): void {
+        const player = context.player;
+        const { amount } = this.generatePropertiesFromContext(context) as IScryProperties;
+        const deckLength = player.drawDeck.length;
+        const actual_amount = amount === -1 ? deckLength : (amount > deckLength ? deckLength : amount);
```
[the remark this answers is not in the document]
CheBato: no usecase whatsover

server/game/gameSystems/ScrySystem.ts
```diff
@@ -0,0 +1,84 @@
...
+    protected override defaultProperties: IScryProperties = {
+        amount: 1,
+    };
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public override eventHandler(event): void { }
+
+    public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext): void {
+        const player = context.player;
+        const { amount } = this.generatePropertiesFromContext(context) as IScryProperties;
```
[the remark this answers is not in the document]
CheBato: you mean the ILookMoverDeckCardsTopOrBottomProperties amirite :D anyways removed it.

server/game/gameSystems/ScrySystem.ts
```diff
@@ -0,0 +1,84 @@
...
+        const actual_amount = amount === -1 ? deckLength : (amount > deckLength ? deckLength : amount);
+        const cards = player.drawDeck.slice(0, actual_amount);
+        let decision_mapper = cards.map((card: Card) => [
+            card.internalName + '-top',
+            card.internalName + '-bottom',
+        ]).reduce((acc, val) => acc.concat(val), []);
+
+        // @ts-ignore
+        const cardHandler = (player, card) => {
+            decision_mapper = decision_mapper.filter((a) => a !== card + '-top' && a !== card + '-bottom');
```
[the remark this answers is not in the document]
CheBato: fixed hopefully :D

server/game/gameSystems/ScrySystem.ts
```diff
@@ -0,0 +1,84 @@
...
+    public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext): void {
+        const player = context.player;
+        const { amount } = this.generatePropertiesFromContext(context) as IScryProperties;
+        const deckLength = player.drawDeck.length;
+        const actual_amount = amount === -1 ? deckLength : (amount > deckLength ? deckLength : amount);
+        const cards = player.drawDeck.slice(0, actual_amount);
+        let decision_mapper = cards.map((card: Card) => [
+            card.internalName + '-top',
+            card.internalName + '-bottom',
+        ]).reduce((acc, val) => acc.concat(val), []);
```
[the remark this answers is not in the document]
CheBato: changed the reduce to .flat()

server/game/gameSystems/ScrySystem.ts
```diff
@@ -0,0 +1,84 @@
...
+            } else {
+                return true;
+            }
+        };
+        context.game.promptWithHandlerMenu(player, {
+            activePromptTitle: 'Select card to move to the top or bottom of the deck',
+            context: context,
+            choices: decision_mapper,
+            choiceHandler: (choice) => {
+                const position = choice.split('-')[choice.split('-').length - 1];
```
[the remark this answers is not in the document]
CheBato: got rid of the first instance of promptWithHandlerMenu and called the cardHandler function

server/game/gameSystems/ScrySystem.ts
```diff
@@ -0,0 +1,84 @@
...
+        const deckLength = player.drawDeck.length;
+        const actual_amount = amount === -1 ? deckLength : (amount > deckLength ? deckLength : amount);
+        const cards = player.drawDeck.slice(0, actual_amount);
+        let decision_mapper = cards.map((card: Card) => [
+            card.internalName + '-top',
+            card.internalName + '-bottom',
+        ]).reduce((acc, val) => acc.concat(val), []);
+
+        // @ts-ignore
+        const cardHandler = (player, card) => {
```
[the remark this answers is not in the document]
CheBato: renamed cardHandler to choiceHandler

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+            });
+
+            it('While playing/defeating lets you look at the top 2 cards of the deck and decide whether to put either them on the bottom or top of deck in any order.', function () {
+                const { context } = contextRef;
+                let preSwapDeck = context.player1.deck;
+
+                // Case 1 on play move top card to bottom
+                context.player1.clickCard(context.infernoFour);
+                context.player1.clickPrompt('pyke-sentinel-bottom');
+                context.player1.clickPrompt('foundling-bottom');
```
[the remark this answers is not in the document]
CheBato: restructured the prompts

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
CheBato: changed to lowercase

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
CheBato: removed excess whitespace in several places

server/game/gameSystems/ScrySystem.ts
CheBato: transformed from choiceHandler to handlers

test/server/cards/01_SOR/units/R2D2IgnoringProtocol.spec.ts
CheBato: added the empty deck case to the R2D2 Ignoring protocol spec

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
CheBato: added logic for look at.

server/game/core/Constants.ts
```diff
@@ -194,6 +194,7 @@ export enum EventName {
     OnRegroupPhaseReadyCards = 'onRegroupPhaseReadyCards',
     OnRoundEnded = 'onRoundEnded',
     OnRoundEndedCleanup = 'onRoundEndedCleanup',
+    OnScry = 'onScry',
```
AMMayberry1, line 197: Missed this in the rename

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -138,6 +139,11 @@ export function lookAt<TContext extends AbilityContext = AbilityContext>(propert
     return new LookAtSystem<TContext>(propertyFactory);
 }
 
+export function LookMoveDeckCardsTopOrBottom<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<ILookMoveDeckCardsTopOrBottomProperties, TContext>): CardTargetSystem<TContext> {
+    return new LookMoveDeckCardsTopOrBottomSystem<TContext>(propertyFactory);
+}
+
+
```
AMMayberry1, line 146: nit: too much whitespace

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export interface ILookMoveDeckCardsTopOrBottomProperties extends ICardTargetSystemProperties {
+    amount: number;
+}
+
+export class LookMoveDeckCardsTopOrBottomSystem<TContext extends AbilityContext = AbilityContext> extends CardTargetSystem<TContext, ILookMoveDeckCardsTopOrBottomProperties> {
+    public override readonly name = 'LookMoveDeckCardsTopOrBottomSystem';
+
+    protected override defaultProperties: ILookMoveDeckCardsTopOrBottomProperties = {
+        amount: 1,
```
AMMayberry1, line 16: Right now the property `amount` is required, so there is no point providing a default because the caller always has to give it a value. We should either change it to `amount?: number` in the properties or remove this default.

The decision depends on whether we think it is useful to make the parameter optional and just do 1 card by default. I would rather be explicit in this case since I don't think there's a default number that is common, but it's up to you.

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+    };
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public override eventHandler(event): void { }
+
+    public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext): void {
+        const player = context.player;
+        const { amount } = this.generatePropertiesFromContext(context);
+        const deckLength = player.drawDeck.length;
+        if (deckLength === 1) {
```
AMMayberry1, line 26: nit: whitespace before this if block starts

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+            events.push(lookAt({
```
AMMayberry1, line 27: We want to try and avoid importing from `GameSystemLibrary` in any GameSystem files because it creates import loops that could cause issues eventually. Let's remove the `lookAt()` import and instead use `new LookAtSystem()`, same for moveCard.

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+                target: player.drawDeck[0],
+                sendChatMessage: true,
+            }).generateEvent(context.target, context));
+        } else {
+            const actual_amount = amount > deckLength ? deckLength : amount;
```
AMMayberry1, line 32: Can just use `Math.min(amount, deckLength)`

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+
+            // @ts-ignore
+            const choiceHandler = (player) => {
+                if (cards.length > 0) {
+                    context.game.promptWithHandlerMenu(player, {
+                        activePromptTitle: 'Select card to move to the top or bottom of the deck',
+                        choices: cards.map((card: Card) => [
+                            'Put ' + card.internalName + ' on top',
+                            'Put ' + card.internalName + ' to bottom',
+                        ]).flat(),
```
AMMayberry1, line 43: This is a bit hard to read, might be easier if you set some of these complicated properties up as variables first and then attach them here

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+            let cards = player.drawDeck.slice(0, actual_amount);
+
+            // @ts-ignore
+            const choiceHandler = (player) => {
+                if (cards.length > 0) {
+                    context.game.promptWithHandlerMenu(player, {
+                        activePromptTitle: 'Select card to move to the top or bottom of the deck',
+                        choices: cards.map((card: Card) => [
+                            'Put ' + card.internalName + ' on top',
+                            'Put ' + card.internalName + ' to bottom',
```
AMMayberry1, line 42: nit: _on_ bottom

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+                                target: card,
+                                bottom: false,
+                                destination: Location.Deck
+                            }).generateEvent(context.target, context));
+                            // get rid of the card from cards
+                            cards = cards.filter((a) => a !== card);
+                            choiceHandler(context.player);
+                        }),
+                        (() => {
+                            events.push(moveCard({
```
AMMayberry1, line 55: Since both the on-top and on-bottom forms of this are so similar, would be good to find a way to reduce the duplicated code. Maybe create a private method that does this and accepts the boolean flag for either top or bottom, then these two handlers can just be like `() => this.pushMoveEvent(bottom: boolean, ...)`

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+        if (deckLength === 1) {
+            events.push(lookAt({
+                target: player.drawDeck[0],
+                sendChatMessage: true,
+            }).generateEvent(context.target, context));
+        } else {
+            const actual_amount = amount > deckLength ? deckLength : amount;
+            let cards = player.drawDeck.slice(0, actual_amount);
+
+            // @ts-ignore
```
AMMayberry1, line 35: What's this for?

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+            const choiceHandler = (player) => {
```
AMMayberry1, line 36: Would be good to add a couple lines of explainer comments here since it's kind of complex. The important thing to communicate is that it recursively removes handlers from the list as it's called.

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+
+            choiceHandler(context.player);
+        }
+    }
+
+    public override getEffectMessage(context: TContext): [string, any[]] {
+        const properties = this.generatePropertiesFromContext(context);
+        const message =
+            properties.amount > 0
+                ? `look at the top ${properties.amount === 1 ? 'card' : `${properties.amount} cards`} of your deck. ${properties.amount === 1 ? 'You may put it on the bottom of your deck' : 'Put any number of them on the bottom of your deck and the rest on top in any order'}` : '';
```
AMMayberry1, line 78: Can you break this into more lines, still very long

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
AMMayberry1: Still uppercase looks like?

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+            const actual_amount = amount > deckLength ? deckLength : amount;
+            let cards = player.drawDeck.slice(0, actual_amount);
+
+            // @ts-ignore
+            const choiceHandler = (player) => {
+                if (cards.length > 0) {
+                    context.game.promptWithHandlerMenu(player, {
+                        activePromptTitle: 'Select card to move to the top or bottom of the deck',
+                        choices: cards.map((card: Card) => [
+                            'Put ' + card.internalName + ' on top',
```
AMMayberry1, line 41: Internal name is mostly for logging and testing and shouldn't be used in anything that the player will see. Use title instead.

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+                        spaceArena: ['tie-advanced'],
+                    }
+                });
+            });
+
+            it('While playing/defeating lets you look at the top 2 cards of the deck and decide whether to put either them on the bottom or top of deck in any order.', function () {
+                const { context } = contextRef;
+                let preSwapDeck = context.player1.deck;
+
+                // Case 1 on play move top card to bottom
```
AMMayberry1, line 21: This comment is incorrect now right? Ditto comment @line 32

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+                expect(context.player1.deck.length).toBe(5);
+
+                // preswap deck: ['foundling', 'pyke-sentinel', 'atst', 'cartel-spacer', 'wampa']
+                // expected after deck: ['atst', 'cartel-spacer', 'wampa', 'pyke-sentinel', 'foundling']
+                expect(context.player1.deck[0]).toEqual(preSwapDeck[1]);
+                expect(context.player1.deck[1]).toEqual(preSwapDeck[2]);
+                expect(context.player1.deck[4]).toEqual(preSwapDeck[0]);
+                expect(context.player2).toBeActivePlayer();
+
+                // restart state
```
AMMayberry1, line 38: nit: restart is a bit misleading here, maybe "record new state." Ditto R2 test.

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+                    phase: 'action',
+                    player1: {
+                        hand: ['inferno-four#unforgetting'],
+                        deck: ['foundling'],
+                    },
+                    player2: {}
+                });
+            });
+
+            it('While playing should only show card and put it back on top of deck since the deck size is 1', function () {
```
AMMayberry1, line 70: For this and the empty deck case in R2, can you confirm that base damage is 0 - need to make sure that we didn't accidentally trigger the rule that damages base when you attempt to draw from an empty deck.

server/game/core/Constants.ts
CheBato: changed it to onLookMoveDeckCardsTopOrBottomSystem

server/game/gameSystems/GameSystemLibrary.ts
CheBato: removed white space

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: remove the default since we always need to specify how many.

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: adde whitespace before if block

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: changed InternalName into title

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
CheBato: changed restart state into record new state

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
CheBato: you are right seems like i missed it then...my mistake

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: Yessir

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: changed to -> on

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: to draw the amount of cards

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: broke into more lines.

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: made it into a variable and then sent to the choices.

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: changed to new LookAtSystem() and new MoveCardSystem

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: added helper method for pushing move events

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: Added two lines for explaination.

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
CheBato: fixed comments to explain more about the cases.

test/server/cards/01_SOR/units/InfernoFourUnforgetting.spec.ts
CheBato: added both expects to the tests.

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
AMMayberry1: Ah sorry, I specifically meant the `// @ts-ignore` comment

CheBato: Ah, it gave out the error that not all routes in the function return a boolean because we recursively call the function.

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -23,59 +20,67 @@ export class LookMoveDeckCardsTopOrBottomSystem<TContext extends AbilityContext
...
+                // setup the choices for each card top and bottom
+                const handlerChoices = cards.map((card: Card) => [
+                    'Put ' + card.title + ' on top',
+                    'Put ' + card.title + ' on bottom',
+                ]).flat();
 
-            choiceHandler(context.player);
+                context.game.promptWithHandlerMenu(player, {
+                    activePromptTitle: 'Select card to move to the top or bottom of the deck',
+                    choices: handlerChoices,
```
AMMayberry1, line 49: pro tip: if you rename `handlerChoices` to be `choices`, you can change this line to just be `choices` (it's syntactic sugar for `choices: choices`)

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -23,59 +20,67 @@ export class LookMoveDeckCardsTopOrBottomSystem<TContext extends AbilityContext
...
+                    handlers: cards.map((card: Card) => [(() => {
+                        this.pushMoveEvent(false, card, cards, events, context, choiceHandler);
+                    }),
+                    (() => {
```
AMMayberry1, line 53: Since each call is just one statement now you can remove the `{ }` blocks and effectively just do `() => this.pushMoveEvent(...)`

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -23,59 +20,67 @@ export class LookMoveDeckCardsTopOrBottomSystem<TContext extends AbilityContext
...
+                        this.pushMoveEvent(true, card, cards, events, context, choiceHandler);
+                    })]).flat()
+                });
+            };
+            choiceHandler(context.player, cards);
```
AMMayberry1, line 58: nit: add whitespace above just to break this apart from the big statement above visually

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -23,59 +20,67 @@ export class LookMoveDeckCardsTopOrBottomSystem<TContext extends AbilityContext
...
+    private pushMoveEvent(bottom: boolean, card: Card, cards: any[], events: GameEvent[], context: TContext,
+        choiceHandler: (player: any, cards: any[]) => boolean) {
+        // create a new card event
+        const moveCardEvent = new MoveCardSystem({
+            target: card,
+            bottom: bottom,
+            destination: Location.Deck
+        }).generateEvent(context.target, context);
+        events.push(moveCardEvent);
+        // get rid of the card from cards
```
AMMayberry1, line 72: nit: whitespace in here to break these blocks up a bit. generally if it's big enough to require a comment it should get its own code block

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -23,59 +20,67 @@ export class LookMoveDeckCardsTopOrBottomSystem<TContext extends AbilityContext
...
+                    })]).flat()
+                });
+            };
+            choiceHandler(context.player, cards);
         }
     }
 
+    // Helper method for pushing the move card event into the events array.
+    private pushMoveEvent(bottom: boolean, card: Card, cards: any[], events: GameEvent[], context: TContext,
+        choiceHandler: (player: any, cards: any[]) => boolean) {
```
AMMayberry1, line 64: nit: when the param lists get long we typically just do one per line. So like:

```typescript
    private pushMoveEvent(
        bottom: boolean, 
        card: Card, 
        cards: any[], 
        events: GameEvent[], 
        context: TContext,
        choiceHandler: (player: any, cards: any[]) => boolean
    ) {
```

server/game/core/Constants.ts
```diff
@@ -194,7 +194,7 @@ export enum EventName {
     OnRegroupPhaseReadyCards = 'onRegroupPhaseReadyCards',
     OnRoundEnded = 'onRoundEnded',
     OnRoundEndedCleanup = 'onRoundEndedCleanup',
-    OnScry = 'onScry',
+    OnLookMoveDeckCardsTopOrBottomSystem = 'onLookMoveDeckCardsTopOrBottomSystem',
```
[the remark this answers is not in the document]
CheBato: got rid of the System and hide in shame.

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
```diff
@@ -23,59 +20,67 @@ export class LookMoveDeckCardsTopOrBottomSystem<TContext extends AbilityContext
...
                 target: player.drawDeck[0],
-                sendChatMessage: true,
-            }).generateEvent(context.target, context));
+                sendChatMessage: true
+            }).generateEvent(context.target, context);
+            events.push(lookAtEvent);
         } else {
-            const actual_amount = amount > deckLength ? deckLength : amount;
-            let cards = player.drawDeck.slice(0, actual_amount);
+            const actual_amount = Math.min(amount, deckLength);
```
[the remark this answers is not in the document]
CheBato: AAAAAA i'm an idiot i just now got what you meant with the snake case. Fixed it up

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: added whitespace

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: yessir!

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: made it like that!

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
CheBato: made it like that! Looks pretty now

server/game/gameSystems/LookMoverDeckCardsTopOrBottomSystem.ts
AMMayberry1: Oh that is a real error, the function should have a consistent return value. If we don't need the bool return value, then change the `return true;` to `return;`. Generally any errors emitted by TS (as opposed to lint issues from eslint) are real and should not be ignored.

CheBato: alrighty good to know i'll keep that in mind moving forward its also one of the reasons why I didn't resolve this convo.

CheBato: also changing from return true to return fixed the issue.

AMMayberry1 (approved):

PR #120, opened 2024-10-18
Fix windows so that "gain ambush" and "gain shielded" work correctly

AMMayberry1:
We had a bug where "gain ambush" and "gain shielded" would not work because units register those abilities when they move into the arena, but the effects that give them abilities resolve after they get to the arena.

Made a few changes to address this. The first is refactoring the event window so that we emit events again after a `resolveGameState` call so that any abilities gained during the event window resolution can successfully trigger.

The other change is refactoring how we handle triggered abilities for keywords. We previously would register keywords at a time point before the event actually resolves (e.g., registering when-played abilities when the unit enters the arena). However, it is possible that gain or lose ability effects can happen in between those two time points, so we need to evaluate whether the unit has the keyword exactly at resolution time.

To facilitate this, refactored how the triggering for keyword abilities happens - instead of pre-registering the abilities at an earlier point, we now check every unit that is played or declares an attack to determine if it has any relevant keywords when the attack / play is resolved, once it is past the point that any effects could add / remove keywords. This guarantees that we will always correctly determine the relevant set of keywords for the card.

### DIFF
--- a/server/game/abilities/keyword/AmbushAbility.ts
+++ b/server/game/abilities/keyword/AmbushAbility.ts
@@ -4,7 +4,7 @@ import TriggeredAbility from '../../core/ability/TriggeredAbility';
 import { TriggeredAbilityContext } from '../../core/ability/TriggeredAbilityContext';
 import { Card } from '../../core/card/Card';
 import { UnitCard } from '../../core/card/CardTypes';
-import { EventName, KeywordName } from '../../core/Constants';
+import { KeywordName } from '../../core/Constants';
 import Game from '../../core/Game';
 import * as Contract from '../../core/utils/Contract';
 import { ITriggeredAbilityProps } from '../../Interfaces';
--- a/server/game/abilities/keyword/AmbushAbility.ts
+++ b/server/game/abilities/keyword/AmbushAbility.ts
@@ -4,7 +4,7 @@ import TriggeredAbility from '../../core/ability/TriggeredAbility';
 import { TriggeredAbilityContext } from '../../core/ability/TriggeredAbilityContext';
 import { Card } from '../../core/card/Card';
 import { UnitCard } from '../../core/card/CardTypes';
-import { KeywordName } from '../../core/Constants';
+import { EventName, KeywordName } from '../../core/Constants';
 import Game from '../../core/Game';
 import * as Contract from '../../core/utils/Contract';
 import { ITriggeredAbilityProps } from '../../Interfaces';
--- a/server/game/cards/02_SHD/units/PrivateerScyk.ts
+++ b/server/game/cards/02_SHD/units/PrivateerScyk.ts
@@ -1,0 +1,23 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { Aspect, KeywordName, WildcardLocation } from '../../../core/Constants';
+
+export default class PrivateerScyk extends NonLeaderUnitCard {
+    protected override getImplementationId () {
+        return {
+            id: '0088477218',
+            internalName: 'privateer-scyk',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addConstantAbility({
+            title: 'While you control another Cunning unit, this unit gains Shielded',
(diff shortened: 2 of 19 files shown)


server/game/abilities/keyword/AmbushAbility.ts
```diff
@@ -4,7 +4,7 @@ import TriggeredAbility from '../../core/ability/TriggeredAbility';
 import { TriggeredAbilityContext } from '../../core/ability/TriggeredAbilityContext';
 import { Card } from '../../core/card/Card';
 import { UnitCard } from '../../core/card/CardTypes';
-import { KeywordName } from '../../core/Constants';
+import { EventName, KeywordName } from '../../core/Constants';
```
VJubert, line 7: necessary ?

server/game/cards/02_SHD/units/PrivateerScyk.ts
```diff
@@ -0,0 +1,23 @@
...
+        return {
+            id: '0088477218',
+            internalName: 'privateer-scyk',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addConstantAbility({
+            title: 'While you control another Cunning unit, this unit gains Shielded',
+            condition: (context) => context.source.controller.getOtherUnitsInPlayWithAspect(context.source, Aspect.Cunning).length > 0,
```
VJubert, line 16: isAspectInPlay

server/game/core/Game.js
```diff
@@ -35,6 +35,8 @@ const { default: Shield } = require('../cards/01_SOR/tokens/Shield.js');
 const { StateWatcherRegistrar } = require('./stateWatcher/StateWatcherRegistrar.js');
 const { DistributeAmongTargetsPrompt } = require('./gameSteps/prompts/DistributeAmongTargetsPrompt.js');
 const HandlerMenuMultipleSelectionPrompt = require('./gameSteps/prompts/HandlerMenuMultipleSelectionPrompt.js');
+const { AmbushAbility } = require('../abilities/keyword/AmbushAbility.js');
```
VJubert, line 38: necessary ?

server/game/cards/02_SHD/units/PrivateerScyk.ts
adbronico: I'm not sure the shorter name is clear. Do we have any units that check their own aspect(s)?

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -34,6 +36,24 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
     const StatsAndDamageClass = WithDamage(WithPrintedPower(BaseClass));
 
     return class AsUnit extends StatsAndDamageClass {
+        public static registerRulesListeners(game: Game) {
```
Ennis0, line 39: Can this not be handled with code belonging to the specific events in question(technically in the System classes I suppose)? At the end of their eventHandlers or something?

AMMayberry1: This is a good question. I think functionally it would work equally well, but I think design-wise I prefer this approach because it separates out the keyword functionality from the systems and keeps the implementation attached to the class it matters for.

Also, it's very unlikely that we would ever need to create something that emits one of the relevant events other than the current systems, but if there was for some reason some other way an OnUnitEntersPlay event could be emitted, we would need to duplicate the ambush / shielded check logic there.

I think realistically it would be fine either way but I lean the direction of keeping it this way

server/game/cards/02_SHD/units/PrivateerScyk.ts
AMMayberry1: Oh that's a new method Val just added, it includes an `ignoreUnit` parameter to filter out a selected card for exactly that reason

AMMayberry1: Oh wait I'm dumb, yeah I just realized we have duplicates of this API.

AMMayberry1: Yeah just putting it together now that there's a bit of redundancy, but we can let it be for now and see if there are distinct use cases over time.

PR #121, opened 2024-10-18
Add C-3PO and list select prompt

AMMayberry1:
Implemented the C-3PO card. As part of that, added a target resolver and prompt for getting the player to choose from a dropdown list of options.

This hit a lot of bugs and weak areas of the code for some reason, had to do a lot of minor bug fixes and cleanup to get it working.


server/game/TargetInterfaces.ts
```diff
@@ -12,11 +12,16 @@ import { PlayerTargetSystem } from './core/gameSystem/PlayerTargetSystem';
 export type ITriggeredAbilityTargetResolver<TContext extends TriggeredAbilityContext = TriggeredAbilityContext> =
   | ICardTargetResolver<TContext>
   | ISelectTargetResolver<TContext>
+  | ISelectStringTargetResolver<TContext>
```
Ennis0, line 15: The existing Select mode is already selecting from a set of strings when you think about it. More descriptive names might include SelectButton and SelectDropdown, if I'm understanding the intended UI

server/game/core/gameSteps/prompts/ChooseFromListPrompt.ts
```diff
@@ -0,0 +1,71 @@
...
+import { UiPrompt } from './UiPrompt';
+
+export interface IChooseFromListPromptProperties extends IPromptPropertiesBase {
+    options: string[];
+    source: Card;
+    choiceHandler: (choice: string) => void;
+}
+
+/**
+ * Prompt for distributing healing or damage among target cards.
```
Ennis0, line 16: Fix comment.

server/game/TargetInterfaces.ts
AMMayberry1: Yeah good callout, I wrestled with this one a bit. I was trying to avoid anything that describes exactly how the FE should work for the sake of decoupling, but I guess we are explicitly providing buttons and stuff now that I think about it. So I'll rename this to mention dropdown. Thanks!

danbastin (approved):

PR #123, opened 2024-10-19
Bunch of easy cards

VJubert:


server/game/cards/02_SHD/events/Outflank.ts
```diff
@@ -0,0 +1,23 @@
...
+        return {
+            id: '0802973415',
+            internalName: 'outflank',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Attack with a unit',
+            initiateAttack: {},
```
AMMayberry1, line 14: nit: they both do the same thing but could we instead do `immediateEffect: AbilityHelper.immediateEffects.attack()`, ditto line 17. I think the meaning of that version is a little clearer in that case for someone new. passing the empty object here might be hard to understand for someone new to js or to the repo

server/game/cards/02_SHD/events/Outflank.ts
```diff
@@ -0,0 +1,23 @@
...
+            then: {
+                title: 'Attack with another unit',
```
AMMayberry1, line 16: Need to add the filter so that the same unit can't be chosen twice, in case the first unit becomes readied again (e.g. Mace Windu ability). See the impl of Maximum Firepower for how to do this with a `then` ability.

test/server/cards/01_SOR/leaders/GrandInquisitorHuntingTheJedi.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+                    },
+                });
+            });
+
+            it('should damage and ready a friendly unit with 3 power or less', function() {
+                const { context } = contextRef;
+                context.scoutBikePursuer.exhausted = true;
+                context.player1.clickCard(context.grandInquisitor);
+                context.player1.clickCard(context.p2Base);
+                expect(context.player1).toBeAbleToSelectExactly([context.scoutBikePursuer, context.deathStarStormtrooper]);
```
AMMayberry1, line 75: nit: can we break this up into smaller blocks

test/server/cards/01_SOR/leaders/IG88RuthlessBountyHunter.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+                        leader: 'ig88#ruthless-bounty-hunter',
+                        resources: 4,
+                    },
+                    player2: {
+                        groundArena: ['viper-probe-droid', 'wampa'],
+                    },
+                });
+            });
+
+            it('should attack with a friendly unit (without buff)', function () {
```
AMMayberry1, line 18: nit: we should have a little more context on how the buff works - e.g., `should attack with a friendly unit and does not buff because there are fewer friendly units than enemy units`. Similar change to the other test below. 

Ideally the goal for test naming is that if the test breaks, someone could come here and read it and understand why the test is set up the way that it is without needing to look at the card

AMMayberry1 (approved):

PR #125, opened 2024-10-20
Bunch of units 2

VJubert:


server/game/cards/01_SOR/units/SpecForceSoldier.ts
```diff
@@ -0,0 +1,26 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { KeywordName, WildcardCardType } from '../../../core/Constants';
+
+export default class SpecForceSoldier extends NonLeaderUnitCard {
```
AMMayberry1, line 5: We may need to pause this card for now, sorry I shouldn't have marked as "implementable." The issue is that tracking lose and gain abilities or keywords is kind of complicated and we haven't implemented it yet. For example, the following situation currently doesn't work:

- Echo Base Defender has sentinel
- SpecForce Soldier is played and takes away sentinel
- Protector is attached to Echo Base Defender, giving it sentinel again

Since this is the only card that involves losing a keyword, we probably should not add it until the whole system is working

server/game/cards/01_SOR/units/TheGhostSpectreHomeBase.ts
```diff
@@ -0,0 +1,29 @@
...
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Give a shield token to another Spectre unit',
+            when: {
+                onAttackDeclared: (event, context) => event.attack.attacker === context.source,
+                onCardPlayed: (event, context) => event.card === context.source
+            },
+            targetResolver: {
+                optional: true,
+                cardCondition: (card, context) => card !== context.source && card.isUnit() && card.hasSomeTrait(Trait.Spectre),
```
AMMayberry1, line 22: Card.isUnit() technically isn't needed here since only units can host upgrades, so the `giveShield()` targeting will automatically filter them out

server/game/cards/02_SHD/units/GeneralRieekanDefensiveStrategist.ts
```diff
@@ -0,0 +1,36 @@
...
+            when: {
+                onAttackDeclared: (event, context) => event.attack.attacker === context.source,
+                onCardPlayed: (event, context) => event.card === context.source
+            },
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                controller: RelativePlayer.Self,
+                immediateEffect: AbilityHelper.immediateEffects.conditional({
+                    condition: (context) => context.target.isUnit() && context.target.hasSomeKeyword(KeywordName.Sentinel),
+                    onTrue: AbilityHelper.immediateEffects.giveExperience((context) => ({ target: context.target })),
```
AMMayberry1, line 25: Does this work if you remove the `target: context.target` from it? That is supposed to happen automatically with target resolvers. If not then maybe there is a bug I need to fix.

server/game/ongoingEffects/OngoingEffectLibrary.ts
```diff
@@ -122,7 +122,11 @@ export = {
     // increaseLimitOnPrintedAbilities: (abilities) =>
     //     OngoingEffectBuilder.card.static(EffectName.IncreaseLimitOnPrintedAbilities, abilities),
     // loseAllNonKeywordAbilities: () => OngoingEffectBuilder.card.static(EffectName.LoseAllNonKeywordAbilities),
-    // loseKeyword: (keyword) => OngoingEffectBuilder.card.static(EffectName.LoseKeyword, keyword),
+    loseKeyword: (keywordOrKeywordProperties: KeywordNameOrProperties) =>
```
AMMayberry1, line 125: Same note here as for SpecForce Soldier - let's comment this back out until we do the full implementation

test/server/cards/01_SOR/units/CountDookuDarthTyranus.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+                        spaceArena: ['green-squadron-awing']
+                    }
+                });
+            });
+
+            it('should defeat a unit with 4 or less remaining HP', function () {
+                const { context } = contextRef;
+                context.player1.clickCard(context.countDooku);
+
+                // also shield
```
AMMayberry1, line 22: What does this mean?

test/server/cards/01_SOR/units/FettsFiresprayPursuingTheBounty.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+                expect(context.player1.countExhaustedResources()).toBe(2);
+                context.player2.passAction();
+
+                // attack with fett's firespray
+                context.player1.clickCard(context.fettsFirespray);
+                context.player1.clickPrompt('Attack');
+                expect(context.fettsFirespray.exhausted).toBeTrue();
+                context.player2.passAction();
+
+                // we still can do his exhaust action even if he's exhausted
```
AMMayberry1, line 168: These are all great! Can we do two additional tests since Firespray ability is one of the few that doesn't have the cost of exhausting the controller, just to test that the action ability cost system is working correctly.

1. Confirm that you can still pay the 2 for the ability even if all units are exhausted
2. If all units are exhausted and the controller does not have 2 resources to spend, ability is not available

test/server/cards/01_SOR/units/GeneralTaggeConcernedCommander.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+            it('should give experience token to up to 3 Trooper unit', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.generalTagge);
+
+                // should be able to select all trooper
+                expect(context.player1).toBeAbleToSelectExactly([context.vanguardInfantry, context.mercenaryCompany, context.battlefieldMarine, context.volunteerSoldier]);
+                context.player1.clickCard(context.vanguardInfantry);
+                context.player1.clickCard(context.mercenaryCompany);
+                context.player1.clickCard(context.battlefieldMarine);
```
AMMayberry1, line 26: Can you add a click on the fourth trooper just to confirm it doesn't work (use `clickCardNonChecking()`)

test/server/cards/01_SOR/units/TheGhostSpectreHomeBase.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.sabineWren, context.kananJarrus]);
+                expect(context.player1).toHaveChooseNoTargetButton();
+                context.player1.clickCard(context.sabineWren);
+
+                // sabine should have a shield
+                expect(context.sabineWren).toHaveExactUpgradeNames(['shield']);
+                expect(context.kananJarrus.isUpgraded()).toBeFalse();
+                expect(context.player2).toBeActivePlayer();
+            });
+        }); describe('The Ghost\'s ability', function () {
```
AMMayberry1, line 33: nit: spacing

test/server/cards/02_SHD/units/TheArmorerSurvivalIsStrength.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+            it('should give shield token to up to 3 Mandalorian unit', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.theArmorer);
+
+                // should be able to select all mandalorian
+                expect(context.player1).toBeAbleToSelectExactly([context.houseKastSoldier, context.mandalorianWarrior, context.concordDawnInterceptors, context.theArmorer, context.supercommandoSquad]);
+                context.player1.clickCard(context.houseKastSoldier);
+                context.player1.clickCard(context.mandalorianWarrior);
+                context.player1.clickCard(context.concordDawnInterceptors);
```
AMMayberry1, line 27: Same here, can we add a 4th non-checking card click

server/game/cards/01_SOR/units/FettsFiresprayPursuingTheBounty.ts
```diff
@@ -0,0 +1,37 @@
...
+            internalName: 'fetts-firespray#pursuing-the-bounty',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'If you control Boba Fett or Jango Fett (as a leader or unit), ready this unit',
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) =>
+                    context.source.controller.leader.name === 'Boba Fett' ||
```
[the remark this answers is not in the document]
VJubert: name came from gameobject, I don't know if I can override it on Card and add deprecation attribute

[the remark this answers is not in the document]
AMMayberry1: Oh huh I thought we already had an override for some reason. But yeah I think that should work.

server/game/gameSystems/PayResourceCostSystem.ts
```diff
@@ -37,7 +37,7 @@ export class PayResourceCostSystem<TContext extends AbilityContext = AbilityCont
 
     public override canAffect(player: Player, context: TContext, additionalProperties = {}): boolean {
         const properties = this.generatePropertiesFromContext(context, additionalProperties);
-        return properties.amount > 0 && player.countSpendableResources() > 0 && super.canAffect(player, context, additionalProperties);
+        return properties.amount > 0 && player.countSpendableResources() > 0 && super.canAffect(player, context, additionalProperties) && player.countSpendableResources() >= properties.amount;
```
AMMayberry1, line 40: Good catch! Can you also do two small bookkeeping favors for me:

1. At the top of this class, put the comment `//TODO: replace this with ExhaustResourcesSystem when we have it`
2. Add an override property named `defaultProperties` that sets `amount: null` and `isCost: true`. You can look at e.g. GiveShieldSystem for an example.

AMMayberry1 (approved):

PR #129, opened 2024-10-23
PlayCardSystem

Ennis0:
Added functionality to play cards as part of abilities. Implemented Admiral Ozzel and Chewbacca leader("Walking Carpet").


test/server/cards/01_SOR/leaders/ChewbaccaWalkingCarpet.spec.js
```diff
@@ -0,0 +1,58 @@
...
+                context.player1.clickCard(context.chewbacca);
+                context.player1.clickPrompt('Play a unit that costs 3 or less. It gains sentinel for this phase');
+                context.player1.clickCard(context.seventhFleetDefender);
+                expect(context.player1.countExhaustedResources()).toBe(5);
+
+                // test sentinel gain again for good measure
+                context.player2.clickCard(context.tielnFighter);
+                expect(context.tielnFighter).toBeInLocation('discard');
+            });
+        });
```
Ennis0, line 56: Should there be tests for the leader side, even though its only text is keywords?

server/game/cards/01_SOR/units/AdmiralOzzelOverconfident.ts
```diff
@@ -0,0 +1,39 @@
...
+    public override setupCardAbilities() {
+        this.addActionAbility({
+            title: 'Play an Imperial unit from your hand. It enters play ready',
+            cost: AbilityHelper.costs.exhaustSelf(),
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.selectCard({
+                    activePromptTitle: 'Play an Imperial unit from your hand. It enters play ready',
+                    cardTypeFilter: CardType.BasicUnit,
+                    // TODO: figure out how to make this assume that the played card must be from hand, unless specified otherwise
+                    locationFilter: Location.Hand,
```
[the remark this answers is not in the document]
AMMayberry1: Actually, thinking on it further, probably that check should be in the play actions themselves since the `PlayType` makes it all the way to them. It was an oversight that they weren't already doing that check.

Can just put it in `PlayCardAction.meetsRequirements` since they inherit from that. I _think_ that should solve this issue and patch a few other holes as well.

server/game/gameSystems/PutIntoPlaySystem.ts
```diff
@@ -71,13 +73,14 @@ export class PutIntoPlaySystem<TContext extends AbilityContext = AbilityContext>
     }
 
     protected override addPropertiesToEvent(event, card: Card, context: TContext, additionalProperties): void {
-        const { controller, overrideLocation } = this.generatePropertiesFromContext(
+        const { controller, overrideLocation, entersReady } = this.generatePropertiesFromContext(
```
AMMayberry1, line 76: Could you do me small favor and a TODO here for "rename this class and all related classes / methods as PutUnitIntoPlay"

test/server/cards/01_SOR/leaders/ChewbaccaWalkingCarpet.spec.js
```diff
@@ -0,0 +1,58 @@
+describe('Chewbacca, Walking Carpet', function() {
+    integration(function(contextRef) {
+        describe('Chewbacca, Walking Carpet\'s undeployed ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['alliance-xwing', 'liberated-slaves', 'seventh-fleet-defender', 'consular-security-force'],
+                        leader: 'chewbacca#walking-carpet',
+                        base: 'administrators-tower'
```
AMMayberry1, line 10: Can we also put some <= 3 cost cards in discard and also some smuggle-able cards in resources to ensure that his ability doesn't become able to target them whenever we add those options.

If you want you can set that up as a separate describe since unfortunately if you want to name any card in the resource pile you need to name _all_ cards in the resource pile.

test/server/cards/01_SOR/units/AdmiralOzzelOverconfident.spec.js
```diff
@@ -0,0 +1,73 @@
...
+                        groundArena: ['admiral-ozzel#overconfident']
+                    },
+                    player2: {
+                        groundArena: [{ card: 'wampa', exhausted: true }],
+                        spaceArena: ['ruthless-raider']
+                    }
+                });
+            });
+
+            it('should be activatable even if the controller can\'t play an imperial unit, and still allow each opponent to ready a unit', function () {
```
AMMayberry1, line 57: Great test!

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+import * as Contract from '../core/utils/Contract';
+import { CardType, PlayType } from '../core/Constants';
+import { isPlayable } from '../core/card/CardTypes';
+import * as GameSystemLibrary from './GameSystemLibrary';
+import { PlayCardAction } from '../core/ability/PlayCardAction';
+import { PlayUnitAction } from '../actions/PlayUnitAction';
+import { PlayUpgradeAction } from '../actions/PlayUpgradeAction';
+import { PlayEventAction } from '../actions/PlayEventAction';
+
+export interface IPlayCardProperties extends ICardTargetSystemProperties {
```
AMMayberry1, line 14: It's kind of an edge case but there is a missing feature here we need to add a flag for. Specifically for the Mon Mothma example in the rule below (this is 1.17.4):

image

AMMayberry1: I haven't thought through what the best way to do this is. I'm not sure if it can be handled automatically or if we need to just have a flag that like "passHiddenInformation" or something. And to avoid a bunch of spurious bug reports, probably the prompt itself should explain _why_ the player is being presented with a pass option even though the ability doesn't say "you may."

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -0,0 +1,81 @@
...
+    ignoredRequirements?: string[];
+
```
AMMayberry1, line 16: One other additional feature I just thought of - we'll need to eventually work out a way to flag whether this action should be nested or not. E.g., for the Vader ability it's not nested but for U-Wing it is. That can be a TODO if you want since it may get kind of tricky, but we should add some comments at least (probably here and on the method in GameSystemLibrary).

test/server/cards/01_SOR/units/AdmiralOzzelOverconfident.spec.js
Ennis0: Thank you.

server/game/cards/01_SOR/units/AdmiralOzzelOverconfident.ts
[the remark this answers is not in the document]
Ennis0: It looks like `PlayCardAction.meetsRequirements` already does that:
`if ( !ignoredRequirements.includes('location') && !context.player.isCardInPlayableLocation(context.source, this.playType)
        ) {
            return 'location';
        }`

server/game/gameSystems/PlayCardSystem.ts
Ennis0: I changed CardTargetResolver to always set itself to optional if it's only targeting cards in hidden zones. I also figured out how to edit the prompt for the player to inform them like you suggested(it was non-obvious because I had to mess with the `selector` member of the CardTargetResolver).

server/game/cards/01_SOR/leaders/ChewbaccaWalkingCarpet.ts
```diff
@@ -0,0 +1,29 @@
...
+    protected override setupLeaderSideAbilities() {
+        this.addActionAbility({
+            title: 'Play a unit that costs 3 or less. It gains sentinel for this phase',
+            cost: AbilityHelper.costs.exhaustSelf(),
+            targetResolver: {
+                cardCondition: (card) => card.isUnit() && card.cost <= 3,
+                locationFilter: Location.Hand,
+                immediateEffect: AbilityHelper.immediateEffects.sequential([
+                    AbilityHelper.immediateEffects.playCard(),
+                    AbilityHelper.immediateEffects.forThisPhaseCardEffect({ effect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Sentinel) })
```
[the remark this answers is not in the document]
Ennis0: I put in ECL. Turns out that replacing the `sequential` system with `simultaneous` works.

server/game/cards/01_SOR/units/AdmiralOzzelOverconfident.ts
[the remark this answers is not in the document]
AMMayberry1: Yeah looks like you're right, commented in chat

server/game/cards/01_SOR/bases/EnergyConversionLab.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public override setupCardAbilities () {
+        this.setEpicActionAbility({
+            title: 'Play a unit that costs 6 or less from your hand. Give it ambush for this phase',
+            targetResolver: {
+                cardCondition: (card) => card.isUnit() && card.cost <= 6,
+                cardTypeFilter: CardType.BasicUnit,
+                locationFilter: Location.Hand,
+                immediateEffect: AbilityHelper.immediateEffects.simultaneous([
```
AMMayberry1, line 20: Very exciting that this worked!! Nice to see timing stuff just work right the first time more often now

server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -17,8 +17,16 @@ export class CardTargetResolver extends TargetResolver<ICardTargetResolver<Abili
...
     public constructor(name: string, properties: ICardTargetResolver<AbilityContext>, ability: PlayerOrCardAbility) {
         super(name, properties, ability);
+
         this.selector = this.getSelector(properties);
 
+        // a player can always choose not to pick a card from a zone that is hidden from their opponents (SWU Comp Rules 2.0 1.17.4)
+        if (this.allZonesAreHidden(this.properties.locationFilter)) {
+            this.properties.optional = true;
+            this.selector.optional = true;
+            this.selector.oldDefaultActivePromptTitle = this.selector.defaultActivePromptTitle;
```
AMMayberry1, line 27: Is this needed?

server/game/core/utils/EnumHelpers.ts
```diff
@@ -38,6 +38,17 @@ export const isAttackableLocation = (location: LocationFilter) => {
     }
 };
 
+export const isHidden = (location: LocationFilter) => {
```
AMMayberry1, line 41: Very useful, good idea. Could you do a me a small favor and update `Card.initializeForCurrentLocation()` to use this helper for setting the `hiddenForOpponent` property

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -146,12 +146,15 @@ export function lookAt<TContext extends AbilityContext = AbilityContext>(propert
...
-// export function playCard(propertyFactory: PropsFactory<PlayCardProperties> = {}): GameSystem {
-//     return new PlayCardAction(propertyFactory);
-// }
+/**
+ * default resetOnCancel = false
+ */
+export function playCardFromHand<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<Omit<IPlayCardProperties, 'playType' | 'optional'>, TContext> = {}): PlayCardSystem<TContext> {
+    // TODO: implement playing with smuggle and from non-standard zones(discard(e.g. Palpatine's Return), top of deck(e.g. Ezra Bridger), etc.) as part of abilties with another function(s)
+    // TODO: implement a "nested" property in PlayCardSystem that controls whether triggered abilities triggered by playing the card resolve after that card play or after the whole ability
+    // playType automatically defaults to PlayFromHand
```
AMMayberry1, line 155: Can we explicitly set the value of the `playType` property to `PlayFromHand` using Object.assign so that we're not taking a dependency on this remaining the default

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -168,7 +171,7 @@ export function putIntoPlay<TContext extends AbilityContext = AbilityContext>(pr
 // export function opponentPutIntoPlay(propertyFactory: PropsFactory<OpponentPutIntoPlayProperties> = {}): GameSystem {
 //     return new OpponentPutIntoPlayAction(propertyFactory, false);
 // }
-export function ready<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IReadySystemProperties, TContext> = {}): GameSystem<TContext> {
+export function ready<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IReadySystemProperties, TContext> = {}): ReadySystem<TContext> {
```
AMMayberry1, line 174: Can just remove the return type completely and it will work. Got a TODO to come do that for all methods in here, iirc

server/game/gameSystems/PlayCardSystem.ts
AMMayberry1: Awesome!! Great work

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -146,12 +146,15 @@ export function lookAt<TContext extends AbilityContext = AbilityContext>(propert
...
-//  * default resetOnCancel = false
-//  */
-// export function playCard(propertyFactory: PropsFactory<PlayCardProperties> = {}): GameSystem {
-//     return new PlayCardAction(propertyFactory);
-// }
+/**
+ * default resetOnCancel = false
+ */
+export function playCardFromHand<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<Omit<IPlayCardProperties, 'playType' | 'optional'>, TContext> = {}): PlayCardSystem<TContext> {
+    // TODO: implement playing with smuggle and from non-standard zones(discard(e.g. Palpatine's Return), top of deck(e.g. Ezra Bridger), etc.) as part of abilties with another function(s)
```
AMMayberry1, line 153: Can we also copy this into `PlayCardSystem`

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -146,12 +146,15 @@ export function lookAt<TContext extends AbilityContext = AbilityContext>(propert
...
-// /**
-//  * default resetOnCancel = false
-//  */
-// export function playCard(propertyFactory: PropsFactory<PlayCardProperties> = {}): GameSystem {
-//     return new PlayCardAction(propertyFactory);
-// }
+/**
+ * default resetOnCancel = false
+ */
+export function playCardFromHand<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<Omit<IPlayCardProperties, 'playType' | 'optional'>, TContext> = {}): PlayCardSystem<TContext> {
```
AMMayberry1, line 152: What's the reason to block out `optional`?

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -0,0 +1,79 @@
...
+        return ['play {0}', [properties.target]];
+    }
+
+    protected override addPropertiesToEvent(event, target, context: TContext, additionalProperties = {}): void {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+
+        super.addPropertiesToEvent(event, target, context, additionalProperties);
+
+        event.playCardAbility = this.generatePlayCardAbility(target, this.properties.playType);
+        event.optional = properties.optional == null ? context.ability.optional : properties.optional;
```
AMMayberry1, line 53: I realized there's a better pattern for this: `properties.optional ?? context.ability.optional`

test/server/cards/01_SOR/units/AdmiralOzzelOverconfident.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+            it('should allow the controller to play an imperial unit from hand, which enters play ready, and allow each opponent to ready a unit', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.admiralOzzel);
+                expect(context.player1).toHaveEnabledPromptButtons(['Play an Imperial unit from your hand. It enters play ready', 'Attack']);
+
+                context.player1.clickPrompt('Play an Imperial unit from your hand. It enters play ready');
+                expect(context.admiralOzzel.exhausted).toBe(true);
+                expect(context.player1).toBeAbleToSelectExactly([context.atst, context.deathStarStormtrooper]);
```
AMMayberry1, line 27: Can we add an expectation for the presence of the pass prompt

server/game/core/ability/abilityTargets/CardTargetResolver.ts
Ennis0: I think so. The different card selectors have different default prompts(`'Choose a ' + this.cardTypeFilter[0];`, `'Choose cards'`, ``Select ${this.numCards} cards``) so if we want a note for the player about why they can choose no targets like you suggested I think it makes sense to just append it to the end. Is there a better way of doing that?

server/game/gameSystems/GameSystemLibrary.ts
Ennis0: Playing from hand will always be optional because it's a hidden zone. Right?

server/game/gameSystems/GameSystemLibrary.ts
Ennis0: That will require checking if `propertyFactory` is a function and altering it differently depending, right?

server/game/gameSystems/PlayCardSystem.ts
Ennis0: I changed it here and in `InitiateAttackSystem`

test/server/cards/01_SOR/units/AdmiralOzzelOverconfident.spec.ts
Ennis0: I got it so that there's a 'Choose no target' button(had to edit SelectCardSystem a bit because it doesn't use the CardTargetResolver code). Does that work?

server/game/core/ability/abilityTargets/CardTargetResolver.ts
AMMayberry1: Ahh sorry, didn't realize these were functions not properties. Makes more sense now. Yeah it's annoying that they're all different, would be much easier if they had a base class so we can override the functionality in one place. Normally I wouldn't want to silently override a function's behavior like this but it's a pretty innocuous place to do it so I think it's fine as a band-aid.

server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -17,8 +17,16 @@ export class CardTargetResolver extends TargetResolver<ICardTargetResolver<Abili
     private selector;
     public constructor(name: string, properties: ICardTargetResolver<AbilityContext>, ability: PlayerOrCardAbility) {
         super(name, properties, ability);
+
         this.selector = this.getSelector(properties);
 
+        // a player can always choose not to pick a card from a zone that is hidden from their opponents (SWU Comp Rules 2.0 1.17.4)
+        if (this.allZonesAreHidden(this.properties.locationFilter)) {
+            this.properties.optional = true;
```
AMMayberry1, line 25: So I did a quick debug and it looks like code execution never makes it inside this block? I think it's because `locationFilter` by default is `WildcardLocation.AnyAttackable` which isn't covered by `EnumHelpers.isHidden`. That's something I should've caught, it needs to handle wildcards correctly as well.

Can you fix and then add a test case that confirms that the prompt title has the appended text from this code block

server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -27,6 +35,13 @@ export class CardTargetResolver extends TargetResolver<ICardTargetResolver<Abili
         this.validateLocationLegalForTarget(properties);
     }
 
+    private allZonesAreHidden(locationFilter): boolean {
+        if (Array.isArray(locationFilter)) {
```
AMMayberry1, line 39: I always forget to use it but there is a helper function called `asArray()` in Helpers.ts that does this as a one-liner

server/game/gameSystems/GameSystemLibrary.ts
AMMayberry1: Oh yeah true. I guess the only difference is that we potentially wouldn't need to show the extra explainer text saying why the play is optional when they might not expect it to be

server/game/gameSystems/GameSystemLibrary.ts
AMMayberry1: Oh yeah fair, that's more work than it's worth

test/server/cards/01_SOR/units/AdmiralOzzelOverconfident.spec.ts
AMMayberry1: Yep! There is also now a `toHaveChooseNoTargetButton()` expectation you can use

AMMayberry1:
Just a note that you have a merge conflict

server/game/gameSystems/SelectCardSystem.ts
```diff
@@ -164,4 +164,11 @@ export class SelectCardSystem<TContext extends AbilityContext = AbilityContext>
         const properties = this.generatePropertiesFromContext(context, additionalProperties);
         return properties.checkTarget && properties.player !== RelativePlayer.Opponent;
     }
+
+    private selectionIsOptional(properties, context): boolean {
+        if (properties.innerSystem.isOptional(context)) {
+            return true;
+        }
+        return CardTargetResolver.allZonesAreHidden(properties.locationFilter);
```
AMMayberry1, line 172: Want to confirm that this behavior is correct in all cases - what about a situation that's like "the opponent reveals their hand, then you choose a card to discard?" Would this logic cause it to become optional when it shouldn't?

My gut feeling is that the selector can't have enough context to determine when the hidden-ness rule is relevant so it will have to be a property or something, but I could be wrong

server/game/core/ability/abilityTargets/CardTargetResolver.ts
Ennis0: It does go into the block in the tests for ECL and Chewie; not Ozzel though, I think because he uses a SelectCardSystem instead of a target resolver. I'll add a test though.

server/game/gameSystems/GameSystemLibrary.ts
Ennis0: Do you mean that we won't want to show the explainer text when the card text says "may"?

server/game/gameSystems/SelectCardSystem.ts
Ennis0: That is a good point(choosing from an opponent's otherwise hidden zone). Do we have any cards implemented yet that do that?
Speculatively, I think the easy thing to do to account for your example would be to add `controller` and `selectingPlayer` arguments to `allZonesAreHidden()`. Then we can say: "if the selecting player is selecting from their opponent's hand, then it's not really a hidden zone".

AMMayberry1: We don't have anything implemented yet, the discard systems aren't in place yet unfortunately, I think your idea should work though. We can put in the logic for now and maybe just leave a TODO comment that not all cases have been tested.

server/game/gameSystems/GameSystemLibrary.ts
AMMayberry1: Ah hey sorry missed your reply here - I realized there isn't actually a case where this would apply so nvm.

server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -58,6 +58,15 @@ export class CardTargetResolver extends TargetResolver<ICardTargetResolver<Abili
     }
 
     protected override resolveInner(context: AbilityContext, targetResults, passPrompt, player: Player) {
+        // a player can always choose not to pick a card from a zone that is hidden from their opponents (SWU Comp Rules 2.0 1.17.4)
+        // TODO: test if picking a card from an opponent's usually hidden zone(e.g. opponent's hand) works as expected(the if block here should be skipped)
+        if (CardTargetResolver.allZonesAreHidden(this.properties.locationFilter, typeof this.properties.controller === 'function' ? this.properties.controller(context) : this.properties.controller)) {
```
AMMayberry1, line 63: Hey unfortunately I realized there's a slight bug here - this condition only applies if there is some restriction on the card being drawn. The rule states that "a player may choose to act as if they have fewer options than they do," but that means if the ability lets you draw _any_ card, you are not revealing hidden information by choosing and you must do it.

That will probably be a slightly bigger work item to distinguish the two cases so I'm fine if for now you just want to put TODOs here and in SelectCardSystem

Ennis0: Would it be too simple to just check if `properties` has a `cardTypeFilter` or `cardCondition`?

AMMayberry1: Yeah that was my thought, would just have to add some computed property on CardSelector like `get hasAnyFilter() { return !!this.cardCondition && !!this.cardTypeFilter; }` or something. I just know this PR review has gone long already so don't want to push you to do more in this one if you'd rather merge here and save it for later.

server/game/gameSystems/SelectCardSystem.ts
```diff
@@ -165,6 +165,10 @@ export class SelectCardSystem<TContext extends AbilityContext = AbilityContext>
         if (properties.innerSystem.isOptional(context)) {
             return true;
         }
-        return CardTargetResolver.allZonesAreHidden(properties.locationFilter, typeof properties.controller === 'function' ? properties.controller(context) : properties.controller);
+        return CardTargetResolver.allZonesAreHidden(properties.locationFilter,
+            typeof properties.controller === 'function'
```
AMMayberry1, line 169: nit: can we do this in a variable for readability

AMMayberry1 (approved):

PR #130, opened 2024-10-23
Ignore Aspect Penalties, Hera, and Karabast

adbronico:


server/game/core/cost/CostAdjuster.ts
```diff
@@ -1,37 +1,57 @@
...
+    amount?: number | ((card: Card, player: Player) => number);
+
+    /** The type of cost adjustment */
+    costAdjustType: CostAdjustType;
+
+    /** The cost floor (for minimum costs) */
     costFloor?: number;
+
+    /** The number of cost reductions permitted */
     limit?: IAbilityLimit;
```
AMMayberry1, line 35: Can you tag this and `playingTypes` as @deprecated with a note that they aren't tested yet

test/server/cards/01_SOR/leaders/HeraSyndullaSpectreTwo.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+
+                context.player1.clickCard(context.karabast);
+                expect(context.player1).toBeAbleToSelectExactly([context.sabineWren, context.battlefieldMarine, context.yoda]);
+                context.player1.clickCard(context.battlefieldMarine);
+
+                expect(context.pykeSentinel).toBeInLocation('discard');
+                expect(context.player1.countExhaustedResources()).toBe(4);
+            });
+
+            it('does not ignore aspect penalties for non-Spectre card', function () {
```
AMMayberry1, line 34: Can we fold this in to the previous test, same for the deployed version

server/game/ongoingEffects/OngoingEffectLibrary.ts
```diff
@@ -209,13 +209,14 @@ export = {
...
-    decreaseCost: (properties: Omit<ICostAdjusterProperties, 'direction'>) => modifyCost(Object.assign(properties, { direction: CostAdjustDirection.Decrease })),
+    decreaseCost: (properties: Omit<ICostAdjusterProperties, 'costAdjustType'>) => modifyCost(Object.assign(properties, { costAdjustType: CostAdjustType.Decrease })),
     // gainActionPhasePriority: () =>
     //     OngoingEffectBuilder.player.detached(EffectName.GainActionPhasePriority, {
     //         apply: (player) => (player.actionPhasePriority = true),
     //         unapply: (player) => (player.actionPhasePriority = false)
     //     }),
-    increaseCost: (properties: Omit<ICostAdjusterProperties, 'direction'>) => modifyCost(Object.assign(properties, { direction: CostAdjustDirection.Increase })),
+    increaseCost: (properties: Omit<ICostAdjusterProperties, 'costAdjustType'>) => modifyCost(Object.assign(properties, { costAdjustType: CostAdjustType.Increase })),
+    ignoreAllAspects: (properties: Omit<ICostAdjusterProperties, 'costAdjustType'>) => modifyCost(Object.assign(properties, { costAdjustType: CostAdjustType.IgnoreAllAspects })),
```
AMMayberry1, line 219: Can this be `ignoreAllAspectPenalties`

server/game/core/cost/CostAdjuster.ts
```diff
@@ -41,14 +61,19 @@ export class CostAdjuster {
...
     ) {
+        this.costAdjustType = properties.costAdjustType;
+        if (this.costAdjustType === CostAdjustType.Increase || this.costAdjustType === CostAdjustType.Decrease) {
+            Contract.assertNotNullLike(properties.amount, 'Amount is required for Increase and Decrease.');
+        }
         this.amount = properties.amount || 1;
         this.costFloor = properties.costFloor || 0;
-        this.direction = properties.direction;
+        // TODO: Implement this further for Rey/Kylo
+        if (this.costAdjustType === CostAdjustType.IgnoreSpecificAspects) {
```
AMMayberry1, line 73: Can we add some whitespace to break this method up a bit

server/game/core/cost/CostAdjuster.ts
```diff
@@ -41,14 +61,19 @@ export class CostAdjuster {
...
+            Contract.assertNotNullLike(this.ignoredAspects, 'Ignored Aspect list is required for IgnoreSpecificAspects');
```
AMMayberry1, line 74: Couple modifications to this:

1. Also need to check that if it is a list that it is non-empty
2. We should validate that ignoredAspects does not have a value if costAdjustType is not IgnoreSpecificAspects

server/game/core/cost/CostAdjuster.ts
```diff
@@ -1,37 +1,57 @@
...
 import type { Card } from '../card/Card';
-import { CardType, PlayType, Aspect, CardTypeFilter } from '../Constants';
+import { PlayType, Aspect, CardTypeFilter, AspectFilter } from '../Constants';
 import type Game from '../Game';
 import type Player from '../Player';
 import * as EnumHelpers from '../utils/EnumHelpers';
+import * as Contract from '../../core/utils/Contract';
 
-export enum CostAdjustDirection {
+export enum CostAdjustType {
```
AMMayberry1, line 10: I like this way of handling it a lot, very clear without being overcomplicated

server/game/core/Constants.ts
```diff
@@ -220,6 +220,8 @@ export enum Aspect {
     Villainy = 'villainy',
 }
 
+export type AspectFilter = Aspect | Aspect[];
```
AMMayberry1, line 223: What's the intent here? This doesn't match the other "Filter" types in here

server/game/core/Player.js
```diff
@@ -657,20 +657,26 @@ class Player extends GameObject {
      * @param {PlayType} playingType
      * @param card DrawCard
      * @param target BaseCard
+     * @param penaltyAspect Aspect that is not present on the current base or leader
      */
     runAdjustersForCostType(playingType, baseCost, card, target, ignoreType = false, penaltyAspect = null) {
```
AMMayberry1, line 662: This method is now mixing cost increase / decrease with aspect penalty negation which is making the logic a little unnecessarily complex. Since there's no situation where those ever overlap right now, probably we should have separate methods with separate logic for regular vs aspect penalty adjustment

server/game/core/cost/CostAdjuster.ts
```diff
@@ -1,37 +1,57 @@
...
-    direction: CostAdjustDirection;
+
+    /** The amount to adjust the cost by */
+    amount?: number | ((card: Card, player: Player) => number);
+
+    /** The type of cost adjustment */
+    costAdjustType: CostAdjustType;
+
+    /** The cost floor (for minimum costs) */
     costFloor?: number;
```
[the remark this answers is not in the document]
adbronico: @AMMayberry1 Do you think it would make sense to leave it in as deprecated or commented out? I could definitely see a cost floor being a thing at some future point.

server/game/core/Constants.ts
adbronico: I was thinking this would be a shorthand for one or more aspects. It might not be used elsewhere so it's probably better to just remove it.

server/game/core/cost/CostAdjuster.ts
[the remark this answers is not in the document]
AMMayberry1: At the barest minimum I would want to comment it out, as we add more features like PlayForFree we would have to figure out how they would interact with this property and it's not worth it for a feature that doesn't exist yet.

I'm actually a little skeptical it will come in anytime soon, I haven't seen stuff like that a lot (though my tcg experience is limited tbf), so I'm fine with just removing it. Up to you though

server/game/core/Constants.ts
AMMayberry1: I think that'd be reasonable if we didn't already have a pattern that "Filter" types don't define arrays but rather mean "a union of an enum type with its corresponding wildcard"

server/game/core/Player.js
```diff
@@ -174,6 +168,14 @@ class Player extends GameObject {
         return this.getArenaCards(arena).filter((card) => card.isUnit() && card !== ignoreUnit && card.hasSomeAspect(aspect) && cardCondition(card));
     }
 
+    /**
+     * @param { String } name the name of the unit to check for control of
+     * @returns { boolean } true if this player controls a unit with the given name
+     */
+    controlsUnitWithName(name) {
```
AMMayberry1, line 175: This name is a little misleading imo since it counts the leader even if it's not currently a unit (i.e. not currently deployed).

test/server/cards/02_SHD/units/OmegaPartOfTheSquad.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+
+                context.player1.clickCard(context.omega);
+                expect(context.player1).toHavePrompt('Select a card to reveal');
+                expect(context.player1).toHaveEnabledPromptButton(context.crosshair.title);
+                expect(context.player1).toHaveDisabledPromptButtons([context.atst.title, context.battlefieldMarine.title, context.cartelSpacer.title, context.pykeSentinel.title]);
+                context.player1.clickPrompt(context.crosshair.title);
+                expect(context.crosshair).toBeInLocation('hand');
+                expect(context.getChatLogs(2)).toContain('player1 takes Crosshair');
+
+                // Omega should cost 4 since it cannot discount itself
```
AMMayberry1, line 27: Here's a fun test - can we add a case where we play an Omega and then next phase play another copy of Omega and confirm that the second one gets the discount. I found some pretty clean patterns for getting references to two copies of a card during test setup, can see in `UniqueRule.spec.ts`

test/server/cards/02_SHD/units/OmegaPartOfTheSquad.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+                expect(context.player1.countExhaustedResources()).toBe(4);
+
+                context.player2.passAction();
```
AMMayberry1, line 30: Can we play a second clone this round just to be extra safe

test/server/cards/02_SHD/units/OmegaPartOfTheSquad.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+                    player1: {
+                        groundArena: ['omega#part-of-the-squad'],
+                        hand: ['crosshair#following-orders', 'wolffe#suspicious-veteran'],
+                        base: 'echo-base',
+                        leader: 'hera-syndulla#spectre-two'
+                    },
+                });
+            });
+
+            it('negates aspect penalties on the first clone played', function () {
```
AMMayberry1, line 51: If we have this test case can we just move all of the aspect penalty stuff from the first one to here? Right now it's kind of spread out between both

test/server/cards/02_SHD/upgrades/TheDarksaber.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+
+                expect(context.concordDawnInterceptors.getPower()).toBe(2);
+                expect(context.concordDawnInterceptors.getHp()).toBe(5);
+
+                expect(context.followerOfTheWay.getPower()).toBe(1);
+                expect(context.followerOfTheWay.getHp()).toBe(3);
+            });
+        });
+
+        describe('The Darksaber\'s ability', function() {
```
AMMayberry1, line 40: Can we add one more test case to confirm that it cannot attach to a vehicle

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -1,12 +1,26 @@
...
 
 // required for mixins to be based on this class
 export type PlayableOrDeployableCardConstructor = new (...args: any[]) => PlayableOrDeployableCard;
 
+export interface IDecreaseEventCostAbilityProps<TSource extends Card = Card> extends Omit<ICostAdjusterProperties, 'cardTypeFilter' | 'match' | 'costAdjustType'> {
+    title: string;
+    condition?: (context: AbilityContext<TSource>) => boolean;
+}
+
+export interface IIgnoreAspectPenaltiesProps<TSource extends Card = Card> extends Omit<ICostAdjusterProperties, 'cardTypeFilter' | 'match' | 'costAdjustType'> {
```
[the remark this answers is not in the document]
adbronico: I'm not sure if they can after the other changes we made. I'll let you take a look to see what you think

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -60,4 +74,70 @@ export class PlayableOrDeployableCard extends Card {
...
+            condition: condition,
+            ongoingEffect: AbilityHelper.ongoingEffects.ignoreAllAspectPenalties(costAdjusterProps)
+        };
+
+        return costAdjustAbilityProps;
+    }
+
+
+    /** Create constant ability props on the card that decreases its cost under the given condition */
+    protected generateIgnoreSpecificAspectPenaltiesAbilityProps(properties: IIgnoreAspectPenaltiesProps<this>): IConstantAbilityProps {
```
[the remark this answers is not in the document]
adbronico: I'm open to suggestions on how to do that. It doesn't seem obvious to me on how to do that

test/server/cards/02_SHD/units/OmegaPartOfTheSquad.spec.ts
adbronico: I was trying to consolidate into two tests instead of three. I combined the play Omega test with the "don't discount a clone played right after Omega" test. I wouldn't be able to verify that in this test since Omega is already on the board

test/server/cards/02_SHD/units/OmegaPartOfTheSquad.spec.ts
AMMayberry1: Sorry, I misread this as `moveToNextActionPhase()`. Could we then add that and have a second action phase to confirm the ability resets? That's what I thought you were doing already, my b

test/server/cards/02_SHD/units/OmegaPartOfTheSquad.spec.ts
AMMayberry1: Oh yeah that's fair, I didn't realize you wanted to make a distinction between the two. That's reasonable though. If you wanted you could combine them down to one using multiple action phases but I'm fine either way

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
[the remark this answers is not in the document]
AMMayberry1: Yeah probably these have to be separate now

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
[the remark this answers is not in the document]
AMMayberry1: Just refactoring the common properties out into methods. This works for me (or at least compiles), though it requires that IOngoingEffectGenerator be exported from Interfaces.ts (it should always have been, just a mistake)

```typescript
    /** Create constant ability props on the card that decreases its cost under the given condition */
    protected generateIgnoreSpecificAspectPenaltiesAbilityProps(properties: IIgnoreSpecificAspectPenaltyProps<this>): IConstantAbilityProps {
        const { title, ignoredAspects, condition, ...otherProps } = properties;

        const costAdjusterProps: ICostAdjusterProperties = {
            ...this.buildCostAdjusterGenericProperties(),
            costAdjustType: CostAdjustType.IgnoreSpecificAspects,
            ignoredAspects: ignoredAspects,
            ...otherProps
        };

        return this.buildCostAdjusterAbilityProps(condition, title, AbilityHelper.ongoingEffects.ignoreSpecificAspectPenalties(costAdjusterProps));
    }

    private buildCostAdjusterGenericProperties() {
        return {
            cardTypeFilter: this.printedType,
            match: (card, adjusterSource) => card === adjusterSource
        };
    }

    private buildCostAdjusterAbilityProps(condition: (context: AbilityContext<this>) => boolean, title: string, ongoingEffect: IOngoingEffectGenerator): IConstantAbilityProps {
        const costAdjustAbilityProps: IConstantAbilityProps = {
            title,
            sourceLocationFilter: WildcardLocation.Any,
            targetController: RelativePlayer.Any,
            condition,
            ongoingEffect
        };

        return costAdjustAbilityProps;
    }
```

test/server/cards/02_SHD/units/OmegaPartOfTheSquad.spec.ts
AMMayberry1: Nvm just saw it has that, ignore me lol

server/game/Interfaces.ts
```diff
@@ -153,7 +182,7 @@ type ITriggeredAbilityAggregateWhenProps<TSource extends Card> = ITriggeredAbili
     aggregateWhen: (events: GameEvent[], context: TriggeredAbilityContext) => boolean;
 };
 
-type IOngoingEffectGenerator = (game: Game, source: Card, props: IOngoingEffectProps) => (OngoingCardEffect | OngoingPlayerEffect);
+export type IOngoingEffectGenerator = (game: Game, source: Card, props: IOngoingEffectProps) => (OngoingCardEffect | OngoingPlayerEffect);
```
AMMayberry1, line 185: Do you mind moving this above the comment line on 170 with the other exported types

server/game/core/Player.js
```diff
@@ -175,6 +169,14 @@ class Player extends GameObject {
         return this.getArenaCards(arena).filter((card) => card.isUnit() && card !== ignoreUnit && card.hasSomeAspect(aspect) && cardCondition(card));
     }
 
+    /**
+     * @param { String } title the title of the unit to check for control of
```
AMMayberry1, line 173: nit: jsdoc should also say "unit or leader" each time

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -60,4 +80,68 @@ export class PlayableOrDeployableCard extends Card {
...
+        const effect = AbilityHelper.ongoingEffects.decreaseCost(costAdjusterProps);
+        return this.buildCostAdjusterAbilityProps(condition, title, effect);
+    }
+
+
+    /** Create constant ability props on the card that decreases its cost under the given condition */
+    protected generateIgnoreAllAspectPenaltiesAbilityProps(properties: IIgnoreAllAspectPenaltiesProps<this>): IConstantAbilityProps {
+        const { title, condition, ...otherProps } = properties;
+
+        const costAdjusterProps: IIgnoreAllAspectsCostAdjusterProperties = {
```
AMMayberry1, line 103: nit, readability: could we make this ICostAdjusterProperties to line up with the other methods (or vice versa)?

PR #134, opened 2024-10-24
Bunch of units 3 (jyn erso, boba daimyo, concord dawn interceptors)

VJubert:


server/game/cards/02_SHD/leaders/BobaFettDaimyo.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addTriggeredAbility({
+            title: 'When you play a unit that has 1 or more keywords: You may exhaust this leader. If you do, give a friendly unit +1/+0 for this phase.',
+            cost: AbilityHelper.costs.exhaustSelf(),
+            optional: true,
+            when: {
+                onCardPlayed: (event, context) =>
+                    event.card.isUnit() && event.card.controller === context.source.controller && event.card.keywords.length > 0
```
adbronico, line 20: We might have a better way to check if a card has a keyword, but I'm not 100% sure. Veld can say for sure

adbronico (commented):
This looks good to me, but I'll leave final review for Veld

server/game/cards/02_SHD/leaders/BobaFettDaimyo.ts
VJubert: for a particular keyword we have method, but for any keyword I think not

AMMayberry1: Yeah pretty sure we don't, but this comes up very infrequently so I don't feel the need to add a bespoke method for it

server/game/cards/02_SHD/leaders/BobaFettDaimyo.ts
```diff
@@ -0,0 +1,43 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '9334480612',
+            internalName: 'boba-fett#daimyo',
+        };
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addTriggeredAbility({
+            title: 'When you play a unit that has 1 or more keywords: You may exhaust this leader. If you do, give a friendly unit +1/+0 for this phase.',
```
AMMayberry1, line 15: Typically the ability is just the description of the effect, we leave out the trigger and the cost. So in this case I would just do `Give a friendly unit +1/+0 for this phase` and then remove the `activePromptTitle` from the target resolver.

We might change that way of doing things once the FE is up but for now let's just keep everything consistent.

test/server/cards/02_SHD/leaders/BobaFettDaimyo.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+                        resources: 4,
+                    },
+                    player2: {
+                        hand: ['outer-rim-headhunter'],
+                        groundArena: ['wampa'],
+                    },
+                });
+            });
+
+            it('should give +1/+0 to a friendly unit when play a unit with keyword (first unit play)', function () {
```
AMMayberry1, line 20: Great job covering all the keywords! A few more tests it would be good to add - for both the deployed and undeployed side, can you add tests with units which have conditional keywords like Hunting Nexu. Need to confirm that the ability handles both cases correctly when the unit does and does not have the keyword active.

For the deployed ability specifically, could you also test a unit with the Heroic Resolve ability to confirm that (a) gaining keyword from another card works right and (b) the +1 will be correctly applied if the keyword is "for this attack."

test/server/cards/02_SHD/leaders/BobaFettDaimyo.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+                        spaceArena: ['green-squadron-awing'],
+                        leader: { card: 'boba-fett#daimyo', deployed: true },
+                    },
+                    player2: {
+                        spaceArena: ['outer-rim-headhunter'],
+                    }
+                });
+            });
+
+            it('should give +1/+0 to other friendly unit with keyword', function () {
```
AMMayberry1, line 121: Just to check that the engine behavior is right, could you add an expectation that cards in hand do _not_ have the +1 if they have a keyword? That way if we ever mess with the targeting for ongoing effects we'll know if we break something

test/server/cards/02_SHD/units/ConcordDawnInterceptors.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+            it('should has +2/+0 while defending', function () {
+                const { context } = contextRef;
+                context.player1.clickCard(context.concordDawnInterceptors);
+                context.player1.clickCard(context.p2Base);
+                expect(context.p2Base.damage).toBe(1);
+                context.player2.clickCard(context.piratedStarfighter);
+                expect(context.concordDawnInterceptors.damage).toBe(3);
+                expect(context.concordDawnInterceptors.getPower()).toBe(1);
+                expect(context.piratedStarfighter.damage).toBe(3);
```
AMMayberry1, line 24: Could you also add a check that it does not gain the +2 when attacking

VJubert: already done few line ahead

test/server/cards/02_SHD/leaders/BobaFettDaimyo.spec.ts
VJubert: I skip the undeployed boba with hunting nexu and an aggression unit test, added keyword test is already covered with my Red Three test

test/server/cards/02_SHD/leaders/BobaFettDaimyo.spec.ts
VJubert: skip as we can't check power modification on hand

AMMayberry1 (approved):

PR #135, opened 2024-10-25
Add card Wedge Antilles (SOR 100)

garrettdc2:
Adds the card Wedge Antilles, Star of the Rebellion (SOR 100). When in play, Wedge gives friendly VEHICLE units Ambush and +1/+1.

SWUDB Link: https://swudb.com/card/SOR/100/wedge-antilles

image


test/server/cards/01_SOR/units/WedgeAntillesStarOfTheRebellion.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+                    },
+                    player2: {
+                        groundArena: ['wampa'],
+                        spaceArena: ['hwk290-freighter'],
+                        hand: ['atst']
+                    }
+                });
+            });
+
+            it('should give Ambush and +1/+1 to a friendly VEHICLE unit', function () {
```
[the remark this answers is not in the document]
garrettdc2: Hey thanks for the comment. You definitely can! The benefit of splitting them up is it given you better insight into what break when it breaks, which is why I opted for a separate test per use case in the suite. For example, if the problem is that new code breaks the matcher and ambush is now given **enemy** units, then the title of the test will make that clear without having to read the specific lines of code.

That being said, it's just a choice that's more in line with what my teams follow at work. If we prefer to club tests I can make that change, but just wanted to given an explanation on why they are separate in this first revision. Just lmk.

server/game/cards/01_SOR/units/WedgeAntillesStarOfTheRebellion.ts
```diff
@@ -0,0 +1,26 @@
...
+            id: '4339330745',
+            internalName: 'wedge-antilles#star-of-the-rebellion',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'Each friendly VEHICLE unit gets +1/+1 and gains Ambush.',
+            targetController: RelativePlayer.Self,
+            matchTarget: (card) => card.hasSomeTrait(Trait.Vehicle),
```
AMMayberry1, line 17: It might be worth adding `cardTypeFilter: WildcardCardType.Unit` just to be safe. Seems pretty unlikely that an upgrade with the Vehicle trait would be printed but you never know

server/game/cards/01_SOR/units/WedgeAntillesStarOfTheRebellion.ts
```diff
@@ -0,0 +1,26 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '4339330745',
+            internalName: 'wedge-antilles#star-of-the-rebellion',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'Each friendly VEHICLE unit gets +1/+1 and gains Ambush.',
```
AMMayberry1, line 15: Our convention for trait and aspect names in player-facing text is just to do PascalCase. I have been meaning to ask the UX folks if they want to keep this convention so I'll do that soon

test/server/cards/01_SOR/units/WedgeAntillesStarOfTheRebellion.spec.ts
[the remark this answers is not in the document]
AMMayberry1: Yeah I'm normally much more in favor of tests being as separated out as possible, but I've been asking the gang to consolidate tests because the jasmine tests are super heavy and slow to start up for some reason and we have finite github actions time with the free tier. It definitely does make the test failures a little harder to diagnose so it's not optimal for sure.

We do have a plan to eventually refactor to a better test framework, but that just didn't make the cut for the first release unfortunately. You can check out other tests like Agent Kallus to see the conventions we use to try and keep the test cases as clean as possible if they end up being bundled.

test/server/cards/01_SOR/units/WedgeAntillesStarOfTheRebellion.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.allianceXwing);
+                expect(context.player1).toHaveExactPromptButtons(['Ambush', 'Pass']);
+                context.player1.clickPrompt('Ambush');
+
+
```
AMMayberry1, line 28: nit: excess whitespace (I need to see if I can get the linter to catch this)

test/server/cards/01_SOR/units/WedgeAntillesStarOfTheRebellion.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+                expect(context.hwk290Freighter.damage).toBe(3);
+            });
+
+            it('should not give Ambush and +1/+1 to a friendly non-VEHICLE unit', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.battlefieldMarine);
+                expect(context.battlefieldMarine).toBeInLocation('ground arena');
+                expect(context.battlefieldMarine.getPower()).toBe(3);
+                expect(context.battlefieldMarine.getHp()).toBe(3);
```
AMMayberry1, line 42: Just to have extra confirmation that the ambush prompt didn't happen, can we add `expect(context.player2).toBeActivePlayer()`. Same for below test

test/server/cards/01_SOR/units/WedgeAntillesStarOfTheRebellion.spec.ts
```diff
@@ -0,0 +1,55 @@
+describe('Wedge Antilles, Star of the Rebellion', function() {
+    integration(function(contextRef) {
+        describe('Wedge Antilles\'s ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['alliance-xwing', 'battlefield-marine', 'clan-saxon-gauntlet'],
+                        groundArena: ['wedge-antilles#star-of-the-rebellion'],
+                        leader: 'hera-syndulla#spectre-two',
```
AMMayberry1, line 10: I think there's some stuff in here that isn't relevant to the test such as the Hera leader and the explicit resource count (it sets at 20 by default). We try to keep the test setup to the bare minimum just for readability, do you mind removing the unused stuff?

server/game/cards/01_SOR/units/WedgeAntillesStarOfTheRebellion.ts
danbastin: I was thinking about this as I was kind of seeing where certain titles and prompts would land. I think it would be nice to be able to render these bold + italic on the frontend so either we have a library of tags and use the casing to identify them for rendering or we do something like wrapping them in brackets to make is easily parsable.

AMMayberry1: Sgtm. I'd prefer some kind of special characters over casing, more obvious to someone reading the backend code (much harder to realize that ALLCAPS is being specially handled). Let's move the discussion over to discord and we can figure out the best option.

server/game/cards/01_SOR/units/WedgeAntillesStarOfTheRebellion.ts
garrettdc2: Makes sense. Adding.

AMMayberry1 (approved):

PR #138, opened 2024-10-26
Add card Admiral Ackbar (SOR 097)

garrettdc2:
SWUDB: https://swudb.com/card/sor/097/admiral-ackbar

image

### DIFF
--- a/server/game/cards/01_SOR/units/AdmiralAckbarBrilliantStrategist.ts
+++ b/server/game/cards/01_SOR/units/AdmiralAckbarBrilliantStrategist.ts
@@ -24,7 +24,7 @@ export default class AdmiralAckbarBrilliantStrategist extends NonLeaderUnitCard 
 
     public override setupCardAbilities() {
         this.addWhenPlayedAbility({
-            title: 'Deal damage to a unit equal to the number of units your control in its arena.'
+            title: 'Deal damage to a unit equal to the number of units your control in its arena.',
             optional: true,
             targetResolver: {
                 cardTypeFilter: WildcardCardType.Unit,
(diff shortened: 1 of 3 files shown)


server/game/cards/01_SOR/units/AdmiralAckbarBrilliantStrategist.ts
```diff
@@ -0,0 +1,37 @@
...
+        };
+    }
+
+    /**
+     * Determines the number of units in the ability's target unit arena.
+     * @param context The context from the ability.
+     * @returns The number of units in target units arena.
+     */
+    private getDamageFromContext(context: AbilityContext): number {
+        const arenaName = context.target._location;
```
AMMayberry1, line 20: Can we use the public `.location` instead just to honor that contract

server/game/cards/01_SOR/units/AdmiralAckbarBrilliantStrategist.ts
```diff
@@ -0,0 +1,37 @@
...
+        const arena = context.source.controller.getUnitsInPlay(arenaName);
+        return arena.length;
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'You may deal damage to a unit equal to the number of units your control in its arena.',
```
AMMayberry1, line 27: nit: for ability titles our convention right now is to just provide the effect description and not include any cost or "you may" wording since that info will be provided by the context of the UX (e.g., offering a "pass" button or something). So in this case it would just be "Deal damage to a unit equal to the number of units your control in its arena"

also nit, typo: "units your control"

AMMayberry1 (approved):

PR #140, opened 2024-10-26
Send gamestate update

danbastin:
Basic support for sending gamestate updates to two connected players. Most of this is simply meant as a starting point for getting to Alpha, predominantly the changes made in GameServer. This in conjunction with the current client is a good way to visualize the communication between the two systems.


README.md
```diff
@@ -28,6 +33,9 @@ npm run get-cards
 # rm -r build/  # needed if files are added / deleted / renamed between builds
 npm run build
 
+# start Gamenode server
+node ./build/server/gamenode
```
AMMayberry1, line 37: I feel like if we add something about gamenode to the readme we should wait until we can give some complete instructions for a working example. Just this line would probably be kind of confusing since it's not clear what's supposed to happen or what gamenode is

AMMayberry1 (approved):

PR #143, opened 2024-10-28
fix null on data.attributes.name

VJubert:


bin/fetchdata.js
```diff
@@ -20,6 +20,10 @@ function getAttributeNames(attributeList) {
         return attributeList.data.map((attr) => attr.attributes.name.toLowerCase());
     }
 
+    if (attributeList.data === null) {
```
AMMayberry1, line 23: This removes the error but doesn't actually solve the problem, which is that we need to generate a correct json file even if the input data is formatted incorrectly. Right now this creates the token files with no card types, which will cause problems when we implement those cards. It looks like the Palpatine / Sidious leader is also missing some basic data that will cause issues.

Unfortunately we may need to add a helper method that checks for special cases and manually fills in values where we know they're missing.

VJubert: What did you see on palp ? To me I do not see anything missing. Probably some null property will break something but it's a weird card 🤷‍♂️

AMMayberry1: Can we remove this check now, I think it's important that if we a property is null it will throw an error. Otherwise we wouldn't have caught this issue in the first place.

bin/fetchdata.js
```diff
@@ -15,6 +15,18 @@ axiosRetry(axios, {
     retryDelay: () => (Math.random() * 2000) + 1000      // jitter retry delay by 1 - 3 seconds
 });
 
+function populateMissingData(attributes, id) {
+    if (id === '3941784506' || id === '3463348370') {
+        attributes.type = {
+            data: {
+                attributes: {
+                    name: 'token unit'
```
AMMayberry1, line 23: Just checking, did you confirm that the generated jsons in the test folder look correct?

bin/fetchdata.js
```diff
@@ -15,6 +15,18 @@ axiosRetry(axios, {
     retryDelay: () => (Math.random() * 2000) + 1000      // jitter retry delay by 1 - 3 seconds
 });
 
+function populateMissingData(attributes, id) {
+    if (id === '3941784506' || id === '3463348370') {
```
AMMayberry1, line 19: Few suggestions:

1. Let's add comments for the names here
2. Can we change to a switch statement. TS supports stacked cases so you can do:

```typescript
switch (id) {
    case '3941784506':  // clone trooper
    case '3463348370':  // battle droid
        // ...code here...
```

I just guessed at those IDs, please confirm if they're right or not 😂

AMMayberry1 (approved):

PR #144, opened 2024-10-28
Safer handling of triggers in event window

AMMayberry1:
We've had a couple bugs due to the fact that sometimes a trigger window is set to pass its triggers to the parent but the parent is not a window that should have triggers and therefore they resolve at the wrong time or not at all.

In order to catch problems like this early, changed the trigger mode from a boolean to an enum with three options: resolves triggers, passes triggers to parent, should not have triggers. The last option is for "meta-events" which are events that exist in our framework as utilities to support proper game function instead of being actual SWU rules events which abilities could respond to.

In making the above changes I caught a similar issue with the regroup phase where triggered abilities would not be resolving correctly (we just didn't have any abilities triggering on that event yet).

Additionally, updated GameSystem so that it forces every child class to explicitly declare an event name as some systems were not doing so and were emitting "Unnamed" by default. Removed the "Unnamed" event completely and added a category of event names for meta-events that should not be triggering any abilities (MetaEventName).


server/game/core/Player.js
```diff
@@ -414,12 +414,12 @@ class Player extends GameObject {
      * @param {number} numCards
      */
     drawCardsToHand(numCards) {
+        // TODO THIS PR: is this logic still needed?
```
Ennis0, line 417: Why wouldn't it be? Is it handled somewhere else now?

server/game/core/ability/PlayerOrCardAbility.js
```diff
@@ -59,9 +60,15 @@ class PlayerOrCardAbility {
         this.uuid = uuidv4();
 
         // TODO: Ensure that nested abilities(triggers resolving during a trigger resolution) are resolving as expected.
-        this.resolveTriggersAfter = properties.resolveTriggersAfter != null
-            ? properties.resolveTriggersAfter
-            : [AbilityType.Triggered, AbilityType.Action].includes(this.type);
+        this.triggerHandlingMode = properties.triggerHandlingMode;
+
+        if (properties.triggerHandlingMode != null) {
```
Ennis0, line 65: Just making sure you know this is `!=` instead of `!==`. I did a bit of research but I'm still not sure which is more appropriate.

server/game/core/ability/PlayerOrCardAbility.js
```diff
@@ -59,9 +60,15 @@ class PlayerOrCardAbility {
         this.uuid = uuidv4();
 
         // TODO: Ensure that nested abilities(triggers resolving during a trigger resolution) are resolving as expected.
-        this.resolveTriggersAfter = properties.resolveTriggersAfter != null
-            ? properties.resolveTriggersAfter
-            : [AbilityType.Triggered, AbilityType.Action].includes(this.type);
+        this.triggerHandlingMode = properties.triggerHandlingMode;
```
Ennis0, line 63: Looks like this line is redundant with the rest of the function.

server/game/core/event/EventWindow.ts
```diff
@@ -0,0 +1,265 @@
...
+import * as Contract from '../utils/Contract';
+
+export enum TriggerHandlingMode {
+
+    /** Any abilities triggered during this event window will not be resolved immediately but passed to the parent window */
+    PassesTriggersToParentWindow = 'passesTriggersToParentWindow',
+
+    /**
+     * Any abilities triggered during this event window or passed up from child windows will be resolved immediately.
+     * Typically `Auto` should be used instead.
```
Ennis0, line 16: What do you mean by "Auto" here?

server/game/core/ability/PlayerOrCardAbility.js
AMMayberry1: The practice that was recommended to me by a TS dev is to use `!=` and `==` when checking null or undefined since they are sometimes used interchangeably

server/game/core/event/EventWindow.ts
AMMayberry1: That was something I added and then removed in the PR, good catch

PR #146, opened 2024-10-29
Feature/card highlighting

CheBato:
Implemented card highlighting and fixed some bugs regarding selection and tests.


server/game/core/gameSteps/ActionWindow.js
```diff
@@ -32,13 +32,12 @@ class ActionWindow extends UiPrompt {
         }
 
         // IMPORTANT: the below code is referenced in the debugging guide (docs/debugging-guide.md). If you make changes here, make sure to update that document as well.
```
AMMayberry1, line 34: Let's move this comment down to `getCardLegalActions` and we can update the wiki to refer there instead, since that's where the important values are being computed now

server/game/core/gameSteps/ActionWindow.js
```diff
@@ -190,6 +190,20 @@ class ActionWindow extends UiPrompt {
...
+        const allPossibleCards = this.game.findAnyCardsInPlay().concat(
+            this.activePlayer.getCardPile(Location.Discard),
+            this.activePlayer.getCardPile(Location.Resource),
+            this.activePlayer.getCardPile(Location.Hand)
+        );
+        this.activePlayer.setSelectableCards(allPossibleCards.filter((card) => this.getCardLegalActions(card, this.activePlayer).length > 0));
+    }
+
+    getCardLegalActions(card, player) {
+        let actions = card.getActions();
```
AMMayberry1, line 203: Since this is a really common debug point let's split this out a little bit to make the values easier to see:

```typescript
getCardLegalActions(card, player) {
    const actions = card.getActions();
    const legalActions = actions.filter((action) => action.meetsRequirements(action.createContext(player)) === '')
    return legalActions;
}
```

That way folks can just break at the return statement and see `actions` and `legalActions` in the debugger

test/server/cards/02_SHD/units/SalaciousCrumbObnoxiousPet.spec.ts
```diff
@@ -62,6 +62,9 @@ describe('Salacious Crumb, Obnoxious Pet', function() {
                 const { context } = contextRef;
 
                 context.salaciousCrumb.exhausted = true;
+                context.player1.passAction();
+                context.player2.clickCard(context.frontierAtrt);
+                context.player2.clickCard(context.p1Base);
```
AMMayberry1, line 67: What's this for?

test/server/core/card/CardHighlighting.spec.ts
```diff
@@ -0,0 +1,67 @@
+describe('Card highlighting', function() {
```
AMMayberry1, line 1: Let's rename this file to be like core/gameSteps/phases/ActionPhase.spec.ts, since it's testing some pieces of the action phase. "Card highlighting" is pretty generic since cards will be highlighted in the FE basically any time a selection prompt happens

test/server/core/card/CardHighlighting.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        resources: ['armed-to-the-teeth',
+                            'collections-starhopper',
+                            'covert-strength',
+                            'chewbacca#pykesbane',
+                            'battlefield-marine',
+                            'moment-of-peace',
```
AMMayberry1, line 13: protip: can use 'atst' for easy copy-paste

test/server/core/card/CardHighlighting.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+                    player2: {
+                        groundArena: ['death-trooper'],
+                        spaceArena: ['mercenary-gunship']
+                    }
+                });
+            });
+
+            it('the prompt before an action and after should be different.', function () {
+                const { context } = contextRef;
+                let currentPossibleActions = context.player1.currentPrompt();
```
AMMayberry1, line 34: Should always use `const` or `var` instead of `let`, not sure why ts and the linter aren't catching that in some cases

test/server/core/card/CardHighlighting.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+                const reset = () => {
+                    currentPossibleActions = context.player1.currentPrompt();
+                };
+
+                // attack action
+                context.player1.clickCard(context.wampa);
+                context.player1.clickCard(context.deathTrooper);
+                context.player2.claimInitiative();
+                expect(context.player1.currentPrompt()).not.toEqual(currentPossibleActions);
```
AMMayberry1, line 44: We should specifically test that the expected action was removed, could use `.filter()` to select items that are in one array but not the other

test/server/core/gameSteps/actionWindow/Initiative.spec.ts
```diff
@@ -22,16 +22,12 @@ describe('Claiming initiative', function() {
                 context.player1.claimInitiative();
                 expect(context.player2).toHaveExactPromptButtons(['Pass']);
                 expect(context.ardentSympathizer.getPower()).toBe(5);
-                // TODO atm we do not list the set of cards with available actions in the action phase to bypass the error
-                // TODO Expected Player prompt to change.
-                context.player2.clickCardNonChecking(context.scoutBikePursuer);
+                context.player2.clickCard(context.scoutBikePursuer);
```
AMMayberry1, line 25: I know this happened at least a few other times with other folks, do you mind briefly investigating other uses of `clickCardNonChecking` to see if they could be switched out as well?

test/server/core/card/CardHighlighting.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+                reset();
+
+                // smuggle action
+                context.player1.clickCard(context.collectionsStarhopper);
+                expect(context.player1.currentPrompt()).not.toEqual(currentPossibleActions);
+
+                reset();
+
+                // play from hand action
+                context.player1.clickCard(context.vanguardInfantry);
```
AMMayberry1, line 55: I thought we discussed that you were having some issue with clicks not working right because something about making the cards selectable broke the test harness, what ended up happening with that?

server/game/core/gameSteps/ActionWindow.js
CheBato: hah, that's actually what I've been doing to see what's happening here. I would love to break it up.

test/server/cards/02_SHD/units/SalaciousCrumbObnoxiousPet.spec.ts
CheBato: Okay so this is an interesting case. Since Player1 is in the action phase SalaciousCrumb is added as a selectableCard before the exhausted = true. Because he is added as a selectable card the prompt before and after expect(context.salaciousCrumb).not.toHaveAvailableActionWhenClickedBy(context.player1);
is different which fails the check. I'm open to any ideas btw.

test/server/core/card/CardHighlighting.spec.ts
CheBato: will convert to var.

test/server/core/card/CardHighlighting.spec.ts
CheBato: alrighty!

test/server/core/gameSteps/actionWindow/Initiative.spec.ts
CheBato: I'd love to.

test/server/core/card/CardHighlighting.spec.ts
CheBato: Well i haven't run into that issue anymore ever since I refactored how clickCards worked. I must have removed some logic and returned it back which fixed the issue.

server/game/core/gameSteps/ActionWindow.js
CheBato: yessir!

test/server/core/card/CardHighlighting.spec.ts
CheBato: will keep that in mind for next time!

test/server/core/card/CardHighlighting.spec.ts
CheBato: do you have an example of this perhaps?

test/server/core/card/CardHighlighting.spec.ts
CheBato: yessir!

test/server/core/card/CardHighlighting.spec.ts
AMMayberry1: oh, huh. I'll take a look at that at some point

test/server/core/card/CardHighlighting.spec.ts
AMMayberry1: Assuming we have arrays for before and after the action, I think the easiest way to do this check without implementing full array difference would be:

```typescript
expect(beforeActions.length - afterActions.length).toBe(1);
expect(beforeActions.filter((beforeAction) => !afterActions.includes(beforeAction));
```

I'm realizing that's a little slow though since it's M * N complexity. so probably easiest just to determine the name of the action and explicitly check that it's not in the list anymore

CheBato: gotcha!

test/server/core/gameSteps/phases/ActionPhase.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+                expect(context.player1.currentActionTargets).not.toContain(context.collectionsStarhopper);
+
+                // play from hand action
+                context.player1.clickCard(context.vanguardInfantry);
+                expect(context.player1.currentActionTargets).not.toContain(context.vanguardInfantry);
+
+                // steal mercenary gunship TODO wait till the gunship is implemented
+                /* expect(context.player1.countSpendableResources()).toBe(4);
+                context.player1.clickCard(context.mercenaryGunship);
+                expect(context.player1.currentPrompt()).not.toEqual(currentPossibleActions);*/
```
AMMayberry1, line 52: One last thing - could you also add a TODO for checking play from discard actions when those are implement

CheBato: resolved

AMMayberry1 (approved):

PR #147, opened 2024-10-29
Clean up methods for event generation

AMMayberry1:
Previously the `GameSystem.generateEvent` method accepted a `target` parameter which would override the `target` property of a system. This made the correct use of the method very confusing since it was unclear which version of `target` should be used, and even moreso since the behavior would be slightly different for each in non-obvious ways such. One such issue was whether or not the value passed to `canAffect()` would be an array or not.

Made two design improvements to improve clarity and usability:
- Renamed `generateEvent()` to `generateRetargetedEvent()` which accepts a target with which to override the system (current behavior), and added a new `generateEvent()` signature which does not accept a target but generates the event with the configured target using `generatePropertiesFromContext`
- Refactored `CardTargetSystem` to enforce that `event.target` must always be a single value at event creation time instead of an array so that calls to `canAffect()` for a card target system are never receiving an array

Also added some additional contract checks to key methods in GameSystem for better fail-fast behavior.


server/game/core/gameSystem/GameSystem.ts
```diff
@@ -196,11 +199,26 @@ export abstract class GameSystem<TContext extends AbilityContext = AbilityContex
...
     }
 
+    /**
+     * Generates one {@link GameEvent} object that will apply the effects of this system to the game state
+     * for the specified target.
+     * The event must be emitted using an {@link EventWindow}, typically via `Game.openEventWindow`.
+     * @param context Context of ability being executed
+     * @param additionalProperties Any additional properties to extend the default ones with
+     */
+    public generateEvent(context: TContext, additionalProperties: any = {}): GameEvent {
```
garrettdc2, line 214: Instead of using just `any` here, we can probably use a more specific type like `Record<string, any>` since we expect a hash.

garrettdc2 (dismissed):

PR #148, opened 2024-10-30
Last starter deck card: Resupply

AMMayberry1:


test/server/cards/01_SOR/events/Resupply.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+        describe('Resupply\'s ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['resupply'],
+                        leader: 'leia-organa#alliance-general'
+                    },
+                    player2: {
+                    }
```
garrettdc2, line 12: Nit (non-blocking): Do we need to define player2 here? I believe there are defaults

garrettdc2 (dismissed):

PR #150, opened 2024-10-30
Bunch of cards 5

VJubert:


server/game/cards/02_SHD/leaders/FennecShandHonoringTheDeal.ts
```diff
@@ -0,0 +1,51 @@
...
+            internalName: 'fennec-shand#honoring-the-deal',
+        };
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addActionAbility({
+            title: 'Play a unit that costs 4 or less from your hand. Give it ambush for this phase',
+            cost: [AbilityHelper.costs.abilityResourceCost(1), AbilityHelper.costs.exhaustSelf()],
+            targetResolver: {
+                cardCondition: (card) => card.isUnit() && card.cost <= 4,
```
AMMayberry1, line 18: Can remove the `isUnit()` from the card condition since you have a filter for it

server/game/cards/02_SHD/upgrades/VambraceFlamethrower.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export default class VambraceFlamethrower extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '6471336466',
+            internalName: 'vambrace-flamethrower',
+        };
+    }
+
+    public override setupCardAbilities() {
```
AMMayberry1, line 13: Need to add the vehicle attach restriction and add a test for it. See any of the lightsabers for an example.

server/game/cards/02_SHD/upgrades/VambraceFlamethrower.ts
```diff
@@ -0,0 +1,27 @@
...
+        this.addGainOnAttackAbilityTargetingAttached({
+            title: 'Deal 3 damage divided as you choose among enemy ground units',
+            immediateEffect: AbilityHelper.immediateEffects.distributeDamageAmong({
+                amountToDistribute: 3,
+                canChooseNoTargets: true,
```
AMMayberry1, line 18: This is a kind of a niche rules thing, but this should only be true in cases where the card says "any number of units" like overwhelming barrage. In this case picking at least one target is required

test/server/cards/01_SOR/units/ZebOrreliosHeadstrongWarrior.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+                    }
+                });
+            });
+
+            it('should deal 4 damage to a ground unit when he kill someone and survives', function () {
+                const { context } = contextRef;
+
+                // kill superlaser technician, should deal 4 to a ground unit
+                context.player1.clickCard(context.zebOrrelios);
+                context.player1.clickCard(context.superlaserTechnician);
```
AMMayberry1, line 22: Can we switch to a unit without a defeat prompt to make this simpler

test/server/cards/01_SOR/units/ZebOrreliosHeadstrongWarrior.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+                context.player2.clickPrompt('Put Superlaser Technician into play as a resource and ready it');
+                expect(context.player1).toBeAbleToSelectExactly([context.zebOrrelios, context.swoopRacer, context.consularSecurityForce, context.steadfastBattalion]);
+                expect(context.player1).toHaveChooseNoTargetButton();
+                context.player1.clickCard(context.consularSecurityForce);
+                expect(context.consularSecurityForce.damage).toBe(4);
+                expect(context.player2).toBeActivePlayer();
+
+                // reset
+                context.zebOrrelios.exhausted = false;
```
AMMayberry1, line 31: To make the test easier to read I recommend using the `reset()` pattern we've started adopting. See tests for "Han Solo, Reluctant Hero" for an example.

test/server/cards/01_SOR/units/ZebOrreliosHeadstrongWarrior.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+                context.setDamage(context.zebOrrelios, 0);
+                context.setDamage(context.consularSecurityForce, 0);
+                context.player2.passAction();
+
+                // consular security force is not defeat, nothing happen
+                context.player1.clickCard(context.zebOrrelios);
+                context.player1.clickCard(context.consularSecurityForce);
+                expect(context.player2).toBeActivePlayer();
```
AMMayberry1, line 39: Few more cases it would be good to have since his trigger is kind of complicated:

- Enemy unit attacks him and dies, ability not triggered
- Another friendly unit defeats an enemy unit, ability not triggered

And one additional one that's kind of complicated: Zeb's trigger doesn't say that _he himself_ has to defeat the unit like Mace Windu does, it just says the unit has to _be defeated_. So technically if the defender is defeated by some "on attack" ability, Zeb's ability should still activate.

It's actually perfect that  you also implemented Vambrace Flamethrower in this PR because that is exactly the way to test this. To summarize - let's have a test where Zeb has Flamethrower attached, attacks a unit and defeats it with Flamethrower ability, and then check that Zeb ability still activates. That could maybe be its own `describe` block since it's a pretty different setup.

test/server/cards/02_SHD/events/LookTheOtherWay.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+            });
+
+            it('should exhaust an enemy unit without choice because opponent does not have sufficient resources', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.lookTheOtherWay);
+                expect(context.player1).toBeAbleToSelectExactly([context.viperProbeDroid, context.wampa]);
+                context.player1.clickCard(context.viperProbeDroid);
+                expect(context.viperProbeDroid.exhausted).toBeTrue();
+                expect(context.player2).toHavePrompt('Choose an action');
```
AMMayberry1, line 94: Let's use `expect(context.player2).toBeActivePlayer();` instead, it's a bit clearer in meaning. Same for any other test doing something similar

test/server/cards/02_SHD/events/LookTheOtherWay.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+                        groundArena: ['viper-probe-droid'],
+                        base: 'echo-base',
+                        leader: 'sabine-wren#galvanized-revolutionary',
+                        hasInitiative: true,
+                        resources: 3,
+                    }
+                });
+            });
+
+            it('should exhaust an enemy unit without choice because opponent does not have sufficient ready resources', function () {
```
AMMayberry1, line 118: Not sure what makes this test different than the one above, can you clarify?

test/server/cards/02_SHD/events/LookTheOtherWay.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.viperProbeDroid, context.wampa]);
+                context.player1.clickCard(context.viperProbeDroid);
+
+                expect(context.player2).toHaveEnabledPromptButtons(['Exhaust Viper Probe Droid', 'Pay 2 resources']);
+                context.player2.clickPrompt('Pay 2 resources');
+                expect(context.viperProbeDroid.exhausted).toBeFalse();
+                expect(context.player2.countExhaustedResources()).toBe(2);
+            });
+
+            it('should does not exhaust unit because controller pays 2 resources (friendly)', function () {
```
AMMayberry1, line 57: Awesome tests!! These cover everything I can think of

test/server/cards/02_SHD/events/TimelyIntervention.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+                        groundArena: ['battlefield-marine'],
+                        resources: ['tech#source-of-insight', 'atst', 'atst', 'atst']
+                    },
+                    player2: {
+                        groundArena: ['isb-agent'],
+                    }
+                });
+            });
+
+            it('should allow the player to play a unit from his hand, paying its cost, and give it ambush', function () {
```
AMMayberry1, line 19: Could we also have a test where we smuggle it just to confirm that works? I don't feel the need to do that with every smuggle card but the timing is a little tricky on this one so I just want some additional confirmation that it works right

test/server/cards/02_SHD/leaders/FennecShandHonoringTheDeal.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+                        groundArena: ['isb-agent'],
+                    }
+                });
+            });
+
+            it('should allow the player to play a unit with cost 4 or less, paying its cost, and give it ambush', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.fennecShand);
+                expect(context.player1).toBeAbleToSelectExactly([context.rebelPathfinder, context.allianceXwing, context.moddedCohort]);
```
AMMayberry1, line 23: Can you change modded cohort to be a red 4-cost card so we can confirm that it's looking at printed cost without aspect penalties

test/server/cards/02_SHD/leaders/FennecShandHonoringTheDeal.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+                context.player1.clickCard(context.p2Base);
+
+                context.player2.passAction();
+
+                // use fennec ability to play a unit with ambush (even if fennec is exhausted)
+                context.player1.clickCard(context.fennecShand);
+                expect(context.player1).toBeAbleToSelectExactly([context.allianceXwing, context.moddedCohort]);
+                context.player1.clickCard(context.moddedCohort);
+
+                // play a unit who already has Ambush
```
AMMayberry1, line 90: These are great! There's one additional test it would be really good to have. This is one of the few actions in the game with no cost, so it gives us a chance to test that our logic for determining whether an action is allowed works right.

Can you add one additional test (could be a separate `describe`) where she is deployed and the player has no cards in hand that are targetable with her ability? Then we need to confirm that when clicked on, the game doesn't give the player the option to use her ability at all. Normally an action ability is always available b/c the cost can change game state, but obviously this is a different situation.

test/server/cards/02_SHD/upgrades/Ruthlessness.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+                    player1: {
+                        groundArena: [{ card: 'battlefield-marine', upgrades: ['ruthlessness'] }],
+                    },
+                    player2: {
+                        groundArena: ['ardent-sympathizer']
+                    }
+                });
+            });
+
+            it('should deal 2 damage to base when defeating a enemy unit (and die)', function () {
```
AMMayberry1, line 67: Would be good to roll these all into one, seems like that should be possible?

test/server/cards/02_SHD/upgrades/Ruthlessness.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+                    player1: {
+                        groundArena: [{ card: 'battlefield-marine', upgrades: ['ruthlessness'] }],
+                    },
+                    player2: {
+                        groundArena: ['specforce-soldier']
+                    }
+                });
+            });
+
+            it('should deal 2 damage to base when defeating a enemy unit (and survives)', function () {
```
AMMayberry1, line 41: We should add the same set of tests that Mace has since the ability is basically the same. There's a couple missing here

test/server/cards/01_SOR/units/ZebOrreliosHeadstrongWarrior.spec.ts
VJubert: No, I want to test if a unit doesn't go to discard the ability still works

test/server/cards/02_SHD/events/LookTheOtherWay.spec.ts
VJubert: I don't know if there is a difference between do not have resource to pay (total resource) and do have total resource to pay but some are exhausted

server/game/cards/02_SHD/leaders/FennecShandHonoringTheDeal.ts
VJubert: I can't, cost is not resolved if I deleted isUnit()

test/server/cards/02_SHD/events/LookTheOtherWay.spec.ts
VJubert: does it check the player is really the active player or does it check only a prompt ? because here he can have a prompt but not the wanted one

AMMayberry1: It checks if they're really the active player

test/server/cards/02_SHD/events/LookTheOtherWay.spec.ts
AMMayberry1: Oh nah not really, I don't think there's enough of a distinction to be worth having separate tests for. If that was a problem then like 100 other tests would break at least

server/game/cards/01_SOR/units/MaceWinduPartyCrasher.ts
```diff
@@ -14,6 +14,7 @@ export default class MaceWinduPartyCrasher extends NonLeaderUnitCard {
             title: 'Ready Mace Windu',
             when: {
                 onCardDefeated: (event, context) =>
+                    // TODO: update trigger condition so that defender being defeated by attacker at any point during the attack will work
```
AMMayberry1, line 17: Sorry I'm realizing I wrote this in an ambiguous way - do you mind updating instances of this to instead say `... being defeated by attacker at the 'on attack' stage will also work`. That way we'll know clearly what the issue is when we come back around to it.

server/game/cards/02_SHD/upgrades/VambraceFlamethrower.ts
```diff
@@ -11,11 +12,14 @@ export default class VambraceFlamethrower extends UpgradeCard {
...
 
     public override setupCardAbilities() {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
         this.addGainOnAttackAbilityTargetingAttached({
             title: 'Deal 3 damage divided as you choose among enemy ground units',
             immediateEffect: AbilityHelper.immediateEffects.distributeDamageAmong({
                 amountToDistribute: 3,
-                canChooseNoTargets: true,
+                optional: true,
```
AMMayberry1, line 21: Ah sorry, I was unclear - I meant for "optional" to be added at the ability level, not inside of `distributeDamageAmong`. Using optional inside of an `immediateEffect` is not well-supported, I need to decide if we just want to remove it altogether.

I think the change you made to `DistributeAmongTargetsSystem` can still work though, see my comment there

server/game/gameSystems/DistributeAmongTargetsSystem.ts
```diff
@@ -61,7 +61,7 @@ export abstract class DistributeAmongTargetsSystem<TContext extends AbilityConte
         const legalTargets = properties.selector.getAllLegalTargets(context);
 
         // auto-select if there's only one legal target and the player isn't allowed to choose 0 targets
-        if (!properties.canChooseNoTargets && legalTargets.length === 1) {
+        if ((!properties.canChooseNoTargets && !properties.optional) && legalTargets.length === 1) {
```
AMMayberry1, line 64: If we make the change I suggested in the Vambrace Flamethrower file, we can still do this if we instead use `context.ability.optional`. Same for change on line 74

test/server/cards/01_SOR/units/ZebOrreliosHeadstrongWarrior.spec.ts
```diff
@@ -28,26 +36,60 @@ describe('Zeb Orrelios, Headstrong Warrior', function () {
...
+                });
+            });
+
+            it('should deal 4 damage to a ground unit when zeb attacks and kill with on attack abilities', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.zebOrrelios);
+                context.player1.clickCard(context.battlefieldMarine);
+
+                // kill battlefield marine with flamethrower, zeb should deal 4 damage to another unit
```
AMMayberry1, line 81: Awesome!! Thanks again for adding this and catching that rules issue, very helpful

test/server/cards/02_SHD/leaders/FennecShandHonoringTheDeal.spec.ts
```diff
@@ -99,5 +99,29 @@ describe('Fennec Shand, Honoring the Deal', function () {
...
+                        resources: 8
+                    },
+                    player2: { }
+                });
+            });
+
+            it('should not allow the player to play a unit with cost 4 or less when he does not have any targetable units in hand', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.fennecShand);
```
AMMayberry1, line 119: Could we simplify / clarify this test a little bit by setting her as exhausted to start (can just `exhausted: true` in setup) and then do `expect(context.fennecShand).not.toHaveAvailableActionWhenClickedBy(context.player1);`

server/game/cards/01_SOR/units/ZebOrreliosHeadstrongWarrior.ts
```diff
@@ -0,0 +1,41 @@
...
+            title: 'If the defender was defeated, you may deal 4 damage to a ground unit',
+            when: {
+                onAttackCompleted: (event, context) => event.attack.attacker === context.source,
+            },
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                locationFilter: Location.GroundArena,
+                optional: true,
+                immediateEffect: AbilityHelper.immediateEffects.conditional({
+                    condition: (context) => context.event.attack.target.isUnit() && this.unitsDefeatedThisPhaseWatcher.getDefeatedUnitsControlledByPlayer(context.source.controller.opponent).includes(context.event.attack.target),
```
AMMayberry1, line 32: nit: can we break this into multiple lines

test/server/cards/02_SHD/upgrades/Ruthlessness.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+                    player1: {
+                        groundArena: [{ card: 'battlefield-marine', upgrades: ['ruthlessness'] }, 'wampa'],
+                    },
+                    player2: {
+                        groundArena: ['atst', 'specforce-soldier', 'wilderness-fighter', 'regional-governor']
+                    }
+                });
+            });
+
+            it('should not deal 2 damage to base when not defeating a enemy unit', function () {
```
AMMayberry1, line 16: The naming for this test seems wrong since we are testing some cases where the ability triggers successfully

test/server/cards/02_SHD/upgrades/VambraceFlamethrower.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                    [context.atst, 2],
+                    [context.specforceSoldier, 1],
+                ]));
+
+                expect(context.atst.damage).toBe(2);
+                expect(context.specforceSoldier.damage).toBe(1);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should be able to put all damage on a single target and exceed its HP total', function () {
```
AMMayberry1, line 35: Since we're adding a new way to make this optional, can we add a specific test that the pass button appears and works right

server/game/cards/02_SHD/upgrades/VambraceFlamethrower.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addGainOnAttackAbilityTargetingAttached({
+            title: 'Deal 3 damage divided as you choose among enemy ground units',
+            immediateEffect: AbilityHelper.immediateEffects.distributeDamageAmong({
+                amountToDistribute: 3,
+                optional: true,
+                canChooseNoTargets: false,
```
AMMayberry1, line 22: Extending the above comment re: the optional parameter. To help people catch this in the future, can you go to GameSystem.ts and change the TODO comment on line 17 to be an @deprecated so that people will see it crossed out:

```typescript
/** @deprecated TODO: evaluate whether to remove this */ 
optional?: boolean;
```

test/server/cards/02_SHD/upgrades/VambraceFlamethrower.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+                context.battlefieldMarine.exhausted = false;
+                context.player2.passAction();
+
+                // battlefield marine attacks
+                context.player1.clickCard(context.battlefieldMarine);
+                context.player1.clickCard(context.p2Base);
+
+                // pass vambrace flamethrower ability
+                expect(context.player1).toHaveExactPromptButtons([flamethrowerPrompt, 'Pass']);
```
AMMayberry1, line 46: Quick question: does this work if you instead do `expect(context.player1).toHavePassAbilityPrompt(flamethrowerPrompt);`? If so I'd prefer to switch over to using that throughout the file, as the intent is clearer

AMMayberry1 (approved):

PR #152, opened 2024-11-01
Add Discard Card System

garrettdc2:
Adds a system to discard cards. Also implements Reinforcement Walker (SOR 119) for testing purposes.

https://swudb.com/card/SOR/119/reinforcement-walker
image


server/game/cards/01_SOR/units/ReinforcementWalker.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    private deckHasCards(context) {
+        return context.source.controller.drawDeck.length > 0;
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Look at the top card of your deck',
+            when: {
+                onCardPlayed: (event, context) => event.card === context.source && this.deckHasCards(context),
```
AMMayberry1, line 22: The `deckHasCards` check should be changed, probably to be some form of `thenCondition` inside the `then` ability like in C-3PO. The reason is that the trigger condition here is evaluated some time before the actual resolution of the ability, so it's possible that game state could change in between.

E.g., in this case, there could be some other (hypothetical) triggered ability resolved before this one that puts cards back in the deck, so that by the time this triggered ability would resolve there are cards again and it actually could resolve correctly but we've bypassed it. So our practice is to just have this condition match the bold trigger text as exactly as possible and put conditionals in the body of the ability to halt resolution if certain conditions aren't met.

server/game/cards/01_SOR/units/ReinforcementWalker.ts
```diff
@@ -0,0 +1,48 @@
...
+                (context) => ({ target: context.source.controller.getTopCardOfDeck() })
+            ),
+            then: (thenContext) => {
+                const topCardOfDeck = thenContext.source.controller.getTopCardOfDeck();
+                return {
+                    title: `Draw "${topCardOfDeck.title}" or discard it and heal 3 damage from your base.`,
+                    targetResolver: {
+                        mode: TargetMode.Select,
+                        choices: {
+                            ['Draw']: AbilityHelper.immediateEffects.draw(),
```
AMMayberry1, line 35: I can't imagine a situation where it would matter, but just to guard against bizarre edge cases it might be better to use `drawSpecificCard()`.

server/game/gameSystems/DiscardCardSystem.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface IDiscardProperties extends ICardTargetSystemProperties {}
+
+export class DiscardCardSystem<TContext extends AbilityContext = AbilityContext> extends CardTargetSystem<TContext, IDiscardProperties> {
+    public override readonly name = 'discardCard';
+    public override readonly eventName = EventName.OnCardsDiscarded;
+
+    public override canAffect(card: Card | Card[], context: TContext, additionalProperties: Record<string, any> = {}): boolean {
+        /* TODO A bug currently leads to this sometimes being an array. Once the bug is fixed, we can expect this to
```
AMMayberry1, line 15: Can you try removing this and see if it's okay now? The most recent PR I merged should fix it, if not then I need to do more investigation and you can proceed with this in

server/game/gameSystems/DiscardCardSystem.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    public override getEffectMessage(context: TContext): [string, any[]] {
+        const properties = this.generatePropertiesFromContext(context);
+        return [
+            'discard {0}',
+            [properties.target]
+        ];
+    }
+
+    public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext, additionalProperties: Record<string, any> = {}): void {
```
AMMayberry1, line 30: Typically systems that only target one card don't need their own implementation of `queueGenerateEventGameSteps` because it's handled in the superclass. Was something specific happening that made you add this (and does it work without it?)

server/game/gameSystems/DiscardCardSystem.ts
```diff
@@ -0,0 +1,47 @@
...
+        if (cards.length === 0) {
+            return;
+        }
+        const event = this.createEvent(null, context, additionalProperties);
+        this.updateEvent(event, target, context, additionalProperties);
+        events.push(event);
+    }
+
+    public eventHandler(event): void {
+        event.card.forEach((card) => {
```
AMMayberry1, line 43: Same general comment here as the other places - ideally the behavior should be fixed now and this should always be a non-array

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -114,9 +115,9 @@ export function deploy<TContext extends AbilityContext = AbilityContext>(propert
 export function defeat<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IDefeatCardProperties, TContext> = {}): CardTargetSystem<TContext> {
     return new DefeatCardSystem<TContext>(propertyFactory);
 }
-// export function discardCard(propertyFactory: PropsFactory<DiscardCardProperties> = {}): CardGameAction {
-//     return new DiscardCardAction(propertyFactory);
-// }
+export function discardCard<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IDiscardProperties, TContext> = {}): DiscardCardSystem<TContext> {
```
AMMayberry1, line 118: Since we're going to end up with a plethora of discard systems we should probably be super specific with the naming. Maybe in the vein of `drawSpecificCard` we make this `discardSpecificCard`? And same for the class itself. Lmk what you think

test/server/core/ability/abilityTargets/CardTargetResolver.spec.ts
```diff
@@ -6,7 +6,7 @@ describe('Target resolvers for cards', function() {
                     phase: 'action',
                     player1: {
                         base: 'energy-conversion-lab',
-                        hand: ['reinforcement-walker', 'rebel-pathfinder', 'alliance-xwing', 'atst']
+                        hand: ['battlefield-marine', 'rebel-pathfinder', 'alliance-xwing', 'atst']
```
AMMayberry1, line 9: Just curious, what happened with this test that broke it?

test/server/cards/01_SOR/units/ReinforcementWalker.spec.ts
```diff
@@ -0,0 +1,118 @@
+import { Location } from '../../../../../server/game/core/Constants';
+
+describe('Reinforcement Walker', function() {
+    integration(function(contextRef) {
+        describe('Reinforcement Walker\'s ability', function() {
+            it('should let the play draw a card on attack when played and on attack', function () {
```
AMMayberry1, line 6: nit, typo: 'on attack when played and on attack'

test/server/cards/01_SOR/units/ReinforcementWalker.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+                        'Draw',
+                        'Discard',
+                    ]);
+
+                    context.player1.clickPrompt('Discard');
+                    expect(context.battlefieldMarine).toBeInLocation(Location.Discard);
+                    expect(context.p1Base.damage).toEqual(4);
+                    expect(context.player2).toBeActivePlayer();
+                }
+            );
```
AMMayberry1, line 92: Great test cases! Just as sort of a mini integration test, do you mind adding one more case where it gains ambush and confirm that the prompts behave as expected

server/game/gameSystems/DiscardCardSystem.ts
```diff
@@ -0,0 +1,47 @@
...
+import { EventName, Location } from '../core/Constants';
+import { GameEvent } from '../core/event/GameEvent';
+import { CardTargetSystem, ICardTargetSystemProperties } from '../core/gameSystem/CardTargetSystem';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface IDiscardProperties extends ICardTargetSystemProperties {}
+
+export class DiscardCardSystem<TContext extends AbilityContext = AbilityContext> extends CardTargetSystem<TContext, IDiscardProperties> {
+    public override readonly name = 'discardCard';
+    public override readonly eventName = EventName.OnCardsDiscarded;
```
AMMayberry1, line 12: Since all forms of discard will probably use this same event type, we should add some additional metadata to the event in case there are ever abilities that care where the card is discarded from. We can add additional properties to the event by overriding `addPropertiesToEvent` (need to be sure to call the superclass method). The we can just do like `event.discardedFromLocation = ...`

server/game/gameSystems/DiscardCardSystem.ts
garrettdc2: Discussed offline. Still an issue

garrettdc2: nvm will be fixed

server/game/gameSystems/GameSystemLibrary.ts
garrettdc2: Yup done

test/server/core/ability/abilityTargets/CardTargetResolver.spec.ts
AMMayberry1: Following up on the discussion from chat - just so we remember to come back to this, can you leave the original version in but commented out with a TODO about the bug?

server/game/gameSystems/DiscardSpecificCardSystem.ts
```diff
@@ -0,0 +1,34 @@
...
+            [properties.target]
+        ];
+    }
+
+    protected override addPropertiesToEvent(event, card: Card, context: TContext, additionalProperties: Record<string, any> = {}): void {
+        event.discardedFromLocation = card.location;
+        super.addPropertiesToEvent(event, card, context, additionalProperties);
+    }
+
+    public eventHandler(event): void {
```
AMMayberry1, line 31: nit: sorry I didn't catch this before, by convention we're putting this as the first method in the class since it's usually the most important

test/server/core/ability/abilityTargets/CardTargetResolver.spec.ts
AMMayberry1: Just a bump on the above comment in case it got missed

garrettdc2: Done

AMMayberry1 (approved):

PR #153, opened 2024-11-01
Chebato/setup phase

CheBato:
- Added the setup phase test suite


test/helpers/IntegrationHelper.js
```diff
@@ -836,22 +840,24 @@ global.integration = function (definitions) {
...
-                this.player1.setSpaceArenaUnits(options.player1.spaceArena, ['removed from game']);
-                this.player2.setSpaceArenaUnits(options.player2.spaceArena, ['removed from game']);
-
-                // Hand + discard
-                this.player1.setHand(options.player1.hand, ['removed from game']);
-                this.player2.setHand(options.player2.hand, ['removed from game']);
-                this.player1.setDiscard(options.player1.discard, ['removed from game']);
-                this.player2.setDiscard(options.player2.discard, ['removed from game']);
 
+                if (options.phase !== 'setup') {
```
AMMayberry1, line 844: This is a good start. One thing we need to consider though is that people might get confused and still try to set some of the configuration settings not realizing that they don't apply in setup phase tests. So it would be good to have some feedback mechanism for letting them know. 

Probably ought to be some kind of parameter checks where we throw an error if they've done something that doesn't apply in this type of test. This would apply not only for putting in cards in arenas like in this if block, but also if they try to set leader status deployed or base damage in the below settings.

I've honestly been meaning for a while to add some validation checks on the `options.playerX` config objects, mostly to catch if someone has a typo in their config and accidentally puts `groundAerna` or something because that would just silently fail. Maybe this is a good chance to add a validation method that can handle the normal checks for action phase tests and then the additional ones for setup phase tests as well.

test/server/core/gameSteps/phases/SetupPhase.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+                // Draw cards step
+                expect(context.player1.handSize).toBe(6);
+                expect(context.player2.handSize).toBe(6);
+                const beforeMulliganHand = context.player1.hand;
+
+                // Mulligan step
+                context.player1.clickPrompt('yes');
+                context.player2.clickPrompt('no');
+                const afterMulliganHand = context.player1.hand;
+                expect(beforeMulliganHand).not.toEqual(afterMulliganHand);
```
AMMayberry1, line 57: Also need to check that player2's hand has not changed

test/server/core/gameSteps/phases/SetupPhase.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+                // select 2 cards to resource
+                context.player1.clickCard(context.player1.hand[0]);
+                context.player1.clickCard(context.player1.hand[1]);
+
+                // Check if selecting any unavailable cards triggers resourcing
+                context.player2.clickCardNonChecking(context.player2.deck[0]);
+                context.player2.clickCardNonChecking(context.player1.hand[0]);
+                context.player2.clickCardNonChecking(context.player1.deck[0]);
+                context.player2.clickCardNonChecking(context.p1Base);
+                context.player2.clickCardNonChecking(context.p2Base);
```
AMMayberry1, line 76: Need to add some expectation after these clicks to confirm they didn't have any effect. Probably something checking the prompt selection state?

test/server/core/gameSteps/phases/SetupPhase.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+                // Select 2 correct cards to resource
+                context.player2.clickCard(context.player2.hand[0]);
+                context.player2.clickCard(context.player2.hand[1]);
```
AMMayberry1, line 80: So this is something I forgot - currently the setup phase resource prompt doesn't have a "Done" button, it just automatically resources once both players have clicked 2 cards. I think that's probably not what we want, it would be better to ask players to click "done" so they can lock in their choice.

Can you make that update to the ResourcePrompt? You can copy some of the logic from VariableResourcePrompt, or maybe even move it completely since ResourcePrompt is the parent class and VariableResourcePrompt will just inherit it.

Note that making this change will also require updating the test setup stuff since now the test setup will need to additionally click "Done" for each player

test/server/core/gameSteps/phases/SetupPhase.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+                // Check if resource length is correct
+                expect(context.player1.resources.length).toBe(2);
+                expect(context.player2.resources.length).toBe(2);
+
+                // Check if resources are correctly set
```
AMMayberry1, line 86: Can we check for both players

test/server/core/gameSteps/phases/SetupPhase.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+                    }
+                });
+            });
+
+            it('should flow normally through each step', function () {
+                const { context } = contextRef;
+
+                // The setup phase is divided into 3 or 4 steps: choose Initiative player, Draw card step, mulligan step (optional), resource step
+                // Choose Initiative Step
+                context.selectInitiativePlayer(context.player1);
```
AMMayberry1, line 46: Can we add a second (smaller) test to confirm that it works if player 2 is the initiative player

test/helpers/IntegrationHelper.js
CheBato: hell yeah i'll write some up and you can check if they're okay cool?

test/server/core/gameSteps/phases/SetupPhase.spec.ts
CheBato: gotcha!

test/server/core/gameSteps/phases/SetupPhase.spec.ts
CheBato: yeah probably we can do that the prompt hasn't changed after each for both players.

test/server/core/gameSteps/phases/SetupPhase.spec.ts
CheBato: sure i can add the Done to the resource phase.

test/server/core/gameSteps/phases/SetupPhase.spec.ts
CheBato: ofcourse

test/server/core/gameSteps/phases/SetupPhase.spec.ts
CheBato: sure no problem :)

test/server/core/gameSteps/phases/SetupPhase.spec.ts
danbastin: Agreed here. In general our approach is that multiple selection will require a confirmation while single selection automatically submits. Might revise that as we see more cases but certainly for this one we want a done button. Do we also have the ability to "un-resource" a card?

test/helpers/IntegrationHelper.js
AMMayberry1: Yup sounds good!

test/server/core/gameSteps/phases/SetupPhase.spec.ts
AMMayberry1: The resourcing doesn't happen until the prompt resolves so shouldn't be needed, if I'm understanding the intent right

CheBato: if i understand @danbastin correctly its whether we support unselecting cards. If we don't, I can make an addition to the ResourcePrompts clickCard function to check whether the card is already in the selectedCards and remove it from there, if that makes sense.

server/game/core/gameSteps/prompts/ResourcePrompt.ts
```diff
@@ -8,18 +8,19 @@ import { AllPlayerPrompt } from './AllPlayerPrompt';
...
     protected selectableCards = new Map<string, Card[]>();
+    protected playersDone = new Map<string, boolean>();
 
     public constructor(
         game: Game,
         private readonly nCardsToResource: number
     ) {
         super(game);
         game.getPlayers().forEach((player) => this.selectedCards[player.name] = []);
+        game.getPlayers().forEach((player) => this.playersDone[player.name] = false);
```
AMMayberry1, line 19: Let's just roll these two into one regular `for` loop

server/game/core/gameSteps/prompts/ResourcePrompt.ts
```diff
@@ -65,10 +66,10 @@ export class ResourcePrompt extends AllPlayerPrompt {
         Contract.assertNotNullLike(player);
         Contract.assertNotNullLike(card);
 
-        if (!this.activeCondition(player)) {
+        if (!this.activeCondition(player) || !player.hand.includes(card) ||
+          this.selectedCards[player.name].length === this.nCardsToResource) {
```
AMMayberry1, line 70: Just for style consistency could you break out the lines like this:

```typescript
  if (
    !this.activeCondition(player) || !player.hand.includes(card) ||
     this.selectedCards[player.name].length === this.nCardsToResource
) {
```

server/game/core/gameSteps/prompts/ResourcePrompt.ts
```diff
@@ -86,7 +87,20 @@ export class ResourcePrompt extends AllPlayerPrompt {
     }
 
     public override menuCommand(player, arg): boolean {
-        Contract.fail(`Unexpected menu command: '${arg}'`);
+        if (arg === 'done') {
```
AMMayberry1, line 90: Can we still have the Contract.fail if the command is something else, since that would mean they somehow clicked a button that doesn't exist

server/game/core/gameSteps/prompts/ResourcePrompt.ts
```diff
@@ -86,7 +87,20 @@ export class ResourcePrompt extends AllPlayerPrompt {
...
+            if (this.completionCondition(player)) {
+                return false;
+            }
+            if (this.selectedCards[player.name].length < this.nCardsToResource) {
+                return false;
+            }
+
+            this.resourceSelectedCards(player);
```
AMMayberry1, line 98: If I'm reading this right, looks like it will resource the cards for a player as soon as they click done regardless of the other player's finished state. I think we should wait to resource until both players are done.

test/helpers/GameFlowWrapper.js
```diff
@@ -56,6 +56,7 @@ class GameFlowWrapper {
     resourceAnyTwo() {
         this.guardCurrentPhase('setup');
         this.allPlayersInInitiativeOrder().forEach((player) => player.clickAnyOfSelectableCards(2));
+        this.allPlayersInInitiativeOrder().forEach((player) => player.clickPrompt('Done'));
```
AMMayberry1, line 59: Similar comment as above - let's just roll these two into one `for` loop

test/helpers/IntegrationHelper.js
```diff
@@ -754,6 +754,42 @@ function generatePromptHelpMessage(player) {
     return `Current prompt for ${player.name}:\n${formatPrompt(player.currentPrompt(), player.currentActionTargets)}`;
 }
 
+function validatePlayerOptions(playerOptions, playerName, startPhase) {
+    // list of approved property names
+    const listOfApprovedPropertiesForAnyOtherPhase = [
```
AMMayberry1, line 759: Let's say "ForNonSetupPhase" so this is understandable without seeing the other variable name. Also, it's personal preference but I think "listOf" in the name is kind of uncessary, if the name is plural it implies a list imo

test/helpers/IntegrationHelper.js
```diff
@@ -754,6 +754,42 @@ function generatePromptHelpMessage(player) {
...
+    ];
+    // list of approved property names for setup phase
+    const listOfApprovedPropertiesForSetupPhase = [
+        'leader',
+        'deck'
+    ];
+
+    // Check for unknown properties
+    if (startPhase !== 'setup') {
+        Object.keys(playerOptions).forEach((prop) => {
```
AMMayberry1, line 779: Per the discussion in discord, for anything that's more than a one-liner I would prefer to use `for` loops since it's better for debugging (the vscode debugger skips directly over forEach loops). For one-line statements I'm more okay with forEach but there's still some discussion on it

test/helpers/IntegrationHelper.js
```diff
@@ -799,6 +835,11 @@ global.integration = function (definitions) {
                 if (!options.player2) {
                     options.player2 = {};
                 }
+
+                // validate supplied parameters
+                validatePlayerOptions(options.player1, 'player1', options.phase);
+                validatePlayerOptions(options.player2, 'player2', options.phase);
```
AMMayberry1, line 841: Awesome!! Thanks so much, been wanting this for a while

test/helpers/IntegrationHelper.js
```diff
@@ -826,35 +867,46 @@ global.integration = function (definitions) {
 
                     // Advance the phases to the specified
                     this.advancePhases(options.phase);
-                }
 
-                // Player stats
-                this.player1.damageToBase = options.player1.damageToBase ?? 0;
-                this.player2.damageToBase = options.player2.damageToBase ?? 0;
+                    // Player stats
+                    this.player1.damageToBase = options.player1.damageToBase ?? 0;
```
AMMayberry1, line 872: Could we handle this inside the base setup method instead of needing to manage it explicitly here? I feel like that'd be cleaner and easier to maintain

test/server/core/gameSteps/phases/SetupPhase.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+                expect(context.player1).toHaveExactPromptButtons(['Done']);
+                expect(context.player2).toHaveExactPromptButtons(['Done']);
+
+                context.player1.clickCard(context.player1.hand[1]);
+                expect(context.player1).toHaveExactPromptButtons(['Done']);
+                expect(context.player2).toHaveExactPromptButtons(['Done']);
+
+                // Check that player1 cannot select any additional cards
+                context.player1.clickCardNonChecking(context.player1.hand[2]);
+                expect(context.player1.selectedCards.length).toBe(2);
```
AMMayberry1, line 88: These tests are awesome!! Do you mind also adding similar checks of the card selection process to the resource step of the setup phase, to the degree it makes sense?

danbastin:
In the `resourceSelectedCards` the `resourceCard` method should be `player.resourceCard(card, false);` to ensure that the cards are ready when resource and available on the first turn. The test should check for this as well.

server/game/core/gameSteps/prompts/ResourcePrompt.ts
CheBato: ofcourse!

server/game/core/gameSteps/prompts/ResourcePrompt.ts
CheBato: Yessir!

server/game/core/gameSteps/prompts/ResourcePrompt.ts
CheBato: sure i'll change it to when they both complete it!

test/helpers/GameFlowWrapper.js
CheBato: yessir!

test/helpers/IntegrationHelper.js
CheBato: ah yes thank you i was thinking how i can rename this sausage of a variable.

test/helpers/IntegrationHelper.js
CheBato: i also made it a bit more clearer and cleaner.

test/helpers/IntegrationHelper.js
CheBato: With pleasure!

test/helpers/IntegrationHelper.js
CheBato: yessir!

test/server/core/gameSteps/phases/SetupPhase.spec.ts
CheBato: do you mean the setupPhase or the regroup phase?

test/helpers/IntegrationHelper.js
CheBato: Now that I have tried this quite a few times, it seems that this.player1.damageToBase needs to be set here; otherwise, some tests fail to end the game. I assume it has something to do with which player1 or player2 variable we are setting here. If you want you can take a look at it.

test/server/core/gameSteps/phases/SetupPhase.spec.ts
CheBato: Added some additional ones in the setupPhase for good measure and some for the regroup Phase.

CheBato:
> In the ResourcePrompt the `resourceSelectedCards` the `resourceCard` method should be `player.resourceCard(card, false);` to ensure that the cards are ready when resource and available on the first turn. The test should check for this as well.

I changed the player.resourceCard(card) -> player.resourceCard(card, false); and added tests to verify they are indeed unexhausted.

test/helpers/IntegrationHelper.js
```diff
@@ -769,24 +769,18 @@ function validatePlayerOptions(playerOptions, playerName, startPhase) {
         'resource',
     ];
     // list of approved property names for setup phase
-    const listOfApprovedPropertiesForSetupPhase = [
+    const ForSetupPhase = [
```
AMMayberry1, line 772: nit: should start with lowercase

AMMayberry1 (approved):

PR #154, opened 2024-11-01
Support for "if you do (not)" clauses

AMMayberry1:
Added support for "do [X], if you do (not) then do [Y]" clauses via the `ifYouDo: ` property which works similarly to the `then :` property. It will key off of the effect in the ability to see if it resolves or not, following the rules of SWU resolution (must change game state or be replaced by an effect that does). If there are multiple effects in the ability, it will key off of the resolution of the last one.

Had to heavily refactor to add a better API for managing events and checking their resolved / resolvable status for this. Added a `mustChangeGameState` parameter to many of the canAffect / hasLegalTarget flows to enforce checking whether events have resolved or not in a way that changes game state so that "if you do" conditions will evaluate correctly.

Added the following cards for validation:
- Boba Fett, Daimyo (updated to use if you do instead of cost)
- Enforced Loyalty
- Toro Calican
- Lom Pyke

Discovered a bug with replacement effect ordering in the Lom Pyke tests. Applied a partial fix but a full one will have to wait for a dedicated PR to build the real replacement effect window. That said, the condition that it applies in is extremely rare so it shouldn't be an issue until then.


VJubert (dismissed):

test/server/cards/02_SHD/events/ToroCalicanAmbitiousUpstart.spec.ts
VJubert, line 1: this is not an event

server/game/core/Game.js
```diff
@@ -903,34 +907,34 @@ class Game extends EventEmitter {
...
-     * Raises a custom event window for checking for any cancels to a card
-     * ability
-     * @param {Object} params
-     * @param {Function} handler - this is an arrow function which is called if
-     * nothing cancels the event
-     */
-    raiseInitiateAbilityEvent(params, handler) {
-        this.raiseMultipleInitiateAbilityEvents([{ params: params, handler: handler }]);
-    }
+    // /**
```
garrettdc2, line 918: Question: Should we have `TODO` comments associated with these?

server/game/core/ability/TriggeredAbility.ts
```diff
@@ -112,6 +115,18 @@ export default class TriggeredAbility extends CardAbility {
         });
     }
 
+    public override checkGameActionsForPotential(context) {
+        const mustChangeGameState = !!this.properties.ifYouDo || !!this.properties.ifYouDoNot;
+
+        return this.immediateEffect.hasLegalTarget(context, {}, mustChangeGameState);
+    }
+
+    public override buildTargetResolver(name: any, properties: any) {
```
garrettdc2, line 124: NIT: Can we type these Params?

server/game/core/ability/TriggeredAbility.ts
```diff
@@ -112,6 +115,18 @@ export default class TriggeredAbility extends CardAbility {
...
+        const propsMustChangeGameState = Object.assign({ mustChangeGameState: this.mustChangeGameState }, properties);
```
garrettdc2, line 125: Thought: This is totally fine and no need to change it. Just want to call out you can use the less verbose spreading as an alternative just in case you aren't familiar with it.

```suggestion
        const propsMustChangeGameState = { mustChangeGameState: this.mustChangeGameState, ...properties};
```

server/game/core/gameSystem/CardTargetSystem.ts
```diff
@@ -142,28 +142,24 @@ export abstract class CardTargetSystem<TContext extends AbilityContext = Ability
     }
 
     public override checkEventCondition(event: any, additionalProperties = {}): boolean {
-        return this.canAffect(event.card, event.context, additionalProperties);
+        return this.canAffect(event.card, event.context, additionalProperties, true);
     }
 
-    public override canAffect(card: Card, context: TContext, additionalProperties = {}): boolean {
+    public override canAffect(card: Card, context: TContext, additionalProperties: any = {}, mustChangeGameState = false): boolean {
```
garrettdc2, line 148: NIT: Record<string, any> would be better here
```suggestion
    public override canAffect(card: Card, context: TContext, additionalProperties: Record<string, any> = {}, mustChangeGameState = false): boolean {
```

test/server/cards/02_SHD/events/EnforcedLoyalty.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.pykeSentinel, context.cartelSpacer]);
+
+                context.player1.clickCard(context.pykeSentinel);
+                expect(context.pykeSentinel).toBeInLocation('discard');
+                expect(context.player1.handSize).toBe(2);
+                expect(context.player2).toBeActivePlayer();
+            });
+        });
+
+        describe('Enforced Loyalty\'s ability', function() {
```
garrettdc2, line 31: NIT: I thin we can avoid adding a new `describe` and just skip using `BeforeEach`. Instead we can setup the test in the it block itself.

test/server/cards/02_SHD/events/EnforcedLoyalty.spec.ts
```diff
@@ -0,0 +1,53 @@
+describe('Enforced Loyalty', function() {
```
garrettdc2, line 1: Suggestion: Can we also add tests to make sure this resolves and does as much as it can in the event we have 0 cards in deck or 1 card in deck?

server/game/core/Game.js
AMMayberry1: While we're in the pre-release timeframe my habit is to just comment out blocks of jinteki stuff that we aren't sure will be needed, so any large block comments like this can be safely assumed to be that. I'm not outright removing in case there's some chance we end up needing it so we can find it again later.

We'll do a cleanup pass for all of this once we get closer to release.

server/game/core/ability/TriggeredAbility.ts
AMMayberry1: Oh yeah good call, I was using Object.assign exclusively for a while and only just recently learned about the spread operator. Definitely want to use it more often now

server/game/core/gameSystem/CardTargetSystem.ts
AMMayberry1: Agreed, but I would prefer to change the type for that property everywhere all at once instead of piecemeal. I have a few noisy changes that are due, I might batch them all together and do them in one loud PR sometime soon just to get them out of the way.

test/server/cards/02_SHD/events/EnforcedLoyalty.spec.ts
AMMayberry1: Yeah for sure. Since we've tested that in a few places already I don't want to do that test for every card that has a draw effect, but in this case since "if you do" is a new thing that may different behaviors around effects resolving correctly I think it's worth testing.

PR #158, opened 2024-11-05
Implementing Gideon Hask Ruthless Loyalist

aruizc:


server/game/cards/01_SOR/units/GideonHaskRuthlessLoyalist.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    public override setupCardAbilities () {
+        this.addTriggeredAbility({
+            title: 'Give an Experience token to a friendly unit',
+            when: {
+                onCardDefeated: (event, context) => event.card.isUnit() && event.card.controller !== context.source.controller
+            },
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardTypeFilter: WildcardCardType.Unit,
```
AMMayberry1, line 21: nit: this type filter is technically not necessary, since `giveExperience()` can only target units there's some logic that will filter everything else out behind the scenes

test/server/cards/01_SOR/units/GideonHaskRuthlessLoyalist.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+                context.player1.clickCard(context.battlefieldMarine);
+                context.player1.clickCard(context.atst);
+
+                expect(context.player2).toBeAbleToSelectExactly([context.specforceSoldier, context.gideonHask, context.atst]);
+                context.player2.clickCard(context.gideonHask);
+
+                expect(context.gideonHask).toHaveExactUpgradeNames(['experience']);
+            });
+
+            it('should give an Experience to a friendly unit when enemy unit is defeated by event', function () {
```
AMMayberry1, line 30: Do you mind consolidating these cases into one? It's bad test practice but we're in a bit of an unfortunate situation because the Jasmine tests are very slow, and we have a limited amount of github actions time in the free tier. So we try to consolidate into as few cases as we reasonably can. We plan to migrate test frameworks eventually to alleviate this issue.

You can look at the tests for Blizzard Assault AT-AT as an example of how we typically set this up.

test/server/cards/01_SOR/units/GideonHaskRuthlessLoyalist.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+                const { context } = contextRef;
+
+                // Defeating an enemy unit using Vanquish
```
AMMayberry1, line 33: Thanks for the good test cases! Could we add a couple more just to be thorough (these are helpful for when we make engine changes to ensure nothing broke):

- Check that when a friendly unit dies, nothing happens (can use `expect(context.player1).toBeActivePlayer();` to confirm that the action has finished resolving with no prompt)
- Confirm that that ability still activates correctly even if Gideon dies in the combat with the enemy unit

test/server/cards/01_SOR/units/GideonHaskRuthlessLoyalist.spec.ts
aruizc: No problem, I moved all test cases into a single unit test

test/server/cards/01_SOR/units/GideonHaskRuthlessLoyalist.spec.ts
aruizc: Addional test cases added.

PR #160, opened 2024-11-06
Add DiscardCardsFromHand System

garrettdc2:
Adds the Player targeted DiscardCardsFromHands system, which allows cards to target a player to discard N cards from hand. Also adds Pillage (SHD 181) in order to test the feature.

SWUDB: https://swudb.com/card/SHD/181/pillage


server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,84 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public override eventHandler(_event): void { }
+
+    public override getEffectMessage(context: TContext): [string, any[]] {
+        const properties = this.generatePropertiesFromContext(context);
+        return ['make {0} discard {1} cards', [properties.target, properties.amount]];
+    }
+
+    public override canAffect(playerOrPlayers: Player | Player[], context: TContext, additionalProperties = {}): boolean {
+        const players = Array.isArray(playerOrPlayers) ? playerOrPlayers : [playerOrPlayers];
```
AMMayberry1, line 31: Just a heads-up if you find it useful, I also have `asArray` in Helper.ts for this situation. So you could directly do `return Helpers.asArray(players).every(...)`

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,84 @@
...
+            return super.canAffect(player, context);
+        });
+    }
+
+    public override queueGenerateEventGameSteps(events: any[], context: TContext, additionalProperties: Record<string, any> = {}): void {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        (properties.target as Player[]).forEach((player) => {
+            const availableHand = player.hand.filter((card) => properties.cardCondition(card, context));
+            const amount = Math.min(availableHand.length, properties.amount);
+            if (amount <= 0) {
```
AMMayberry1, line 48: Definitely shouldn't be < 0. Actually, could we do a `Contract.assertTrue` check that `amount > 0`? Should have at least one somewhere in the class for safety.

test/server/cards/02_SHD/events/Pillage.spec.ts
```diff
@@ -0,0 +1,92 @@
+import { Location } from '../../../../../server/game/core/Constants';
```
AMMayberry1, line 1: Unfortunately this seems to be causing test issues for me sporadically - it causes jasmine failures for some reason, though not in all cases and only if I'm running one specific file and not the whole suite. Jasmine is a pain to debug so I haven't been able to find a cause. Until we get it resolved, would you mind defaulting to just using the string values?

test/server/cards/02_SHD/events/Pillage.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+                const { player1, player2, pillage, allianceXwing, battlefieldMarine, imperialInterceptor, wampa } = context;
+
+                player1.clickCard(pillage);
+                player1.clickPrompt('Opponent');
+                expect(player2).toBeAbleToSelectExactly([
+                    allianceXwing, battlefieldMarine, imperialInterceptor, wampa
+                ]);
+
+                player2.clickCard(imperialInterceptor);
+                player2.clickCard(wampa);
```
AMMayberry1, line 27: Since we're having some card selection bugs, do you mind adding a third clicked card (with `clickCardNonChecking()` to avoid getting an error) to ensure that everything still works right

test/server/cards/02_SHD/events/Pillage.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+                expect(wampa).toBeInLocation(Location.Discard);
+                expect(imperialInterceptor).toBeInLocation(Location.Discard);
+
+                expect(allianceXwing).toBeInLocation(Location.Hand);
+                expect(battlefieldMarine).toBeInLocation(Location.Hand);
+
+                expect(player2).toBeActivePlayer();
+            });
+
+            it('should let the player target the opponent, and automatically discard the only card in the opponents hand', function() {
```
AMMayberry1, line 39: Can you add another test where the opponent has 0 cards in hand and confirm that they are still selectable

test/server/cards/02_SHD/events/Pillage.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+            it('should let the player target the themselves, and let the player discard 2 cards from their hand', function() {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['pillage', 'alliance-xwing', 'battlefield-marine', 'imperial-interceptor', 'wampa'],
+                    },
+                });
+
+                const { context } = contextRef;
+                const { player1, player2, pillage, allianceXwing, battlefieldMarine, imperialInterceptor, wampa } = context;
```
AMMayberry1, line 70: Not a suggestion per se but just a note - the test repo had a pattern like this originally and I moved us away from it b/c it's annoying to have to make a change in one more place if we change cards involved in the test. That said, it's highly personal what people find more and less annoying, so if you prefer declaring up front like this then go for it

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,84 @@
+import { EventName, Location, RelativePlayer, TargetMode } from '../core/Constants';
+import { AbilityContext } from '../core/ability/AbilityContext';
+import type Player from '../core/Player';
+import { IPlayerTargetSystemProperties, PlayerTargetSystem } from '../core/gameSystem/PlayerTargetSystem';
+import { Card } from '../core/card/Card';
+import { DiscardSpecificCardSystem } from './DiscardSpecificCardSystem';
+
+export interface IDiscardCardsFromHandProperties extends IPlayerTargetSystemProperties {
+    amount?: number;
```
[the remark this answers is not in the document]
garrettdc2: I like that. Will change it

server/game/gameSystems/DiscardCardsFromHand.ts
garrettdc2: Mainly just added the least than to be super safe and ensure we don't error out if somehow we end up with a negative number. But I am fine adding a contract if we def want to error out 👍

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,84 @@
...
+            if (amount <= 0) {
+                return;
+            }
+
+            if (amount >= availableHand.length) {
+                this.generateEventsForCards(availableHand, context, events);
+                return;
+            }
+
+            if (context.choosingPlayerOverride && context.choosingPlayerOverride !== player) {
```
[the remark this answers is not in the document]
garrettdc2: Holdover from L5R. I think it's not needed and we wouldn't want to do this now that you mention it. I'll remove it

server/game/gameSystems/DiscardCardsFromHand.ts
AMMayberry1: Yeah I would prefer that

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export interface IDiscardCardsFromHandProperties extends IPlayerTargetSystemProperties {
+    amount?: number;
+    cardCondition?: (card: Card, context: AbilityContext) => boolean;
+}
+
+export class DiscardCardsFromHand<TContext extends AbilityContext = AbilityContext> extends PlayerTargetSystem<TContext, IDiscardCardsFromHandProperties> {
+    protected override defaultProperties: IDiscardCardsFromHandProperties = {
+        amount: 1,
+        cardCondition: () => true
```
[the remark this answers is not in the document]
garrettdc2: I personally would prefer to default to a type like what we expect, like what happening here. I think this will often lead to less errors and allow us to implement logic a bit easier cuz we don't have check for null constantly. Null can also. Granted once we are in a full TS system the error justification probably is not a problem anymore cuz TS should catch type issues at compile.

But that being said I don't think using TS/JS really adds much to the conversation. Its more up to preference. I think both are fine, and TS will catch issues either way.

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,84 @@
...
+                onSelect: (_player, cards) => {
+                    this.generateEventsForCards(cards, context, events);
+                    return true;
+                }
+            });
+        });
+    }
+
+    private generateEventsForCards(cards: Card[], context: TContext, events: any[]): void {
+        cards.forEach((card) => {
```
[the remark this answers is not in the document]
garrettdc2: Given this makes me wonder if we should revert back to having a single event that iterates through the cards and moves them. That would solve this issue, correct?

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -239,9 +240,9 @@ export function reveal<TContext extends AbilityContext = AbilityContext>(propert
 // /**
 //  * default amount = 1
 //  */
-// export function chosenDiscard(propertyFactory: PropsFactory<ChosenDiscardProperties> = {}): GameSystem {
-//     return new ChosenDiscardAction(propertyFactory);
-// }
+export function discardCardsFromHand<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IDiscardCardsFromHandProperties, TContext> = {}): DiscardCardsFromHand<TContext> {
```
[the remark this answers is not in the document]
garrettdc2: Discussed offline and will go will discardFromOwnHand for now.

test/server/cards/02_SHD/events/Pillage.spec.ts
garrettdc2: I prefer it, and some even consider it a best practice to destructure object properties that are use more than once to reduce code clutter and clearly denote what is used up front. AirBnb style guide is what my team follows (pretty popular outside my team as well) is probably why I picked up this habit, and they recommend it and even enforce it in their eslint plugins: https://github.com/airbnb/javascript?tab=readme-ov-file#destructuring--object

Of course this is just a style preference so totally just a preference thing at the end of the day. Both are functionally equivalent.

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,84 @@
...
+        return ['make {0} discard {1} cards', [properties.target, properties.amount]];
+    }
+
+    public override canAffect(playerOrPlayers: Player | Player[], context: TContext, additionalProperties = {}): boolean {
+        const players = Array.isArray(playerOrPlayers) ? playerOrPlayers : [playerOrPlayers];
+        return players.every((player) => {
+            const properties = this.generatePropertiesFromContext(context, additionalProperties);
+            const availableHand = player.hand.filter((card) => properties.cardCondition(card, context));
+
+            if (availableHand.length === 0 || properties.amount === 0) {
```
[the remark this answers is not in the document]
garrettdc2: Have some questions on `mustChangeGameState` that I will hit you up tomorrow about offline.

server/game/gameSystems/DiscardCardsFromHand.ts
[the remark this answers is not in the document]
garrettdc2: Thinking on this more I wouldn't solve it cuz we still need to do as much as we can and that would be a successful even. So yeah back to brainstorming or doing one of the above

server/game/gameSystems/DiscardCardsFromHand.ts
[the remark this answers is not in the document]
AMMayberry1: Sounds good to me!

server/game/gameSystems/DiscardCardsFromHand.ts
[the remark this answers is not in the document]
AMMayberry1: We could make some extensions to where it would accomplish what we want - like right now there is a "resolved" state for GameEvent, we could add a "partially resolved" state as well without much issue.

The bigger problem with using one event is that it will cause inconsistencies for abilities that interact with discard, which I think I mentioned elsewhere. If the multi-discard case has `event.cards` as an array of the discarded cards but the card target system that discard a single card just uses `event.target`, but both emit the `onDiscardCards` event type, then any system that reads that event needs to know to check both properties.

There are ways around that as well, but imo that's the bigger issue here

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+export class DiscardCardsFromHand<TContext extends AbilityContext = AbilityContext> extends PlayerTargetSystem<TContext, IDiscardCardsFromHandProperties> {
+    protected override defaultProperties: IDiscardCardsFromHandProperties = {
+        amount: 1,
+        cardCondition: () => true
+    };
+
+    public override name = 'discard';
+    public override eventName = EventName.OnCardsDiscardedFromHand;
```
AMMayberry1, line 27: I think we should make sure that the event names for this and `DiscardSpecificCardSystem` are really clear in terms what the distinction is. I think this one is currently good, but we might want to change the specific card event to `OnCardDiscarded` singular

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,102 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public override eventHandler(_event): void { }
+
+    public override getEffectMessage(context: TContext): [string, any[]] {
+        const properties = this.generatePropertiesFromContext(context);
+        return ['make {0} discard {1} cards', [properties.target, properties.amount]];
+    }
+
+    public override canAffect(playerOrPlayers: Player | Player[], context: TContext, additionalProperties = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        for (const player of asArray(playerOrPlayers)) {
```
AMMayberry1, line 38: nit: for helper functions like this our convention so far is to import the namespace so we can use C# / Java-style import naming. In this case, `Helper.asArray`. Same for the methods out of Contract and EnumHelpers. It just makes it clearer that it's an import and not a local method imo.

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,102 @@
...
+            const properties = this.generatePropertiesFromContext(context, additionalProperties);
+            const availableHand = player.hand.filter((card) => properties.cardCondition(card, context) && cardTypeMatches(card.type, properties.cardTypeFilter));
+
+            if (mustChangeGameState !== GameStateChangeRequired.None && (availableHand.length === 0 || properties.amount === 0)) {
```
AMMayberry1, line 42: All checks look perfect!

server/game/gameSystems/DiscardCardsFromHand.ts
```diff
@@ -0,0 +1,102 @@
...
+        }
+        return true;
+    }
+
+    public override queueGenerateEventGameSteps(events: any[], context: TContext, additionalProperties: Record<string, any> = {}): void {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        for (const player of properties.target as Player[]) {
+            const availableHand = player.hand.filter((card) => properties.cardCondition(card, context));
+
+            assertTrue(properties.amount >= 0, 'Cards to discard must be a non-negative integer');
```
AMMayberry1, line 62: nit: can use `assertNonNegative` for simplicity, which has an equivalent error message baked in

AMMayberry1 (approved):

PR #161, opened 2024-11-06
Support for "gain Bounty" abilities

AMMayberry1:
Discovered a timing bug in the original Bounty PR that prevented "gain Bounty" abilities on upgrades from working correctly because the defeat of the attached upgrade was removing the ability before the defeat of the unit could trigger it.

Reordered the timing of ability and trigger registration / de-registration to not happen until after all events in the window have finished resolving. This required some tweaks in various places that were making assumptions about game state and card location.

Additionally, noticed some opportunities to improve the target selection logic to cut back on unnecessary prompts, particularly when related to exhausting / readying cards. In some cases prompts were being emitted to ready / exhaust cards when all available targets were already in that state, in which case the user's selection is pointless. Improved the targeting logic and several related test cases.

Fixed all of the above and added the Wanted and Top Target cards.


server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -5,7 +5,7 @@ import * as EnumHelpers from '../core/utils/EnumHelpers';
 import { type ICardTargetSystemProperties, CardTargetSystem } from '../core/gameSystem/CardTargetSystem';
 
 export interface IMoveCardProperties extends ICardTargetSystemProperties {
-    destination?: Location;
+    destination?: Location.Discard | Location.Deck;
```
garrettdc2, line 8: Question: Could this also be hand, in the event we are retrieving a card from the discard pile?

server/game/cards/02_SHD/upgrades/TopTarget.ts
```diff
@@ -0,0 +1,28 @@
...
+            id: '4282425335',
+            internalName: 'top-target',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addGainKeywordTargetingAttached({
+            keyword: KeywordName.Bounty,
+            ability: {
+                title: 'Heal 4 damage from a unit or base. If the Bounty unit is unique, heal 6 damage instead.',
```
garrettdc2, line 17: Suggestion: This can also heal a unit so we need logic allow targeting of units.

test/server/cards/02_SHD/upgrades/TopTarget.spec.ts
```diff
@@ -0,0 +1,53 @@
+describe('Top Target', function() {
+    integration(function(contextRef) {
+        describe('Top Target\'s Bounty ability', function() {
+            it('should heal 4 damage from a base if the attached unit is not unique', function () {
```
garrettdc2, line 4: Suggestion: I think we may also want to test the scenario where the owner of the bounty has it placed on its own unit, and ensure the bounty is claimed by the opponent. Just to be sure bounty ownership does not affect anything.

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -302,6 +292,8 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
          * Checks if the unit currently has any keywords with a "when defeated" effect and registers them if so.
          * Also adds a listener to remove the registered abilities after the effect resolves.
          */
+        // TODO THIS PR: refactor this into a separate method so "checkRegisterWhenCaptured" will be simple
+        // TODO THIS PR: can this be easily extended for Bossk?
```
Ennis0, line 296: Should get these cleared away before merging.

server/game/gameSystems/MoveCardSystem.ts
Ennis0, line 8: This feels odd to me. If this system is only for a subset of card moves it should probably have a different name.

server/game/cards/02_SHD/upgrades/TopTarget.ts
```diff
@@ -0,0 +1,28 @@
...
+                targetResolver: {
+                    controller: RelativePlayer.Any,
+                    cardTypeFilter: CardType.Base,
+                    immediateEffect: AbilityHelper.immediateEffects.heal((context) => ({ amount: context.source.unique ? 6 : 4 })),
```
VJubert, line 21: it is not context.source.parentCard.unique ? like all other upgrades ?

server/game/gameSystems/MoveCardSystem.ts
AMMayberry1: Yeah the issue is that we have two different systems that sort of cover that case (the other is ReturnToHandSystem) which is why I was filtering that out here. I have an open work item for sorting that out, so maybe for now I'll just leave this in the state it is and add a TODO since making this change makes the purpose kind of unclear as Ennis noted in their comment.

server/game/cards/02_SHD/upgrades/TopTarget.ts
AMMayberry1: In this case it's because the ability is owned by the parent card. So e.g. if Top Target is attached to Wampa, then context.source here is Wampa because the Bounty ability technically belongs to Wampa.

PR #162, opened 2024-11-06
Add 01_SOR Power of the Dark Side

jasimmons:


test/server/cards/01_SOR/events/PowerOfTheDarkSide.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+                context.player2.clickCard(context.fleetLieutenant);
+                expect(context.fleetLieutenant).toBeInLocation('discard');
+            });
+
+            it('forces opponent to defeat a leader unit if nothing else is available', function () {
+                const { context } = contextRef;
+
+                context.player2.setActivePlayer();
+
+                // Attack with Fleet Lieutenant, choose Veers as target
```
AMMayberry1, line 35: nit: for readability, instead of requiring player actions to set up the desired gamestate could we have a separate setupTest block that declares it

AMMayberry1 (dismissed):

AMMayberry1 (approved):

PR #165, opened 2024-11-06
Implement Pirated Starfighter

jan-preisig:
Implementation for pirated starfighter including two tests

### DIFF
--- a/test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
+++ b/test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
@@ -35,8 +35,8 @@ describe('Pirated Starfighter', function () {
                 context.player1.clickCard(context.piratedStarfighter);
                 expect(context.player1).toBeAbleToSelectExactly([context.piratedStarfighter, context.pykeSentinel]);
                 context.player1.clickCard(context.pykeSentinel);
-                expect(context.pykeSentinel).toBeInLocation('hand', context.player1);
-                expect(context.piratedStarfighter).toBeInLocation('space arena', context.player1);
+                expect(context.pykeSentinel).toBeInLocation('hand');
+                expect(context.piratedStarfighter).toBeInLocation('space arena');
             });
         });
     });
--- a/test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
+++ b/test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
@@ -18,13 +18,14 @@ describe('Pirated Starfighter', function () {
                 phase: 'action',
                 player1: {
                     hand: ['pirated-starfighter'],
-                    groundArena: ['pyke-sentinel'],
+                    groundArena: ['pyke-sentinel', 'seasoned-shoretrooper'],
+                    leader: { card: 'leia-organa#alliance-general', deployed: true }
                 }
             });
             const { context } = contextRef;
 
             context.player1.clickCard(context.piratedStarfighter);
-            expect(context.player1).toBeAbleToSelectExactly([context.piratedStarfighter, context.pykeSentinel]);
+            expect(context.player1).toBeAbleToSelectExactly([context.piratedStarfighter, context.pykeSentinel, context.seasonedShoretrooper]);
             context.player1.clickCard(context.pykeSentinel);
             expect(context.pykeSentinel).toBeInLocation('hand');
             expect(context.piratedStarfighter).toBeInLocation('space arena');
--- a/test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
+++ b/test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
@@ -1,0 +1,43 @@
+describe('Pirated Starfighter', function () {
+    integration(function (contextRef) {
+        describe('Pirated Starfighter\'s ability', function () {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['pirated-starfighter'],
+                    }
+                });
+            });
+
+            it('should not prompt player if no units are available to return to hand', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.piratedStarfighter);
+                expect(context.player2).toBeActivePlayer();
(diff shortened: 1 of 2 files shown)


test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['pirated-starfighter'],
+                    }
+                });
+            });
+
+            it('should not prompt player if no units are available to return to hand', function () {
```
AMMayberry1, line 13: nit: just a readability improvement, if a setupTest() call is used only for one test case then we can move the setup into the test case itself and consolidate down to one describe block. This is a pattern we've only recently started doing, you can see the Reinforcement Walker tests as an example.

test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            });
+        });
+
+        describe('Pirated Starfighter\'s ability', function () {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['pirated-starfighter'],
+                        groundArena: ['pyke-sentinel'],
```
AMMayberry1, line 27: Can you add some additional units to ensure that the card filter was correctly implemented. Some enemy units and a friendly deployed leader would be great. The syntax for deployed leader is just `leader: { card: 'leia-organa#alliance-general', deployed: true }`

test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+                });
+            });
+
+            it('should be able to return self or other unit to hand', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.piratedStarfighter);
+                expect(context.player1).toBeAbleToSelectExactly([context.piratedStarfighter, context.pykeSentinel]);
+                context.player1.clickCard(context.pykeSentinel);
+                expect(context.pykeSentinel).toBeInLocation('hand', context.player1);
```
AMMayberry1, line 38: You can remove the `context.player1` from these, that's the default value

test/server/cards/01_SOR/units/PiratedStarfighter.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['pirated-starfighter'],
+                        groundArena: ['pyke-sentinel'],
+                    }
+                });
+            });
+
+            it('should be able to return self or other unit to hand', function () {
```
AMMayberry1, line 32: Do you mind adding one more test case where pirated starfighter is the only friendly unit and therefore automatically bounces itself back to hand? Just to make sure the logic for all of that is working correctly

jan-preisig: Ah so then I just noticed i might have found a bug, because if you play Pirated Starfighter and have no units out, I believe you have to return Pirated Starfighter to your hand. But when I tried to write a test for this it didn't work.

jan-preisig: So with the following code
`
contextRef.setupTest({
                phase: 'action',
                player1: {
                    hand: ['pirated-starfighter'],
                }
            });

            context.player1.clickCard(context.piratedStarfighter);
            expect(context.player1).toBeAbleToSelectExactly([context.piratedStarfighter]);
            context.player1.clickCard(context.piratedStarfighter);
            expect(context.piratedStarfighter).toBeInLocation('hand');
`
I get the following error:
`
 Message:
    TestSetupError: Expected player prompt state to change after clicking pirated-starfighter but it did not. Current prompt:
    Waiting for opponent to take an action or pass

    Pirated Starfighter

  Stack:
    error properties: Object({ statusCode: undefined })
    TestSetupError: Expected player prompt state to change after clicking pirated-starfighter but it did not. Current prompt:
    Waiting for opponent to take an action or pass
    Pirated Starfighter
        at PlayerInteractionWrapper.clickCard (C:\DEV\01_Privat\forceteki\test\helpers\PlayerInteractionWrapper.js:607:23)
        at UserContext.<anonymous> (C:\DEV\01_Privat\forceteki\test\server\cards\01_SOR\units\PiratedStarfighter.spec.ts:14:29)
        at <Jasmine>

`

AMMayberry1: Followed up in chat but I figured it out. I was going too fast, apologies - your first test already is the thing I'm looking for. The test you wrote is failing because the pirated starfighter ability completely resolves in the first click and it ends up back in hand.

So when you go to do the second click, it is already player 2's action, and the error is letting you know that nothing happened when player 1 clicked on the card (which is correct).

The change I would make to the first test is this

```typescript
    it('should not prompt player if no units are available to return to hand', function () {
        contextRef.setupTest({
            phase: 'action',
            player1: {
                hand: ['pirated-starfighter'],
            },
            base: 'chopper-base'
        });

        context.player1.clickCard(context.piratedStarfighter);
        expect(context.piratedStarfighter).toBeInLocation('hand');
        expect(context.player1.countExhaustedResources()).toBe(2);
    });
```

jan-preisig: That makes sense, I'll adjust it this way then

jan-preisig: Adjusting the test this way, says Expected 4 to be 2. Any Ideas on that?

AMMayberry1: Commented in discord (that's the better way to reach me for followups like this), but that would suggest that there's an aspect penalty being paid. Just to confirm, did you add the yellow base to the test setup?

AMMayberry1 (approved):

PR #166, opened 2024-11-06
Keyword losing effects

Ennis0:
Implemented keyword losing effects including some tests to cover complex situations involving when a keyword should and shouldn't be regained.

Implemented SpecForce Soldier.

### DIFF
--- a/server/game/core/GameObject.ts
+++ b/server/game/core/GameObject.ts
@@ -25,6 +25,12 @@ export abstract class GameObject {
     }
 
     public addOngoingEffect(ongoingEffect: IOngoingCardEffect) {
+        // logic for "lose keyword" abilites: they need to register all effects that are giving the keyword at the time they are applied, so that they can nullify only those
+        if (ongoingEffect.type === EffectName.LoseKeyword) {
+            const currentKeywordGrantingEffects = this.ongoingEffects.filter((currentEffect) => currentEffect.type === EffectName.GainKeyword && currentEffect.getValue(this).name === ongoingEffect.getValue(this).name);
+            Object.assign(ongoingEffect.getValue(this), { effectsToCancel: currentKeywordGrantingEffects });
+        }
+
         this.ongoingEffects.push(ongoingEffect);
     }
 
--- a/server/game/core/GameObject.ts
+++ b/server/game/core/GameObject.ts
@@ -106,7 +106,7 @@ export abstract class GameObject {
         return effects[effects.length - 1];
     }
 
-    protected getOngoingEffects() {
+    public getOngoingEffects() {
         const suppressEffects = this.ongoingEffects.filter((ongoingEffect) => ongoingEffect.type === EffectName.SuppressEffects);
         const suppressedEffects = suppressEffects.reduce((array, ongoingEffect) => array.concat(ongoingEffect.getValue(this)), []);
         return this.ongoingEffects.filter((ongoingEffect) => !suppressedEffects.includes(ongoingEffect));
--- a/server/game/core/GameObject.ts
+++ b/server/game/core/GameObject.ts
@@ -100,7 +100,7 @@ export abstract class GameObject {
         return effects[effects.length - 1];
     }
 
-    public getOngoingEffects() {
+    protected getOngoingEffects() {
         const suppressEffects = this.ongoingEffects.filter((ongoingEffect) => ongoingEffect.type === EffectName.SuppressEffects);
         const suppressedEffects = suppressEffects.reduce((array, ongoingEffect) => array.concat(ongoingEffect.getValue(this)), []);
         return this.ongoingEffects.filter((ongoingEffect) => !suppressedEffects.includes(ongoingEffect));
--- a/server/game/core/GameObject.ts
+++ b/server/game/core/GameObject.ts
@@ -25,12 +25,6 @@ export abstract class GameObject {
     }
 
     public addOngoingEffect(ongoingEffect: IOngoingCardEffect) {
-        // logic for "lose keyword" abilites: they need to register all effects that are giving the keyword at the time they are applied, so that they can nullify only those
-        if (ongoingEffect.type === EffectName.LoseKeyword) {
-            const currentKeywordGrantingEffects = this.ongoingEffects.filter((currentEffect) => currentEffect.type === EffectName.GainKeyword && currentEffect.getValue(this).name === ongoingEffect.getValue(this).name);
-            Object.assign(ongoingEffect.getValue(this), { effectsToCancel: currentKeywordGrantingEffects });
-        }
-
         this.ongoingEffects.push(ongoingEffect);
     }
 
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -4,18 +4,16 @@ import PlayerOrCardAbility from '../ability/PlayerOrCardAbility';
 import OngoingEffectSource from '../ongoingEffect/OngoingEffectSource';
 import type Player from '../Player';
 import * as Contract from '../utils/Contract';
-import { AbilityRestriction, AbilityType, Arena, Aspect, CardType, Duration, EffectName, EventName, KeywordName, Location, RelativePlayer, Trait, WildcardLocation } from '../Constants';
+import { AbilityRestriction, Aspect, CardType, Duration, EffectName, EventName, KeywordName, Location, RelativePlayer, Trait, WildcardLocation } from '../Constants';
 import * as EnumHelpers from '../utils/EnumHelpers';
-import AbilityHelper from '../../AbilityHelper';
-import * as Helpers from '../utils/Helpers';
 import { AbilityContext } from '../ability/AbilityContext';
 import { CardAbility } from '../ability/CardAbility';
 import type Shield from '../../cards/01_SOR/tokens/Shield';
-import { KeywordInstance, KeywordWithAbilityDefinition, KeywordWithCostValues } from '../ability/KeywordInstance';
+import { KeywordInstance, KeywordWithCostValues } from '../ability/KeywordInstance';
 import * as KeywordHelpers from '../ability/KeywordHelpers';
 import { StateWatcherRegistrar } from '../stateWatcher/StateWatcherRegistrar';
 import type { EventCard } from './EventCard';
-import type { CardWithExhaustProperty, CardWithTriggeredAbilities, CardWithConstantAbilities, TokenCard, UnitCard, CardWithDamageProperty } from './CardTypes';
+import type { TokenCard, UnitCard, CardWithDamageProperty } from './CardTypes';
 import type { UpgradeCard } from './UpgradeCard';
 import type { BaseCard } from './BaseCard';
 import type { LeaderCard } from './LeaderCard';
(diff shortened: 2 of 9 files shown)


server/game/core/GameObject.ts
```diff
@@ -25,6 +25,12 @@ export abstract class GameObject {
     }
 
     public addOngoingEffect(ongoingEffect: IOngoingCardEffect) {
+        // logic for "lose keyword" abilites: they need to register all effects that are giving the keyword at the time they are applied, so that they can nullify only those
+        if (ongoingEffect.type === EffectName.LoseKeyword) {
```
AMMayberry1, line 29: This works but it's not the best practice to add complexity in base classes like this which is specific for certain situations with certain subclasses. Separating the implementation from the context can make it harder to avoid bugs, and also if folks are working on the code they may not realize stuff is happening at this layer (or they may not know where to find this implementation if they're looking for it).

Typically when an ongoing effect has some complex logic like this we tend to create a class implementation for handling the complexity (e.g. GainAbility or CostAdjuster), I think there's enough here to justify creating a small one.

The other benefit of that would be offering maybe slightly better documentation and typing for the properties that we're stuffing into the effect value, since that system is untyped it can be a little hard to understand what the contents of an effect are in each situation so having it clearly documented is a big win.

At the barest minimum though this logic should be moved into an override of `addOngoingEffect` in UnitProperties since that's the only class it applies to afaik

server/game/core/card/Card.ts
```diff
@@ -359,16 +357,20 @@ export class Card extends OngoingEffectSource {
     // ******************************************* KEYWORD HELPERS *******************************************
     /** Helper method for {@link Card.keywords} */
     protected getKeywords() {
-        const keywords = [...this.printedKeywords];
-
-        for (const gainedKeyword of this.getOngoingEffectValues(EffectName.GainKeyword)) {
-            keywords.push(gainedKeyword);
+        let keywords = [...this.printedKeywords];
```
AMMayberry1, line 360: This works but I want to offer an alternative design suggestion for your consideration based on what I was thinking of doing for ability blanking. Keyword blanking is much simpler so there's definitely flexibility in the approaches and this does work, just want to run a suggestion past you.

One of the design principles I tried to use in this repo (with varying levels of success) is isolating out the complexity of specific mechanics into their own classes as much as possible so that entities using those mechanisms didn't have to worry about the mechanics. This code is a big mechanics-heavy, I think there's a way we could isolate it a bit so that the code here is really simple. It does have some tradeoffs though.

A way you could do this differently would be to add a list called `BlankingEffects` to the `KeywordInstance` class. Then, whenever a blanking effect is added to a unit, it adds itself to the BlankingEffects list of each keyword on that unit (even if already blanked). It could also have an `isBlank` property that's just a boolean returning whether BlankingEffects is empty or not.

Then here the check is just filtering for keywords that have `isBlank = false`. Obviously we're not reducing complexity that much overall since the effect has to add itself to the unit's keywords which implies some new methods and functionality, and it has to do bookkeeping to remove itself from those keywords when it's unapplied. But overall I think it would make the functionality here a little easier to read and consolidate the core logic into one place.

That said, some of the same benefits come if we just refactor this code into a few more sub-methods. I'm at peace with whichever you prefer.

test/scenarios/constantEffectInteractions/losingAndGainingKeywords.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+                context.player1.clickCard(context.wampa);
+
+                // attack target chosen automatically due to sentinel
+                expect(context.player2).toBeActivePlayer();
+                expect(context.wampa).toBeInLocation('discard');
+                expect(context.vigilantHonorGuards.damage).toBe(4);
+            });
+
+            it('it should not regain the keyword from that source after the source\'s condition toggles off then on', function() {
```
AMMayberry1, line 35: I am... actually unsure of what the SWU rules say should happen here? I may ask in the judge chat, it's not clear to me whether this situation counts as applying a new effect or not (obviously if it is a new effect then this test is wrong)

test/scenarios/constantEffectInteractions/losingAndGainingKeywords.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+                        groundArena: ['wampa', 'crafty-smuggler']
+                    },
+                    player2: {
+                        groundArena: ['vigilant-honor-guards'],
+                        hand: ['protector', 'repair']
+                    }
+                });
+            });
+
+            it('it should be able to gain the keyword again from a different source', function () {
```
AMMayberry1, line 18: Could we add a test of multiple instances of gain and then lose to make sure that it works for more than one iteration, that tends to be the issue with a lot of potential ways of implementing the blanking.

So like specifically, after the current test have another round of specforce soldier and then protector. Can see UniqueUnit.spec.ts for good patterns for how to handle duplicate cards

test/scenarios/constantEffectInteractions/losingAndGainingKeywords.spec.ts
AMMayberry1: Ah okay, looks like the test is right: https://discord.com/channels/1265343874105081856/1285715527548796968

Good job with this, would not have thought of this test at all

server/game/core/GameObject.ts
Ennis0: Thank you. I figured there was a more isolated way to define effects like this but between OngoingEffect itself, IOngoingCardEffect, OngoingEffectImpl, and the ValueWrapper(which I don't think I even noticed before) I kind of had no idea what was what.

AMMayberry1: Yeah that's totally fair, the ongoing effect system is one of the most tangled parts of the codebase. They cheated _a lot_ with types so it's going to be a pretty big project to unravel at all at some point. The general idea is that OngoingEffect holds an OngoingEffectImpl, which in turn holds a ValueWrapper. IOngoingCardEffect is an interface that partially matches OngoingEffect that they use to hide some of the bad stuff they did with types.

Ennis0: I made a LoseKeyword class extending OngoingEffectValueWrapper(like GainAbility, which you mentioned), but it looks like I won't be able to access anything in it in getKeywords() because in that context all the ongoingEffects are IOngoingCardEffect, which doesn't know anything about ValueWrappers. I don't want to mess with the underlying stuff to change that because I'm still new to this system and don't really understand it. Does that make sense and do you have any idea what I should do?

AMMayberry1: Oh yeah so I was thinking that it would operate very similar to the way GainAbility does in which the ongoing effect acts on the unit instead of the unit reading from the effect. So e.g. for GainAbility it tells the unit to add a new ability to itself, with the blanking it would tell the unit to apply the blanking effect to all of its current keywords using something like the algorithm I described in the comment on Card.ts (though there are several options there).

If you did want to continue your current route and instead have the card read from the effect, then you should take a look at the stat modification system where we use the modifyStats() effect to wrap a StatsModifier object which then gets unwrapped in UnitProperties.getPower() / getHp().

Ennis0: Okay. I didn't think of just adding a method on Card, good idea.

server/game/core/card/Card.ts
Ennis0: I gave your idea a shot. It works and it does indeed seem very clean.

server/game/core/GameObject.ts
Ennis0: I ended up not needing to add a method actually, but the idea of the LoseKeyword asking Card for information rather than the other way around was helpful.

server/game/core/GameObject.ts
```diff
@@ -106,7 +100,7 @@ export abstract class GameObject {
         return effects[effects.length - 1];
     }
 
-    protected getOngoingEffects() {
+    public getOngoingEffects() {
```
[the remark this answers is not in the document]
Ennis0: Must be a leftover from experimenting with the implementation. Good catch.

server/game/core/ongoingEffect/effectImpl/LoseKeyword.ts
```diff
@@ -0,0 +1,22 @@
+import { KeywordName } from '../../Constants';
+import { OngoingEffectValueWrapper } from './OngoingEffectValueWrapper';
+import type { InPlayCard } from '../../card/baseClasses/InPlayCard';
+import { KeywordInstance } from '../../ability/KeywordInstance';
+
+export class LoseKeyword extends OngoingEffectValueWrapper<KeywordName> {
+    private blankedKeywordInstances: KeywordInstance[] = [];
+
+    public override apply(target: InPlayCard) {
+        const keywordInstancesToApplyTo = target.keywords.filter((keywordInstance) => keywordInstance.name === this.value);
```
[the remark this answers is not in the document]
Ennis0: Based on the rulings I've read I believe that a blanking effect will only "see" effects that are currently actually active on the unit.

test/scenarios/constantEffectInteractions/losingAndGainingKeywords.spec.ts
```diff
@@ -5,52 +5,109 @@ describe('Losing and gaining keywords', function() {
...
 
-                context.player1.clickCard(context.wampa);
-                expect(context.player1).toBeAbleToSelectExactly([context.vigilantHonorGuards, context.p2Base]);
+                context.player1.clickCard(context.consularSecurityForce);
+                // attack target chosen automatically due to sentinel
+                expect(context.player2).toBeActivePlayer();
+                expect(context.consularSecurityForce.damage).toBe(6);
             });
+            // TODO: once overall ability blanking is implemented(Imprisoned):
+            // Find out the rules and make tests for interaction between that and SpecForce(e.g. Imprisoned unit with sentinel is targeted by SpecForce, then Imprisoned is removed. What happens?)
```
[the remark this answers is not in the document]
Ennis0: The difference is that Imprisoned could fall off at any time, "uncovering" a sentinel granting effect that the specforce ability wasn't able to "see".

server/game/core/ongoingEffect/effectImpl/LoseKeyword.ts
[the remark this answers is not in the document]
AMMayberry1: Do you mind sharing the rulings? I couldn't find anything specifically related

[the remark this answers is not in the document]
Ennis0: You linked it before: https://discord.com/channels/1265343874105081856/1285715527548796968
"If an ability causes a unit to “lose” a keyword, the unit ceases to have the specified keyword and any abilities granting it that keyword **when the “‘lose” effect is resolved** cease granting the keyword for the duration of the “lose” effect. Unless otherwise specified, the unit can regain that keyword through new effects."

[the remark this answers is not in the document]
AMMayberry1: Ah that's fair, I didn't think about the implications of the highlighted part. I'm starting to think you're right, though I'll probably still start a thread in there just to confirm and to ask about the Imprisoned example. For now let's assume it works this way, good discussion!

AMMayberry1 (approved):

PR #167, opened 2024-11-06
Adds Strike True implementation

aruizc:


test/server/cards/01_SOR/events/StrikeTrue.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+                        groundArena: ['wampa', 'battlefield-marine']
+                    },
+                    player2: {
+                        groundArena: ['specforce-soldier', 'atst'],
+                        spaceArena: ['tieln-fighter']
+                    }
+                });
+            });
+
+            it('should deal damage to an enemy unit equals to a friendly unit power', function () {
```
AMMayberry1, line 18: I think we're okay without this test, it's more testing that the defeat due to damage works correctly than validating the implementation of the card. We do mix tests like that in this repo sometimes but in this case there's already a ton of tests validating this interaction.

I'd rather keep the below test where the unit survives so we can validate that the amount is correct (can't do that if the unit is defeated obviously)

test/server/cards/01_SOR/events/StrikeTrue.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+                        hand: ['strike-true']
+                    },
+                    player2: {
+                        groundArena: ['specforce-soldier', 'atst'],
+                        spaceArena: ['tieln-fighter']
+                    }
+                });
+            });
+
+            it('should be playable even if there is no friendly units ', function () {
```
AMMayberry1, line 57: Good test! Can we add one more variant confirming that nothing happens if there are friendly units but no enemy units?

test/server/cards/01_SOR/events/StrikeTrue.spec.ts
aruizc: sounds good, I just removed it

test/server/cards/01_SOR/events/StrikeTrue.spec.ts
aruizc: I've added the test case. Thanks for the feedback!

test/server/cards/01_SOR/events/StrikeTrue.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+        });
+
+        describe('Strike True\'s ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['strike-true']
+                    },
+                    player2: {
```
AMMayberry1, line 38: Looks like the player 2 cards aren't relevant for this case, can we just remove the player 2 object so we're only declaring cards relevant to the test. Same for the below test

test/server/cards/01_SOR/events/StrikeTrue.spec.ts
```diff
@@ -0,0 +1,76 @@
+describe('Strike True', function() {
+    integration(function(contextRef) {
+        describe('Strike True\'s ability', function() {
+            beforeEach(function () {
```
AMMayberry1, line 4: Just noticed this - for cases where each test setup is used for only one test case we have a more compact way of doing the setup inside the test case. See Reinforcement Walker tests for an example.

Not required at all, just letting you know the option if you find it more convenient

AMMayberry1 (approved):

PR #168, opened 2024-11-06
Add card piles to gamestate

danbastin:
Adds additional piles to the gamestate. Also updates some properties on cardSummary so that they can be properly rendered and suppress mouseclick events when not selectable.

### DIFF
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -1276,7 +1276,10 @@ class Game extends EventEmitter {
                 playerState[player.name] = player.getState(activePlayer);
             }
 
+            this.currentActionWindow?.activePrompt();
+
             return {
+                playerUpdate: activePlayer.name,
                 id: this.id,
                 manualMode: this.manualMode,
                 name: this.name,
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -1293,8 +1293,6 @@ class Game extends EventEmitter {
                 playerState[player.name] = player.getState(activePlayer);
             }
 
-            this.currentActionWindow?.activePrompt();
-
             return {
                 playerUpdate: activePlayer.name,
                 id: this.id,
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -738,6 +738,19 @@ export class Card extends OngoingEffectSource {
         const isActivePlayer = activePlayer === this.controller;
         const selectionState = activePlayer.getCardSelectionState(this);
 
+        // add exhausted state if the card can be exhausted and is in play
+        let exhaustedState = {};
+        if (this.canBeExhausted() && this.canBeInPlay() && ((this as unknown as InPlayCard).isInPlay() || this.location === Location.Resource)) {
+            const exhausted = (this as unknown as PlayableOrDeployableCard).exhausted;
+            exhaustedState = { exhausted: exhausted };
+        }
+
+        // let damageState = {};
+        // if (this.canBeDamaged() && this.canBeInPlay() && ((this as unknown as InPlayCard).isInPlay() || this.location === Location.Base)) {
+        //     const damage = (this as unknown as NonLeaderUnitCard).damage;
+        //     damageState = { damage: damage };
+        // }
+
         // This is my facedown card, but I'm not allowed to look at it
         // OR This is not my card, and it's either facedown or hidden from me
         if (
@@ -752,8 +765,9 @@ export class Card extends OngoingEffectSource {
                 location: this.location,
                 uuid: isActivePlayer ? this.uuid : undefined
             };
-            return Object.assign(state, selectionState);
+            return { ...state, ...selectionState, ...exhaustedState };
         }
+
 
         const state = {
             id: this.cardData.id,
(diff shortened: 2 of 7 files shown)


server/game/core/Game.js
```diff
@@ -1276,7 +1276,10 @@ class Game extends EventEmitter {
                 playerState[player.name] = player.getState(activePlayer);
             }
 
+            this.currentActionWindow?.activePrompt();
```
AMMayberry1, line 1279: What is this for?

server/game/core/card/Card.ts
```diff
@@ -738,6 +738,19 @@ export class Card extends OngoingEffectSource {
         const isActivePlayer = activePlayer === this.controller;
         const selectionState = activePlayer.getCardSelectionState(this);
 
+        // add exhausted state if the card can be exhausted and is in play
+        let exhaustedState = {};
+        if (this.canBeExhausted() && this.canBeInPlay() && ((this as unknown as InPlayCard).isInPlay() || this.location === Location.Resource)) {
```
AMMayberry1, line 743: Probably better to override `getSummary()` in `PlayableOrDeployableCard` and have it append the exhausted property to the state. That way we don't need to jump through all these hoops here to see if the property exists or not, if we do it from `PlayableOrDeployableCard` it has direct access to the `_exhausted` field and can just drop its value right in. If it's in an invalid location for being exhausted, the value will just be null.

Separately, as a tip for future use (it's not relevant here if we make the above change), you shouldn't need to make the cast of `this` to `InPlayCard` in this situation. `this.canBeInPlay()` has return type `this is InPlayCard` which means that after `this.canBeInPlay()` is determined to be true, typescript will assume that `this` has all of the properties of `InPlayCard`.

server/game/core/gameSteps/prompts/ResourcePrompt.ts
```diff
@@ -92,7 +92,7 @@ export class ResourcePrompt extends AllPlayerPrompt {
     protected resourceSelectedCards(player: Player) {
         if (this.selectedCards[player.name].length > 0) {
             for (const card of this.selectedCards[player.name]) {
-                player.resourceCard(card);
+                player.resourceCard(card, false);
```
AMMayberry1, line 95: Might be better to just ping Chebato to add this to his PR so there isn't a merge collision. Also that way he can add tests for it.

server/game/core/Game.js
danbastin: This is what is adding pass and claim to the player's action prompts. I need to figure out how exactly these are getting into the playerSummary because without it we don't get these buttons, but I had this in the actual gameState and the options were added to both the game and player.

AMMayberry1: Huh, that's really weird because it looks like the method just returns an object and we aren't doing anything with it. It doesn't have any side effects as far as I can tell

AMMayberry1: I'll point at it when I get a chance

danbastin: Entirely possible that it's actually working without this. I had to do some weird hopping bath and forth from Chebato's branch to mine to work on changes since his PR is required to get through the setup phase. I'll retry it once that's merged in.

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -77,6 +78,19 @@ export class PlayableOrDeployableCard extends Card {
         return true;
     }
 
+    public override canBeInPlay(): this is InPlayCard {
+        return false;
+    }
+
+    public override getSummary(activePlayer: Player, hideWhenFaceup: boolean) {
+        const summary = super.getSummary(activePlayer, hideWhenFaceup);
+        if ((this.canBeInPlay() && this.isInPlay()) || this.location === Location.Resource) {
```
AMMayberry1, line 87: Just a note - I think you can simplify this further, it could just be a one liner as:

```typescript
return { ...super.getSummary(activePlayer, hideWhenFaceup), exhausted: this._exhausted };
```

It'll just return `exhausted: null` in cases where exhausted doesn't apply but that's basically the default value that TS gives you when accessing a property that doesn't exist anyway, so it doesn't impact behavior really at all. I guess the difference is that whatever library is sending the data over the wire might still include the `exhausted: null` field in the state which could cause issues if you're not expecting it. Up to you which is better.

server/game/core/Game.js
danbastin: Unsurprisingly, this line was not needed.

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
danbastin: The issue I ran into with this and damage was that I ran into some contract issues when cards were in hand and inelligable attributes were being accessed.

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -77,6 +78,19 @@ export class PlayableOrDeployableCard extends Card {
         return true;
     }
 
+    public override canBeInPlay(): this is InPlayCard {
```
AMMayberry1, line 81: Pretty sure this isn't needed since it's declared this way in Card already

PR #169, opened 2024-11-07
Add Doctor Pershing implementation

Frugghi:
image

### DIFF
--- a/server/game/core/Constants.ts
+++ b/server/game/core/Constants.ts
@@ -35,6 +35,13 @@ export enum StatType {
     Power = 'power'
 }
 
+export enum DamageType {
+    Combat = 'combat',
+    Ability = 'ability',
+    Excess = 'excess',
+    Overwhelm = 'overwhelm'
+}
+
 export enum EffectName {
     AbilityRestrictions = 'abilityRestrictions',
     AdditionalAction = 'additionalActions',
--- a/server/game/gameSystems/SelectCardSystem.ts
+++ b/server/game/gameSystems/SelectCardSystem.ts
@@ -36,6 +36,7 @@ export interface ISelectCardProperties<TContext extends AbilityContext = Ability
  * Functions the same as a targetResolver and used in situations where one can't be created (e.g., costs).
  */
 export class SelectCardSystem<TContext extends AbilityContext = AbilityContext> extends CardTargetSystem<TContext, ISelectCardProperties<TContext>> {
+    public override readonly name: string = 'select-card';
     protected override readonly eventName: MetaEventName.SelectCard;
     protected override readonly defaultProperties: ISelectCardProperties<TContext> = {
         cardCondition: () => true,
--- a/server/game/gameSystems/SelectCardSystem.ts
+++ b/server/game/gameSystems/SelectCardSystem.ts
@@ -36,7 +36,7 @@ export interface ISelectCardProperties<TContext extends AbilityContext = Ability
  * Functions the same as a targetResolver and used in situations where one can't be created (e.g., costs).
  */
 export class SelectCardSystem<TContext extends AbilityContext = AbilityContext> extends CardTargetSystem<TContext, ISelectCardProperties<TContext>> {
-    public override readonly name: string = 'select-card';
+    public override readonly name: string = 'selectCard';
     protected override readonly eventName: MetaEventName.SelectCard;
     protected override readonly defaultProperties: ISelectCardProperties<TContext> = {
         cardCondition: () => true,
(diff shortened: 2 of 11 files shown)


server/game/core/Constants.ts
```diff
@@ -35,6 +35,13 @@ export enum StatType {
     Power = 'power'
 }
 
+export enum DamageType {
+    Combat = 'combat',
+    Ability = 'ability',
+    Excess = 'excess',
+    Overwhelm = 'overwhelm'
+}
```
Frugghi, line 43: I moved this here because importing `DamageSystem` in `CostLibrary` was causing an import cycle.

AMMayberry1: Ugh yeah I hate that that's an issue, glad you were able to resolve it. Thanks for the note

server/game/gameSystems/SelectCardSystem.ts
```diff
@@ -36,6 +36,7 @@ export interface ISelectCardProperties<TContext extends AbilityContext = Ability
  * Functions the same as a targetResolver and used in situations where one can't be created (e.g., costs).
  */
 export class SelectCardSystem<TContext extends AbilityContext = AbilityContext> extends CardTargetSystem<TContext, ISelectCardProperties<TContext>> {
+    public override readonly name: string = 'select-card';
```
AMMayberry1, line 39: nit, styling: we've been using camelCase for descriptor strings generally (the card names we use for tests are kind of a weird exception)

test/server/cards/02_SHD/units/DoctorPershingExperimentingWithLife.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+                // Trigger the ability again
+                reset();
+                context.player1.clickCard(context.doctorPershing);
+                context.player1.clickPrompt('Draw a card');
+                expect(context.player1).toBeAbleToSelectExactly([context.doctorPershing, context.wampa]);
+                expect(context.player1).not.toHaveChooseNoTargetButton();
+
+                // Target a ground friendly unit
+                context.player1.clickCard(context.wampa);
```
AMMayberry1, line 47: No need to make changes here but typically it's safe to assume for testing purposes that any unit which appears as selectable will work correctly and don't need to test all cases. Only exceptions are when there might be distinct behavior differences depending on the target (e.g. targeting self is usually a worthwhile extra case), but if it's just e.g. arena or controller differences it's usually fine. Really appreciate the thorough tests.

test/server/cards/02_SHD/units/DoctorPershingExperimentingWithLife.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+                // Target self
+                context.player1.clickCard(context.doctorPershing);
+                expect(context.doctorPershing.exhausted).toBeTrue();
+                expect(context.doctorPershing.damage).toBe(1);
+                expect(context.player1.hand.length).toBe(3);
+                expect(context.player2.hand.length).toBe(0);
+            });
+
+            it('should not be available if Doctor Pershing is exhausted', function () {
```
AMMayberry1, line 68: It looks like you've picked up this pattern already but just a note - we have this unfortunate issue where the Jasmine tests are really slow and heavy, and we have finite Github Actions time in the free tier. So it's bad test practice but we tend to consolidate tests into as few as reasonably possible. In this case I think this test could be easily appended to the above one?

test/server/cards/02_SHD/units/DoctorPershingExperimentingWithLife.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+                        spaceArena: ['tieln-fighter']
+                    },
+                    player2: {
+                        groundArena: ['frontier-atrt'],
+                        spaceArena: ['cartel-spacer']
+                    }
+                });
+            });
+
+            it('should deal 1 damage to any selected friendly unit and draw a card', function () {
```
AMMayberry1, line 18: Since this is our first instance of damage as a cost, do you mind also adding an integration test case for me? Specifically I want to confirm that if he does his damage to a unit with a shield that it still satisfies the condition. We haven't had a lot of opportunities to confirm that replacement effects work correctly for paying costs yet.

If it has any issues you can just reach out to me or start a different PR if you're interested to do it yourself (either is fine). It's too big an issue to be worth including in this PR if there's a problem there.

test/server/cards/02_SHD/units/DoctorPershingExperimentingWithLife.spec.ts
Frugghi: Makes sense! I gave the Wampa a shield now.

test/server/cards/02_SHD/units/DoctorPershingExperimentingWithLife.spec.ts
Frugghi: I've added a new test scenario to check this!

AMMayberry1 (approved):

PR #170, opened 2024-11-07
Implement Ma Klounkee

jan-preisig:
image

### DIFF
--- a/server/game/cards/02_SHD/events/MaKlounkee.ts
+++ b/server/game/cards/02_SHD/events/MaKlounkee.ts
@@ -16,7 +16,7 @@ export default class MaKlounkee extends EventCard {
             targetResolver: {
                 controller: RelativePlayer.Self,
                 cardTypeFilter: WildcardCardType.NonLeaderUnit,
-                cardCondition: (card, context) => card.hasSomeTrait(Trait.Underworld),
+                cardCondition: (card, _context) => card.hasSomeTrait(Trait.Underworld),
                 immediateEffect: AbilityHelper.immediateEffects.returnToHandFromPlay()
             },
             ifYouDo: {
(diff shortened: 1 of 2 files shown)


server/game/cards/02_SHD/events/MaKlounkee.ts
```diff
@@ -0,0 +1,33 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Return a friendly non-leader Underworld unit to its owner\'s hand',
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                cardCondition: (card, context) => card.hasSomeTrait(Trait.Underworld),
```
AMMayberry1, line 19: nit: since we don't use the context variable you can change it to `_context` to signify that

test/server/cards/02_SHD/events/MaKlounkee.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+                }
+            });
+            const { context } = contextRef;
+
+            context.player1.clickCard('ma-klounkee');
+            expect(context.pykeSentinel).toBeInLocation('hand');
+            expect(context.academyDefenseWalker.damage).toBe(3);
+        });
+
+        it('if no underworld units are in play, nothing happens, but resources get exhausted', function () {
```
AMMayberry1, line 42: Great tests! Since the "if you do" system is a little new, do you mind adding one more where there's just one friendly unit in play to confirm that it works right even though there's no target for the damage

AMMayberry1 (approved):

PR #171, opened 2024-11-07
Add Bombing Run and Outmaneuver implementation

Frugghi:
| Bombing Run | Outmaneuver |
| ------- | ------- |
| Bombing run | Outmaneuver |


AMMayberry1 (commented):
Hi! Thanks a ton for the PR, and great job diving headfirst into some of the systems stuff. Before I do a review, could you reach out on discord to me (Veld)? I'll get you added to the contributor role in the Karabast discord so you can join the design discussions.

server/game/core/ability/abilityTargets/ArenaTargetResolver.ts
```diff
@@ -0,0 +1,54 @@
+import { TargetResolver } from './TargetResolver';
+import { IArenaTargetResolver } from '../../../TargetInterfaces';
+import { AbilityContext } from '../AbilityContext';
+import type Player from '../../Player';
+import PlayerOrCardAbility from '../PlayerOrCardAbility';
+import { Location } from '../../Constants';
+import { isArray } from 'underscore';
+import { isArena } from '../../utils/EnumHelpers';
+
+export class ArenaTargetResolver extends TargetResolver<IArenaTargetResolver<AbilityContext>> {
```
AMMayberry1, line 10: I think it may be a bit overkill to create a whole target resolver for this situation, handful of reasons:

- Only two cards total that use it
- Doesn't align super well with the design of the other target resolvers in that it looks like a player or card resolver but the target "arena" isn't really an input to a system the same way that player or card is and therefore can't be handled implicitly in the same way
- I think the functionality is already pretty well covered by the select target resolver since it should just be two buttons each of which calls a system (see Leia unit or Reinforcement Walker for examples of use)

test/server/cards/01_SOR/events/BombingRun.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+                    expect(context.atst.damage).toBe(3);
+                    expect(context.cartelSpacer.damage).toBe(0);
+                    expect(context.wampa.damage).toBe(3);
+                    expect(context.allianceXwing.damage).toBe(0);
+                    expect(context.bobaFett.damage).toBe(3);
+                    expect(context.lukeSkywalker.damage).toBe(3);
+                });
+            });
+
+            describe('when space arena is chosen', function() {
```
AMMayberry1, line 38: Can we add one more case where there's no units in the selected arena, just b/c target resolvers will often omit options with no available target and we want to make sure that's not happening here. Don't need the test in both cards, just once should be fine

Frugghi: One of the reasons for having `ArenaTargetResolver` was to cover that scenario but I managed to find a way to rewrite it using the existing `SelectTargetResolver` instead.

AMMayberry1 (approved):

PR #173, opened 2024-11-07
#Setup-Phase Mulligan

CheBato:
- Added mulligan simultaneus for both players.
- Added tests to the SetupPhase to support MulliganPrompt.ts

### DIFF
--- a/server/game/core/gameSteps/prompts/MulliganPrompt.ts
+++ b/server/game/core/gameSteps/prompts/MulliganPrompt.ts
@@ -5,11 +5,12 @@ import * as Contract from '../../utils/Contract';
 
 export class MulliganPrompt extends AllPlayerPrompt {
     protected playersDone = new Map<string, boolean>();
-
+    protected playerMulligan = new Map<string, boolean>();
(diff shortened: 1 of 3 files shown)


server/game/core/gameSteps/prompts/MulliganPrompt.ts
```diff
@@ -0,0 +1,57 @@
...
+    public override menuCommand(player, arg): boolean {
+        if (arg === 'yes') {
+            if (this.completionCondition(player)) {
+                return false;
+            }
+            for (const card of player.hand) {
+                player.moveCard(card, 'deck bottom');
+            }
+
+            player.shuffleDeck();
```
AMMayberry1, line 44: Similar to the resourcing, we should wait until both players have committed to their choice before doing anything. This will prevent people from trying to base their choice on whether the other person mulligans

CheBato: You got it!

AMMayberry1 (approved):

PR #174, opened 2024-11-08
Card: Smuggling Compartment 

tony-pacheco:


test/server/cards/01_SOR/upgrades/SmugglingCompartment.spec.ts
```diff
@@ -0,0 +1,51 @@
+describe('Smuggling Compartment', function() {
+    integration(function(contextRef) {
+        describe('Smuggling Compartment\'s on attack ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: [{ card: 'atst', upgrades: ['smuggling-compartment'] }, { card: 'wampa', upgrades: ['smuggling-compartment'] }],
```
VJubert, line 8: why do you have wampa with this upgrades ?

test/server/cards/01_SOR/upgrades/SmugglingCompartment.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+            });
+        });
+
+        describe('Smuggling Compartment', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['smuggling-compartment'],
+                        groundArena: ['snowspeeder', 'battlefield-marine']
```
VJubert, line 36: maybe add some units on space arena and on player 2 to check arena & controller ?

test/server/cards/01_SOR/upgrades/SmugglingCompartment.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+                    }
+                });
+            });
+
+            it('should ready a resource on attack', function () {
+                const { context } = contextRef;
+
+                context.player1.exhaustResources(2);
+
+                const exhaustedResourcesBeforeAbility = context.player1.countExhaustedResources();
```
AMMayberry1, line 21: Don't really need this variable in this case since we know from the above that the number of exhausted resources is 2

test/server/cards/01_SOR/upgrades/SmugglingCompartment.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        describe('Smuggling Compartment', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['smuggling-compartment'],
+                        groundArena: ['snowspeeder', 'battlefield-marine']
+                    },
+                    player2: {
```
AMMayberry1, line 38: Can just omit the `player2` object completely if it's empty

test/server/cards/01_SOR/upgrades/SmugglingCompartment.spec.ts
AMMayberry1: I'm never opposed to more tests but I think it's fine without that, we have tests for checking general attachment rules like that in Upgrade.spec.ts so there's no real need to have those tests included in suites for individual upgrades

test/server/cards/01_SOR/upgrades/SmugglingCompartment.spec.ts
tony-pacheco: This was just left over bad code. Thanks!

test/server/cards/01_SOR/upgrades/SmugglingCompartment.spec.ts
tony-pacheco: Good point!

AMMayberry1 (approved):

PR #176, opened 2024-11-08
Add ISB Agent implementation

Frugghi:
Adds the implementation for ISB Agent and slightly changes the logic for the "Choose no target" button to avoid showing it alongside the "Done" button when only one card needs to be selected.

image

### DIFF
--- a/server/game/core/ability/abilityTargets/CardTargetResolver.ts
+++ b/server/game/core/ability/abilityTargets/CardTargetResolver.ts
@@ -122,7 +122,12 @@ export class CardTargetResolver extends TargetResolver<ICardTargetResolver<Abili
                 passPrompt.hasBeenShown = true;
             }
             if (this.selector.optional) {
-                buttons.push({ text: 'Choose no target', arg: 'noTarget' });
+                // If the selector is for a single card and it will automatically fire on selection,
+                // uses the 'done' arg so that the prompt doesn't show both 'Choose no target' and 'Done' buttons.
+                buttons.push({
+                    text: 'Choose no target',
+                    arg: this.selector.numCards === 1 && this.selector.automaticFireOnSelect(context) ? 'done' : 'noTarget'
+                });
             }
         }
         const mustSelect = legalTargets.filter((card) =>
--- a/test/server/cards/01_SOR/units/ISBAgent.spec.ts
+++ b/test/server/cards/01_SOR/units/ISBAgent.spec.ts
@@ -1,0 +1,75 @@
+describe('ISB Agent', function() {
+    integration(function(contextRef) {
+        describe('ISB Agent\' ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['confiscate', 'waylay', 'isb-agent'],
+                        groundArena: ['atst'],
+                        spaceArena: ['cartel-spacer']
+                    },
+                    player2: {
+                        hand: ['disarm'],
+                        groundArena: ['wampa'],
+                        spaceArena: ['alliance-xwing']
+                    }
+                });
+            });
+
+            describe('when event is revealed', function() {
+                it('should deal 1 damage to a unit', function () {
+                    const { context } = contextRef;
+
+                    context.player1.clickCard(context.isbAgent);
+
+                    expect(context.player1).toBeAbleToSelectExactly([context.confiscate, context.waylay]);
+                    expect(context.player1).toHaveExactPromptButtons(['Choose no target']);
+                    context.player1.clickCard(context.confiscate);
+
+                    expect(context.getChatLogs(3)).toEqual([
+                        'player1 plays ISB Agent',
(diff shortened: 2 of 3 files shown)


test/server/cards/01_SOR/units/ISBAgent.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+            });
+
+            describe('when event is revealed', function() {
+                it('should deal 1 damage to a unit', function () {
+                    const { context } = contextRef;
+
+                    context.player1.clickCard(context.isbAgent);
+
+                    expect(context.player1).toBeAbleToSelectExactly([context.confiscate, context.waylay]);
+                    expect(context.player1).toHaveExactPromptButtons(['Choose no target']);
```
VJubert, line 27: there is a method for this, toHaveChooseNoTarget or something like that

test/server/cards/01_SOR/units/ISBAgent.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+                    },
+                    player2: {
+                        hand: ['disarm'],
+                        groundArena: ['wampa'],
+                        spaceArena: ['alliance-xwing']
+                    }
+                });
+            });
+
+            describe('when event is revealed', function() {
```
VJubert, line 20: describe & it can be merged on a one `it`

test/server/cards/01_SOR/units/ISBAgent.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+                    expect(context.cartelSpacer.damage).toBe(0);
+                    expect(context.wampa.damage).toBe(0);
+                    expect(context.allianceXwing.damage).toBe(0);
+
+                    expect(context.player2).toBeActivePlayer();
+                });
+            });
+        });
+    });
+});
```
VJubert, line 75: maybe add a test without any event on hand ?

test/server/cards/01_SOR/units/ISBAgent.spec.ts
Frugghi: `toHaveChooseNoTarget` doesn't do exactly the same thing because this also checks that there are no other buttons.

server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -122,7 +122,12 @@ export class CardTargetResolver extends TargetResolver<ICardTargetResolver<Abili
                 passPrompt.hasBeenShown = true;
             }
             if (this.selector.optional) {
-                buttons.push({ text: 'Choose no target', arg: 'noTarget' });
+                // If the selector is for a single card and it will automatically fire on selection,
+                // uses the 'done' arg so that the prompt doesn't show both 'Choose no target' and 'Done' buttons.
+                buttons.push({
```
AMMayberry1, line 127: Nice! Thanks

test/server/cards/01_SOR/units/ISBAgent.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+                        'player1 reveals Confiscate due to ISB Agent',
+                    ]);
+
+                    expect(context.player1).toBeAbleToSelectExactly([context.isbAgent, context.atst, context.cartelSpacer, context.wampa, context.allianceXwing]);
+                    context.player1.clickCard(context.wampa);
+
+                    expect(context.isbAgent).toBeInLocation('ground arena');
+                    expect(context.isbAgent.damage).toBe(0);
+                    expect(context.atst.damage).toBe(0);
+                    expect(context.cartelSpacer.damage).toBe(0);
```
AMMayberry1, line 42: Never opposed to verbose test cases, but if you prefer writing fewer lines we generally don't bother checking that units except the target have damage. We have enough confirmation that the targeting + damage pipeline works across the test suite

test/server/cards/01_SOR/units/ISBAgent.spec.ts
AMMayberry1: Could we add an additional parameter to it like `allowOtherButtons = true` or something? Though the other similar methods would also need to be extended

test/server/cards/01_SOR/units/ISBAgent.spec.ts
Frugghi: Yeah, makes sense! Removed :)

test/server/cards/01_SOR/units/ISBAgent.spec.ts
Frugghi: I guess ideally `toHaveExactPromptButton` should be able to accept custom matchers (e.g. `chooseNoTargetButton`, `passAbilityButton`, etc.).

I changed it to:
```Typescript
expect(context.player1).toHaveChooseNoTargetButton();
expect(context.player1).not.toHavePassAbilityButton();
```
because it was quicker :)

AMMayberry1 (approved):

PR #177, opened 2024-11-08
implement emperors legion

jan-preisig:
image


jan-preisig:
Please take a look at my second Test. It didn't work without double pass action, any idea why that happens? Does the setupTest count as played this phase?

test/server/cards/01_SOR/events/TheEmperorsLegion.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+                player2: {
+                    hand: ['superlaser-blast'],
+                    discard: ['tactical-advantage', 'guerilla-attack-pod']
+                }
+            });
+            const { context } = contextRef;
+
+            context.player1.passAction();
+            context.player2.clickCard(context.superlaserBlast);
+            expect(context.pykeSentinel.location).toBe('discard');
```
AMMayberry1, line 21: We typically only use `expect()` to check the results of the card being tested. In this case, we don't need to check that cards correctly end up in discard from SLB since that would be more of a test of that card's implementation.

test/server/cards/01_SOR/events/TheEmperorsLegion.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+            expect(context.seasonedShoretrooper.location).toBe('discard');
+            expect(context.blizzardAssaultAtat.location).toBe('discard');
+            expect(context.cartelSpacer.location).toBe('discard');
+            context.player1.clickCard(context.theEmperorsLegion);
```
AMMayberry1, line 25: Do you mind breaking this into shorter code blocks for readability, 4 - 6 lines to a block is ideal

test/server/cards/01_SOR/events/TheEmperorsLegion.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+            expect(context.pykeSentinel.location).toBe('hand');
+            expect(context.seasonedShoretrooper.location).toBe('hand');
+            expect(context.blizzardAssaultAtat.location).toBe('hand');
+            expect(context.cartelSpacer.location).toBe('hand');
+            expect(context.player2).toBeActivePlayer();
+            // Expect earlier defeated units to still be in the discard pile
```
AMMayberry1, line 31: Actually we may need to set this test up slightly differently. When the test harness populates the discard pile, it kind of cheats the cards in so they were never actually "defeated" from a game perspective.

Ideally we would want to test two phases where we defeat some cards explicitly in the first phase using attacks or something, use `context.moveToNextActionPhase()` to move to a second phase, then do this test with SLB.

We should do something similar for the below test except in the second phase we don't defeat any units.

test/server/cards/01_SOR/events/TheEmperorsLegion.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+            contextRef.setupTest({
+                phase: 'action',
+                player1: {
+                    hand: ['the-emperors-legion'],
+                    groundArena: ['pyke-sentinel', 'seasoned-shoretrooper', 'blizzard-assault-atat'],
+                    spaceArena: ['cartel-spacer'],
+                    discard: ['keep-fighting', 'green-squadron-awing', 'disarm']
+                },
+                player2: {
+                    hand: ['superlaser-blast'],
```
AMMayberry1, line 13: Can we add some units for player2 and that get defeated and confirm that they are not returned by Emperor's Legion

test/server/cards/01_SOR/events/TheEmperorsLegion.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+            context.player2.clickCard(context.superlaserBlast);
+            expect(context.pykeSentinel.location).toBe('discard');
+            expect(context.seasonedShoretrooper.location).toBe('discard');
+            expect(context.blizzardAssaultAtat.location).toBe('discard');
+            expect(context.cartelSpacer.location).toBe('discard');
+            context.player1.clickCard(context.theEmperorsLegion);
+            expect(context.pykeSentinel.location).toBe('hand');
+            expect(context.seasonedShoretrooper.location).toBe('hand');
+            expect(context.blizzardAssaultAtat.location).toBe('hand');
+            expect(context.cartelSpacer.location).toBe('hand');
```
AMMayberry1, line 29: Can we add some units for player2 that get defeated at the same time and confirm that they are not returned by Emperor's Legion

AMMayberry1 (approved):

PR #178, opened 2024-11-08
Add Lieutenant Childsen implementation

Frugghi:
image


server/game/cards/01_SOR/units/LieutenantChildsenDeathStarPrisonWarden.ts
```diff
@@ -0,0 +1,35 @@
...
+            title: 'Reveal up to 4 Vigilance cards from your hand',
+            targetResolver: {
+                mode: TargetMode.UpTo,
+                numCards: 4,
+                controller: RelativePlayer.Self,
+                locationFilter: Location.Hand,
+                cardCondition: (card) => card.hasSomeAspect(Aspect.Vigilance),
+                immediateEffect: AbilityHelper.immediateEffects.reveal(),
+            },
+            then: (context) => ({
```
AMMayberry1, line 24: nit: by convention we call this `thenContext` to differentiate from any other context parameters used inside

test/server/cards/01_SOR/units/LieutenantChildsenDeathStarPrisonWarden.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+                        'player1 reveals Pyke Sentinel due to Lieutenant Childsen',
+                    ]);
+
+                    expect(context.lieutenantChildsen).toHaveExactUpgradeNames(['experience']);
+                    expect(context.player2).toBeActivePlayer();
+
+                    reset();
+
+                    // Reveal 2 Vigilance cards
+                    context.player1.clickCard(context.lieutenantChildsen);
```
AMMayberry1, line 59: We can simplify this test by just checking some key points in terms of number of cards - 0, 1, 4

test/server/cards/01_SOR/units/LieutenantChildsenDeathStarPrisonWarden.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+                    reset();
+
+                    // Reveal 4 Vigilance cards
+                    context.player1.clickCard(context.lieutenantChildsen);
+                    expect(context.player1).toBeAbleToSelectExactly([context.pykeSentinel, context.systemPatrolCraft, context.cargoJuggernaut, context.resilient, context.vanquish]);
+                    context.player1.clickCard(context.cargoJuggernaut);
+                    context.player1.clickCard(context.vanquish);
+                    context.player1.clickCard(context.resilient);
+                    context.player1.clickCard(context.pykeSentinel);
```
AMMayberry1, line 99: Can we click a fifth Vigilance card using `clickCardNonChecking()` to confirm the upper bound is held

test/server/cards/01_SOR/units/LieutenantChildsenDeathStarPrisonWarden.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+                    context.player1.clickCard(context.lieutenantChildsen);
+                    expect(context.player1).toBeAbleToSelectExactly([context.pykeSentinel, context.systemPatrolCraft, context.cargoJuggernaut, context.resilient, context.vanquish]);
+                    context.player1.clickPrompt('Done');
+
+                    expect(context.lieutenantChildsen).toHaveExactUpgradeNames([]);
+                    expect(context.player2).toBeActivePlayer();
+                });
+            });
+
+            describe('when hand does not contain any Vigilance card', function() {
```
AMMayberry1, line 122: Now that we have Childsen, there's a scenario we get asked about all the time it would be great to have a test for if you don't mind. Could you go to `DefeatTiming.spec.ts` and add a case confirming that if there is an enemy Snoke on the field, that Childsen dies despite his ability? I'm confident it will work but it comes up so much we might as well have a test we can point to.

Frugghi: I didn't even know it worked like that! Good to know!

AMMayberry1 (approved):

PR #180, opened 2024-11-09
Add I Had No Choice implementation

Frugghi:
image

### DIFF
--- a/server/game/gameSystems/MoveCardSystem.ts
+++ b/server/game/gameSystems/MoveCardSystem.ts
@@ -92,7 +92,6 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
               (!card.hasRestriction(EffectName.TakeControl, context) &&
                 !card.anotherUniqueInPlay(context.player))) &&
                 (!destination || context.player.isLegalLocationForCardType(card.type, destination)) &&
-                !EnumHelpers.isArena(card.location) &&
                 super.canAffect(card, context)
         );
     }
--- a/test/server/cards/01_SOR/events/IHadNoChoice.spec.ts
+++ b/test/server/cards/01_SOR/events/IHadNoChoice.spec.ts
@@ -1,0 +1,91 @@
+describe('I Had No Choice', function() {
+    integration(function(contextRef) {
+        describe('I Had No Choice\'s ability', function() {
+            describe('when there are two or more non-leader units in play', function() {
+                beforeEach(function () {
+                    contextRef.setupTest({
+                        phase: 'action',
+                        player1: {
+                            hand: ['i-had-no-choice'],
+                            deck: ['foundling', 'pyke-sentinel', 'atst', 'cartel-spacer', 'battlefield-marine'],
+                            groundArena: ['wampa', 'warzone-lieutenant'],
+                        },
+                        player2: {
+                            groundArena: ['viper-probe-droid'],
+                            leader: { card: 'darth-vader#dark-lord-of-the-sith', deployed: true }
+                        }
+                    });
+                });
+
+                it('returns the unit chose by the opponent to its owner hand and put the other ones at the bottom of the deck', function () {
+                    const { context } = contextRef;
+
+                    context.player1.clickCard(context.iHadNoChoice);
+                    expect(context.player1).toBeAbleToSelectExactly([context.wampa, context.warzoneLieutenant, context.viperProbeDroid]);
+                    context.player1.clickCard(context.wampa);
+                    context.player1.clickCard(context.viperProbeDroid);
+                    context.player1.clickPrompt('Done');
+
+                    expect(context.player2).toBeAbleToSelectExactly([context.wampa, context.viperProbeDroid]);
+                    context.player2.clickCard(context.viperProbeDroid);
+
+                    expect(context.viperProbeDroid).toBeInLocation('hand', context.player2);
+                    expect(context.wampa).toBeInLocation('deck', context.player1);
+                    expect(context.player1.deck[5]).toEqual(context.wampa);
+                    expect(context.player2).toBeActivePlayer();
+                });
+            });
+
+            describe('when there is only one non-leader units in play', function() {
+                beforeEach(function () {
+                    contextRef.setupTest({
+                        phase: 'action',
+                        player1: {
+                            hand: ['i-had-no-choice'],
+                            deck: ['foundling', 'pyke-sentinel', 'atst', 'cartel-spacer', 'battlefield-marine'],
+                            groundArena: ['wampa'],
+                        },
+                        player2: {
+                            groundArena: [],
+                            leader: { card: 'darth-vader#dark-lord-of-the-sith', deployed: true }
+                        }
+                    });
+                });
+
+                it('returns the unit to its owner hand', function () {
+                    const { context } = contextRef;
+
+                    context.player1.clickCard(context.iHadNoChoice);
+
+                    expect(context.wampa).toBeInLocation('hand', context.player1);
+                    expect(context.player2).toBeActivePlayer();
+                });
+            });
+
+            describe('when there are no non-leader units in play', function() {
+                beforeEach(function () {
+                    contextRef.setupTest({
+                        phase: 'action',
+                        player1: {
+                            hand: ['i-had-no-choice'],
+                            deck: ['foundling', 'pyke-sentinel', 'atst', 'cartel-spacer', 'battlefield-marine'],
+                            groundArena: [],
+                        },
+                        player2: {
+                            groundArena: [],
+                            leader: { card: 'darth-vader#dark-lord-of-the-sith', deployed: true }
+                        }
+                    });
+                });
+
+                it('does nothing', function () {
+                    const { context } = contextRef;
+
+                    context.player1.clickCard(context.iHadNoChoice);
+
+                    expect(context.player2).toBeActivePlayer();
+                });
+            });
+        });
+    });
+});
--- a/server/game/cards/01_SOR/events/IHadNoChoice.ts
+++ b/server/game/cards/01_SOR/events/IHadNoChoice.ts
@@ -1,0 +1,51 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { Location, RelativePlayer, TargetMode, WildcardCardType, WildcardLocation } from '../../../core/Constants';
+
+export default class IHadNoChoice extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '2715652707',
+            internalName: 'i-had-no-choice',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Choose up to 2 non-leader units. An opponent chooses 1 of those units. Return that unit to its owner’s hand and put the other on the bottom of its owner’s deck.',
+            targetResolvers: {
+                targetUnits: {
+                    mode: TargetMode.UpTo,
+                    numCards: 2,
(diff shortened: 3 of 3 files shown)


server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -92,7 +92,6 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
               (!card.hasRestriction(EffectName.TakeControl, context) &&
                 !card.anotherUniqueInPlay(context.player))) &&
                 (!destination || context.player.isLegalLocationForCardType(card.type, destination)) &&
-                !EnumHelpers.isArena(card.location) &&
```
Frugghi, line 95: This check prevents moving cards that are in play, I don't have any context about this so I'm not really sure if it's correct to remove this. The alternative would be to introduce a new system to move cards in play into the deck.

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -1,11 +1,12 @@
 import type { AbilityContext } from '../core/ability/AbilityContext';
 import type { Card } from '../core/card/Card';
-import { CardType, EffectName, EventName, Location, WildcardCardType } from '../core/Constants';
+import { CardType, EffectName, EventName, GameStateChangeRequired, Location, LocationFilter, WildcardCardType } from '../core/Constants';
 import * as EnumHelpers from '../core/utils/EnumHelpers';
 import { type ICardTargetSystemProperties, CardTargetSystem } from '../core/gameSystem/CardTargetSystem';
 
 export interface IMoveCardProperties extends ICardTargetSystemProperties {
     destination?: Location;
```
AMMayberry1, line 8: I think it would be good to constrain the acceptable values for location since there are more specific systems that should be used in some cases. Specifically, this should never be used to move a card into discard, resources, or play arenas as we have more specific systems for those.

Probably would be good to add short jsdoc on the property and on the system itself briefly explaining why it's set up that way as well.

test/server/cards/01_SOR/events/IHadNoChoice.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+                    expect(context.player1).toBeAbleToSelectExactly([context.wampa, context.warzoneLieutenant, context.viperProbeDroid]);
+                    context.player1.clickCard(context.wampa);
+                    context.player1.clickCard(context.viperProbeDroid);
+                    context.player1.clickPrompt('Done');
+
+                    expect(context.player2).toBeAbleToSelectExactly([context.wampa, context.viperProbeDroid]);
+                    context.player2.clickCard(context.viperProbeDroid);
+
+                    expect(context.viperProbeDroid).toBeInLocation('hand', context.player2);
+                    expect(context.wampa).toBeInBottomOfDeck(context.player1, 6);
```
AMMayberry1, line 33: I think the second parameter should be 1 here, since we expect it to be the single bottom-most card

test/server/cards/01_SOR/events/IHadNoChoice.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    integration(function(contextRef) {
+        describe('I Had No Choice\'s ability', function() {
+            describe('when there are two or more non-leader units in play', function() {
+                beforeEach(function () {
+                    contextRef.setupTest({
+                        phase: 'action',
+                        player1: {
+                            hand: ['i-had-no-choice'],
+                            deck: ['foundling', 'pyke-sentinel', 'atst', 'cartel-spacer', 'battlefield-marine'],
+                            groundArena: [{ card: 'wampa', upgrades: ['experience', 'resilient'] }, 'warzone-lieutenant'],
```
AMMayberry1, line 11: What's the reason for the experience on the wampa?

test/server/cards/01_SOR/events/IHadNoChoice.spec.ts
```diff
@@ -0,0 +1,92 @@
+describe('I Had No Choice', function() {
+    integration(function(contextRef) {
+        describe('I Had No Choice\'s ability', function() {
+            describe('when there are two or more non-leader units in play', function() {
+                beforeEach(function () {
+                    contextRef.setupTest({
+                        phase: 'action',
+                        player1: {
+                            hand: ['i-had-no-choice'],
+                            deck: ['foundling', 'pyke-sentinel', 'atst', 'cartel-spacer', 'battlefield-marine'],
```
AMMayberry1, line 10: In the future it's fine to just use duplicates of 'atst' as a filler card if you need to pad stuff out, just to save typing time

test/server/cards/01_SOR/events/IHadNoChoice.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+                    context.player2.clickCard(context.viperProbeDroid);
+
+                    expect(context.viperProbeDroid).toBeInLocation('hand', context.player2);
+                    expect(context.wampa).toBeInBottomOfDeck(context.player1, 6);
+                    expect(context.resilient).toBeInLocation('discard', context.player1);
+                    expect(context.player2).toBeActivePlayer();
+                });
+            });
+
+            describe('when there is only non-leader units in play', function() {
```
AMMayberry1, line 39: nit: not sure what this test name means

test/server/cards/01_SOR/events/IHadNoChoice.spec.ts
Frugghi: Ah, I thought the second parameter was supposed to indicate the size of the deck!

test/server/cards/01_SOR/events/IHadNoChoice.spec.ts
Frugghi: Just to ensure that token upgrades are destroyed, `MoveCardSystem` didn't support that previously.

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -81,19 +108,35 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
+            return false;
+        }
+
+        // Ensure that the card has no restrictions that would prevent it from being moved to a different player
+        if (changePlayer && (card.hasRestriction(EffectName.TakeControl, context) || card.anotherUniqueInPlay(context.player))) {
+            return false;
+        }
+
+        // Ensure that if the card is in play, it can only be moved to the hand or deck
+        if (EnumHelpers.isArena(card.location) && destination !== Location.Hand && destination !== Location.Deck) {
```
AMMayberry1, line 130: Looking at these again, I'm kind of conflicted as to whether these should silently fail or actually throw a contract fail error. I generally lean towards the latter so that we can catch static configuration glitches early, unless there's a strong case to be made for why it should be allowable to attempt to perform an action that fails (which we do sometimes). Just can't think of a strong reason to here.

test/server/cards/01_SOR/events/IHadNoChoice.spec.ts
AMMayberry1: Makes sense. Trying to figure out if there's an actual check we could do that the behavior is correct. Or was it just throwing an error previously?

Frugghi: It was throwing an error saying something like "you cannot move experience tokens to discard pile".

server/game/gameSystems/MoveCardSystem.ts
Frugghi: `PlayEventAction` uses `MoveCardSystem` to move cards to the discard so we won't be able to remove that location.

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -1,19 +1,27 @@
...
+ *
+ * @property destination - The target location for the card. Excludes discard pile, space arena, ground arena, and resources.
+ * @property shuffle - Indicates whether the card should be shuffled into the destination.
+ * @property bottom - Indicates whether the card should be placed at the bottom of the destination.
+ */
 export interface IMoveCardProperties extends ICardTargetSystemProperties {
-    destination?: Location;
-    switch?: boolean;
-    switchTarget?: Card;
+    destination?: Exclude<Location, Location.SpaceArena | Location.GroundArena | Location.Resource>;
```
AMMayberry1, line 22: Discard also should be excluded since that's managed by the various discard systems

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -24,76 +32,86 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
-                !card.anotherUniqueInPlay(context.player))) &&
-                (!destination || context.player.isLegalLocationForCardType(card.type, destination)) &&
-                !EnumHelpers.isArena(card.location) &&
-                super.canAffect(card, context)
-        );
+    public override canAffect(card: Card, context: TContext, additionalProperties = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        const { destination } = this.generatePropertiesFromContext(context, additionalProperties) as IMoveCardProperties;
+
+        // Ensure that we have a valid destination and that the card can be moved there
+        if (!destination || !context.player.isLegalLocationForCardType(card.type, destination)) {
```
AMMayberry1, line 104: Honestly I think this one should probably also be a contract fail

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -24,76 +32,86 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
+            return false;
+        }
+
+        // Ensure that if the card is returning to the hand, it must be in the discard pile or in play or be a resource
+        if (destination === Location.Hand && !EnumHelpers.cardLocationMatches(card.location, Location.Discard) && !EnumHelpers.cardLocationMatches(card.location, Location.Resource) && !EnumHelpers.isArena(card.location)) {
```
AMMayberry1, line 109: Don't need to use cardLocationMatches for these, that's more for when one side or the other could be a wildcard (`WildcardLocation`) instead of a concrete location (`Location`). In this case `card.location` will always be `Location`, and the values we're checking against are also `Location`. So just can do e.g. `[Location.Discard, Location.Resurce].includes(card.location)`

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -24,76 +32,86 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
+            Contract.fail(`Cannot use MoveCardSystem to return a card to hand from ${card.location}`);
```
AMMayberry1, line 110: Instead of having an if / else that calls Contract.fail you can just use one of the assertion forms that tests a condition:

```typescript
Contract.assertFalse(destination === Location.Hand && <rest of statement>, `Cannot use MoveCardSystem to return a card to hand from ${card.location});
```

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -24,76 +32,86 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
+            return false;
```
AMMayberry1, line 111: The `Contract` methods throw an exception so this isn't needed

server/game/gameSystems/MoveCardSystem.ts
Frugghi: Unfortunately it can't be done because of this:
https://github.com/SWU-Karabast/forceteki/blob/bc614093334ce3356e4071b94a428deabf701dff/server/game/actions/PlayEventAction.ts#L38-L41

AMMayberry1 (approved):

PR #181, opened 2024-11-10
Abilities that play cards with discounts.

Ennis0:
PlayCardSystem can now accept a CostAdjuster that will apply to that card play.
Created a CostAdjusterLibrary included in AbilityHelper.
Added the `Free` CostAdjustType.
Implemented Galactic Ambition and Alliance Dispatcher as tests of the new functionality.


server/game/core/ongoingEffect/OngoingEffectEngine.ts
```diff
@@ -152,6 +152,10 @@ export class OngoingEffectEngine {
         this.effectsChangedSinceLastCheck = this.unapplyAndRemove((effect) => effect.duration === Duration.UntilEndOfRound);
     }
 
+    private onActionEnded() {
```
AMMayberry1, line 155: I may not have time to do a full review right now but wanted to give one piece of top-level feedback. We might want to couple the cost adjuster a little more tightly to the specific ability that's being resolved. I worry that if we activate it for the duration of the action we still may run into issues like e.g. if someone triggers the ability but then hits the cancel button at the target stage to revert back to action selection.

An approach that might be useful both as a future feature and for general bug protection would be to make it so that the cost adjust can be configured to only match the specific play card action that we're intending it for. Haven't thought about this deeply but maybe a way to solve this is that cost adjusters don't have to be bound to the player, but could also be provided from the ability itself? That seems like a fairly elegant way to provide these one-off adjusters that are tied directly to the executing action. Could just add it as a new constructor parameter / property for PlayCardAction.

Ennis0: I think that is a good idea. I finally figured out where to put things for the implementation of it, but I ran into a really strange error: "AbilityContext is not a constructor" when it's called in PlayerAction. I didn't change PlayerAction or AbilityContext at all so I'm not sure what's wrong.

server/game/core/ability/PlayCardAction.ts
```diff
@@ -6,14 +6,22 @@ import { KeywordName, PhaseName, PlayType, Stage } from '../Constants';
...
-    protected playType: PlayType;
-
-    public constructor(card: Card, title: string, playType: PlayType, additionalCosts: ICost[] = [], targetResolver: IActionTargetResolver = null) {
-        super(card, PlayCardAction.getTitle(title, playType), additionalCosts.concat(CostLibrary.payPlayCardResourceCost(playType)), targetResolver);
+    public constructor(
+        card: Card,
+        title: string,
+        protected playType: PlayType,
+        triggerHandlingMode: TriggerHandlingMode = TriggerHandlingMode.ResolvesTriggers,
+        public costAdjuster: CostAdjuster = null,
```
AMMayberry1, line 20: Needs readonly, so does `playType` actually. Also I'm realizing I don't love declaring properties in the ctor params like this, it's actually kind of hard to notice them imo and decreases overall readability. If you agree then would you mind creating these as explicit properties and just doing like `this.playType = playType` in the ctor? If you like it this way then feel free to keep it.

server/game/costs/PlayCardResourceCost.ts
```diff
@@ -22,9 +22,11 @@ export class PlayCardResourceCost<TContext extends AbilityContext = AbilityConte
             return false;
         }
 
+        const costAdjusterFromAbility = 'costAdjuster' in context.ability ? context.ability.costAdjuster : null;
```
AMMayberry1, line 25: This is a bit brittle, if we change the name of this property on PlayCardAction then it will silently break. If we do `Contract.assertTrue(context.ability.isCardPlayed())` then type narrowing should make `context.ability.costAdjuster` accessible

server/game/core/cost/CostAdjuster.ts
```diff
@@ -18,12 +18,12 @@ export enum CostAdjustType {
 interface ICostAdjusterPropertiesBase {
 
     /** The type of cards that can be reduced */
-    cardTypeFilter: CardTypeFilter;
+    cardTypeFilter?: CardTypeFilter;
 
     /** The type of cost adjustment */
     costAdjustType: CostAdjustType;
 
-    /** @deprecated (not yet tested) The number of cost reductions permitted */
```
AMMayberry1, line 26: Did you end up testing this while you were developing? So long as you saw it work at least once I'm fine to remove the deprecated comment

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -33,15 +40,37 @@ export class PlayCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
+        /* if (this.properties.costAdjusterProperties) {
+            this.queueApplyCostAdjusterGameSteps(this.properties.costAdjusterProperties, newContext);
+        }*/
         event.context.game.queueStep(new AbilityResolver(event.context.game, newContext, event.optional));
     }
 
+    private queueApplyCostAdjusterGameSteps(costAdjusterProperties: ICostAdjusterProperties, context: AbilityContext) {
+        const applyCostAdjusterSystem = new PlayerLastingEffectSystem({
+            duration: Duration.UntilEndOfAction,
+            effect: modifyCost(Object.assign(costAdjusterProperties, { limit: AbilityLimit.perGame(1), playingTypes: context.playType }))
```
AMMayberry1, line 60: nit: a little cleaner to use object assignment syntax instead of Object.assign imo - e.g. `{ ...costAdjusterProperties, limit: AbilityLimit.perGame(1), playingTypes: context.playType }`

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -33,15 +40,37 @@ export class PlayCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
+        });
+        const effectEvents = [];
+        applyCostAdjusterSystem.queueGenerateEventGameSteps(effectEvents, context, {});
+        context.game.queueSimpleStep(() => context.game.openEventWindow(effectEvents), 'open event window for application of cost adjuster');
+    }
+
+    private makeCostAdjuster(properties: ICostAdjusterProperties, context: AbilityContext) {
+        if (properties) {
+            return new CostAdjuster(context.game, context.source, properties);
```
AMMayberry1, line 69: Could do all in one line with a ternary

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -33,15 +40,37 @@ export class PlayCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
+            duration: Duration.UntilEndOfAction,
+            effect: modifyCost(Object.assign(costAdjusterProperties, { limit: AbilityLimit.perGame(1), playingTypes: context.playType }))
+        });
+        const effectEvents = [];
+        applyCostAdjusterSystem.queueGenerateEventGameSteps(effectEvents, context, {});
+        context.game.queueSimpleStep(() => context.game.openEventWindow(effectEvents), 'open event window for application of cost adjuster');
+    }
+
+    private makeCostAdjuster(properties: ICostAdjusterProperties, context: AbilityContext) {
+        if (properties) {
```
AMMayberry1, line 68: If null is an acceptable value of properties then we should add `| null` to the type for that parameter

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -74,11 +103,12 @@ export class PlayCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
      * Generate a play card ability for the specified card.
      */
-    private generatePlayCardAbility(card: Card, playType: PlayType) {
+    private generatePlayCardAbility(card: Card, playType: PlayType, costAdjusterForThisAction = null) {
+        const triggerHandlingMode = this.properties.nested ? TriggerHandlingMode.ResolvesTriggers : TriggerHandlingMode.PassesTriggersToParentWindow;
         switch (card.type) {
-            case CardType.BasicUnit: return new PlayUnitAction(card, playType, this.properties.entersReady);
-            case CardType.BasicUpgrade: return new PlayUpgradeAction(card, playType);
-            case CardType.Event: return new PlayEventAction(card, playType);
+            case CardType.BasicUnit: return new PlayUnitAction(card, playType, this.properties.entersReady, triggerHandlingMode, costAdjusterForThisAction);
```
AMMayberry1, line 109: Just a heads-up this will have a merge conflict with the code Chebato just merged, he passes a property object to this method now so we can probably just append to that (if I'm remembering right)

server/game/core/cost/CostAdjuster.ts
Ennis0: It was already tested as far as I can tell. Bendu's effect, for example, successfully discounts only the next card played.

server/game/costs/PlayCardResourceCost.ts
Ennis0: Doesn't work: `Property 'costAdjuster' does not exist on type 'PlayerOrCardAbility'`

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -18,7 +22,10 @@ export interface IPlayCardProperties extends ICardTargetSystemProperties {
     optional?: boolean;
     entersReady?: boolean;
     playType?: PlayType;
-    // TODO: implement a "nested" property that controls whether triggered abilities triggered by playing the card resolve after that card play or after the whole ability
+    costAdjusterProperties?: ICostAdjusterProperties;
```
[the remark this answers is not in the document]
Ennis0: Ideally I would like to use the same methods that exist for creating cost adjusters, that's the idea that I started with. I think the problem is that those methods are all about creating ongoing effects with cost adjusters, not directly making cost adjusters. It might be possible to use those methods(so PlayCardSystem would accept an IOngoingEffectGenerator technically because that's what they return), but extract the cost adjuster from the generated ongoing effect(without actually applying the effect to the game at all) and pass that to the places that need it. What do you think?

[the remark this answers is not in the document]
AMMayberry1: Oh yeah I don't think we can directly use the methods for the reason you say, it would be more about using the same naming convention and expected set of properties

server/game/costs/PlayCardResourceCost.ts
AMMayberry1: Ahh right because PlayerOrCardAbility is still a JS file. I actually just tried something out and made a discovery - I thought that it wasn't possible to use type predicates like this correctly with a JS class but it turns out you can. If you can make the following tweak to the code the above will work.

In PlayerOrCardAbility change `isCardPlayed()` to be this (including the jsdoc):

```typescript
// TODO: refactor the other methods to also be type predicates
/**
 * Indicates whether a card is played as part of the resolution this ability
 * @returns {this is PlayCardAction}
 */
isCardPlayed() {
    return false;
}
```

then make a similar change in PlayCardAction:

```typescript
public override isCardPlayed(): this is PlayCardAction {
    return true;
}
```

Then it should work!

Ennis0: That does let us use `context.ability.costAdjuster` directly, but unfortunately importing `PlayCardAction` into PlayerOrCardAbility.js so we can declare the type predicate causes (what I think is) an import error:
`TypeError: Class extends value [object Object] is not a constructor or null
    at Object.<anonymous> (/home/ennis/MyStuff/misc/programming/Github/forceteki/server/game/core/ability/PlayerAction.js:6:28)`

server/game/gameSystems/PlayCardSystem.ts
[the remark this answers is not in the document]
Ennis0: Alright, I changed PlayCardSystem to directly take a CostAdjuster, and created a CostAdjusterLibrary which I included in AbilityHelper.

server/game/costs/PlayCardResourceCost.ts
Ennis0: Figured it out. I took a hint from isAttackAction() further down and did an import right in the return type declaration.

server/game/core/ability/CardAbilityStep.js
```diff
@@ -135,7 +135,7 @@ class CardAbilityStep extends PlayerOrCardAbility {
         let effectShouldResolve;
 
         if (this.properties.ifYouDo) {
-            ifAbility = this.properties.ifYouDo;
+            ifAbility = typeof this.properties.ifYouDo === 'function' ? this.properties.ifYouDo(context) : this.properties.ifYouDo;
```
AMMayberry1, line 138: I think this change isn't needed anymore due to the fix we did on line 147 in a recent merge, can just do `ifAbility = this.properties.ifYouDo;`

server/game/core/ability/PlayCardAction.ts
```diff
@@ -9,21 +9,34 @@ import PlayerAction from './PlayerAction';
...
-        public costAdjuster: CostAdjuster = null,
-        targetResolver: IActionTargetResolver = null,
-        additionalCosts: ICost[] = []
-    ) {
-        super(card, PlayCardAction.getTitle(title, playType), additionalCosts.concat(CostLibrary.payPlayCardResourceCost(playType)), targetResolver, triggerHandlingMode);
+    protected readonly playType: PlayType;
+    public readonly costAdjuster: CostAdjuster;
+
+    public constructor(properties: IPlayCardActionProperties) {
+        const propertiesWithDefaults = Object.assign({}, { title: 'Play this card', playType: PlayType.PlayFromHand, triggerHandlingMode: TriggerHandlingMode.ResolvesTriggers, additionalCosts: [] }, properties);
```
AMMayberry1, line 29: If you do end up keeping this as Object.assign then I don't think the first `{}` is necessary

server/game/core/ability/PlayerOrCardAbility.js
```diff
@@ -8,6 +8,7 @@ const { v4: uuidv4 } = require('uuid');
 const { PlayerTargetResolver } = require('./abilityTargets/PlayerTargetResolver.js');
 const { DropdownListTargetResolver } = require('./abilityTargets/DropdownListTargetResolver.js');
 const { TriggerHandlingMode } = require('../event/EventWindow.js');
+// const { PlayCardAction } = require('./PlayCardAction.js');
```
AMMayberry1, line 11: Can remove this looks like

server/game/actions/PlayEventAction.ts
```diff
@@ -1,16 +1,9 @@
...
-import { Card } from '../core/card/Card';
 import * as Contract from '../core/utils/Contract.js';
 import { PlayCardContext, PlayCardAction } from '../core/ability/PlayCardAction.js';
-import { AbilityContext } from '../core/ability/AbilityContext.js';
-import { MoveCardSystem } from '../gameSystems/MoveCardSystem.js';
 import { GameEvent } from '../core/event/GameEvent.js';
 
 export class PlayEventAction extends PlayCardAction {
-    public constructor(card: Card, playType: PlayType = PlayType.PlayFromHand) {
-        super(card, 'Play this event', playType);
```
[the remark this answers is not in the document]
Ennis0: Good catch

server/game/core/ability/PlayCardAction.ts
Ennis0: I actually didn't realize you could do property overwrites with the spread operator. I'm fine using that if that's going to be the preferred style for the project. This line would become `const propertiesWithDefaults = { ...{ title: 'Play this card', playType: PlayType.PlayFromHand, triggerHandlingMode: TriggerHandlingMode.ResolvesTriggers, additionalCosts: [] }, ...properties };`, right?

server/game/AbilityHelper.ts
```diff
@@ -3,11 +3,13 @@ import Effects from './ongoingEffects/OngoingEffectLibrary';
...
+import * as costAdjusters from './costs/CostAdjusterLibrary';
 
 export = {
     limit: AbilityLimit,
     ongoingEffects: Effects,
     costs: Costs,
     immediateEffects: GameSystems,
-    stateWatchers: StateWatcherLibrary
+    stateWatchers: StateWatcherLibrary,
+    costAdjusters: costAdjusters
```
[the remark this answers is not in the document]
AMMayberry1: Yeah good call, probably right tbh. That sounds good to me

server/game/core/ability/PlayCardAction.ts
AMMayberry1: I think all the properties can be flat: `const propertiesWithDefaults = { title: 'Play this card', playType: PlayType.PlayFromHand, triggerHandlingMode: TriggerHandlingMode.ResolvesTriggers, additionalCosts: [], ...properties };`

Lmk if that doesn't seem to be overwriting correctly

AMMayberry1 (approved):

PR #182, opened 2024-11-10
Two cards implemented

Bluffkin:
Implemented cards and tests for:
Rallying Cry (SOR)
Gentle Giant (SHD)


server/game/cards/02_SHD/units/GentleGiant.ts
```diff
@@ -0,0 +1,29 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '8190373087',
+            internalName: 'gentle-giant'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'You may heal damage from another unit equal to the damage on this unit',
```
AMMayberry1, line 15: nit: we typically don't include the "You may" in ability titles since it will be implied by the presence of a "Pass" button. We aim for titles just to describe the potential effect

server/game/cards/02_SHD/units/GentleGiant.ts
```diff
@@ -0,0 +1,29 @@
...
+            optional: true,
+            targetResolver: {
+                cardCondition: (card, context) => card !== context.source,
+                cardTypeFilter: WildcardCardType.Unit,
+                immediateEffect: AbilityHelper.immediateEffects.heal((context) => ({
```
AMMayberry1, line 20: nit: just for readability, could put this all in one since there's just one property. if you like it spread out though feel free to keep it that way

test/server/cards/02_SHD/units/GentleGiant.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+                // Healing Target
+                expect(context.player1).toBeAbleToSelectExactly([context.r2d2, context.c3po, context.wampa]);
+                context.player1.clickCard(context.c3po);
+
+                // Confirm Results
+                expect(context.gentleGiant.exhausted).toBe(true);
+                expect(context.c3po.damage).toBe(0);
+            });
+
+            it('should heal 1 damage from a unit', function () {
```
AMMayberry1, line 37: Typically we aggregate multiple test cases into one if easily possible. This is bad test practice but we have a practical problem where the test framework we use is very slow and heavy and we have limited free github actions time in the free tier. So we need to consolidate a bit.

Typically with on-attack abilities like this we'll chain them together with a reset function that un-exhausts and heals relevant units, with comments for each case. Take a look at the tests for Rukh as an example of how we typically set things up.

test/server/cards/02_SHD/units/GentleGiant.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+                            { card: 'r2d2#ignoring-protocol', damage: 3 },
+                            { card: 'c3po#protocol-droid', damage: 1 }],
+                    },
+                    player2: {
+                        groundArena: [{ card: 'wampa', damage: 1 }]
+                    }
+                });
+            });
+
+            it('should heal a target with 1 damage to full', function () {
```
AMMayberry1, line 19: This case isn't really needed at this point, we have a lot of tests already confirming that healing to full works. Can remove it just to keep this test suite simple and easy to read

test/server/cards/02_SHD/units/GentleGiant.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+                // Healing Target
+                expect(context.player1).toBeAbleToSelectExactly([context.r2d2, context.c3po, context.wampa]);
+                context.player1.clickCard(context.r2d2);
+
+                // Confirm Results
+                expect(context.gentleGiant.exhausted).toBe(true);
+                expect(context.r2d2.damage).toBe(2);
+            });
+
+            it('should be able to heal an enemy unit', function () {
```
AMMayberry1, line 55: This one also is probably unnecessary, typically it's enough to confirm that a card appears in the target selection list. We assume in most cases that if it works for one target it would work for all targets, unless there's some reason there would be behavior differences (e.g. if a unit can target itself for stuff we usually test that explicitly).

test/server/cards/01_SOR/events/RallyingCry.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+            it('should give Raid 2 to friendly units', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.rallyingCry);
+                context.player2.passAction();
+                context.player1.clickCard(context.deathTrooper);
+                expect(context.player1).toBeAbleToSelectExactly([context.p2Base, context.ruggedSurvivors]);
+                context.player1.clickCard(context.p2Base);
+                expect(context.p2Base.damage).toBe(5);
```
AMMayberry1, line 27: Can we also confirm that imperial interceptor gets the buff by doing an attack. Other cases look great!

test/server/cards/01_SOR/events/RallyingCry.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['rallying-cry'],
+                        groundArena: ['death-trooper'],
+                        spaceArena: ['imperial-interceptor']
+                    },
+                    player2: {
+                        groundArena: ['rugged-survivors'],
```
AMMayberry1, line 13: Seems like we can remove rugged survivors for this test, it's not being used for anything. Be aware that will change the Death Trooper test since it will just automatically select the base as the only target, so the attack will happen with just one click.

test/server/cards/02_SHD/units/GentleGiant.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                    if (passAction) {
+                        context.player2.passAction();
+                    }
+                };
+
+                // CASE 1: Heal a target
+                // Attack
+                context.player1.clickCard(context.gentleGiant);
+                expect(context.gentleGiant).toBeInLocation('ground arena');
+                expect(context.player1).toBeAbleToSelectExactly([context.p2Base, context.wampa]);
```
AMMayberry1, line 34: Don't need these checks since they're not really testing the card implementation. We can safely assume that test setup and basic stuff like attack just works since we have other tests that cover them, want to keep card tests as streamlined as possible

test/server/cards/02_SHD/units/GentleGiant.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.r2d2, context.c3po, context.wampa]);
+                context.player1.clickCard(context.r2d2);
+
+                // Confirm Results
+                expect(context.gentleGiant.exhausted).toBe(true);
+                expect(context.r2d2.damage).toBe(2);
+
+                reset();
+
+                // CASE 2: Should be able to pass
```
AMMayberry1, line 47: Instead of adding this case you can just add `expect(context.player1).toHavePassAbilityButton();` at line 39 below the selection prompt. Note that the button sometimes has different names (we have a work item to fix that) so it may need to be like `toHaveChooseNoTargetsButton()` or something depending on what you see in the error message.

If the pass button appears we can safely assume it works without testing it explicitly.

test/server/cards/02_SHD/units/GentleGiant.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: [
+                            { card: 'gentle-giant', damage: 1 },
+                            { card: 'r2d2#ignoring-protocol', damage: 3 },
+                            { card: 'c3po#protocol-droid', damage: 1 }],
+                    },
+                    player2: {
+                        groundArena: [{ card: 'wampa', damage: 1 }]
```
AMMayberry1, line 14: Let's remove the damage from anything that isn't actually going to be targeted for the heal effect just to make it super obvious what the test will be doing. That way if anyone accidentally breaks this test and needs to see what's going on they can tell really fast.

test/server/cards/02_SHD/units/GentleGiant.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                    }
+                });
+            });
+
+            it('heal a target', function () {
```
AMMayberry1, line 19: nit: can we add a bit more detail in the test case name in case someone unfamiliar with the card ends up needing to look at this test. Also note that the test name is concatenated between `describe()` and `it()`. So in this case the name would be: `Gentle Giant's ability heal a target.` Ideally it would be something like `Gentle Giant's ability can heal a uniot for the amount of damage it has`.

If you want to see the case name you can just force the test to fail

AMMayberry1:
Two requests for future PRs:
1. When you've finished updating, please resolve all comments unless any have open questions you need me to answer. That makes it easier to track when a PR is ready for review again.
2. Please include the card(s) image in the PR description so I can more easily double-check the implementation

Thanks!

AMMayberry1 (approved):

PR #183, opened 2024-11-10
Public Enemy

Bluffkin:
Implemented card and tests


server/game/cards/02_SHD/upgrades/PublicEnemy.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.addGainKeywordTargetingAttached({
+            keyword: KeywordName.Bounty,
+            ability: {
+                title: 'Give a Shield token to a unit.',
+                targetResolver: {
+                    controller: RelativePlayer.Any,
+                    cardTypeFilter: [WildcardCardType.Unit],
```
AMMayberry1, line 20: Doesn't need to be an array

test/server/cards/02_SHD/upgrades/PublicEnemy.spec.ts
```diff
@@ -0,0 +1,52 @@
+describe('Public Enemy', function() {
+    integration(function(contextRef) {
+        describe('Public Enemy\'s Bounty ability', function() {
+            it('should give a unit a shield', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: [
+                            { card: 'battlefield-marine' },
```
AMMayberry1, line 9: If you aren't adding any special properties like damage or upgrades you can just provide these as string names. E.g.:

```typescript
groundArena: ['battlefield-marine', 'atst']
```

Can mix and match both styles as needed

test/server/cards/02_SHD/upgrades/PublicEnemy.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                            { card: 'wampa', upgrades: ['public-enemy'] },
+                            { card: 'republic-tactical-officer' }
+                        ]
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.atst);
+                expect(context.player1).toBeAbleToSelectExactly([context.p2Base, context.wampa, context.republicTacticalOfficer]);
```
AMMayberry1, line 24: Can skip this, we don't need to test that combat targets are working right

test/server/cards/02_SHD/upgrades/PublicEnemy.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                // Destroys Wampa
+                context.player1.clickCard(context.wampa);
+                expect(context.atst.damage).toBe(4);
+                // Checking Wampa is destroyed
+                expect(context.wampa).toBeInLocation('discard');
+                // Selecting Unit to give shield
```
AMMayberry1, line 30: The comments are good, do you mind also adding a little whitespace to break up the code block for readability

test/server/cards/02_SHD/upgrades/PublicEnemy.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                expect(context.player1).toHavePrompt('Choose a unit');
```
AMMayberry1, line 31: Can remove this as well, it's enough to check that the pass button and the targets are there

test/server/cards/02_SHD/upgrades/PublicEnemy.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                expect(context.player1).toHavePassAbilityButton();
+                expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine, context.atst, context.republicTacticalOfficer]);
+                context.player1.clickCard(context.battlefieldMarine);
+                // Unit should have shield
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['shield']);
+
+                // New round
```
AMMayberry1, line 38: You can end the test here, there's not a need to confirm that shields are working correctly in this test. Typically we just add tests to confirm that what we wrote in that specific card's implementation file is done right, so e.g. target filters and having the right effect

test/server/cards/02_SHD/upgrades/PublicEnemy.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.atst);
+
+                // Destroys Wampa
+                context.player1.clickCard(context.wampa);
+
+                // Damange from Wampa to be assigned
+                expect(context.atst.damage).toBe(4);
```
AMMayberry1, line 26: Can skip this check since it's not really related to the card ability being tested

PR #184, opened 2024-11-10
Add DiscardHandSystem and Black One implementation

Frugghi:
image


server/game/gameSystems/DiscardHandSystem.ts
```diff
@@ -0,0 +1,32 @@
...
+    public override readonly costDescription: string = 'discard hand';
+    public override readonly effectDescription: string = 'discard hand';
+
+    public override eventHandler(event): void {
+        const players = Helpers.asArray(event.player) as Player[];
+
+        players.forEach((player) =>
+            player
+                .getCardPile(Location.Hand)
+                .forEach((card) => player.moveCard(card, Location.Discard))
```
AMMayberry1, line 25: We just settled on a pattern for how we want to handle an effect that discards multiple cards, check out DiscardCardsFromHand.generateEventsForCards. We emit one individual discard event for each card and then an aggregate event indicating whether the entire discard effect resolved or not. This makes it possible for the discarding of individual cards to be reacted to much more easily.

server/game/gameSystems/DiscardHandSystem.ts
```diff
@@ -0,0 +1,32 @@
...
+        );
+    }
+
+    public override defaultTargets(context: TContext): Player[] {
```
AMMayberry1, line 29: I'd rather not override the default target behavior unless it's on a system where it is super clear that the natural default should be different, just so people don't get tripped up on needing to check what the default is for each system

server/game/cards/01_SOR/units/BlackOneScourgeOfStarkillerBase.ts
```diff
@@ -0,0 +1,29 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '8429598559',
+            internalName: 'black-one#scourge-of-starkiller-base'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'You may discard your hand',
```
AMMayberry1, line 14: nit: our convention is to omit the "you may" since that's implied by the presence of the pass button and just state the ability effect

server/game/gameSystems/DiscardHandSystem.ts
```diff
@@ -0,0 +1,32 @@
...
+import type Player from '../core/Player';
+import { IPlayerTargetSystemProperties, PlayerTargetSystem } from '../core/gameSystem/PlayerTargetSystem';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface IDiscardHandSystemProperties extends IPlayerTargetSystemProperties {}
+
+/**
+ * A {@link GameSystem} which discards a player's entire hand
+ */
+export class DiscardHandSystem<TContext extends AbilityContext = AbilityContext> extends PlayerTargetSystem<TContext, IDiscardHandSystemProperties> {
```
AMMayberry1, line 13: nit: just to be brutally clear let's maybe call it like `DiscardWholeHand` or `DiscardEntireHand`

test/server/cards/01_SOR/units/BlackOneScourgeOfStarkillerBase.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+            });
+
+            it('draw 3 cards if hand is discarded', function () {
+                const { context } = contextRef;
+
+                // Player 1 plays Black One and activates When Played, drawing cards
+                context.player1.clickCard(context.blackOne);
+
+                expect(context.player1.handSize).toBe(0);
+                expect(context.player1).toHaveEnabledPromptButton('You may discard your hand');
```
AMMayberry1, line 24: Can we use `toHavePassAbilityPrompt` which will confirm the name and that there's a pass button

test/server/cards/01_SOR/units/BlackOneScourgeOfStarkillerBase.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+                context.player2.clickCard(context.vanquish);
+
+                // Player 1 actives When Defeated, drawing cards
+                expect(context.player1).toHaveEnabledPromptButton('You may discard your hand');
+                context.player1.clickPrompt('You may discard your hand');
+
+                expect(context.player1.handSize).toBe(3);
+                expect(context.cartelSpacer).toBeInLocation('hand', context.player1);
+                expect(context.wampa).toBeInLocation('hand', context.player1);
+                expect(context.disarm).toBeInLocation('hand', context.player1);
```
AMMayberry1, line 43: Can we also confirm that the previous cards are in discard

test/server/cards/01_SOR/units/BlackOneScourgeOfStarkillerBase.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+                // Player 2 plays Rival's Fall and defeats Black One
+                expect(context.player2).toBeActivePlayer();
+                context.player2.clickCard(context.rivalsFall);
+
+                // Player 1 actives When Defeated, drawing cards
+                context.player1.clickPrompt('You may discard your hand');
+
+                expect(context.player1.handSize).toBe(0);
+                expect(context.p1Base.damage).toBe(15);
```
AMMayberry1, line 63: This last test feels redundant since we've already confirmed that draw damage happens, unless there's something I'm missing

Frugghi: Yeah, it's a bit redundant but the other one was testing the "when played" ability, this is the "when defeated" one. I would keep both to be safe.

server/game/gameSystems/DiscardHandSystem.ts
Frugghi: The problem is that there isn't an easy way to pass a player to the system. This, by default, target the opponent player because `PlayerTargetSystem` is using that:
```Typescript
immediateEffect: AbilityHelper.immediateEffects.discardEntireHand()
```

So you would need to do something like this:
```Typescript
immediateEffect: AbilityHelper.immediateEffects.discardEntireHand({ target: context.player })
```

But you can't because you don't have a context.

The alternatives would be:
1. To add a way to set `immediateEffect` with a function that takes a `context`.
2. To add a way to set up a player target resolver where you can return the player you want.

I guess (2) would be the best one?

AMMayberry1: Oh yeah so (1) is already implemented in the engine. Every GameSystem (immediateEffect) can accept either a properties object or a handler that constructs one from a context object. If the latter, then the engine will automatically call it to create the properties object at the appropriate time with the right context object.

This works for both card target and player target systems - see Smuggling Compartment for an example of the latter.

Frugghi: Of course it does, I already use that many times!

 For some reason my brain was trying to do:
```Typescript
(context) => AbilityHelper.immediateEffects.discardEntireHand({ target: context.player })
```

Instead of:
```Typescript
AbilityHelper.immediateEffects.discardEntireHand((context) => ({ target: context.player }))
```

server/game/cards/01_SOR/units/BlackOneScourgeOfStarkillerBase.ts
```diff
@@ -11,13 +11,13 @@ export default class BlackOneScourgeOfStarkillerBase extends NonLeaderUnitCard {
...
         this.addTriggeredAbility({
-            title: 'You may discard your hand',
+            title: 'Discard your hand',
             when: {
                 onCardPlayed: (event, context) => event.card === context.source,
                 onCardDefeated: (event, context) => event.card === context.source,
             },
             optional: true,
-            immediateEffect: AbilityHelper.immediateEffects.discardHand(),
+            immediateEffect: AbilityHelper.immediateEffects.discardEntireHand((context) => ({ target: context.player })),
```
AMMayberry1, line 20: I have to confess I'm not 100% sure how `context.player` is set in all situations so we typically use an explicit reference to the host card's controller via `context.source.controller`

server/game/gameSystems/DiscardEntireHandSystem.ts
```diff
@@ -0,0 +1,44 @@
+import { EventName, Location } from '../core/Constants';
+import { AbilityContext } from '../core/ability/AbilityContext';
+import * as Helpers from '../core/utils/Helpers.js';
+import type Player from '../core/Player';
+import { IPlayerTargetSystemProperties, PlayerTargetSystem } from '../core/gameSystem/PlayerTargetSystem';
+import { DiscardSpecificCardSystem } from './DiscardSpecificCardSystem';
+
+export interface IDiscardEntireHandSystemProperties extends IPlayerTargetSystemProperties {
+    target: Player | Player[];
```
AMMayberry1, line 9: Oh interesting, I didn't realize this wasn't being explicitly overridden in IPlayerTargetSystemProperties. Do you mind changing it there to be:

```typescript
export interface IPlayerTargetSystemProperties extends IGameSystemProperties {
    target?: Player | Player[];
}
```

Then this can just be an empty class (we always define one even if empty just for consistency)

server/game/gameSystems/DiscardEntireHandSystem.ts
```diff
@@ -0,0 +1,44 @@
...
+    public override eventHandler(): void {
+        // Do nothing
+    }
+
+    public override queueGenerateEventGameSteps(events: any[], context: TContext, additionalProperties: Record<string, any> = {}): void {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+
+        for (const player of Helpers.asArray(properties.target) as Player[]) {
+            // Ensure that the player can be affected by this system
+            if (!this.canAffect(player, context, additionalProperties)) {
```
AMMayberry1, line 30: The way the pipeline is written we can assume that `canAffect` has been used to filter targets by the time we get to `queueGenerateEventGameSteps` or `eventHandler`, so we can assume that every target is already valid and don't need to call `canAffect` here

Frugghi: The implementation of `queueGenerateEventGameSteps` in `GameSystem` checks for `canAffect` so that's why I added it. Happy to remove it if you think it's not needed.

server/game/gameSystems/DiscardEntireHandSystem.ts
Frugghi: I'll add it to `IPlayerTargetSystemProperties` too but I would keep it here too without the optionality to force you to pass a target. I think the default behaviour of discarding the opponent hand is a bit counterintuitive and it'd be better to always specify the target.

server/game/gameSystems/DiscardEntireHandSystem.ts
AMMayberry1: Ohh good catch, hmm. Let's leave it then, there's probably a reason.

server/game/gameSystems/DiscardEntireHandSystem.ts
AMMayberry1: I agree but unfortunately the pattern we use with target resolvers requires that the target property is always optional. Since target resolvers will populate the target field at resolution time, we don't provide it when we initialize the system. So right now this system couldn't be used inside a PlayerTargetResolver

Frugghi: Wouldn't you be able to do:
```Typescript
AbilityHelper.immediateEffects.discardEntireHand((context) => ({ target: context.target }))
```
when using `PlayerTargetResolver`?

server/game/gameSystems/DiscardEntireHandSystem.ts
Frugghi: Done!

AMMayberry1 (approved):

PR #185, opened 2024-11-11
DiscardFromDeck (mill) System, Kanan Jarrus, Chopper

adbronico:
image
image


server/game/core/Player.js
```diff
@@ -424,6 +432,21 @@ class Player extends GameObject {
         return null;
     }
 
+
+    /**
+     * Returns ths top cards of the player's deck
+     * @returns {import('./card/CardTypes').PlayableCard[] |null} the Card,© or null if the deck is empty
```
AMMayberry1, line 438: If an array return value is expected anyway then it probably makes things more streamlined in some cases on the calling side if we return an empty array instead of null? That way you can do like a for loop over the results without checking

server/game/gameSystems/DiscardFromDeckSystem.ts
```diff
@@ -0,0 +1,82 @@
...
+    // }
+
+    public override getEffectMessage(context: TContext): [string, any[]] {
+        const properties = this.generatePropertiesFromContext(context);
+        return ['discard ' + (properties.amount + (properties.amount > 1 ? ' cards' : ' card') + ' from deck'), []];
+    }
+
+    public override canAffect(player: Player, context: TContext, additionalProperties = {}): boolean {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        return properties.amount !== 0 && super.canAffect(player, context);
```
AMMayberry1, line 38: Oh yeah there's a lot of nuance here about whether this should report back true or false if there aren't enough cards in the deck, because in some situations it's okay if an effect resolves and in some cases it isn't. Feel free to ask in chat if you want details but the tl;dr is you can just reuse the logic from DiscardCardsFromHand in terms of mustChangeGameState and isCost.

server/game/gameSystems/DiscardFromDeckSystem.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public override defaultTargets(context: TContext): Player[] {
```
AMMayberry1, line 41: I would prefer to avoid overriding default targets so that card implementers can assume they know what the default is without having to look it up

server/game/cards/01_SOR/units/ChopperMetalMenace.ts
```diff
@@ -0,0 +1,45 @@
...
+        this.addConstantAbility({
+            title: 'While you control another Spectre unit, Chopper gains Raid 1',
+            condition: (context) => context.source.controller.getOtherUnitsInPlayWithTrait(context.source, Trait.Spectre).length > 0,
+            ongoingEffect: AbilityHelper.ongoingEffects.gainKeyword({ keyword: KeywordName.Raid, amount: 1 })
+        });
+    }
+
+    private isDiscardedCardEvent(events: any[]): boolean {
+        // If nothing was discarded, return false
+        if (events.length === 0) {
```
AMMayberry1, line 38: Could do this all in one line with a ternary, and even just do that directly on line 22 instead of having the method

server/game/cards/01_SOR/units/KananJarrusRevealedJedi.ts
```diff
@@ -0,0 +1,40 @@
...
+                    target: context.source.controller.base,
+                    amount: this.getAspectCountFromEvents(context.events)
+                }))
+            ])
+        });
+    }
+
+    private getAspectCountFromEvents(events: any[]): number {
+        const aspects = new Set<Aspect>();
+        const discardedCards = events.filter((event) => event.name === 'onCardDiscarded').map((event) => event.card);
```
AMMayberry1, line 32: Can we use `EventName.OnCardDiscarded` instead of the string

server/game/cards/02_SHD/events/LookTheOtherWay.ts
```diff
@@ -25,7 +25,7 @@ export default class LookTheOtherWay extends EventCard {
                         [`Exhaust ${context.targets.targetUnit.title}`]: AbilityHelper.immediateEffects.exhaust({
                             target: context.targets.targetUnit,
                         }),
-                        ['Pay 2 resources']: AbilityHelper.immediateEffects.payResourceCost({
+                        ['Pay 2 resources']: AbilityHelper.immediateEffects.exhaustResources({
```
AMMayberry1, line 28: In this case I think it should be `payResourceCost`, unless that caused some issue

server/game/costs/CostLibrary.ts
```diff
@@ -255,8 +255,8 @@ export function payPlayCardResourceCost<TContext extends AbilityContext = Abilit
 export function abilityResourceCost<TContext extends AbilityContext = AbilityContext>(amount: number | ((context: TContext) => number)): ICost<TContext> {
     return new GameSystemCost<TContext>(
         typeof amount === 'function'
-            ? GameSystems.payResourceCost<TContext>((context) => ({ target: context.player, amount: amount(context) }))
-            : GameSystems.payResourceCost<TContext>((context) => ({ target: context.player, amount }))
+            ? GameSystems.payResourceCost<TContext>((context) => ({ target: context.player, amount: amount(context), isCost: true }))
```
AMMayberry1, line 258: I think the isCost shouldn't be needed for these the way the defaults are set up now?

server/game/core/Player.js
```diff
@@ -424,6 +443,21 @@ class Player extends GameObject {
         return null;
     }
 
+
+    /**
+     * Returns ths top cards of the player's deck
+     * @returns {import('./card/CardTypes').PlayableCard[]} the Card,© or null if the deck is empty
+     */
+    getTopCardsOfDeck(numCard = 1) {
```
AMMayberry1, line 451: Do you think the default adds value here? I feel like this form would mostly be used when the value is something other than 1

server/game/gameSystems/DiscardFromDeckSystem.ts
```diff
@@ -0,0 +1,89 @@
...
+export interface IDiscardFromDeckProperties extends IPlayerTargetSystemProperties {
+    amount?: number;
+}
+
+export class DiscardFromDeckSystem<TContext extends AbilityContext = AbilityContext> extends PlayerTargetSystem<TContext, IDiscardFromDeckProperties> {
+    public override readonly name = 'discardFromDeck';
+    public override readonly eventName = EventName.OnDiscardFromDeck;
+
+    protected override defaultProperties: IDiscardFromDeckProperties = {
+        amount: 1
```
AMMayberry1, line 18: I always go back and forth and whether to include default values for stuff like this. I think the question boils down to, does `AbilityHelper.immediateEffects.discardFromDeck()` intuitively read as discarding 1 card? I could genuinely go either way, just wanted to raise it

server/game/gameSystems/DiscardFromDeckSystem.ts
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public override eventHandler(_event): void { }
+
+    // public eventHandler(event): void {
```
AMMayberry1, line 24: Old comment

server/game/gameSystems/DiscardFromDeckSystem.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    public override getEffectMessage(context: TContext): [string, any[]] {
+        const properties = this.generatePropertiesFromContext(context);
+        return ['discard ' + (properties.amount + (properties.amount > 1 ? ' cards' : ' card') + ' from deck'), []];
+    }
+
+    public override canAffect(player: Player, context: TContext, additionalProperties = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+
+        const availableDeck = player.drawDeck || []; // Without this | [], sometimes this is undefined, causing issues below
```
AMMayberry1, line 39: ....huh. Did you check to see what the value of `player` is in those situations? I think it legit can be an array here sometimes, which isn't true for CardTargetSystem extensions. We definitely shouldn't need this workaround, something is really wrong if so.

server/game/gameSystems/DiscardFromDeckSystem.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        if ((properties.isCost || GameStateChangeRequired.MustFullyResolve) && availableDeck.length < properties.amount) {
+            return false;
+        }
+
+        if (!super.canAffect(player, context, additionalProperties)) {
+            return false;
+        }
+
+        return properties.amount !== 0 && super.canAffect(player, context);
```
AMMayberry1, line 53: Looks we can shortcut pretty early if amount === 0, maybe let's put that as its own if check pretty high in the function? That way we could remove the check from e.g. line 41 so we're not checking it twice

server/game/gameSystems/DiscardFromDeckSystem.ts
```diff
@@ -0,0 +1,89 @@
...
+        super.addPropertiesToEvent(event, player, context, additionalProperties);
+        event.amount = amount;
+    }
+
+    public override queueGenerateEventGameSteps(events: any[], context: TContext, additionalProperties: Record<string, any> = {}): void {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        for (const player of properties.target as Player[]) {
+            const availableDeck = player.drawDeck;
+
+            Contract.assertNonNegative(properties.amount);
```
AMMayberry1, line 67: Actually sorry now that I think about it, could we move this to `canAffect` since that gets hit earlier and more often

server/game/core/Player.js
```diff
@@ -424,6 +443,21 @@ class Player extends GameObject {
...
+        const deckLength = this.drawDeck.length;
+        const cardsToGet = Math.min(numCard, deckLength);
```
AMMayberry1, line 453: Can we also do a contract check that numCard is > 0 (can use `assertPositiveNonZero`). Those checks have actually saved me a couple of times

server/game/gameSystems/DiscardFromDeckSystem.ts
```diff
@@ -0,0 +1,89 @@
...
+import { IPlayerTargetSystemProperties, PlayerTargetSystem } from '../core/gameSystem/PlayerTargetSystem';
+import Player from '../core/Player';
+import { DiscardSpecificCardSystem } from './DiscardSpecificCardSystem';
+import * as Contract from '../core/utils/Contract';
+
+export interface IDiscardFromDeckProperties extends IPlayerTargetSystemProperties {
+    amount?: number;
+}
+
+export class DiscardFromDeckSystem<TContext extends AbilityContext = AbilityContext> extends PlayerTargetSystem<TContext, IDiscardFromDeckProperties> {
```
AMMayberry1, line 13: Do you think `DiscardFromTopOfDeck` would be worth it for the added clarity?

test/server/cards/01_SOR/units/ChopperMetalMenace.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+                const { context } = contextRef;
+
+                expect(context.player2.deck.length).toBe(5);
+                expect(context.player2.countSpendableResources()).toBe(5);
+
+                context.player1.clickCard(context.chopper);
+                expect(context.player2.base.damage).toBe(1);
+
+                // Check mill and heal
+                expect(context.player2.deck.length).toBe(4);
```
AMMayberry1, line 28: Can we confirm that the correct card was milled since this is the first test of this system

test/server/cards/01_SOR/units/ChopperMetalMenace.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+                        leader: 'hera-syndulla#spectre-two',
+                    },
+                    player2: {
+                        deck: ['the-chaos-of-war', 'battlefield-marine', 'pyke-sentinel', 'underworld-thug', 'volunteer-soldier'],
+                        resources: 5
+                    }
+                });
+            });
+
+            it('should discard a card on attack and exhaust an enemy resource if the card is an event; should also have Raid 1 from friendly Spectre', function () {
```
AMMayberry1, line 48: Probably we should have an empty deck test as well, including confirming that the base draw damage doesn't happen (not really possible I know but it protects us in case of future refactoring)

test/server/cards/01_SOR/units/KananJarrusRevealedJedi.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+                context.player1.clickCard(context.kananJarrus);
+                expect(context.player1).toHavePassAbilityPrompt('Discard a card from the defending player\'s deck for each Spectre you control. Heal 1 damage for each aspect among the discarded cards.');
+                context.player1.clickPrompt('Discard a card from the defending player\'s deck for each Spectre you control. Heal 1 damage for each aspect among the discarded cards.');
+                expect(context.player2.base.damage).toBe(4);
+
+                // Check mill and heal
+                expect(context.player2.deck.length).toBe(2);
+
+                // Should heal 4 for Command, Heroism, Vigilance, Villainy
```
AMMayberry1, line 64: Could we adjust this so that there are two copies of one aspect to confirm it doesn't double-count it

test/server/cards/01_SOR/units/KananJarrusRevealedJedi.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                expect(context.player1.base.damage).toBe(1);
+            });
```
AMMayberry1, line 66: Would be good to have one more test where there aren't enough cards in deck to fully satisfy the condition, to make sure (a) draw damage doesn't happen and (b) the effect still works right

server/game/costs/CostLibrary.ts
adbronico: @AMMayberry1 Wouldn't they be needed now that we've removed the default true?

server/game/gameSystems/DiscardFromDeckSystem.ts
adbronico: @AMMayberry1 AFAIK there isn't currently any way to discard from anywhere else in the deck, so I think it's fine for now. If we ever have other discard from deck options, we can add some sort of a DiscardMode to this. Does that sound good?

AMMayberry1: Yup makes sense

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -188,8 +192,12 @@ export function playCardFromHand<TContext extends AbilityContext = AbilityContex
     // playType automatically defaults to PlayFromHand
     return new PlayCardSystem(propertyFactory);
 }
-export function payResourceCost<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IPayResourceCostProperties, TContext>): GameSystem<TContext> {
-    return new PayResourceCostSystem<TContext>(propertyFactory);
+export function exhaustResources<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IExhaustResourcesProperties, TContext>): GameSystem<TContext> {
+    return new ExhaustResourcesSystem<TContext>(Object.assign({ ...propertyFactory, isCost: false }));
```
AMMayberry1, line 196: Now that we removed the default we can remove the `isCost: false` here and instead we need to do `isCost: true` in payResourceCost so that it has the right behavior.

Also, just noticed that this assignment is being done wrong. Since the `propertyFactory` parameter can be a function sometimes (i.e. when we use the `(context) => ({ ... })` form) this won't always work. We need to use `GameSystem.appendToPropertiesOrPropertyFactory` as you can see in e.g. `moveToDeck`

server/game/costs/CostLibrary.ts
AMMayberry1: Added a comment in GameSystemLibrary, ideally I think this method's default should be set to isCost: true and this shouldn't be needed. Otherwise there's no real reason to have `payResourceCost` as its own method imo

server/game/gameSystems/DiscardFromDeckSystem.ts
```diff
@@ -36,21 +25,27 @@ export class DiscardFromDeckSystem<TContext extends AbilityContext = AbilityCont
     public override canAffect(player: Player, context: TContext, additionalProperties = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
         const properties = this.generatePropertiesFromContext(context, additionalProperties);
 
-        const availableDeck = player.drawDeck || []; // Without this | [], sometimes this is undefined, causing issues below
+        const availableDeck = Array.isArray(player) ? player[0].drawDeck : player.drawDeck;
```
AMMayberry1, line 28: Ah yeah this needs to be a for loop then, if multiple player targets are provided it needs to act on all of them. Same as it's done for DiscardCardsFromHand

AMMayberry1 (approved):

PR #187, opened 2024-11-11
Adding more cards

Bluffkin:
ATAT Suppressor
Ruthless Raider


test/server/cards/01_SOR/units/RuthlessRaider.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+            it('should deal 2 damage to an enemy unit and enemy base.', function () {
+                const { context } = contextRef;
+
+                // Play Ruthless Raider from hand
+                context.player1.clickCard(context.ruthlessRaider);
+
+                // Select Enemy Unit and Base. Not able to select friendly units
+                expect(context.player1).toBeAbleToSelectExactly([context.greenSquadronAwing, context.wampa, context.player2.base]);
+                context.player1.clickCard(context.greenSquadronAwing);
+                context.player1.clickCard(context.p2Base);
```
AMMayberry1, line 28: This click shouldn't be here, the base will be automatically selected for `theirBase` since it's the only legal target that fits the filters

server/game/cards/01_SOR/units/RuthlessRaider.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Deal 2 damage to an enemy base and 2 damage to an enemy unit',
+            when: {
+                onCardPlayed: (event, context) => event.card === context.source,
+                onCardDefeated: (event, context) => event.card === context.source
+            },
+            targetResolvers: {
```
AMMayberry1, line 20: This is kind of getting into the weeds a little and into a place where the card implementation tools need some improvement, but let's remove the `targetResolvers` property and replace it with an `immediateEffect` property that does this:

```typescript
      AbilityHelper.immediateEffects.selectCard({
          controller: RelativePlayer.Opponent,
          innerSystem: AbilityHelper.immediateEffects.damage({ amount: 2 }),
      }),
      AbilityHelper.immediateEffects.damage((context) => ({ amount: 2, target: context.source.controller.opponent.base }))
  ])
```

The reason is that `theirBase` shouldn't really be a target resolver since target resolvers are for when players will have to choose a target, but that isn't true for `theirBase` since it will always be the opponent's base specifically.

So we need to mix a target resolver (`theirUnit`) with a system that doesn't need a target resolver and that's kind of ugly to do right now, as you see above. Basically the `selectCard()` call acts like a target resolver in place `theirUnit` and the `theirBase` handles the base damage by selecting that card explicitly.

Feel free to follow up in discord if you have questions

server/game/cards/01_SOR/units/StarWingScout.ts
```diff
@@ -0,0 +1,25 @@
+import AbilityHelper from '../../../AbilityHelper';
```
AMMayberry1, line 1: Just a note - in the future please don't add new cards to an already open PR. New code probably means more rounds of review so the PR will kind of drag on for a while, and also it can make it harder to know what's been reviewed and what's new.

Folks doing lots of card implementations typically batch a handful of cards into a PR and then move on to do a new batch while they wait for review, that's a good pattern to follow.

test/server/cards/01_SOR/units/StarWingScout.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+                    // Player 2 plays Rival's Fall, targeting Star Wing Scout
+                    context.player2.clickCard(context.rivalsFall);
+                    expect(context.starWingScout.location).toBe('discard');
+
+                    // Player 1 draws 2 cards
+                    expect(context.player1.handSize).toBe(2);
+                });
+            });
+
+            describe('when the player has initiative', function () {
```
AMMayberry1, line 34: nit: should say "when the player does not have initiative"

server/game/cards/01_SOR/units/StarWingScout.ts
Bluffkin: Understood.

AMMayberry1 (approved):

PR #188, opened 2024-11-12
Implementing Rich Reward bounty

aruizc:
image


test/server/cards/02_SHD/upgrades/RichReward.spec.ts
```diff
@@ -0,0 +1,95 @@
+describe('Rich Reward', function() {
+    integration(function(contextRef) {
+        describe('Rich Reward\'s Bounty ability', function() {
+            it('should give Experience token up to two units', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: [
+                            { card: 'battlefield-marine', upgrades: ['rich-reward'] },
+                            { card: 'atst' },
```
AMMayberry1, line 10: You can mix object and string forms in the same list. So this would be simpler as

`groundArena: [{ card: 'battlefield-marine', upgrades: ['rich-reward'] }, 'atst', 'yoda#old-master']`

test/server/cards/02_SHD/upgrades/RichReward.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.battlefieldMarine);
+                context.player1.clickCard(context.wampa);
+
+                expect(context.player2).toBeAbleToSelectExactly([context.wampa, context.atst, context.yoda, context.concordDawnInterceptors]);
+                expect(context.player2).toHavePassAbilityButton();
+                expect(context.player2).toHavePrompt('Select 2 cards');
```
AMMayberry1, line 27: Can you break this block up with whitespace a little for readability, same note for other tests

test/server/cards/02_SHD/upgrades/RichReward.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+                context.player2.clickCard(context.wampa);
+                context.player2.clickCard(context.concordDawnInterceptors);
```
AMMayberry1, line 29: We should check that no more than 2 cards can be selected. The way to do that is to click a third card using `clickCardNonChecking()` instead of `clickCard()`, and then confirm that the third clicked unit does not get an experience

test/server/cards/02_SHD/upgrades/RichReward.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+                expect(context.player2).toBeAbleToSelectExactly([context.wampa, context.atst, context.yoda, context.concordDawnInterceptors]);
+                expect(context.player2).toHavePassAbilityButton();
+                expect(context.player2).toHavePrompt('Select 2 cards');
+                context.player2.clickCard(context.yoda);
+                context.player2.clickPrompt('Done');
+                expect(context.yoda).toHaveExactUpgradeNames(['experience']);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should be able to not collect the bounty', function () {
```
AMMayberry1, line 66: We can remove this test, we have other tests that confirm that Bounties are optional so we don't need to do it on the individual cards

test/server/cards/02_SHD/upgrades/RichReward.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+                expect(context.player2).toHavePrompt('Select 2 cards');
+                context.player2.clickCard(context.wampa);
+                context.player2.clickCard(context.concordDawnInterceptors);
+                context.player2.clickPrompt('Done');
+                expect(context.wampa).toHaveExactUpgradeNames(['experience']);
+                expect(context.concordDawnInterceptors).toHaveExactUpgradeNames(['experience']);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should give Experience token to one unit', function () {
```
AMMayberry1, line 36: We haven't used `TargetMode.UpTo` a lot, would you mind adding an extra test confirming that if there are no available targets it just passes (can use `expect(context.player2).toBeActivePlayer();` to confirm that player1 got no prompt)

test/server/cards/02_SHD/upgrades/RichReward.spec.ts
aruizc: done

test/server/cards/02_SHD/upgrades/RichReward.spec.ts
aruizc: done

test/server/cards/02_SHD/upgrades/RichReward.spec.ts
aruizc: I've added a test for that case, please let me know if I'm missing something

AMMayberry1:
Quick note - please mark conversations resolved once they've been addressed, makes it way easier to tell when it's ready for review

test/server/cards/02_SHD/upgrades/RichReward.spec.ts
AMMayberry1: Oh we don't need a whole extra case for that, especially since we're trying to consolidate to as few individual cases as possible. Apologies for my being unclear, but do you mind just adding the click here and removing the other test? Unless there's something else that test is covering that this test isn't

aruizc: my bad, I should simplify the unit test. Done

AMMayberry1 (approved):

PR #189, opened 2024-11-12
Implement Reputable Hunter (SHD)

arjandepooter:


test/server/cards/02_SHD/units/ReputableHunter.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+            });
+
+            it('should cost 2 if opponent has a unit with a dedicated bounty in play', () => {
+                const { context } = contextRef;
+
+                context.player1.passAction();
+                context.player2.clickCard(context.hylobonEnforcer);
+                context.player1.clickCard(context.reputableHunter);
+
+                expect(context.player1.countExhaustedResources()).toBe(2);
```
VJubert, line 49: I think you can omit beforeEach to have more lighter and precise setup. See Cartel Turncoat to see example

VJubert (approved):

AMMayberry1:
In the future, would you mind including the card image in the PR description? Makes it easier to double-check the text. Thanks!

test/server/cards/02_SHD/units/ReputableHunter.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+                });
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.reputableHunter);
+
+                expect(context.player1.countExhaustedResources()).toBe(3);
+            });
+
+
+            it('should cost 2 after we play a bounty on a unit', () => {
```
AMMayberry1, line 20: Just a note - for tests of similar abilities in the future we won't need to check that dynamically gained bounties work correctly, we can assume it works. That system is relatively new so the first few units we added had an explicit check for it just to confirm that it works, and it's also fine to include in this one as well. (Not great separation of test concerns but we do what we can)

test/server/cards/02_SHD/units/ReputableHunter.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+                context.player1.clickCard(context.topTarget);
+                const exhaustedResourcesBeforePlay = context.player1.countExhaustedResources();
+                context.player2.passAction();
+                context.player1.clickCard(context.reputableHunter);
+
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['top-target']);
+                expect(context.player1.countExhaustedResources() - exhaustedResourcesBeforePlay).toBe(2);
+            });
+
+            it('should cost 2 if opponent has a unit with a dedicated bounty in play', () => {
```
AMMayberry1, line 42: Can we have a third test to confirm that it costs 3 if there's not a bounty present. If we set it up in the same test so that a friendly unit does have a bounty, we can kill two birds with one stone and confirm that it only checks opponent units.

arjandepooter: Added Hylobon on the board in the test that checks if it costs 3.

AMMayberry1: Ah sorry, missed that first test. Thanks for the update

AMMayberry1 (approved):

PR #190, opened 2024-11-12
feat: implement outlaw corona unit and its unit tests

jgordijo:
image


server/game/cards/02_SHD/units/OutlawCorona.ts
```diff
@@ -0,0 +1,22 @@
...
+        return {
+            id: '3503780024',
+            internalName: 'outlaw-corona'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addBountyAbility({
+            title: 'Put the top card of your deck into play as a resource.',
+            targetResolver: {
```
VJubert, line 15: I think you can omit targetResolver and write only immediateEffect

jgordijo: Yeah, great catch! Fixed it.

server/game/cards/02_SHD/units/OutlawCorona.ts
```diff
@@ -0,0 +1,20 @@
...
+        return {
+            id: '3503780024',
+            internalName: 'outlaw-corona'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addBountyAbility({
+            title: 'Put the top card of your deck into play as a resource.',
+            immediateEffect: AbilityHelper.immediateEffects.resourceCard((context) => ({ target: context.player.getTopCardOfDeck() }))
```
AMMayberry1, line 15: Let's use `context.source.controller` instead of `context.player`, I'm not 100% sure they're interchangeable in all cases and we know the former works correctly

julio-cesar-andela: Hey @AMMayberry1, this is a bounty so it doesn't goes to the card controller, but I'll keep that in mind for future implementations!

AMMayberry1 (approved):

PR #191, opened 2024-11-12
Bunch of cards 8

VJubert:


server/game/cards/01_SOR/units/GreedoSlowOnTheDraw.ts
```diff
@@ -0,0 +1,36 @@
...
+            optional: true,
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.discardFromDeck((context) => ({
+                    amount: 1,
+                    target: context.source.controller
+                })),
+                AbilityHelper.immediateEffects.conditional((context) => ({
+                    // There will be one event for the discard system overall plus one per card, so we need to ensure at least two exist
+                    condition: context.events.length < 2 ? false : !context.events[0].card.isUnit(),
+                    onTrue: AbilityHelper.immediateEffects.selectCard({
```
adbronico, line 25: I'm not sure if there's a better way to do this than the selectCard. Can't we just have a target resolver on the damage itself? @AMMayberry1 ?

server/game/cards/02_SHD/units/ChainCodeCollector.ts
```diff
@@ -0,0 +1,29 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '7171636330',
+            internalName: 'chain-code-collector'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'Each friendly unit named Zuckuss gets +1/+1 and gains Ambush',
```
adbronico, line 15: wrong description here

server/game/cards/02_SHD/units/ChainCodeCollector.ts
```diff
@@ -0,0 +1,29 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => context.event.attack.target.hasSomeKeyword(KeywordName.Bounty),
```
adbronico, line 17: Can we add a check that the target is a unit before the target keyword check?

server/game/cards/02_SHD/upgrades/GuildTarget.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.addGainKeywordTargetingAttached({
+            keyword: KeywordName.Bounty,
+            ability: {
+                title: 'Deal 2 damage to a base. If the Bounty unit is unique, deal 3 damage instead',
+                targetResolver: {
+                    cardTypeFilter: CardType.Base,
+                    immediateEffect: AbilityHelper.immediateEffects.damage((context) => ({ amount: context.source.unique ? 3 : 2 })),
```
adbronico, line 20: @AMMayberry1 Do we want to have a cleaner way to do this?

adbronico (changes requested):

server/game/cards/02_SHD/units/ChainCodeCollector.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'If the defender has a Bounty, it gets –4/–0 for this attack',
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => {
+                    const target = context.event.attack.target;
+                    return target.isUnit() && target.hasSomeKeyword(KeywordName.Bounty);
+                },
+                onTrue: AbilityHelper.immediateEffects.cardLastingEffect((context) => ({
```
AMMayberry1, line 21: I think it would be good to have a `cardAttackLastingEffectSystem` similar to `cardPhaseLastingEffectSystem` but it will require some mild refactoring (very easy). Would you mind just creating a copy of CardPhaseLastingEffectSystem.ts named CardAttackLastingEffectSystem.ts (and then add a new function for it in GameSystemLibrary named `forThisAttackCardEffect`).

Other than renaming stuff, the only change needed would be to add this function:

```typescript
public override canAffect(card: Card, context: TContext) {
    if (!card.isUnit()) {
        return false;
    }

    Contract.assertNotNullLike(card.activeAttack, `Attempting to apply an attack lasting effect to ${card.internalName} but it is not actively attacking`);

    return super.canAffect(card, context);
}
```

server/game/cards/02_SHD/units/DoctorEvazanWantedOnTwelveSystems.ts
```diff
@@ -0,0 +1,22 @@
...
+        return {
+            id: '6135081953',
+            internalName: 'doctor-evazan#wanted-on-twelve-systems'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addBountyAbility({
+            title: 'Ready up to 12 resources',
+            // force resolution of this just to skip prompting the player. we can safely assume they will always want to ready  all resources
```
AMMayberry1, line 15: Agreed!

server/game/cards/02_SHD/units/TheClientDictatedByDiscretion.ts
```diff
@@ -0,0 +1,36 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '1090660242',
+            internalName: 'the-client#dictated-by-discretion'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addActionAbility({
+            title: 'Choose a unit. For this phase, it gains: "Bounty — Heal 5 damage from a base."',
```
AMMayberry1, line 15: nit: can remove the "choose a unit" part since the player will already be looking at a target selection prompt. Might need to reword it grammatically to make sense though, maybe `For this phase, target unit gains ...`

test/server/cards/01_SOR/events/NoGoodToMeDead.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+                context.player2.clickCard(context.keepFighting);
+                context.player2.clickCardNonChecking(context.battlefieldMarine);
+                expect(context.battlefieldMarine.exhausted).toBeTrue();
+
+                // battlefield marine should not be ready on regroup phase
+                context.moveToNextActionPhase();
+                expect(context.battlefieldMarine.exhausted).toBeTrue();
+            });
+
+            it('should exhaust attached unit and restrict it to be ready (until the end of round)', function () {
```
AMMayberry1, line 42: Could we consolidate down to one test with two copies of Keep Fighting? See UniqueRule.spec.ts for clean patterns on how to handle duplicates

test/server/cards/01_SOR/units/GreedoSlowOnTheDraw.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+                // top card is an upgrade, deal 2 damage to a ground unit
+                expect(context.protector.location).toBe('discard');
+                expect(context.player2).toBeAbleToSelectExactly([context.wampa, context.battlefieldMarine, context.atst]);
+
+                context.player2.clickCard(context.battlefieldMarine);
+                expect(context.battlefieldMarine.damage).toBe(2);
+            });
+
+            it('should not deal 2 damage to a ground unit if the discarded card is a unit', function () {
```
AMMayberry1, line 37: If we do end up making the change I suggested for the Greedo implementation then we should add a third case with an empty deck to confirm the resolution check works

test/server/cards/02_SHD/units/ChainCodeCollector.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+                // kill hylobon enforcer
+                context.player1.clickCard(context.chainCodeCollector);
+                context.player1.clickCard(context.hylobonEnforcer);
+
+                // pass bounty from hylobon enforcer
+                context.player1.clickPrompt('Pass');
+
+                // chain code collector did not take damage (hylobon gets -4/-0)
+                expect(context.chainCodeCollector.damage).toBe(0);
```
AMMayberry1, line 24: Can we attack unit with at least 5 power so we can confirm the number

test/server/cards/02_SHD/units/ChainCodeCollector.spec.ts
```diff
@@ -0,0 +1,39 @@
+describe('Chain Code Collector', function () {
+    integration(function (contextRef) {
+        describe('Chain Code Collector\'s on attack ability', function () {
+            it('should give -4/-0 on unit who have bounty', function () {
+                const { context } = contextRef;
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['chain-code-collector'],
```
AMMayberry1, line 9: I don't think there's any major concern but just for an integration test it might be good to start the unit in hand so we can confirm the ability works right with the Ambush attack

test/server/cards/02_SHD/units/DoctorEvazanWantedOnTwelveSystems.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+                        resources: 13
+                    },
+                    player2: {
+                        groundArena: ['doctor-evazan#wanted-on-twelve-systems']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                // play wampa (4 resources exhausted)
```
AMMayberry1, line 20: There's also the option of `context.player1.exhaustResources(13);`

server/game/cards/01_SOR/units/GreedoSlowOnTheDraw.ts
AMMayberry1: I just realized the change hadn't been merged in yet, just did it

AMMayberry1 (approved):

PR #192, opened 2024-11-12
feat: implement no bargain event

jgordijo:
image


adbronico (dismissed):
This looks good to me. Wait on Veld to do a final review, though.

test/server/cards/02_SHD/events/NoBargain.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+                    player1: {
+                        hand: ['no-bargain', 'wampa'],
+                    },
+                    player2: {
+                        hand: ['pillage', 'crafty-smuggler'],
+                    }
+                });
+            });
+
+            it('should draw a card and make opponent discard one', function () {
```
AMMayberry1, line 16: If you could do me a favor - the discard system is new, we'd like to add a few extra tests on the first few cards that use it. Could you add one more case where the opponent has no cards in hand and confirm that the draw still happens?

jgordijo: @AMMayberry1 Done!

AMMayberry1 (approved):

PR #193, opened 2024-11-12
Add Forced Surrender implementation

Frugghi:
image

### DIFF
--- a/server/game/gameSystems/DrawSystem.ts
+++ b/server/game/gameSystems/DrawSystem.ts
@@ -52,10 +52,17 @@ export class DrawSystem<TContext extends AbilityContext = AbilityContext> extend
             const contingentEvents = [];
             if (event.amount > event.player.drawDeck.length) {
                 const damageAmount = 3 * (event.amount - event.player.drawDeck.length);
+
+                // Here we generate a the damage event with a new context that contains just the player,
+                // this way the damage is attributed to the player and not the card that triggered the draw (or its controller).
+                // As per rules, the player that is drawing is also the player that is causing the damage and
+                // this is important for cards like Forced Surrender.
+                // The downside is that we lose any connection with the original card that triggered the draw,
+                // which shouldn't matter for any of the existing cards.
                 contingentEvents.push(new DamageSystem({
                     target: event.player.base,
                     amount: damageAmount
-                }).generateEvent(context));
+                }).generateEvent(context.game.getFrameworkContext(event.player)));
             }
             return contingentEvents;
         });
--- a/server/game/gameSystems/DrawSystem.ts
+++ b/server/game/gameSystems/DrawSystem.ts
@@ -56,7 +56,7 @@ export class DrawSystem<TContext extends AbilityContext = AbilityContext> extend
                 // Here we generate a the damage event with a new context that contains just the player,
                 // this way the damage is attributed to the player and not the card that triggered the draw (or its controller).
                 // As per rules, the player that is drawing is also the player that is causing the damage and
-                // this is important for cards like Forced Surrender.
+                // this is important for cards like Forced Surrender. (FFG ruling confirms this)
                 // The downside is that we lose any connection with the original card that triggered the draw,
                 // which shouldn't matter for any of the existing cards.
                 contingentEvents.push(new DamageSystem({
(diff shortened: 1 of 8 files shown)


adbronico (approved):
This looks good to me but wait for @AMMayberry1 to review & merge

server/game/stateWatchers/DamageDealtThisPhaseWatcher.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export class DamageDealtThisPhaseWatcher extends StateWatcher<IDamageDealtThisPhase> {
+    public constructor(
+        registrar: StateWatcherRegistrar,
+        card: Card
+    ) {
+        super(StateWatcherName.DamageDealtThisPhase, registrar, card);
+    }
+
+    public getOpponentsWhoseBasesWereDamagedByPlayer(player: Player) {
```
AMMayberry1, line 24: Could we make this a little more generic since there is only one card that needs this specific check? In keeping with the other watchers we could add something like `getDamageDealtByPlayer` which accepts a filter and then provide the additional logic in the call from the Forced Surrender impl.

test/server/cards/01_SOR/events/ForcedSurrender.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+
+                context.player1.clickCard(context.forcedSurrender);
+                expect(context.player2).toBeActivePlayer();
+                expect(context.player1.handSize).toBe(4);
+                expect(context.player2.handSize).toBe(2);
+
+                reset();
+
+                // Scenario 3: Damage dealt with overwhelm
+                context.player2.clickCard(context.mazKanata);
```
AMMayberry1, line 68: What's the reason for doing the Maz Kanata attack every time?

Frugghi: Just to ensure that we have damaged dealt to both bases.

AMMayberry1: Still a little unclear - what is that specifically testing? Feels like we could do it one time in the case where we have dealt no damage to opponent base and should be good

test/server/cards/01_SOR/events/ForcedSurrender.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+
+                context.player2.passAction();
+
+                context.player1.moveCard(context.daringRaid, 'hand');
+                context.player1.clickCard(context.daringRaid);
+                context.player1.clickCard(context.p2Base);
+
+                context.player2.passAction();
+
+                const p2BaseDamageBeforeForceSurrender = context.p2Base.damage;
```
AMMayberry1, line 136: Sorry for the late comment, just thought of one other case it would be great to have as more of an integration test - can you add one more case where both players try to draw from an empty deck and take damage? Then we confirm that forced surrender does nothing. The damage attribution in that situation is a special case so it would be great to have some coverage for it. Probably easiest to set up as its own case with its own setup.

Frugghi: How should the damage attribution work in that case? I tried having Yoda triggering the draw and the damage was attributed to Yoda's ability, so counting for Forced Surrender.

AMMayberry1: Replied in chat

server/game/gameSystems/DrawSystem.ts
```diff
@@ -52,10 +52,17 @@ export class DrawSystem<TContext extends AbilityContext = AbilityContext> extend
             const contingentEvents = [];
             if (event.amount > event.player.drawDeck.length) {
                 const damageAmount = 3 * (event.amount - event.player.drawDeck.length);
+
+                // Here we generate a the damage event with a new context that contains just the player,
+                // this way the damage is attributed to the player and not the card that triggered the draw (or its controller).
+                // As per rules, the player that is drawing is also the player that is causing the damage and
+                // this is important for cards like Forced Surrender.
```
AMMayberry1, line 59: Could you add a small note like `(FFG ruling confirms this)`. Unfortunately they haven't added the ruling into the wiki yet so we can't link it directly

AMMayberry1 (approved):

PR #194, opened 2024-11-13
Add K2SO implementation

Frugghi:
image

### DIFF
--- a/server/game/cards/01_SOR/units/K2SOCassiansCounterpart.ts
+++ b/server/game/cards/01_SOR/units/K2SOCassiansCounterpart.ts
@@ -1,0 +1,34 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { TargetMode } from '../../../core/Constants';
+
+export default class K2SOCassiansCounterpart extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '3232845719',
+            internalName: 'k2so#cassians-counterpart',
+        };
+    }
+
+    public override setupCardAbilities() {
+        // TODO: This doesn't support Twin Sun correctly
+        this.addWhenDefeatedAbility({
+            title: 'Deal 3 damage to your opponent\'s base or your opponent discards a card from their hand',
+            targetResolver: {
+                mode: TargetMode.Select,
+                choices: {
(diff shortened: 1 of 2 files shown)


test/server/cards/01_SOR/units/K2SOCassiansCounterpart.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+                context.player1.clickCard(context.k2so);
+                context.player1.clickCard(context.kraytDragon);
+                expect(context.p2Base.damage).toBe(2);
+
+                expect(context.player1).toHaveExactPromptButtons(['Deal 3 damage to opponent\'s base', 'The opponent discards a card']);
+                expect(context.player1).not.toHavePassAbilityButton();
+
+                context.player1.clickPrompt('Deal 3 damage to opponent\'s base');
+                expect(context.p2Base.damage).toBe(5);
```
VJubert, line 49: I think you should add a test with no hand for opponent to see if they is not prompt to choose between both choice

Frugghi: Lines 29-36 are covering the case where the opponent doesn't have any card. I'm not sure the option should be auto selected though, wouldn't be possible to pick "discard a card" even if the opponent has no card? I don't think there is a requirement to change the game state here.

server/game/cards/01_SOR/units/K2SOCassiansCounterpart.ts
```diff
@@ -0,0 +1,34 @@
...
+export default class K2SOCassiansCounterpart extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '3232845719',
+            internalName: 'k2so#cassians-counterpart',
+        };
+    }
+
+    public override setupCardAbilities() {
+        // TODO: This doesn't support Twin Sun correctly
```
AMMayberry1, line 14: Fine to leave this here, but for the future - there are so many places where this is true that we definitely shouldn't try to mark all of them

test/server/cards/01_SOR/units/K2SOCassiansCounterpart.spec.ts
AMMayberry1: Yeah the option should still be available

AMMayberry1 (approved):

PR #195, opened 2024-11-13
Add 02_SHD leader Finn

jasimmons:
This was more complicated than I anticipated. The tests might be a little verbose, happy to distill it down to fewer actions per test if that makes sense.

### DIFF
--- a/server/game/cards/02_SHD/leaders/FinnThisIsARescue.ts
+++ b/server/game/cards/02_SHD/leaders/FinnThisIsARescue.ts
@@ -1,0 +1,48 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { LeaderUnitCard } from '../../../core/card/LeaderUnitCard';
+import { RelativePlayer, WildcardCardType } from '../../../core/Constants';
+
+export default class FinnThisIsARescue extends LeaderUnitCard {
+    protected override getImplementationId () {
+        return {
+            id: '9596662994',
+            internalName: 'finn#this-is-a-rescue',
+        };
+    }
+
+    private buildFinnAbilityTargetEffect() {
+        const upgradedCard = (context) => ({ target: context.target.parentCard });
+        return AbilityHelper.immediateEffects.sequential([
+            AbilityHelper.immediateEffects.giveShield(upgradedCard),
+            AbilityHelper.immediateEffects.defeat(),
+        ]);
+    }
+
+    protected override setupLeaderSideAbilities () {
+        this.addActionAbility({
+            title: 'Defeat a friendly upgrade on a unit. If you do, give a Shield token to that unit',
+            cost: [AbilityHelper.costs.exhaustSelf()],
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardTypeFilter: WildcardCardType.Upgrade,
+                cardCondition: (card, context) => card.controller === context.source.controller,
+                immediateEffect: this.buildFinnAbilityTargetEffect(),
+            }
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities () {
+        this.addOnAttackAbility({
+            title: 'You may defeat a friendly upgrade on a unit. If you do, give a Shield token to that unit',
+            optional: true,
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardTypeFilter: WildcardCardType.Upgrade,
+                cardCondition: (card, context) => card.controller === context.source.controller,
+                immediateEffect: this.buildFinnAbilityTargetEffect(),
+            }
+        });
+    }
+}
+
+FinnThisIsARescue.implemented = true;
--- a/server/game/cards/02_SHD/leaders/FinnThisIsARescue.ts
+++ b/server/game/cards/02_SHD/leaders/FinnThisIsARescue.ts
@@ -12,7 +12,7 @@ export default class FinnThisIsARescue extends LeaderUnitCard {
 
     protected override setupLeaderSideAbilities () {
         this.addActionAbility({
-            title: 'Defeat a friendly upgrade on a unit. If you do, give a Shield token to that unit',
+            title: 'Defeat a friendly upgrade on a unit',
             cost: [AbilityHelper.costs.exhaustSelf()],
             targetResolver: {
                 cardTypeFilter: WildcardCardType.Upgrade,
@@ -35,7 +35,7 @@ export default class FinnThisIsARescue extends LeaderUnitCard {
 
     protected override setupLeaderUnitSideAbilities () {
         this.addOnAttackAbility({
-            title: 'Defeat a friendly upgrade on a unit. If you do, give a Shield token to that unit',
+            title: 'Defeat a friendly upgrade on a unit',
             optional: true,
             targetResolver: {
                 cardTypeFilter: WildcardCardType.Upgrade,
--- a/server/game/cards/02_SHD/leaders/FinnThisIsARescue.ts
+++ b/server/game/cards/02_SHD/leaders/FinnThisIsARescue.ts
@@ -10,37 +10,47 @@ export default class FinnThisIsARescue extends LeaderUnitCard {
         };
     }
 
-    private buildFinnAbilityTargetEffect() {
-        const upgradedCard = (context) => ({ target: context.target.parentCard });
-        return AbilityHelper.immediateEffects.sequential([
-            AbilityHelper.immediateEffects.giveShield(upgradedCard),
-            AbilityHelper.immediateEffects.defeat(),
-        ]);
-    }
-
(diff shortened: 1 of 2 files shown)


server/game/cards/02_SHD/leaders/FinnThisIsARescue.ts
```diff
@@ -0,0 +1,50 @@
...
+            internalName: 'finn#this-is-a-rescue',
+        };
+    }
+
+    private buildFinnAbilityTargetEffect() {
+        const upgradedCard = context => ({target: context.target.parentCard});
+        return AbilityHelper.immediateEffects.sequential([
+            AbilityHelper.immediateEffects.giveShield(upgradedCard),
+            AbilityHelper.immediateEffects.defeat(),
+        ]);
```
jasimmons, line 18: I had trouble resolving the parent card of an upgrade after it was defeated. As a result, I ended up with giveShield before defeat. Happy to change this to be more accurate if there's a way to do it that I missed.

test/server/cards/02_SHD/leaders/FinnThisIsARescue.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+                        resources: 5
+                    }
+                });
+            });
+
+            it('should defeat a friendly upgrade and give a shield token', function () {
+                const { context } = contextRef;
+
+                // Equip a friendly upgrade to battlefield marine
+                context.player1.clickCard(context.jediLightsaber);
```
adbronico, line 25: You can remove this and have the test start with the lightsaber equipped. See my above comment.

test/server/cards/02_SHD/leaders/FinnThisIsARescue.spec.ts
```diff
@@ -0,0 +1,127 @@
+describe('Finn, This is a Rescue', function () {
+    integration(function (contextRef) {
+        describe('Finn\'s undeployed ability', function () {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['jedi-lightsaber'],
+                        leader: { card: 'finn#this-is-a-rescue', deployed: false },
+                        groundArena: ['battlefield-marine'],
```
adbronico, line 10: You can have this as:
`groundArena: [ { card: 'battlefield-marine', upgrades: [ 'jedi-lightsaber' ] } ]`

test/server/cards/02_SHD/leaders/FinnThisIsARescue.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+                        resources: 5
+                    },
+                    player2: {
+                        hand: ['top-target'],
+                        groundArena: ['wampa'],
```
adbronico, line 15: I would add an upgrade to the enemy Wampa and ensure it isn't selectable. Also add a TODO for a friendly upgrade on an enemy unit

adbronico (dismissed):

server/game/cards/02_SHD/leaders/FinnThisIsARescue.ts
AMMayberry1: We definitely need to keep them in the right order to ensure game state stays correct in between. Not sure exactly how that could cause an issue but weird edge cases always come up.

That said, I think I can make a small tweak to `sequential()` so that we could do something like this:

```typescript
  immediateEffect: AbilityHelper.immediateEffects.sequential((context) => {
      const upgradedCard = context.target.parentCard;
      return [
          AbilityHelper.immediateEffects.defeat(),
          AbilityHelper.immediateEffects.giveShield(upgradedCard),
      ];
  }) });
```

I'll do the patch for that later today.

AMMayberry1: Okay I did a fix and it passes your current tests. Probably easier for you to just merge the branch into yours, could you merge `ammayberry1/finn-bug-fix` (branch link: https://github.com/SWU-Karabast/forceteki/tree/ammayberry1/finn-bug-fix)

AMMayberry1: Ah okay so I did all of the above before really reviewing the code and realized that this is not the right implementation of his ability. Any time a card says "if you do" we need to use the ifYouDo property because there's some nuance around effect resolution that it handles correctly.

The `sequential` operation will do both the defeat and the giveShield operation nomatter what, but if something prevented the defeat from resolving then giveShield should also not resolve. Again, I'm not 100% if there's a case where that would come up currently, but best to use the right tool for the job. Take a look at Enforced Loyalty or Hondo Ohnaka for examples of how to use the `ifYouDo` property.

test/server/cards/02_SHD/leaders/FinnThisIsARescue.spec.ts
AMMayberry1: Note: I left a comment asking just to implement that case outright, it requires the explicit action to play an upgrade out to an opponent's unit but I think it's fine to just do that for now so we can have this test done

test/server/cards/02_SHD/leaders/FinnThisIsARescue.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+                // Opponent equips an upgrade to battlefield marine
+                context.player2.clickCard(context.topTarget);
+                context.player2.clickCard(context.battlefieldMarine);
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['jedi-lightsaber', 'top-target']);
+
+                // Unable to use Finn's ability on non-friendly upgrade
+                context.player1.clickCard(context.finn);
+                context.player1.clickPrompt('Defeat a friendly upgrade on a unit. If you do, give a Shield token to that unit');
+                expect(context.player1).not.toBeAbleToSelect(context.topTarget);
```
AMMayberry1, line 55: We try to avoid testing like this just because it's pretty nonspecific, there's unexpected things that could happen which would still allow this test to pass. Since we're not expecting any target prompt at all, instead better to test `expect(context.player2).toBeActivePlayer();` to confirm that there was no prompt of any kind for player1 and control passed immediately. That plus the upgrade check below should be a good check for this case.

test/server/cards/02_SHD/leaders/FinnThisIsARescue.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['jedi-lightsaber']);
+
+                // Opponent equips an upgrade to battlefield marine
+                context.player2.clickCard(context.topTarget);
+                context.player2.clickCard(context.battlefieldMarine);
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['jedi-lightsaber', 'top-target']);
+
+                // Unable to use Finn's ability on non-friendly upgrade
+                context.player1.clickCard(context.finn);
+                context.player1.clickPrompt('Defeat a friendly upgrade on a unit. If you do, give a Shield token to that unit');
```
AMMayberry1, line 54: Protip: you can set `resources: 4` for player1 so that there aren't enough resources to deploy and you won't get this prompt since his leader action is the only available on-click action

test/server/cards/02_SHD/leaders/FinnThisIsARescue.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+                expect(context.player1).not.toBeAbleToSelect(context.topTarget);
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['top-target', 'shield']);
+            });
```
AMMayberry1, line 57: We also need to test that the exhaust cost was set up correctly, could you add a test after this which confirms that a click on Finn does nothing since he is exhausted (just make sure it's player1's action and not player2's): `expect(context.finn).not.toHaveAvailableActionWhenClickedBy(context.player1);`

test/server/cards/02_SHD/leaders/FinnThisIsARescue.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+            it('should not defeat an opponent\'s upgrade', function () {
+                const { context } = contextRef;
+
+                context.player1.passAction();
+
+                context.player2.clickCard(context.topTarget);
+                context.player2.clickCard(context.battlefieldMarine);
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['jedi-lightsaber', 'top-target']);
```
AMMayberry1, line 53: We can join this into the above test by having player2 attach the Top Target at the beginning and confirm that it's never selectable. Same note for the deployed version of this test

AMMayberry1 (approved):

PR #197, opened 2024-11-13
feat: implements republic tactical officer unit

jgordijo:
image


AMMayberry1:
Left a comment in the discord but also putting it here - we're not doing set 3 cards for now, we want to finish up set 1 first. Please close this PR but feel free to leave the branch up and leave your name on the card in the sheet, once we open up set 3 you can immediately open this PR again.

PR #199, opened 2024-11-14
Snapshot Reflexes implementation

aruizc:
image

### DIFF
--- a/server/game/cards/01_SOR/upgrades/SnapshotReflexes.ts
+++ b/server/game/cards/01_SOR/upgrades/SnapshotReflexes.ts
@@ -11,7 +11,7 @@ export default class SnapshotReflexes extends UpgradeCard {
 
     public override setupCardAbilities() {
         this.addWhenPlayedAbility({
-            title: 'You may attack with attached unit.',
+            title: 'Attack with attached unit',
             optional: true,
             immediateEffect: AbilityHelper.immediateEffects.attack((context) => ({
                 target: context.source.parentCard
--- a/server/game/cards/01_SOR/upgrades/SnapshotReflexes.ts
+++ b/server/game/cards/01_SOR/upgrades/SnapshotReflexes.ts
@@ -1,0 +1,27 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { Card } from '../../../core/card/Card';
+import { UpgradeCard } from '../../../core/card/UpgradeCard';
+import { PlayType, Trait } from '../../../core/Constants';
+
+export default class SnapshotReflexes extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '9985638644',
+            internalName: 'snapshot-reflexes'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
+            title: 'You may attack with attached unit.',
+            optional: true,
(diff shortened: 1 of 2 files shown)


test/server/cards/01_SOR/upgrades/SnapshotReflexes.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+                        groundArena: ['battlefield-marine'],
+                        hand: ['snapshot-reflexes']
+                    },
+                    player2: {
+                        groundArena: ['specforce-soldier']
+                    }
+                });
+            });
+
+            it('attached unit may attack when played', function () {
```
Ennis0, line 17: The string argument for each `it` should form a complete sentence when added to the string from the `describe` call above. That sentence is output when a test fails, so it should describe what the test is supposed to prove so that whoever reads it gets an idea of what failed.

server/game/cards/01_SOR/upgrades/SnapshotReflexes.ts
```diff
@@ -0,0 +1,23 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '9985638644',
+            internalName: 'snapshot-reflexes'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'You may attack with attached unit.',
```
AMMayberry1, line 14: nit: for ability titles, we don't include "You may" since there will be "pass" button making it clear that the effect is optional. Also we shouldn't have a period at the end (we haven't been consistent about that, I need to go through and clean up at some point)

test/server/cards/01_SOR/upgrades/SnapshotReflexes.spec.ts
AMMayberry1: +1. If you want to see how the test name looks, you can just force the test to fail and it will print it out in the error message

aruizc: I modified the texts to be more self descriptive and describe the upgrade behaviour

server/game/cards/01_SOR/upgrades/SnapshotReflexes.ts
aruizc: done, thanks for clarification

test/server/cards/01_SOR/upgrades/SnapshotReflexes.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.snapshotReflexes);
+                context.player1.clickCard(context.battlefieldMarine);
+
+                context.player1.clickPrompt('You may attack with attached unit.');
```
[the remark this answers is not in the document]
aruizc: done

test/server/cards/01_SOR/upgrades/SnapshotReflexes.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+                context.player1.clickCard(context.snapshotReflexes);
+                context.player1.clickCard(context.battlefieldMarine);
+
+                context.player1.passAction();
+
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['snapshot-reflexes']);
+                expect(context.battlefieldMarine.exhausted).toBe(false);
+                expect(context.player2).toBeActivePlayer();
+            });
```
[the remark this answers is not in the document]
aruizc: done

AMMayberry1 (approved):

PR #202, opened 2024-11-14
Volunteer Soldier and Outflank 

aruizc:
- Volunteer Soldier Implementation 
- Outflank implementation
- Adds unit test for Rey Keeping The Past when Kylo Ren is leader

image

image


server/game/cards/02_SHD/events/Outflank.ts
```diff
@@ -0,0 +1,29 @@
...
+        return {
+            id: '0802973415',
+            internalName: 'outflank',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setEventAbility({
+            title: 'Attack with two units (one at a time)',
+            initiateAttack: {
```
AMMayberry1, line 16: Just for clarity let's do this as the below instead of using an empty `initiateAttack`. They are both equivalent in function and the form below is a little easier to understand than an empty `initiateAttack` imo.

```typescript
immediateEffect: AbilityHelper.immediateEffects.attack()
```

The attack inside of `then` can stay the way it is

test/server/cards/02_SHD/events/Outflank.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.outflank);
+
+                expect(context.pykeSentinel.exhausted).toBe(true);
+                expect(context.p2Base.damage).toBe(2);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should be playable even if there is no units', function () {
```
AMMayberry1, line 51: Can remove this test, we have a lot of cases confirming that events with no target are playable already

server/game/cards/02_SHD/events/Outflank.ts
aruizc: done

AMMayberry1 (approved):

PR #204, opened 2024-11-15
Bunch of cards 9

VJubert:


test/server/cards/01_SOR/units/MillenniumFalconPieceOfJunk.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+                context.player1.clickPrompt('Done');
+                context.player2.clickPrompt('Done');
+
+                // return falcon to hand
+                expect(context.player1).toHaveEnabledPromptButtons(['Pay 1 resource', 'Return this unit to her owner\'s hand']);
+                context.player1.clickPrompt('Return this unit to her owner\'s hand');
+
+                expect(context.player1).toBeActivePlayer();
+                expect(context.millenniumFalcon).toBeInZone('hand');
+            });
```
AMMayberry1, line 63: Can we add a TODO for checking that it enters play ready when played from somewhere other than hand

test/server/cards/02_SHD/units/EchoRestored.spec.ts
```diff
@@ -0,0 +1,96 @@
+describe('Echo, Restored', function () {
+    integration(function (contextRef) {
+        describe('Echo\'s ability', function () {
+            const prompt = 'Discard a card from your hand. Give 2 Experience tokens to a unit in play with the same name as the discarded card.';
+            it('can heal a target unit for the amount of damage it has', function () {
```
AMMayberry1, line 5: nit: wrong test name

test/server/cards/02_SHD/units/EchoRestored.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+                context.player1.clickCard(context.echo);
+                context.player1.clickPrompt(prompt);
+                context.player1.clickCard(handWampa);
+
+                // 2 units with the same title, can choose between them
+                expect(context.player1).toBeAbleToSelectExactly([friendlyGroundWampa, enemyGroundWampa]);
+                expect(context.player1).not.toHaveChooseNoTargetButton();
+                context.player1.clickCard(friendlyGroundWampa);
+
+                // friendly wampa should hvae 2 experiences tokens
```
AMMayberry1, line 84: Fantastic job with these tests! Covers every case I can think of

test/server/cards/02_SHD/units/EmboStoicAndResolute.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                context.player1.clickCard(context.embo);
+                context.player1.clickCard(context.battlefieldMarine);
+                expect(context.player1).toBeAbleToSelectExactly([context.embo, context.echoBaseDefender, context.consularSecurityForce, context.swoopRacer, context.greenSquadronAwing, context.superlaserTechnician]);
+                expect(context.player1).toHaveChooseNoTargetButton();
+                context.player1.setDistributeHealingPromptState(new Map([
+                    [context.echoBaseDefender, 2],
+                ]));
+
+                // heal echo base defender
+                expect(context.echoBaseDefender.damage).toBe(0);
```
AMMayberry1, line 38: Can we set it up with a unit with 3 damage to confirm the exact amount of healing

test/server/cards/02_SHD/units/EmboStoicAndResolute.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                context.player2.clickPrompt('Put Superlaser Technician into play as a resource and ready it');
+                expect(context.player1).toBeAbleToSelectExactly([context.embo, context.echoBaseDefender, context.consularSecurityForce, context.swoopRacer, context.greenSquadronAwing]);
+                expect(context.player1).toHaveChooseNoTargetButton();
+                context.player1.setDistributeHealingPromptState(new Map([
+                    [context.echoBaseDefender, 2],
+                ]));
+
+                reset(false);
+
+                // embo does not complete attack, nothing happen
```
AMMayberry1, line 62: Couple more cases it would be good to have:

- Enemy unit attacks Embo and is defeated, ability doesn't trigger
- Defender is defeated before combat damage due to some effect like Vambrace Flamethrower, Embo ability still triggers (this can be a separate setupTest potentially since it involves an upgrade)

Also re: the comment on Embo, we should check the following more complicated case (could also be its own setupTest). These are the steps:

1. Enemy unit is defeated by Vanquish or something
2. Opponent uses Emperor's Legion to return it to hand
3. Opponent plays the unit
4. Embo attacks and does not defeat the unit
5. Confirm that his ability does not trigger since it was not defeated during this attack (even though it was defeated earlier in the phase)

server/game/cards/02_SHD/units/XanaduBloodCadBanesReward.ts
```diff
@@ -0,0 +1,38 @@
...
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardCondition: (card, context) => card !== context.source && card.isUnit() && card.hasSomeTrait(Trait.Underworld),
+                immediateEffect: AbilityHelper.immediateEffects.returnToHand()
+            },
+            ifYouDo: {
+                title: 'Exhaust an enemy unit or resource',
+                targetResolver: {
+                    controller: RelativePlayer.Opponent,
+                    zoneFilter: [ZoneName.Resource, ZoneName.GroundArena, ZoneName.SpaceArena],
```
AMMayberry1, line 30: We don't want to force the player to pick a specific resource to exhaust if they want to choose that option, so we should set this up like the Leia unit where they have a selection with two options.

It actually would be really good to have a filter that will prevent the system from being able to target stuff in the resource zone. Do you mind going to `ExhaustOrReadySystem.canAffect` and removing line 17 (resource zone check), I think that will do it

VJubert: removing this line break some system so I revert change on it
see : https://github.com/SWU-Karabast/forceteki/actions/runs/11900082668

AMMayberry1 (approved):

PR #206, opened 2024-11-15
Add Smoke and Cinders and Cripple Authority implementation

Frugghi:
| Smoke and Cinders | Cripple Authority  |
| --- | ---- |
| image | image |


server/game/cards/01_SOR/events/SmokeAndCinders.ts
```diff
@@ -0,0 +1,27 @@
...
+            internalName: 'smoke-and-cinders',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Each player discards all but 2 cards from their hand',
+            immediateEffect: AbilityHelper.immediateEffects.simultaneous(
+                (context) => context.game.getPlayers().map((player) =>
+                    AbilityHelper.immediateEffects.discardCardsFromOwnHand({
```
AMMayberry1, line 17: This is a really clever way to pull this off, I think it might be better though if we could add something that directly supports this pattern as a feature for the sake of people who might have trouble parsing this level of complexity.

We had a false start on doing a Qira leader implementation a while back (long story), but we added there an option to the damage system for amount to be a function that accepts a card that could be computed dynamically. I think it would be good to do the same for the discard system as it would read much easier (though in this case it would accept a player obviously): https://github.com/SWU-Karabast/forceteki/blob/51a7527b4197176f3aaed36c31786979278c0ce6/server/game/gameSystems/DamageSystem.ts#L10

server/game/cards/02_SHD/events/CrippleAuthority.ts
```diff
@@ -0,0 +1,35 @@
...
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Draw a card. Each opponent who controls more resources than you discards a card from their hand',
+            immediateEffect: AbilityHelper.immediateEffects.simultaneous(
+                (context) => [
+                    AbilityHelper.immediateEffects.draw({
+                        target: context.source.controller,
+                        amount: 1
+                    }),
+                    ...context.game.getPlayers()
```
AMMayberry1, line 21: Since there's only one opponent to consider could we just do this as a conditional? I know Twin Suns is theoretically a thing we could do at some point but even if so it's so far off that I'd rather just keep implementations as clean and simple as possible on the assumption of one opponent for now, and if we do get there I'll want to add a lot of helpers to avoid needing to write up complex logic like this directly in the card impls. As much as possible the effects should be one-liners, and any complexity hidden in the systems (except for the cards that have really wacky one-off mechanics).

test/server/cards/01_SOR/events/SmokeAndCinders.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+            });
+
+            it('discards all but 2 cards from each player hand', function () {
+                const { context } = contextRef;
+
+                const reset = () => {
+                    context.player2.passAction();
+
+                    context.player1.moveCard(context.smokeAndCinders, 'hand');
+                    context.player1.claimInitiative();
```
AMMayberry1, line 23: Can also just do `context.moveToNextActionPhase();`

server/game/cards/02_SHD/events/CrippleAuthority.ts
Frugghi: Sure, I changed it! The conditional is a bit verbose though, it takes the same amount of code to target a single player.

server/game/cards/01_SOR/events/SmokeAndCinders.ts
```diff
@@ -0,0 +1,27 @@
...
+            id: '6702266551',
+            internalName: 'smoke-and-cinders',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Each player discards all but 2 cards from their hand',
+            immediateEffect: AbilityHelper.immediateEffects.simultaneous(
+                (context) => context.game.getPlayers().map((player) =>
```
AMMayberry1, line 16: With the amount function we can simplify by not using `simultaneous()` and just making `target` an array with both players, I believe (if not then we should aim for that, I think that'd be the ideal syntax here)

Frugghi: Yes, that's much better 😄

AMMayberry1 (approved):

PR #208, opened 2024-11-16
Slavers Freighter card and tests

andtrobs:


server/game/cards/02_SHD/units/SlaversFreighter.ts
```diff
@@ -0,0 +1,26 @@
...
+        };
+    }
+
+    protected override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'You may ready another unit with power equal to or less than the number of upgrades on enemy units',
+            optional: true,
+            targetResolver: {
+                locationFilter: WildcardLocation.AnyArena,
+                cardCondition: (card, context) => card.isUnit() && card.getPower() <= context.source.controller.opponent.getUnitsInPlay().reduce((total, unit) => total + unit.upgrades.length, 0),
```
AMMayberry1, line 19: Nice! One nit though, do you mind putting the upgrade check on its own line just for improved readability

test/server/cards/02_SHD/units/SlaversFreighter.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+                        groundArena: [{ card: 'cell-block-guard', exhausted: true }, { card: 'count-dooku#darth-tyranus', exhausted: true }],
+                    },
+                    player2: {
+                        groundArena: [{ card: 'wampa', upgrades: ['fallen-lightsaber', 'electrostaff'], exhausted: true }],
+                        spaceArena: [{ card: 'green-squadron-awing', upgrades: ['resilient'] }]
+                    }
+                });
+            });
+
+            it('should allow a friendly unit with power less than or equal to the number of upgrades on enemy units to be readied.', function () {
```
AMMayberry1, line 18: nit: shouldn't say "friendly unit" in the test title, just "unit"

test/server/cards/02_SHD/units/SlaversFreighter.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.slaversFreighter);
+
+                expect(context.player1).toHavePassAbilityButton();
+                expect(context.player1).toBeAbleToSelectExactly([context.cellBlockGuard, context.greenSquadronAwing]);
```
AMMayberry1, line 24: Could we bump the number of upgrades to 5 and confirm that slaver's freighter can target itself. Also, after that could we replace Dooku with a 6-cost just to confirm that the upgrade count cutoff is exactly correct

test/server/cards/02_SHD/units/SlaversFreighter.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+                context.player1.clickCard(context.cellBlockGuard);
+
+                expect(context.cellBlockGuard.exhausted).toBe(false);
+
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should be optional', function () {
```
AMMayberry1, line 33: We have a practical issue with our tests where the Jasmine tests are super slow and we have finite Github Actions time in the free tier, so we have to be pretty sparing in our tests and try to consolidate down as much as we can (even though it's not great test practice obviously).

In this case we typically don't add a whole test for the pass ability button, our practice is to assume that if it's present it's working (and we do have tests that check that).

test/server/cards/02_SHD/units/SlaversFreighter.spec.ts
andtrobs: I just checked the wording of the card and it says ready another unit, so shouldn't be able to target itself, but your point still stands, I should test this with the higher number of upgrades, I will update

AMMayberry1: Ahh good catch, I missed that. Thanks

AMMayberry1 (approved):

PR #213, opened 2024-11-17
Ongoing Effects TS Conversion

CDMayberry:
Started to dig into card vs player effects but turned into just converting the OnGoingEffects classes into TS, so I made a PR for the conversion alone.

NOTES:
- Some of the JS files didn't like importing OngoingEffectSource as a default export and I'm not sure why, so for now I've removed the default export.
- Moved targetController assignment logic to the base OngoingEffect class, as both derived classes did the exact same thing.
- Removed unused property.matchTarget assignment in OngoingCardEffect, it's assigning to the properties object, but the base class already does this logic and does it before this line.
- Still has some "any" types, OngoingEffect's "ability" field is seemingly assigned to two completely different types, as noted. A few things I set as "GameObject" because I was unsure, but they could maybe just be Card.
- Re-ordered OngoingCardEffect.isValidTarget checks and made the non-function matchTarget check always return, as it's expected to be a function by the end.

### DIFF
--- a/server/game/core/ongoingEffect/OngoingCardEffect.ts
+++ b/server/game/core/ongoingEffect/OngoingCardEffect.ts
@@ -1,11 +1,19 @@
-const OngoingEffect = require('./OngoingEffect.js');
-const { RelativePlayer, WildcardLocation, WildcardCardType } = require('../Constants.js');
-const EnumHelpers = require('../utils/EnumHelpers.js');
-const Contract = require('../utils/Contract.js');
-const Helpers = require('../utils/Helpers.js');
+import OngoingEffect from './OngoingEffect';
+import { RelativePlayer, WildcardLocation, WildcardCardType, CardTypeFilter, Location } from '../Constants';
+import * as EnumHelpers from '../utils/EnumHelpers';
+import * as Contract from '../utils/Contract';
+import * as Helpers from '../utils/Helpers';
+import Game from '../Game';
+import { Card } from '../card/Card';
+import { IOngoingEffectProps } from '../../Interfaces';
+import { OngoingEffectImpl } from './effectImpl/OngoingEffectImpl';
 
 class OngoingCardEffect extends OngoingEffect {
-    constructor(game, source, properties, effect) {
+    public targetsSourceOnly: boolean;
+    public targetLocationFilter: WildcardLocation;
+    public targetCardTypeFilter: CardTypeFilter[];
+
+    public constructor(game: Game, source: Card, properties: IOngoingEffectProps, effect: OngoingEffectImpl<any>) {
         super(game, source, properties, effect);
 
         if (!properties.matchTarget) {
@@ -14,8 +22,6 @@ class OngoingCardEffect extends OngoingEffect {
                 this.targetsSourceOnly = true;
                 return;
             }
-
-            properties.matchTarget = () => true;
         }
 
         this.targetsSourceOnly = false;
@@ -28,8 +34,6 @@ class OngoingCardEffect extends OngoingEffect {
             this.targetLocationFilter = properties.targetLocationFilter;
         }
 
-        this.targetController = properties.targetController || RelativePlayer.Self;
-
         // TODO: rework getTargets() so that we can provide an array while still not searching all cards in the game every time
         Contract.assertFalse(Array.isArray(properties.targetLocationFilter), 'Target location filter for an effect definition cannot be an array');
 
@@ -37,14 +41,18 @@ class OngoingCardEffect extends OngoingEffect {
     }
 
     /** @override */
-    isValidTarget(target) {
-        if (target === this.matchTarget) {
-            // This is a hack to check whether this is a lasting effect
-            return true;
-        }
-
+    public override isValidTarget(target) {
         if (this.targetsSourceOnly) {
             return target === this.context.source;
+        }
+
+        if (typeof this.matchTarget !== 'function') {
+            if (target === this.matchTarget) {
+                // This is a hack to check whether this is a lasting effect
+                return true;
+            }
+
+            return false;
         }
 
         return (
@@ -57,7 +65,7 @@ class OngoingCardEffect extends OngoingEffect {
     }
 
     /** @override */
-    getTargets() {
+    public override getTargets() {
         if (this.targetsSourceOnly) {
             return [this.context.source];
         } else if (this.targetLocationFilter === WildcardLocation.Any) {
@@ -69,4 +77,4 @@ class OngoingCardEffect extends OngoingEffect {
     }
 }
 
-module.exports = OngoingCardEffect;
+export default OngoingCardEffect;
--- a/server/game/core/ongoingEffect/OngoingCardEffect.ts
+++ b/server/game/core/ongoingEffect/OngoingCardEffect.ts
@@ -29,7 +29,7 @@ export class OngoingCardEffect extends OngoingEffect {
 
         this.targetsSourceOnly = false;
         this.targetController = properties.targetController || RelativePlayer.Self;
-        Contract.assertArrayIncludes(relativePlayerValues, this.targetController, "target controller must be a RelativePlayer enum.");
+        Contract.assertArrayIncludes(relativePlayerValues, this.targetController, 'target controller must be a RelativePlayer enum.');
 
         if (!properties.targetLocationFilter) {
             this.targetLocationFilter = properties.sourceLocationFilter === WildcardLocation.Any
--- a/server/game/core/ongoingEffect/OngoingCardEffect.ts
+++ b/server/game/core/ongoingEffect/OngoingCardEffect.ts
@@ -1,19 +1,22 @@
 import { OngoingEffect } from './OngoingEffect';
-import { RelativePlayer, WildcardLocation, WildcardCardType, CardTypeFilter, LocationFilter } from '../Constants';
+import { RelativePlayer, WildcardLocation, WildcardCardType, CardTypeFilter, LocationFilter, RelativePlayerValues } from '../Constants';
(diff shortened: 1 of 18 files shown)


server/game/core/ongoingEffect/OngoingEffect.ts
```diff
@@ -148,4 +174,4 @@ class OngoingEffect {
     }
 }
 
-module.exports = OngoingEffect;
+export default OngoingEffect;
```
AMMayberry1, line 177: We haven't really been using export default a lot anyway, could we just change to using `export class OngoingEffect` and then update relevant imports to be `{ OngoingEffect }`? This would be the only TS file in the repo that has the export at the end of the file like this

server/game/core/ongoingEffect/OngoingCardEffect.ts
```diff
@@ -1,11 +1,19 @@
...
+import Game from '../Game';
+import { Card } from '../card/Card';
+import { IOngoingEffectProps } from '../../Interfaces';
+import { OngoingEffectImpl } from './effectImpl/OngoingEffectImpl';
 
 class OngoingCardEffect extends OngoingEffect {
-    constructor(game, source, properties, effect) {
+    public targetsSourceOnly: boolean;
+    public targetLocationFilter: LocationFilter;
+    public targetCardTypeFilter: CardTypeFilter[];
```
AMMayberry1, line 14: We should make these readonly if possible

server/game/core/ongoingEffect/OngoingEffect.ts
```diff
@@ -29,7 +34,24 @@ const EnumHelpers = require('../utils/EnumHelpers');
  *                        and the numerical value of the effect, if any.
  */
 class OngoingEffect {
-    constructor(game, source, properties, effectImpl) {
+    public game: Game;
+    public source: Card;
+    // TODO: Can we make GameObject more specific? Can we add generics to the class for AbilityContext?
+    public matchTarget: GameObject | ((target: GameObject, context: AbilityContext) => boolean);
```
AMMayberry1, line 40: I think the most specific we can (and should) do is `Player | Card`, and then we could specify just one or the other in an override in the child classes

server/game/core/ongoingEffect/OngoingEffect.ts
```diff
@@ -29,7 +34,24 @@ const EnumHelpers = require('../utils/EnumHelpers');
...
+    public duration?: Duration;
+    public until: WhenType;
+    public condition: (context?: AbilityContext) => boolean;
+    public sourceLocationFilter: LocationFilter;
+    public canChangeZoneOnce: boolean;
+    public canChangeZoneNTimes: number;
```
AMMayberry1, line 46: We can outright remove these canChangeZone properties; tbh I thought I already had

server/game/core/ongoingEffect/OngoingEffect.ts
```diff
@@ -29,7 +34,24 @@ const EnumHelpers = require('../utils/EnumHelpers');
...
+    public impl: OngoingEffectImpl<any>;
+    // ISSUE: refreshContext sets ability to IOngoingEffectProps, but the listed type for context is PlayerOrCardAbility. Why is there a mismatch? Are we just overriding it in the context of OngoingEffects and everywhere else it acts as PlayerOrCardAbility?
```
AMMayberry1, line 48: That was an oversight most likely, constant abilities are just an interface (a set of properties really) and don't actually derive from PlayerOrCardAbility. So it almost certainly is supposed to always be IOngoingEffectProps

server/game/core/ongoingEffect/OngoingEffect.ts
```diff
@@ -42,54 +64,56 @@ class OngoingEffect {
         this.impl = effectImpl;
         this.ability = properties;
         this.targets = [];
+        this.targetController = properties.targetController || RelativePlayer.Self;
```
AMMayberry1, line 67: Wondering if this should be here? `targetController` is referring to the controller of a card which wouldn't be meaningful if the target is a player

server/game/core/ongoingEffect/OngoingCardEffect.ts
```diff
@@ -1,11 +1,19 @@
...
+import { RelativePlayer, WildcardLocation, WildcardCardType, CardTypeFilter, LocationFilter } from '../Constants';
+import * as EnumHelpers from '../utils/EnumHelpers';
+import * as Contract from '../utils/Contract';
+import * as Helpers from '../utils/Helpers';
+import Game from '../Game';
+import { Card } from '../card/Card';
+import { IOngoingEffectProps } from '../../Interfaces';
+import { OngoingEffectImpl } from './effectImpl/OngoingEffectImpl';
 
 class OngoingCardEffect extends OngoingEffect {
```
AMMayberry1, line 11: One thing that might be worth looking at would be seeing if you can remove or integrate IOngoingCardEffect. There's a comment note on it but the tl;dr is in jigoku, IOngoingCardEffect is used in a lot of places in place of OngoingCardEffect to hide the details of type stuff so they wouldn't have to deal with the fact that the inner effect impl could be any type at all.

I tried removing it once but very quickly discovered it was a very deep rabbit hole, if you find yourself doing the same we can leave it for now and the current change is still a really useful incremental step forward.

server/game/core/ongoingEffect/OngoingEffect.ts
CDMayberry: Can do, I had to do that for OngoingEffectSource anyway.

server/game/core/ongoingEffect/OngoingCardEffect.ts
CDMayberry: This definitely derived from trying to untangle IOngoingCardEffect, I think it and IOngoingPlayerEffect could be consolidated but it was a bit of digging, and then I reached the classes and realized they weren't TS yet. So this initial pass will just be getting the classes up to snuff, then I can work getting the interfaces next.

server/game/core/ongoingEffect/OngoingEffect.ts
AMMayberry1: Yeah go ahead, lmk if it breaks anything. I'm pretty sure they're not used anywhere but would be good to know if I'm wrong

server/game/core/ongoingEffect/OngoingEffect.ts
AMMayberry1: Yeah that's a little bit of a hack but I believe it's one propagated forward from jigoku. The only place we use it currently I believe is in `GainAbility.ts` which is the effect type used when a card gains an ability from the effect of another card, e.g. when an upgrade says "attached unit gains: `When played: ...`"

In there we have to refer back to the source ability that caused the gain to happen, and since context.ability (iirc) isn't naturally set for a constant ability / ongoing effect, we explicitly add it on a refresh. I _think_ that's what's going on from my memory, if that doesn't line up feel free to ping me about it, and if you can see a better pattern for accomplishing that then please definitely feel free to implement it

server/game/core/ongoingEffect/OngoingEffect.ts
AMMayberry1: Oh hmm, let's move it back into just `OngoingCardEffect` and see what happens? Definitely it shouldn't be set this way imo. Just a lot of messiness in the effects in general (as you've seen) so it may have just been a copy-paste error

server/game/core/ongoingEffect/OngoingCardEffect.ts
AMMayberry1: Yup sgtm

server/game/core/ongoingEffect/OngoingEffect.ts
CDMayberry: Aaaah ok, so it is used by both, but in the context of OngoingPlayerEffect it means "Target Controller" rather than "Target's Controller". OngoingPlayerEffect can also have it be a literal player object in addition to being a RelativePlayer, whereas OngoingCardEffect can only be a RelativePlayer, at least from what I can infer. I'll move it up and update the types appropriately.

server/game/core/ongoingEffect/OngoingEffect.ts
CDMayberry: Ok so I've rewritten this comment ~~4~~5 times due to different rabbit holes, but looking at jigoku it definitely wasn't propagated forward. this field was originally `this.ability = properties.ability` rather than the properties object itself, and refreshContext explicitly had `if(this.ability) { this.context.ability = this.ability }`. 

That said, the system definitely relies on its current design, I see the process of building the ongoing effect essentially makes the properties object ability-like. I'm just gonna leave type cast with a note and put this under "needs a look at another time".

server/game/core/ongoingEffect/OngoingEffect.ts
CDMayberry: Addendum: So it's a little harder to enforce this on the ability creation side, the I added separate interfaces for IPlayerOngoingEffectProps and ICardOngoingEffectProps, but the ability creation process just has it's own set of interfaces it loosely tacks on which include targetController. I've still separated it, but added a enforcement in OngoingCardEffect to ensure it's always a relative player.

server/game/core/ongoingEffect/OngoingEffect.ts
CDMayberry: Didn't break anything, so we're good. Probably.

server/game/core/ongoingEffect/OngoingEffect.ts
AMMayberry1: Ahh okay, finally found an explanation in the comments. The naming for this property is just horrible:
image

AMMayberry1 (approved):

PR #214, opened 2024-11-17
Qira leader and tests

andtrobs:
Heal and Damage systems also modified to allow functions to be specified for the amount to heal and ability damage


server/game/gameSystems/DamageSystem.ts
```diff
@@ -83,7 +83,14 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext, TPro
 
     private getDamageAmountFromEvent(event: any): number {
         if (event.amount != null) {
-            return event.amount;
+            switch (typeof event.amount) {
```
AMMayberry1, line 86: I think it was set up this way in the reference I gave you as well, which was an oversight on our part. The tl;dr is this should go in `addPropertiesToEvent`, detailed explanation below if you're interested. We should do the same for Heal just to be consistent.

Basically there are some replacement effects that could check the damage amount before it reaches resolution, i.e. Boba Fett's Armor. For that to work we need to determine the concrete amount of damage when the event is actually created, which is when `addPropertiesToEvent` is called. If you're interested you can see the event resolution steps in `EventWindow.ts`.

Overall the event creation / resolution flow is kind of disjointed and I want to take some time to clean it up post-release, so apologies in advance for any further onboarding pains related to that

test/server/cards/02_SHD/leaders/QiraIAloneSurvived.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.scoutBikePursuer, context.deathStarStormtrooper]);
+                context.player1.clickCard(context.scoutBikePursuer);
+
+                // check damage and shield are applied
+                expect(context.scoutBikePursuer.damage).toBe(2);
+                expect(context.scoutBikePursuer).toHaveExactUpgradeNames(['shield']);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should damage (and kill) a friendly unit with 3 power or less', function() {
```
AMMayberry1, line 31: This is definitely an instance where we could consolidate tests to reduce run time. Could we merge this one into the above by setting `context.qira.exhausted = false;` in between and just have a comment explaining the goal of each case. Unfortunately ugly but necessary for now

server/game/gameSystems/DamageSystem.ts
andtrobs: Don't worry about it, just gets me more familiar with the code. I'll take a look at this now.

PR #215, opened 2024-11-18
Ammayberry1/zone refactor

AMMayberry1:


server/game/core/card/Card.ts
```diff
@@ -434,17 +443,30 @@ export class Card extends OngoingEffectSource {
...
         if (originalZone === targetZone) {
             return;
         }
 
         this.cleanupBeforeMove(targetZone);
-        const prevZone = this._zoneName;
-        this._zoneName = targetZone;
-        this.initializeForCurrentZone(prevZone);
+
+        const prevZone = this._zone;
```
adbronico, line 457: Are there any effects/cards like Gar Saxon that would want us to store the previous zone on the card after it has moved?

server/game/core/card/Card.ts
```diff
@@ -455,11 +477,79 @@ export class Card extends OngoingEffectSource {
...
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    protected initializeForStartZone(): void {}
+
+    private addSelfToZone(zoneName: MoveZoneName) {
+        switch (zoneName) {
+            case ZoneName.Base:
+                this._zone = this.owner.baseZone;
+                Contract.assertTrue(this.isLeader());
```
adbronico, line 496: Would the base itself get into this block?

server/game/core/Player.js
```diff
@@ -159,7 +182,7 @@ class Player extends GameObject {
      * @param { WildcardZoneName.AnyArena | ZoneName.GroundArena | ZoneName.SpaceArena } arena Arena to select units from
      */
     getOtherUnitsInPlayWithTrait(ignoreUnit, trait, arena = WildcardZoneName.AnyArena) {
-        return this.getArenaCards(arena).filter((card) => card.isUnit() && card !== ignoreUnit && card.hasSomeTrait(trait));
+        return this.getArenaCards({ otherThan: ignoreUnit, trait, arena }).filter((card) => card.isUnit() && card !== ignoreUnit && card.hasSomeTrait(trait));
```
adbronico, line 185: Do we have any sort of intellisense for these filter properties? I saw them in the other file where they were declared, and this is how i wish the arena/zone filter/searching had worked, I just want to make sure it's obvious what we have available to us.

server/game/core/zone/ConcreteArenaZone.ts
```diff
@@ -0,0 +1,72 @@
+import { InPlayCard } from '../card/baseClasses/InPlayCard';
+import { Card } from '../card/Card';
+import { ZoneName } from '../Constants';
+import Game from '../Game';
+import Player from '../Player';
+import * as Contract from '../utils/Contract';
+import { ConcreteOrMetaArenaZone, IArenaZoneCardFilterProperties } from './ConcreteOrMetaArenaZone';
+import { IAddRemoveZone } from './ZoneAbstract';
+
+export abstract class ConcreteArenaZone extends ConcreteOrMetaArenaZone implements IAddRemoveZone {
```
adbronico, line 10: Can we add some docs on each of these classes describing what they hold/do? Doesn't have to be super in-depth, but just to explain the hierarchy a bit

server/game/gameSystems/DefeatCardSystem.ts
```diff
@@ -35,12 +36,11 @@ export class DefeatCardSystem<TContext extends AbilityContext = AbilityContext,
 
         if (event.card.isToken()) {
             // move the token out of the play area so that effect cleanup happens, then remove it from all card lists
-            event.card.owner.moveCard(event.card, ZoneName.OutsideTheGame, event.options || {});
-            event.context.game.removeTokenFromPlay(event.card);
+            event.card.moveTo(ZoneName.OutsideTheGame);
```
adbronico, line 39: Would it make any sense to put a .defeat() method on the card that moves itself based on type?

adbronico (changes requested):
Begun, the Zone Wars have

server/game/Interfaces.ts
```diff
@@ -157,6 +165,16 @@ export type IKeywordProperties =
 
 export type KeywordNameOrProperties = IKeywordProperties | NonParameterKeywordName;
 
+export type Zone =
+  | BaseZone
```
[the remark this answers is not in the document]
AMMayberry1: Yeah I think code-wise that would make more sense but the rules are pretty clear that both leader and base are in the same zone, so I want to stick with that

server/game/core/card/Card.ts
AMMayberry1: Yeah, for defeat purposes we'll need that. But that's kind of a separate task, haven't really figured out how that will be stored yet

server/game/core/card/Card.ts
AMMayberry1: With the current architecture the base is never actually moved, it is initialized already inside the base zone

adbronico (approved):
LGTM

PR #217, opened 2024-11-18
principled outlaw card and tests

thrazznos:


server/game/cards/02_SHD/units/PrincipledOutlaw.ts
```diff
@@ -0,0 +1,26 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '9040137775',
+            internalName: 'principled-outlaw'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'Choose a ground unit to exhaust',
```
AMMayberry1, line 15: nit: our style for ability titles here would be `'Exhaust a ground unit'`. We try not to include "choose", "you may", etc. since that will be implied by the presence of the corresponding prompts

thrazznos:
@AMMayberry1 fixed your comment, go ahead and merge if you are ready.

AMMayberry1 (approved):

PR #220, opened 2024-11-18
#Added Additional on-play tests

CheBato:
- All keywords except Smuggle (under test/core/abilities/keyword)
- two when-played ability, one on-attack ability (ideally something simple like draw a card)
- One constant ability like Snoke or Dodonna Something like Reputable Hunter that gives itself a play discount effect

- a generic "when defeated" ability test where we defeat, recall with Emperor's Legion, play, then defeat again
- a test for General Krell, there's already a comment note there. basically we want to start him in play, defeat, recall to hand, play again, and then confirm that friendly units draw exactly 1 card when they die (the concern is that they might gain two instances of his ability)

### DIFF
--- a/test/server/cards/01_SOR/units/GeneralKrellHeartlessTactician.spec.ts
+++ b/test/server/cards/01_SOR/units/GeneralKrellHeartlessTactician.spec.ts
@@ -98,7 +98,60 @@ describe('General Krell, Heartless Tactician', function() {
                 expect(context.player1.handSize).toBe(startingHandSize + 1);
             });
         });
+        describe('Krell\'s ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['the-emperors-legion'],
+                        groundArena: ['battlefield-marine', 'general-krell#heartless-tactician', 'reputable-hunter']
+                    },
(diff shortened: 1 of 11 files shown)


test/server/cards/01_SOR/units/GeneralKrellHeartlessTactician.spec.ts
```diff
@@ -98,7 +98,60 @@ describe('General Krell, Heartless Tactician', function() {
...
+                context.player1.clickCard(context.wampa);
+
+                expect(context.player1).toHavePrompt('Trigger the ability \'Draw a card\' or pass');
+                context.player1.clickPrompt('Draw a card');
+                expect(context.player2).toBeActivePlayer();
+
+                // its the same as the starting hand since we play 1 card
+                expect(context.player1.handSize).toBe(startingHandSize);
+
+                // CASE 2: we Waylay krell and return him to hand and play him again
```
AMMayberry1, line 140: For the tests of basic systems (as opposed to card) I think are the tests where I'm okay with having more cases, just because there aren't that many of them and they're so important so we want to know exactly which case failed quickly if something breaks. So for all of the tests in here except Krell I think we can break things out into individual cases a little more than we normally would.

AMMayberry1 (approved):

PR #222, opened 2024-11-18
feat: implement frontline shuttle

jgordijo:
image

### DIFF
--- a/test/server/cards/01_SOR/units/FrontlineShuttle.spec.ts
+++ b/test/server/cards/01_SOR/units/FrontlineShuttle.spec.ts
@@ -6,7 +6,7 @@ describe('Frontline Shuttle', function() {
                     phase: 'action',
                     player1: {
                         spaceArena: ['frontline-shuttle'],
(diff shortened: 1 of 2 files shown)


test/server/cards/01_SOR/units/FrontlineShuttle.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+                    player2: {
+                        groundArena: ['bail-organa#rebel-councilor', 'sundari-peacekeeper'],
+                    }
+                });
+            });
+
+            it('should be able to activate action ability and attack with exhausted unit', function () {
+                const { context } = contextRef;
+
+                context.battlefieldMarine.exhausted = true;
```
AMMayberry1, line 20: Can declare in this test setup like so:

```typescript
groundArena: [{ card: 'battlefield-marine', exhausted: true }, 'r2d2#ignoring-protocol', 'c3po#protocol-droid']
```

jgordijo: @AMMayberry1 Nice, I updated this test and will keep the same pattern in future implementations.

AMMayberry1 (approved):

PR #224, opened 2024-11-19
FrontierTrader/KihraxzHeavyFighter/RazorCrest

jan-preisig:
Implementation of Frontier Trader, Kihraxz Heavy Fighter and Razor Crest
image
image
image


server/game/cards/02_SHD/units/FrontierTrader.ts
```diff
@@ -0,0 +1,30 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'You may return a resource you control to your hand. If you do, you may put the top card of your deck into play as a resource.',
+            optional: true,
+            targetResolver: {
+                zoneFilter: ZoneName.Resource,
+                immediateEffect: AbilityHelper.immediateEffects.returnToHand()
```
AMMayberry1, line 19: Need to add `controller: RelativePlayer.Self` or else it might be possible to select the opponent's resources as well

test/server/cards/02_SHD/units/FrontierTrader.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+                    }
+                });
+            });
+
+            it('should be able to take resource back to hand and replace it with top card from deck', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.frontierTrader);
+                expect(context.player1).toHavePassAbilityButton();
+                context.player1.clickCard(context.collectionsStarhopper);
```
AMMayberry1, line 29: Any time we are creating a `targetResolver` we need to have at least one instance of `toBeAbleToSelectExactly()` to confirm that the targeting filters are set up correctly (in this case there's some concern that it might be able to target opponent's resources)

test/server/cards/02_SHD/units/FrontierTrader.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+                context.player1.clickPrompt('Put the top card of your deck into play as a resource.');
```
AMMayberry1, line 30: Similar note here - we should confirm that this is being presented as an optional prompt with `expect(context.player1).toHavePassAbilityPrompt('Put the top card of your deck into play as a resource.');`

server/game/cards/02_SHD/units/KihraxzHeavyFighter.ts
```diff
@@ -0,0 +1,34 @@
...
+            id: '4721657243',
+            internalName: 'kihraxz-heavy-fighter'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addOnAttackAbility({
+            title: 'You may exhaust another friendly unit. If you do, this unit gets +3/+0 for this attack.',
+            targetResolver: {
+                optional: true,
```
AMMayberry1, line 17: We try to put the `optional` as high as possible in the property hierarchy, in this case it can go directly on the ability (same level as `title` and `targetResolver`)

test/server/cards/02_SHD/units/KihraxzHeavyFighter.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+            });
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.kihraxzHeavyFighter);
+            expect(context.player1).toBeAbleToSelectExactly(['battlefield-marine', 'pyke-sentinel']);
+            context.player1.clickCard(context.battlefieldMarine);
+            expect(context.devastatorInescapable.damage).toBe(6);
+        });
+
+        it('should do normal attack if no unit can be exhausted', () => {
```
AMMayberry1, line 22: This test could be merged into the above by setting exhausted = false. See Rukh tests for a good example of how we set that up (we have to combine test cases because we need to be careful about our usage of Github Actions time).

Combining them also gives us the advantage of confirming that the +3 goes away after the first attack as expected.

test/server/cards/02_SHD/units/KihraxzHeavyFighter.spec.ts
```diff
@@ -0,0 +1,38 @@
+describe('Kihraxz Heavy Fighter', function () {
+    integration(function (contextRef) {
+        it('should gain +3 on attack if you exhaust another friendly unit and self is not selectable', function () {
```
AMMayberry1, line 3: nit: this and the Razor Crest tests are missing the `describe(). Even if the setupTest is in the individual cases, we still include the prefix "Kihraxz Heavy Fighter's ability"

server/game/cards/02_SHD/units/KihraxzHeavyFighter.ts
jan-preisig: Does this really make sense in this case? If I move it up, it won't auto skip if there are no units to tap

AMMayberry1 (approved):

PR #226, opened 2024-11-19
feat: implement punishing one

farisT:
image


server/game/cards/02_SHD/units/PunishingOne.ts
```diff
@@ -0,0 +1,30 @@
...
+
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'You may ready this unit. Use this ability only once each round.',
+            when: {
+                onCardDefeated: (event, context) =>
+                    event.card.isUnit() &&
+                    event.card.isInPlay() &&
+                    event.card.upgrades.length > 0 &&
```
AMMayberry1, line 20: nit: can use `event.card.isUpgraded()`

server/game/cards/02_SHD/units/PunishingOne.ts
```diff
@@ -0,0 +1,30 @@
...
+        return {
+            id: '8687233791',
+            internalName: 'punishing-one#dengars-jumpmaster',
+        };
+    }
+
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'You may ready this unit. Use this ability only once each round.',
```
AMMayberry1, line 15: nit: we try to keep ability titles to just the part that describes the effect since this may be displayed in a handful of different contexts. We typically don't include limits or the "you may" at the front since there will be a pass button that indicates that. So in this case just `Ready this unit` is good

test/server/cards/02_SHD/units/PunishingOne.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+                        hand: ['vanquish'],
+                        spaceArena: [{ card: 'punishing-one#dengars-jumpmaster', exhausted: true }],
+                    },
+                    player2: {
+                        groundArena: [{ card: 'wampa', upgrades: ['experience'] }],
+                    }
+                });
+            });
+
+            it('should be able to ready when an upgraded enemy unit is defeated', function () {
```
AMMayberry1, line 17: A few more test cases would be good to confirm we have the conditions set up right. Note that unfortunately, due to the fact that Jasmine tests are really heavy and we have finite Github Actions time in the free tier, we have to kind of chain tests together in single cases. See the Rukh tests for the typical pattern we use for this to keep it clean.

Below are the tests it would be good to extend this case with. Since the ability limit is once per round, you can use `context.moveToNextActionPhase()` in between to reset the limit.

- Does not ready when a non-upgraded enemy unit is defeated
- Does not ready when an upgraded friendly unit is defeated
- Test the ability limit by defeating a second upgraded enemy unit in the same phase after the ability has been triggered once already
- An interesting test would be to have Punishing One defeat an upgraded unit in combat, use its ability on itself, then confirm it is ready

AMMayberry1 (approved):

PR #227, opened 2024-11-19
restock

VJubert:


server/game/cards/01_SOR/events/Restock.ts
```diff
@@ -0,0 +1,39 @@
...
+            id: '6087834273',
+            internalName: 'restock'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Choose up to 4 cards in a discard pile. Put them on the bottom of their owner\'s deck in a random order',
+            targetResolver: {
+                mode: TargetMode.Select,
```
AMMayberry1, line 17: Eventually I think we will need a selector type that restricts the list of options based on what's already selected. For example, here we could just make all discard cards selectable and then once one is selected it would restrict the choices to be only cards in that same pile. Here it's optional, but I think we'll have to have it for Bold Resistance in set 3.

All that to say, could we add a TODO here for creating a conditional select type

server/game/cards/01_SOR/events/Restock.ts
```diff
@@ -0,0 +1,39 @@
...
+                        zoneFilter: ZoneName.Discard,
+                        controller: RelativePlayer.Self,
+                        innerSystem: AbilityHelper.immediateEffects.moveToBottomOfDeck()
+                    }),
+                    ['Opponent discard']: AbilityHelper.immediateEffects.selectCard({
+                        mode: TargetMode.UpTo,
+                        numCards: 4,
+                        zoneFilter: ZoneName.Discard,
+                        controller: RelativePlayer.Opponent,
+                        innerSystem: AbilityHelper.immediateEffects.moveToBottomOfDeck()
```
AMMayberry1, line 31: I'll raise this with Caz for FE stuff at some point soon, but I wonder if it would be good to have an extra option for this that makes it so that the player can see the order in which they selected cards. That way they can easily remember what order they'll be put in the bottom of the deck.

test/server/cards/01_SOR/events/Restock.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+                expect(context.player1).toHaveExactPromptButtons(['Your discard', 'Opponent discard']);
+                context.player1.clickPrompt('Your discard');
+
+                expect(context.player1).toBeAbleToSelectExactly([context.pykeSentinel, context.atst, context.battlefieldMarine, context.resupply, context.restock]);
+
+                // select cards
+                context.player1.clickCard(context.resupply);
+                context.player1.clickCard(context.restock);
+                context.player1.clickCard(context.battlefieldMarine);
+                context.player1.clickCard(context.atst);
```
AMMayberry1, line 32: We should confirm the 4 card restriction with `clickCardNonChecking()` and confirm that the fifth clicked card stays in the deck

server/game/cards/01_SOR/events/Restock.ts
VJubert: Restock need to be at random order

server/game/cards/01_SOR/events/Restock.ts
VJubert: Do you think it's a good idea ? That maybe could represent many cards and that could lead to user error if there is no sufficient indication between our card and opponent cards. I think it will be necessary for Bold Resolution but here I don't know when any players will select opponent discard

server/game/cards/01_SOR/events/Restock.ts
AMMayberry1: LOL yep thanks

server/game/cards/01_SOR/events/Restock.ts
AMMayberry1: Will continue this discussion in chat to unblock this PR

server/game/cards/01_SOR/events/Restock.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Choose up to 4 cards in a discard pile. Put them on the bottom of their owner\'s deck in a random order',
+            targetResolver: {
+                mode: TargetMode.UpTo,
+                numCards: 4,
+                zoneFilter: ZoneName.Discard,
+                controller: WildcardRelativePlayer.Any,
```
AMMayberry1, line 20: I think we need to add `canChooseNoCards: true` like in Medal Ceremony. To confirm, let's add a test case for selecting 0 cards

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -16,11 +25,12 @@ import * as Contract from '../core/utils/Contract';
  *
  * @property destination - The target zone for the card. Excludes discard pile, space arena, ground arena, and resources.
  * @property shuffle - Indicates whether the card should be shuffled into the destination.
- * @property bottom - Indicates whether the card should be placed at the bottom of the destination.
+ * @property shuffleTargets - Indicates whether all targets should be shuffled before added into the destination.
```
AMMayberry1, line 28: I think the person writing the card doesn't think in terms of targets, maybe this should be called shuffleCards or shuffleMovedCards?

server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -113,4 +123,11 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
         // Call the super implementation
         return super.canAffect(card, context, additionalProperties, mustChangeGameState);
     }
+
+    protected override processTargets(target: Card | Card[]) {
+        if (this.properties.shuffleTargets && Array.isArray(target)) {
+            shuffleArray(target);
```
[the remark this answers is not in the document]
VJubert: doc of shuffle array : /* Randomize array in-place using Durstenfeld shuffle algorithm */

AMMayberry1 (approved):

PR #228, opened 2024-11-19
defeat resource + entreprising lackeys + wrecker

VJubert:


test/server/cards/02_SHD/units/WreckerBoom.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+            it('should defeat a resource and deal 5 damage to a ground unit', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.wrecker);
+
+                // select a resource to defeat
+                expect(context.player1).toBeAbleToSelectExactly([context.wampa, context.battlefieldMarine, context.wildRancor, context.protector, context.devotion, context.restoredArc170]);
+                expect(context.player1).toHavePassAbilityButton();
+                expect(context.player1).toHaveChooseNoTargetButton();
+                context.player1.clickCard(context.devotion);
```
joshfullmer, line 29: Can we also do a resource count after this? Or something else that confirms that the resource has been destroyed?

server/game/gameSystems/DefeatCardSystem.ts
```diff
@@ -98,6 +94,8 @@ export class DefeatCardSystem<TContext extends AbilityContext = AbilityContext,
     }
 
     protected override updateEvent(event, card: Card, context: TContext, additionalProperties): void {
+        // if (card.zoneName !== ZoneName.Resource) {
```
AMMayberry1, line 97: I think the commented out code here is actually what we should do - `addLeavesPlayPropertiesToEvent` shouldn't be called on defeated resources

test/server/cards/02_SHD/units/EnterprisingLackeys.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+                expect(context.player1).toHavePassAbilityButton();
+                expect(context.player1).toHaveChooseNoTargetButton();
+
+                context.player1.clickCard(context.wampa);
+
+                // wampa should be defeated and lackeys should be in resource
+                expect(context.wampa).toBeInZone('discard');
+                expect(context.enterprisingLackeys).toBeInZone('resource');
+                expect(context.player1).toBeActivePlayer();
```
AMMayberry1, line 34: Frugghi just added logic so that if a resource is moved out of the resource zone when it is not exhausted, the game will attempt to exhaust it and ready another resource if possible. That way the player isn't losing a ready resource for stuff like this or Smuggle.

Can we test this here by exhausting one resource that isn't being defeated and confirm that the available resource count stays the same?

test/server/cards/02_SHD/units/WreckerBoom.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+                // select a ground unit to deal 5 damage
+                expect(context.player1).toBeAbleToSelectExactly([context.greedo, context.wrecker, context.atst]);
+                context.player1.clickCard(context.atst);
+
+                expect(context.atst.damage).toBe(5);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should deal 5 damage to a ground unit if we do not defeat a resource', function () {
```
AMMayberry1, line 39: nit: should _not_ deal 5 damage

server/game/gameSystems/DefeatCardSystem.ts
VJubert: on OnCardDefeated event is empty if I uncomment these code

AMMayberry1: Oh I know why. It's because `addLeavesPlayPropertiesToEvent` includes a call to `super.updateEvent` which is required. So we need to explicitly add that in the case where we don't call `addLeavesPlayPropertiesToEvent`

AMMayberry1 (approved):

PR #233, opened 2024-11-20
Delayed Effects (Zorii Bliss / Change of Heart) + Ownership Test Refactors

adbronico:
https://swudb.com/cards/SHD/203.png
https://swudb.com/cards/SOR/224.png


server/game/core/ongoingEffect/OngoingEffectEngine.ts
```diff
@@ -63,18 +63,19 @@ export class OngoingEffectEngine {
             const context = effect.context;
             const targets = effect.targets;
             return {
-                title: context.source.name + '\'s effect' + (targets.length === 1 ? ' on ' + targets[0].name : ''),
+                title: context.source.title + '\'s effect' + (targets.length === 1 ? ' on ' + targets[0].name : ''),
                 handler: () => {
-                    properties.gameAction.setDefaultTarget(() => targets);
-                    if (properties.message && properties.gameAction.hasLegalTarget(context)) {
+                    // TODO The below line was setting the target to the card originating the effect, even for player effects
```
AMMayberry1, line 68: Still true? I assumed this was fixed by switching to a player target effect

test/server/cards/02_SHD/units/ZoriiBlissValiantSmuggler.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+                const { context } = contextRef;
+
+                // Attack with Zorii and draw a card; create delayed discard
+                context.player1.clickCard(context.zoriiBliss);
+                expect(context.player1.hand.length).toBe(2);
+                expect(context.battlefieldMarine).toBeInZone('hand', context.player1);
+
+                // Pass to regroup phase
+                context.player2.claimInitiative();
+                context.player1.passAction();
```
AMMayberry1, line 25: Can use `context.moveToRegroupPhase()`

server/game/core/ongoingEffect/OngoingEffectEngine.ts
adbronico: @AMMayberry1 So it does still set the default target to Zorii Bliss but the DiscardFromHand system doesn't use it, so I'm curious how this would work with a card system vs. a player system

AMMayberry1: Ping me on this in chat, got a few questions. Specifically wondering how `effect.targets` is getting set, seems like there's a bug there if it's presenting a card target default for a player target system

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -0,0 +1,71 @@
...
+        const delayedEffect = AbilityHelper.ongoingEffects.delayedEffect({
+            title: 'Discard a card from your hand',
+            when: when,
+            immediateEffect: immediateEffect,
+            limit: limit
+        });
+
+        const renamedProperties = Object.assign(otherProperties, { ongoingEffect: delayedEffect });
+
+        event.context.source[duration](() => renamedProperties);
```
AMMayberry1, line 48: Looks like player target lasting effects and card target lasting effects do this differently. The card version is explicitly registering with the OngoingEffectEngine but the player version is doing this, which is calling a method on `OngoingEffectSource` based on the duration.

As an aside, probably this should be something more explicit like a switch since there are some values of Duration that don't have a corresponding method (and shouldn't), and also b/c this is impossible to read. Took me a while to piece together what it was doing.

test/server/cards/02_SHD/units/ZoriiBlissValiantSmuggler.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+                        hand: ['volunteer-soldier'],
+                        deck: ['battlefield-marine', 'wampa', 'pyke-sentinel']
+                    },
+                    player2: {
+                        hand: ['vanquish']
+                    }
+                });
+            });
+
+            it('draws a card on attack and discards a card at the start of the regroup phase even if zorii dies', function () {
```
AMMayberry1, line 18: Can we add one more case where the player has no cards to discard at the end of the phase and confirm that nothing breaks

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -0,0 +1,85 @@
...
+            limit: limit
+        });
+
+        const renamedProperties = Object.assign(otherProperties, { ongoingEffect: delayedEffect });
+
+        // If target is set, this targeting a card; otherwise, it is targeting a player
+        const delayedEffectTarget = event.context.target || event.context.source;
+
+        // Add more Durations as we implement
+        switch (duration) {
```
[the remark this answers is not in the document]
AMMayberry1: Actually I'm looking at the effect engine for something else and I think that the persistent case works fine as a lasting effect if we just use Duration.Persistent

server/game/cards/01_SOR/events/ChangeOfHeart.ts
```diff
@@ -0,0 +1,51 @@
...
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                immediateEffect: AbilityHelper.immediateEffects.simultaneous([
+                    AbilityHelper.immediateEffects.takeControlOfUnit((context) => ({
+                        newController: context.source.controller
+                    })),
+                    AbilityHelper.immediateEffects.delayedEffect((context) => ({
+                        when: {
+                            onPhaseStarted: (context) => context.phase === PhaseName.Regroup
+                        },
+                        duration: Duration.UntilEndOfRound,
```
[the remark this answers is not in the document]
adbronico: I've found that removing this and using the Persistent logic means the delayed effect never resolves. Any ideas?

server/game/gameSystems/DelayedEffectSystem.ts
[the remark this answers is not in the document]
adbronico: See above comment, but I'm having some issues with the method just using the duration. I think maybe we need it to be a constant ability after all?

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public override generatePropertiesFromContext(context: TContext, additionalProperties = {}) {
+        const properties = super.generatePropertiesFromContext(context, additionalProperties);
+        return properties;
+    }
+
+    public override hasLegalTarget(context: TContext, additionalProperties = {}): boolean {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        return properties.immediateEffect != null;
```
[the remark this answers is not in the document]
adbronico: If I do what you said here it starts skipping over the resolution of the delayed effect. I'm not 100% sure why

test/scenarios/constantEffectInteractions/SimultaneousDelayedEffects.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                        deck: ['battlefield-marine', 'wampa']
+                    },
+                    player2: {
+                        hand: ['vanquish'],
+                        groundArena: ['pyke-sentinel']
+                    }
+                });
+            });
+
+            it('player may choose order to resolve simultaneous delayed effects', function () {
```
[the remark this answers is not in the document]
adbronico: Possibly not. Should I remove the whole file or just comment it out with a TODO?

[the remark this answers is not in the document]
AMMayberry1: I think TODO is the better option since we know we'll want it eventually

test/helpers/DeckBuilder.js
```diff
@@ -35,6 +35,57 @@ class DeckBuilder {
         return cards;
     }
 
+    getOwnedCards(playerNumber, playerOptions, oppOptions, arena = null) {
+        let { groundArena, spaceArena, ...playerCards } = playerOptions;
+
+        let opponentAttachedUpgrades = [];
+
+        if ((arena === 'groundArena' || arena == null) && playerOptions.groundArena) {
+            const playerControlled = playerOptions.groundArena?.filter((card) => !card.hasOwnProperty('owner') && !card.owner?.endsWith(playerNumber));
```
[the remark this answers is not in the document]
adbronico: These aren't the real cards, these are the JSON representations of the cards defined in the test setup

server/game/core/ongoingEffect/OngoingEffectEngine.ts
```diff
@@ -49,7 +49,7 @@ export class OngoingEffectEngine {
             } else {
                 const triggeringEvents = events.filter((event) => properties.when[event.name]);
                 if (triggeringEvents.length > 0) {
-                    if (!properties.multipleTrigger && effect.duration !== Duration.Persistent) {
+                    if (properties.limit.isAtMax(effect.source.owner)) {
```
AMMayberry1, line 52: Shouldn't this check come after the limit is incremented?

server/game/core/ongoingEffect/OngoingEffectEngine.ts
```diff
@@ -93,12 +95,25 @@ export class OngoingEffectEngine {
 
     public removeLastingEffects(card: OngoingEffectSource) {
         this.unapplyAndRemove(
-            (effect) =>
-                effect.matchTarget === card &&
-                effect.duration !== Duration.Persistent
+            (effect) => this.shouldRemove(card, effect)
```
AMMayberry1, line 98: If this is the only place `shouldRemove()` is called we could maybe just make it a code block here? Not a strong preference, I think it works equally fine either way

server/game/core/ongoingEffect/OngoingEffectSource.ts
```diff
@@ -11,6 +11,22 @@ export class OngoingEffectSource extends GameObject {
...
+    /**
+     * Applies an effect which lasts until the end of the round.
+     */
+    public persistent(propertyFactory) {
+        const properties = propertyFactory(AbilityHelper);
+        this.addEffectToEngine(Object.assign({ duration: Duration.Persistent, zoneFilter: WildcardZoneName.Any }, properties));
+    }
+
+    /**
+     * Applies an effect which lasts until the end of the phase.
```
AMMayberry1, line 23: nit: says 'end of phase'

server/game/core/ongoingEffect/OngoingEffectSource.ts
```diff
@@ -11,6 +11,22 @@ export class OngoingEffectSource extends GameObject {
         super(game, name);
     }
 
+    /**
+     * Applies an effect which lasts until the end of the round.
```
AMMayberry1, line 15: nit: says 'end of the round'

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -24,10 +28,12 @@ export class DelayedEffectSystem<TContext extends AbilityContext = AbilityContex
     public override readonly effectDescription: string = 'apply a delayed effect';
 
     protected override defaultProperties: IDelayedEffectSystemProperties = {
+        title: '',
```
AMMayberry1, line 31: Might be better to just put it as `null` since that's what we typically check for as the 'empty value'

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -37,36 +43,38 @@ export class DelayedEffectSystem<TContext extends AbilityContext = AbilityContex
...
+        const renamedProperties = { ...otherProperties, ongoingEffect:
+            AbilityHelper.ongoingEffects.delayedEffect({
+                title,
+                when,
+                immediateEffect,
+                limit
+            }) };
 
-        const renamedProperties = Object.assign(otherProperties, { ongoingEffect: delayedEffect });
+        const delayedEffectTarget = properties.effectType === DelayedEffectType.Card ? event.context.target : event.context.source;
```
AMMayberry1, line 56: I realized that calling it `delayedEffectTarget` is a little misleading since it's really the card that is registering the ability, not the target of any effects. Maybe `delayedEffectSource`?

Also, this made me wonder - have we done any tests where the `delayedEffectTarget` unit gets defeated? I'm trying to understand why it's even relevant which card is registering the effect, I'm _pretty_ sure it's not because the effect goes away if the unit leaves the field but it's worth confirming that.

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -0,0 +1,93 @@
...
+export class DelayedEffectSystem<TContext extends AbilityContext = AbilityContext> extends GameSystem<TContext, IDelayedEffectSystemProperties> {
+    public override readonly name: string = 'applyDelayedEffect';
+    public override readonly eventName: EventName = EventName.OnEffectApplied;
+    public override readonly effectDescription: string = 'apply a delayed effect';
+
+    protected override defaultProperties: IDelayedEffectSystemProperties = {
+        title: '',
+        when: null,
+        duration: Duration.Persistent,
+        limit: AbilityHelper.limit.perGame(1),
```
AMMayberry1, line 34: We should avoid importing AbilityHelper outside of card implementations, it imports a huge amount of other files so it really increases the likelihood of import loops. We can directly import `PerGameAbilityLimit` and `OngoingEffectLibrary`

test/server/cards/02_SHD/units/ZoriiBlissValiantSmuggler.spec.ts
```diff
@@ -0,0 +1,103 @@
+describe('Zorii Bliss', function() {
+    integration(function(contextRef) {
+        describe('Zorii Bliss\'s ability', function() {
+            beforeEach(function () {
```
AMMayberry1, line 4: nit: if there's only one case for the setupTest then we can just move it inside the test case, see Zillo Beast. Unless you prefer it like this

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -37,36 +43,38 @@ export class DelayedEffectSystem<TContext extends AbilityContext = AbilityContex
...
+                delayedEffectTarget.untilEndOfPhase(() => renamedProperties);
+                break;
             case Duration.UntilEndOfRound:
-                // TODO - Is this how we want to implement these? If no duration is provided, what do we set here?
                 delayedEffectTarget.untilEndOfRound(() => renamedProperties);
                 break;
+            case Duration.Custom:
+                throw new Error(`Duration ${duration} not implemented yet`);
             default:
                 Contract.fail('Invalid Duration for DelayedEffect');
```
AMMayberry1, line 74: Can we include the enum value in the error message

test/helpers/PlayerInteractionWrapper.js
```diff
@@ -307,6 +320,15 @@ class PlayerInteractionWrapper {
         });
     }
 
+    attachOpponentOwnedUpgrades(opponentOwnedUpgrades = []) {
```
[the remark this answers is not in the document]
AMMayberry1: That's reasonable, there's a ton in this PR already. For the meantime though, how feasible would it be to rename the property from `owner` to `ownerAndController`? Just for upgrades

test/helpers/PlayerInteractionWrapper.js
```diff
@@ -309,6 +317,15 @@ class PlayerInteractionWrapper {
...
+        opponentOwnedUpgrades.forEach((upgrade) => {
+            const upgradeCard = this.findCardByName(upgrade.card, 'any', 'opponent');
```
AMMayberry1, line 322: I worry that using `findCardByName` in a lot of places will cause things to break if there are multiple cards in the field with the same name. Any reason we're using strings here instead of just passing around explicit references to the card objects like with some of the other test methods?

test/helpers/PlayerInteractionWrapper.js
```diff
@@ -309,6 +317,15 @@ class PlayerInteractionWrapper {
         });
     }
 
+    attachOpponentOwnedUpgrades(opponentOwnedUpgrades = []) {
+        opponentOwnedUpgrades.forEach((upgrade) => {
```
AMMayberry1, line 321: Any chance we could use an explicit `for()` loop, `forEach` is a headache to debug

test/helpers/PlayerInteractionWrapper.js
```diff
@@ -309,6 +317,15 @@ class PlayerInteractionWrapper {
...
+            const upgradeCard = this.findCardByName(upgrade.card, 'any', 'opponent');
+            const attachedCardAlsoOpponentControlled = upgrade.hasOwnProperty('attachedToOwner') && upgrade.attachedToOwner !== this.player.nameField;
```
AMMayberry1, line 323: Having trouble following - is the `attachedToOwner` property getting added to the card object itself, or is this some other wrapper object or something? I'd like to avoid modifying the card properties if possible

test/server/cards/02_SHD/leaders/FinnThisIsARescue.spec.ts
```diff
@@ -99,27 +86,26 @@ describe('Finn, This is a Rescue', function () {
                     }
                 };
 
+                // TODO - should we remove this comment now that ownership is handled in setup?
```
AMMayberry1, line 89: I think we can remove the comment yeah, it's unclear who it's even for since the tests are already done

server/game/core/ongoingEffect/OngoingEffectEngine.ts
adbronico: I think it could, but it would require some refactoring. I'm guessing that right now the effect is sitting around until the next regroup phase. What do you think?

server/game/core/ongoingEffect/OngoingEffectEngine.ts
adbronico: I was thinking it may need to be reused elsewhere like where the other check for the limit is. Do you think we could use this method in place of that if we move it down after the increment like you proposed?

server/game/gameSystems/DelayedEffectSystem.ts
adbronico: I think I need to add an export on PerGameAbilityLimit? It isn't currently referenceable I think

test/helpers/PlayerInteractionWrapper.js
adbronico: This is because the opponentOwnedUpgrades array is not actual cards but is instead just the JSON representation of the setup info. We have to map that to the actual card using findCardByName

test/helpers/PlayerInteractionWrapper.js
adbronico: 'upgrade' is the json version, 'upgradeCard' is the actual card

test/helpers/PlayerInteractionWrapper.js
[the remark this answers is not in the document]
adbronico: I can do that. We can split it back out later

server/game/core/ongoingEffect/OngoingEffectEngine.ts
AMMayberry1: I think it would be better to make sure it gets removed as soon as the limit is hit, (a) just to keep things clean for debugging purposes and (b) to prevent any weird potential bugs due to having the triggers still sitting around

server/game/gameSystems/DelayedEffectSystem.ts
AMMayberry1: Oh sorry, you can just import `perGame()` looks like

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        const renamedProperties = { ...otherProperties, ongoingEffect:
+            AbilityHelper.ongoingEffects.delayedEffect({
+                title,
+                when,
+                immediateEffect,
+                limit
+            }) };
+
+        const delayedEffectSource = properties.effectType === DelayedEffectType.Card ? event.context.target : event.context.source;
```
AMMayberry1, line 56: Was thinking on this some more, I think right now it will fail if used outside of a target resolver. Like if we just have it as an `immediateEffect` instead of `targetResolver` there will be no `event.context.target`.

I think what we actually need is to override `addPropertiesToEvent()`, determine the source card there, and attach it to the event as `event.sourceCard` or something. Then we would reference it here.

We may also want to just put this logic in a separate method so it could be called from other places like `hasLegalTarget`, see comment there

server/game/gameSystems/DelayedEffectSystem.ts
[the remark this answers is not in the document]
AMMayberry1: Took a look at this and yeah I think we can forget about the `canAffect()` idea, it won't work easily. So we can keep it roughly like this, except I would change the following:

- Have `hasLegalTarget` determine the source card (above comment suggests moving logic that to its own method for this reason) and validate that it's non-null
- If we're already overriding `addPropertiesToEvent()` as suggested above we can add the check for immediateEffect not being null there

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -0,0 +1,93 @@
...
+        title: null,
+        when: null,
+        duration: Duration.Persistent,
+        limit: AbilityHelper.limit.perGame(1),
+        immediateEffect: null,
+        effectType: null
+    };
+
+    public eventHandler(event: any, additionalProperties: any): void {
+        const properties = this.generatePropertiesFromContext(event.context, additionalProperties);
```
AMMayberry1, line 40: I missed this earlier - I realized that the design intent for these systems is that we shouldn't access properties directly here but instead put everything relevant on the event and then read from there when we get to eventHandler. I'm planning on slowly working my way through and fixing that in older systems.

Below comments suggest overriding addPropertiesToEvent anyway so we could just add everything needed there

server/game/gameSystems/DelayedEffectSystem.ts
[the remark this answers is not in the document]
adbronico: I've made all these changes except the first bullet point above - the hasLegalTarget method doesn't have the event as a parameter, and the new method needs it to get the source. What should I do there?

If I just remove the hasLegalTarget override, delayed effects don't work

server/game/gameSystems/DelayedEffectSystem.ts
adbronico: @AMMayberry1 I already had a test for this in Change of Heat, I think. LMK if you think an additional test is needed or what scenario we could use to further test this

server/game/gameSystems/DelayedEffectSystem.ts
adbronico: So we would want to move the switch logic here to that method? Would I even need the eventHandler method anymore?

AMMayberry1: Most of the logic can stay here, but anything you would read off of properties should instead be added to the event via `addPropertiesToEvent`

server/game/gameSystems/DelayedEffectSystem.ts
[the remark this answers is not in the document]
AMMayberry1: Ah fair. It's probably fine as is then.

server/game/gameSystems/DelayedEffectSystem.ts
AMMayberry1: Would be great to also do one with Zorii if we don't have it already, where she gets defeated before the end of phase effect happens

adbronico: Ah yeah that has been there since the beginning

server/game/gameSystems/DelayedEffectSystem.ts
```diff
@@ -43,18 +43,11 @@ export class DelayedEffectSystem<TContext extends AbilityContext = AbilityContex
...
-                title,
-                when,
-                immediateEffect,
-                limit
-            }) };
-
         const delayedEffectSource = event.sourceCard;
 
+        const renamedProperties = event.renamedProperties;
+        const duration = properties.duration;
```
AMMayberry1, line 49: Looks like we're still using `properties` here - the goal is to remove the `generatePropertiesFromContext` call completely and put everything relevant on the event itself

server/game/core/Game.js
```diff
@@ -26,7 +26,8 @@ const { cards } = require('../cards/Index.js');
 // const ConflictFlow = require('./gamesteps/conflict/conflictflow');
 // const MenuCommands = require('./MenuCommands');
 
-const { EventName, ZoneName, TokenName, Trait, WildcardZoneName } = require('./Constants.js');
+const { EventName, ZoneName, Trait, WildcardZoneName, TokenUpgradeName, TokenUnitName } = require('./Constants.js');
+const { BaseStepWithPipeline } = require('./gameSteps/BaseStepWithPipeline.js');
```
AMMayberry1, line 30: Is this import needed? Normally I wouldn't care but if the really basic classes import each other that creates more possibility for import loops

server/game/core/ongoingEffect/OngoingEffectEngine.ts
```diff
@@ -63,36 +60,64 @@ export class OngoingEffectEngine {
...
         }
     }
 
     public removeLastingEffects(card: OngoingEffectSource) {
         this.unapplyAndRemove(
-            (effect) =>
-                effect.matchTarget === card &&
-                effect.duration !== Duration.Persistent
+            (effect) => {
+                if (effect.duration !== Duration.Persistent) {
```
AMMayberry1, line 108: I think we might need to do the delayed effect check before this one - since this method is called when a unit leaves the field (or at least should be, currently there's a bug) I believe this check would cause any delayed effect with duration != persistent to be removed when the unit dies. So we should do the delayed effect check first if I'm reading it right.

AMMayberry1 (approved):

PR #234, opened 2024-11-20
death mark

thrazznos:


server/game/cards/02_SHD/upgrades/DeathMark.ts
```diff
@@ -0,0 +1,26 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addGainKeywordTargetingAttached({
+            keyword: KeywordName.Bounty,
+            ability: {
+                title: 'Draw 2 cards',
+                immediateEffect: AbilityHelper.immediateEffects.draw({
+                    amount: 2
```
AMMayberry1, line 19: nit: since this has only one parameter you could do it all in one line. personal preference tho

PR #237, opened 2024-11-20
Add Palpatine's Return and You're My Only Hope implementation

Frugghi:
| Palpatine's Return | You're My Only Hope |
| --- | ---- |
| image | image |


server/game/cards/01_SOR/events/YoureMyOnlyHope.ts
```diff
@@ -0,0 +1,40 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Look at the top card of your deck',
+            immediateEffect: AbilityHelper.immediateEffects.lookAt((context) => ({
+                target: context.source.controller.getTopCardOfDeck()
+            })),
+            then: (thenContext) => ({
```
AMMayberry1, line 19: It works either way, but we can do a slight optimization here by using `ifYouDo: (ifYouDoContext)` instead of `then` in which case it will shortcut evaluating the stuff inside the `then` if the deck is empty because the lookAt didn't happen

server/game/cards/01_SOR/events/YoureMyOnlyHope.ts
```diff
@@ -0,0 +1,40 @@
...
+                title: thenContext.source.controller.base.remainingHp <= 5
+                    ? 'Play it for free'
+                    : 'Play it, it costs 5 less',
```
AMMayberry1, line 22: Might be good to insert the name of the card into the title to help the player remember the card they saw

test/server/cards/02_SHD/events/PalpatinesReturn.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+                    context.player2.passAction();
+                };
+
+                // Scenario 1: Play a force unit for 8 less
+                context.player1.clickCard(context.palpatinesReturn);
+                expect(context.player1).toBeAbleToSelectExactly([context.supremeLeaderSnoke, context.avenger]);
+
+                let exhaustedResourcesBeforeAction = context.player1.exhaustedResourceCount;
+                context.player1.clickCard(context.supremeLeaderSnoke);
+                expect(context.player1.exhaustedResourceCount).toBe(exhaustedResourcesBeforeAction);
```
AMMayberry1, line 33: Could we play an out-of-aspect unit so we can get the cost above 8 and confirm the exact amount of the discount?

test/server/cards/01_SOR/events/YoureMyOnlyHope.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                // Scenario 1: Do nothing
+                context.player1.clickCard(context.youreMyOnlyHope);
+                // TODO: we need a 'look at' prompt for secretly revealing, currently chat logs go to all players
+                expect(context.getChatLogs(1)).toContain('You\'re My Only Hope sees AT-ST');
+                expect(context.player1).toHavePassAbilityPrompt('Play AT-ST, it costs 5 less');
+
+                const deckBeforeAction = context.player1.deck;
+                context.player1.passAction();
+                expect(context.player1.deck).toEqual(deckBeforeAction);
+                expect(context.atst).not.toBeInZone('groundArena', context.player1);
```
adbronico, line 36: I think it would be easier to check that the ATST is in the deck and remove the deckbeforeAction completely

AMMayberry1 (dismissed):

AMMayberry1 (approved):

PR #239, opened 2024-11-21
#199 Coruscant Dissident implementation

sethhelman:


server/game/cards/02_SHD/units/CoruscantDissident.ts
```diff
@@ -0,0 +1,25 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '9115773123',
+            internalName: 'coruscant-dissident'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'On attack: You may ready a resource',
```
VJubert, line 15: on title you can omit "may" part as he will have the pass ability button and "on attack"

sethhelman: I have updated this to say 'Ready a resource'. Please let me know if you want that worded differently and if there are any other issues with the implementation code

AMMayberry1 (approved):

PR #240, opened 2024-11-21
SHD# 233 Evacuate implementation

brianries:
Added evacuate logic and unit test.  Made sure upgrades get defeated/discarded while units return to hand

### DIFF
--- a/test/server/cards/02_SHD/events/Evacuate.spec.ts
+++ b/test/server/cards/02_SHD/events/Evacuate.spec.ts
@@ -37,6 +37,6 @@ describe('Evacuate', function() {
 
                 // TODO CAPTURE: test that captured units are returned to the arena and not to hand
             });
-        }); 
+        });
     });
 });
--- a/test/server/cards/02_SHD/events/Evacuate.spec.ts
+++ b/test/server/cards/02_SHD/events/Evacuate.spec.ts
@@ -1,7 +1,7 @@
 describe('Evacuate', function() {
     integration(function(contextRef) {
         describe('Evacuate\'s ability', function() {
-            it('should return all units cards to hand', function () {
+            it('should return all unit cards to hand and discard upgrades and ignore deployed leaders', function () {
                 contextRef.setupTest({
                     phase: 'action',
                     player1: {
@@ -13,8 +13,8 @@ describe('Evacuate', function() {
                     player2: {
                         hand: ['death-star-stormtrooper'],
                         groundArena: ['wampa'],
-                        spaceArena: ['imperial-interceptor'],
-                        leader: { card: 'grand-moff-tarkin#oversector-governor', deployed: false },
+                        spaceArena: [{ card: 'imperial-interceptor', upgrades: ['academy-training'] }],
+                        leader: { card: 'grand-moff-tarkin#oversector-governor', deployed: true },
                     }
                 });
                 const { context } = contextRef;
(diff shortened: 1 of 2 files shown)


test/server/cards/02_SHD/events/Evacuate.spec.ts
```diff
@@ -0,0 +1,104 @@
+describe('Evacuate', function() {
+    integration(function(contextRef) {
+        it('returning all units cards, no leaders', function () {
```
jan-preisig, line 3: this is missing the inner describe(). Even if the setupTest is in the individual cases, we still include the prefix "Evacuate's ability"
I know its missing in a few tests, but I think it's a general best practice.
Also the text in the it clause shoud form a sentence with the text from the describe clause. So I recommend to changing it to something like ```should return all unit cards, except leaders to hand```
AT-ATSuppressor.spec.ts has a great example on this

test/server/cards/02_SHD/events/Evacuate.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+            expect(context.player1.groundArena).toHaveSize(0);
+
+            expect(context.player2.hand).toHaveSize(3);
+            expect(context.player2.hand).toContain(context.deathStarStormtrooper);
+            expect(context.player2.hand).toContain(context.wampa);
+            expect(context.player2.hand).toContain(context.imperialInterceptor);
+            expect(context.player1.groundArena).toHaveSize(0);
+        });
+
+        it('return all unit cards to hand and discard upgrades', function () {
```
jan-preisig, line 35: To save a lot of duplicate code, you could probably merge test 2 and 3 together or at least reduce the code a bit so you don't have to check so many things. So if you don't need the leader for the test, I wouldn't have him in the test setup.
Also would only have one friendly and one enemy unit in this test case, because you already check that multiple units get bounced with the first test case.

jan-preisig (changes requested):
Nice work on the Event Ability. I left some suggestions for the tests :)

test/server/cards/02_SHD/events/Evacuate.spec.ts
brianries: collapsed test 3 into test 2, removed a few cards to keep it smaller

test/server/cards/02_SHD/events/Evacuate.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+                expect(context.player1.groundArena).toHaveSize(0);
+
+                expect(context.player2.hand).toHaveSize(3);
+                expect(context.player2.hand).toContain(context.deathStarStormtrooper);
+                expect(context.player2.hand).toContain(context.wampa);
+                expect(context.player2.hand).toContain(context.imperialInterceptor);
+                expect(context.player1.groundArena).toHaveSize(0);
+            });
+
+            it('should return all unit cards to hand and discard upgrades and ignore deployed leaders', function () {
```
AMMayberry1, line 36: These tests are great! One request though, unfortunately we have a practical issue where the Jasmine tests are slow and heavy and we have limited Github Actions time in the free tier, so as a consequence we have to try and consolidate our tests as much as possible (though it's not great practice for sure). We plan to address it some time in the future.

For now though, I think these two test cases could be consolidated into one by just adding the upgrades and deployed leaders to the above case.

test/server/cards/02_SHD/events/Evacuate.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+                expect(context.player1.groundArena).toHaveSize(1);
+
+                expect(context.player2.hand).toHaveSize(2);
+                expect(context.player2.hand).toContain(context.deathStarStormtrooper);
+                expect(context.player2.hand).toContain(context.imperialInterceptor);
+                expect(context.academyTraining).toBeInZone('discard');
+                expect(context.grandMoffTarkin).toBeInZone('groundArena');
+                expect(context.player1.groundArena).toHaveSize(1);
+            });
+        });
```
AMMayberry1, line 66: Could you do me a small favor and add this comment to remind us to revisit this for Capture:

`// TODO CAPTURE: test that captured units are returned to the arena and not to hand`

brianries: Is this implemented yet? captured? I can add to the test, but I don't think I saw that under the leaves play handler, I'm guessing thats pending?

test/server/cards/02_SHD/events/Evacuate.spec.ts
brianries: consolidated the test, actually caught a bug too, I checked player1 ground arena account twice

test/server/cards/02_SHD/events/Evacuate.spec.ts
brianries: Added comment

AMMayberry1 (approved):

PR #244, opened 2024-11-21
Migs Mayfeld - Triggerman implementation and tests

oaj2011:


server/game/cards/02_SHD/units/MigsMayfeldTriggerman.ts
```diff
@@ -0,0 +1,29 @@
...
+        this.addTriggeredAbility({
+            title: 'You may deal 2 damage to a unit or base.',
+            limit: AbilityHelper.limit.perRound(1),
+            when: {
+                onCardsDiscardedFromHand: (event, context) => true,
+            },
+            targetResolver: {
+                optional: true,
+                immediateEffect: AbilityHelper.immediateEffects.damage(() => ({
+                    amount: 2,
```
AMMayberry1, line 22: nit: since there's just one parameter it would be fine all in one line, but personal preference

server/game/cards/02_SHD/units/MigsMayfeldTriggerman.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'You may deal 2 damage to a unit or base.',
+            limit: AbilityHelper.limit.perRound(1),
+            when: {
+                onCardsDiscardedFromHand: (event, context) => true,
+            },
+            targetResolver: {
+                optional: true,
```
AMMayberry1, line 20: We try to keep the `optional` at the ability level unless it only applies to one effect in a longer sequence (note that this will change the name of the pass button from 'choose no targets' to 'pass ability' which is relevant for which expectation you use; see comment below)

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+                context.player1.clickPrompt('Opponent');
+                context.player2.clickCard(context.confiscate);
+                context.player2.clickCard(context.daringRaid);
+                context.player2.clickPrompt('Done');
+                expect(context.player1).toBeAbleToSelectExactly([context.groguIrresistible, context.migsMayfeldTriggerman, context.wampa, context.systemPatrolCraft, context.chopperBase, context.commandCenter]);
+                context.player1.clickCard(context.chopperBase);
+                expect(context.chopperBase.damage).toBe(2);
+                expect(context.player2).toBeActivePlayer();
+
+                // CASE 2: Cannot use the ability twice in a round
```
AMMayberry1, line 35: Can we add a third case confirming that it's usable in the next round

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+                        base: 'chopper-base'
+                    }
+                });
+            });
+
+            it('can deal two damage to a unit or base after a card is discarded from hand', function () {
+                const { context } = contextRef;
+
+                // CASE 1: Can deal two damage to a unit or base after a card is discarded from hand
+                context.player1.clickCard(context.pillage);
```
AMMayberry1, line 25: We have No Bargain implemented which would simplify the setup here a bit and make it a little easier to parse what's going on. Though if we add the third case for next round (see below comment) then it would be good to keep pillage and target the controlling player instead to confirm that the ability can trigger for either player

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
```diff
@@ -0,0 +1,44 @@
+describe('Migs Mayfeld, Triggerman', function() {
+    integration(function(contextRef) {
+        describe('Migs Mayfeld\'s ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['pillage', 'forced-surrender'],
+                        groundArena: [{ card: 'grogu#irresistible' }, { card: 'migs-mayfeld#triggerman' }],
```
AMMayberry1, line 9: If there's no `exhausted`, `upgrades`, etc can we just do the string name instead of an object with a `card` property. just to keep the test setup as clean as possible

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+            it('can deal two damage to a unit or base after a card is discarded from hand', function () {
+                const { context } = contextRef;
+
+                // CASE 1: Can deal two damage to a unit or base after a card is discarded from hand
+                context.player1.clickCard(context.pillage);
+                context.player1.clickPrompt('Opponent');
+                context.player2.clickCard(context.confiscate);
+                context.player2.clickCard(context.daringRaid);
+                context.player2.clickPrompt('Done');
+                expect(context.player1).toBeAbleToSelectExactly([context.groguIrresistible, context.migsMayfeldTriggerman, context.wampa, context.systemPatrolCraft, context.chopperBase, context.commandCenter]);
```
AMMayberry1, line 30: We have helpers named `context.p1Base` and `context.p2Base` you can use. That way you can remove the explicit base declarations, we typically only use those if we're testing something related to aspect penalty

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+                context.player1.clickCard(context.chopperBase);
```
AMMayberry1, line 31: Since the ability is optional, before clicking we should have `expect(context.player1).toHavePassAbilityButton();`

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['pillage', 'forced-surrender'],
+                        groundArena: [{ card: 'grogu#irresistible' }, { card: 'migs-mayfeld#triggerman' }],
+                        base: 'command-center'
+                    },
+                    player2: {
+                        hand: ['confiscate', 'daring-raid', 'krayt-dragon'],
+                        groundArena: [{ card: 'wampa', upgrades: ['shield'] }],
```
AMMayberry1, line 14: Any reason for this to have a shield?

server/game/cards/02_SHD/units/MigsMayfeldTriggerman.ts
oaj2011: I come from the land of C# where separate lines are The Way. I'll move it to one line to meet js style norms.

server/game/cards/02_SHD/units/MigsMayfeldTriggerman.ts
oaj2011: Do you know of an example off the top where it is correct to have Optional inside targetResolver? Is Krayt Dragon correct in having it inside targetResolver? I saw in both ways, and I used Krayt Dragon as the template for Migs. In any event, moved up to the ability level.

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
oaj2011: Yes we can!

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
oaj2011: I added No Bargain for the third discard case. Once we have Force Throw, I'll use that instead of Forced Surrender or Pillage so the test case looks cleaner.

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
oaj2011: Got it!

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
oaj2011: Got it!

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
oaj2011: Got it!

test/server/cards/02_SHD/units/MigsMayfeldTriggerman.spec.ts
oaj2011: I plan to test Force Throw (when implemented) with Migs and ensure that the Force Throw damage hits before the Migs trigger. Force Throw Krayt Dragon into shield should pop shield and then Migs trigger should leave Wampa with just 2 damage.

server/game/cards/02_SHD/units/MigsMayfeldTriggerman.ts
AMMayberry1: It would have to be a situation where one effect in an ability is optional but the others aren't. We haven't been 100% consistent about this rule in the past but we're trying to now, if you spot other abilities that don't do it this way feel free to update them

AMMayberry1 (approved):

PR #245, opened 2024-11-22
SHD# Kintan Intimidator card

c0rrupt3d-d4t4:
kintan-intimidator

### DIFF
--- a/server/game/cards/02_SHD/units/KintanIntimidator.ts
+++ b/server/game/cards/02_SHD/units/KintanIntimidator.ts
@@ -1,0 +1,22 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+
+export default class KintanIntimidator extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '9725921907',
+            internalName: 'kintan-intimidator',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'Exhaust the defender when attacking a unit.',
+            immediateEffect: AbilityHelper.immediateEffects.exhaust((context) => {
+                return { target: context.source.activeAttack?.target };
+            })
+        });
+    }
+}
+
+KintanIntimidator.implemented = true;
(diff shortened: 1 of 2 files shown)


server/game/cards/02_SHD/units/KintanIntimidator.ts
```diff
@@ -0,0 +1,22 @@
...
+            id: '9725921907',
+            internalName: 'kintan-intimidator',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'Exhaust the defender when attacking a unit.',
+            immediateEffect: AbilityHelper.immediateEffects.exhaust((context) => {
+                return { target: context.source.activeAttack?.target };
```
AMMayberry1, line 16: Instead of using return, you can write the effect like this:

```typescript
immediateEffect: AbilityHelper.immediateEffects.exhaust((context) => ({ target: context.source.activeAttack?.target }))
```

test/server/cards/02_SHD/units/KintanIntimidator.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+                        groundArena: ['battlefield-marine']
+                    }
+                });
+            });
+
+            it('should exhaust the defender when attacking a unit', function () {
+                const { context } = contextRef;
+
+                // Select Kintan Intimidator
+                expect(context.player1).toBeAbleToSelectExactly([context.kintanIntimidator]);
```
AMMayberry1, line 20: We should only use `expect` if we're testing the card ability. This check and the check on line 22 are validating basic stuff like attack that we don't need to check here, it's checked in other places.

server/game/cards/02_SHD/units/KintanIntimidator.ts
```diff
@@ -0,0 +1,20 @@
...
+        return {
+            id: '9725921907',
+            internalName: 'kintan-intimidator',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'Exhaust the defender when attacking a unit.',
+            immediateEffect: AbilityHelper.immediateEffects.exhaust((context) => ({ target: context.source.activeAttack?.target }))
```
VJubert, line 15: what happened when you attack a base ?

VJubert (approved):

AMMayberry1 (approved):

PR #246, opened 2024-11-22
han solo 2

VJubert:
013-portrait

### DIFF
--- a/test/server/cards/02_SHD/leaders/HanSoloWorthTheRisk.spec.ts
+++ b/test/server/cards/02_SHD/leaders/HanSoloWorthTheRisk.spec.ts
@@ -28,7 +28,7 @@ describe('Han Solo, Worth the Risk', function () {
                 context.player2.passAction();
 
                 context.player1.clickCard(context.hanSolo);
-                // play vanguard infantry automatically
+                context.player1.clickCard(context.vanguardInfantry)
 
                 // choose which unit to give an experience token (vanguard infantry when defeated ability)
                 expect(context.player1).toBeAbleToSelectExactly([context.cantinaBraggart, context.colonelYularen]);
@@ -44,7 +44,7 @@ describe('Han Solo, Worth the Risk', function () {
         });
 
         describe('Han Solo\'s leader deployed ability', function () {
-            beforeEach(function () {
+            it('should play a unit from our hand, it costs 1 resource less and take 2 damage', function () {
                 contextRef.setupTest({
                     phase: 'action',
                     player1: {
@@ -54,11 +54,8 @@ describe('Han Solo, Worth the Risk', function () {
                         base: { card: 'echo-base', damage: 5 },
                     },
                 });
-            });
 
-            it('should play a unit from our hand, it costs 1 resource less and take 2 damage', function () {
                 const { context } = contextRef;
-
 
                 context.player1.clickCard(context.hanSolo);
                 context.player1.clickPrompt('Attack');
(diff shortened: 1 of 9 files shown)


test/server/cards/02_SHD/leaders/HanSoloWorthTheRisk.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                expect(context.player1.exhaustedResourceCount).toBe(0);
+                expect(context.cantinaBraggart).toBeInZone('groundArena');
+                expect(context.cantinaBraggart.damage).toBe(2);
+                expect(context.hanSolo.exhausted).toBeTrue();
+
+                context.hanSolo.exhausted = false;
+                context.player2.passAction();
+
+                context.player1.clickCard(context.hanSolo);
+                // play automatically twin pod cloud car
```
AMMayberry1, line 31: This is a good ability to confirm some window stuff. Could we set it up so that there is both a when played and when defeated ability that is triggerd by this? Looks like there's not a 1 or 2 hp unit that has both when played and when defeated, so maybe we need to use like Yularen or Gideon Hask or something.

test/server/cards/02_SHD/leaders/HanSoloWorthTheRisk.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                context.player1.clickCard(context.cantinaBraggart);
+
+                expect(context.player1.exhaustedResourceCount).toBe(0);
+                expect(context.cantinaBraggart).toBeInZone('groundArena');
+                expect(context.cantinaBraggart.damage).toBe(2);
+
+                context.player2.passAction();
+
+                context.player1.clickCard(context.hanSolo);
+                // play automatically twin pod cloud car
```
AMMayberry1, line 68: This ability is another good test case because it's one of the few that is an action without a cost. At the end of this test when there are no cards in hand, can you confirm that Han has no ability available to activate? Want to make sure that players can't pseudo-pass by using his ability repeatedly with no cards in hand

test/server/cards/02_SHD/leaders/HanSoloWorthTheRisk.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+                expect(context.hanSolo.exhausted).toBeTrue();
+
+                context.hanSolo.exhausted = false;
+                context.player2.passAction();
+
+                context.player1.clickCard(context.hanSolo);
+                context.player1.clickCard(context.vanguardInfantry);
+
+                // choose which unit to give an experience token (vanguard infantry when defeated ability)
+                expect(context.player1).toBeAbleToSelectExactly([context.cantinaBraggart, context.colonelYularen]);
```
AMMayberry1, line 34: I'm pretty sure this is incorrect due to an open issue with PlayCardSystem. If both triggered abilities from Vanguard Infantry and Yularen go into the same window, there's supposed to be a prompt asking which one to do first (see TriggeredAbilityWindow.spec.ts). If they're happening automatically in a certain order, that means they're not going into the same trigger window like they're supposed to.

Probably what's happening is that the play card ability is resolving as if it were nested, so it's resolving its triggers immediately instead of at the end of the card ability. Ennis I think is planning to correctly add that feature sometime soon, probably we should ask them in the discord what their plan is and go from there. We probably can leave this as a TODO for Ennis to finish.

test/server/cards/02_SHD/leaders/HanSoloWorthTheRisk.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+                context.player2.passAction();
+                context.player1.clickCard(context.greenSquadronAwing);
+                context.player2.passAction();
+
+                // hand is empty, can not use han solo ability to soft pass
+                expect(context.hanSolo).not.toHaveAvailableActionWhenClickedBy(context.player1);
+                expect(context.player1).toBeActivePlayer();
+            });
+
+            it('should play a unit from our hand, it costs 1 resource less and take 2 damage', function () {
```
AMMayberry1, line 97: Looks like this test has the same name as the above. Also, what is it testing for that's different from the previous test? Having trouble figuring it out

AMMayberry1 (approved):

PR #247, opened 2024-11-22
Feature/saw gerrera

VJubert:


server/game/core/ability/PlayCardAction.ts
```diff
@@ -93,6 +93,19 @@ export abstract class PlayCardAction extends PlayerAction {
...
 
+    public override getCosts(context, playCosts = true) {
+        const costs = super.getCosts(context, playCosts);
+        if (!context.subResolution && playCosts && context.player.hasOngoingEffect(EffectName.AdditionalPlayCost)) {
+            const additionalPlayCosts = context.player
+                .getOngoingEffectValues(EffectName.AdditionalPlayCost)
+                .map((effect) => effect(context))
+                // filter out any undefined or null cost
+                .filter((cost) => cost);
+            return costs.concat(...additionalPlayCosts);
```
AMMayberry1, line 104: Don't need the `...` here I think, concat will handle the array naturally

test/server/cards/01_SOR/units/SawGerreraExtremist.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+            context.player1.passAction();
+
+            // opponent play a unit : nothing happen
+            context.player2.clickCard(context.battlefieldMarine);
+            expect(context.battlefieldMarine).toBeInZone('groundArena');
+            expect(context.p2Base.damage).toBe(0);
+
+            context.player1.passAction();
+
+            // opponent play a unit and kill saw gerrera : 2 damage on base
```
AMMayberry1, line 42: nit, typo: play an _event_

test/server/cards/01_SOR/units/SawGerreraExtremist.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+                player2: {
+                    hand: ['resupply', 'vanquish', 'battlefield-marine'],
+                    resources: ['smugglers-aid', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst', 'atst'],
+                    hasInitiative: true,
+                }
+            });
+            const { context } = contextRef;
+
+            // opponent play an event : 2 damage on base
+            context.player2.clickCard(context.resupply);
```
AMMayberry1, line 19: Could we replace this with an event that has a sub-ability step like `ifYouDo` or `then` to confirm that it's not somehow triggering on each CardAbilityStep instead of the whole CardAbility

server/game/core/ability/CardAbility.ts
```diff
@@ -113,12 +113,6 @@ export class CardAbility extends CardAbilityStep {
                 .map((effect) => effect(context));
```
AMMayberry1, line 113: Does anything break if we just removed this method now? I think the other stuff it doesn't is legacy l5r rules that we don't have in swu, so if we can remove it that would increase the clarity of the code overall. Sorry, I meant to ask for you to investigate that as part of cleaning up the parameters of `getCosts` in the other class but I wasn't clear

AMMayberry1 (approved):

PR #255, opened 2024-11-22
Implemented Don't Get Cocky

Ennis0:

### DIFF
--- a/server/game/cards/01_SOR/events/DontGetCocky.ts
+++ b/server/game/cards/01_SOR/events/DontGetCocky.ts
@@ -1,7 +1,7 @@
 import AbilityHelper from '../../../AbilityHelper';
 import { AbilityContext } from '../../../core/ability/AbilityContext';
 import { EventCard } from '../../../core/card/EventCard';
-import { TargetMode, WildcardCardType } from '../../../core/Constants';
+import { GameStateChangeRequired, TargetMode, WildcardCardType } from '../../../core/Constants';
 import { ISelectTargetResolver } from '../../../TargetInterfaces';
 import { GameSystem } from '../../../core/gameSystem/GameSystem';
 import { IThenAbilityPropsWithSystems } from '../../../Interfaces';
@@ -20,7 +20,7 @@ export default class DontGetCocky extends EventCard {
             title: 'Choose a unit',
             targetResolver: {
                 cardTypeFilter: WildcardCardType.Unit,
-                mustChangeGameState: false
+                mustChangeGameState: GameStateChangeRequired.None
             },
             then: (unitChosenContext) => ({
                 title: 'Reveal the top card of your deck',
--- a/server/game/cards/01_SOR/events/DontGetCocky.ts
+++ b/server/game/cards/01_SOR/events/DontGetCocky.ts
@@ -2,7 +2,6 @@ import AbilityHelper from '../../../AbilityHelper';
 import { AbilityContext } from '../../../core/ability/AbilityContext';
 import { EventCard } from '../../../core/card/EventCard';
 import { GameStateChangeRequired, TargetMode, WildcardCardType } from '../../../core/Constants';
-import { ISelectTargetResolver } from '../../../TargetInterfaces';
 import { GameSystem } from '../../../core/gameSystem/GameSystem';
 import { IThenAbilityPropsWithSystems } from '../../../Interfaces';
 import * as Contract from '../../../core/utils/Contract';
--- a/server/game/cards/01_SOR/events/DontGetCocky.ts
+++ b/server/game/cards/01_SOR/events/DontGetCocky.ts
@@ -1,0 +1,100 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { AbilityContext } from '../../../core/ability/AbilityContext';
+import { EventCard } from '../../../core/card/EventCard';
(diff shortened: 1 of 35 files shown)


server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -87,11 +87,12 @@ export class CardTargetResolver extends TargetResolver<ICardTargetResolver<Abili
         // (unless there are dependent targets that might care about the targeting result)
         if (
             !this.dependentTarget &&
+            this.properties.mustChangeGameState !== GameStateChangeRequired.None &&
```
AMMayberry1, line 90: If possible I'd like to find a solution that doesn't require explicitly setting `mustChangeGameState` in the target resolver properties like in DontGetCocky.ts, that property was mostly intended for internal engine use and is pretty technical so most people probably wouldn't understand how and when to set it. Could we get around this by just skipping this check if `this.getGameSystems()` is empty (i.e. no immediate effect in the target resolver)?

server/game/cards/01_SOR/events/DontGetCocky.ts
```diff
@@ -0,0 +1,74 @@
...
+                title: 'Reveal the top card of your deck',
+                immediateEffect: AbilityHelper.immediateEffects.reveal((context) => ({ target: context.source.controller.getTopCardOfDeck() })),
+                then: this.thenAfterReveal(1, unitChosenContext)
+            })
+        });
+    }
+
+    private thenAfterReveal(cardsRevealedCount: number, contextWithUnitTarget: AbilityContext): IThenAbilityPropsWithSystems<AbilityContext> {
+        Contract.assertTrue(cardsRevealedCount > 0 && cardsRevealedCount < 8, `Error in Don\'t Get Cocky implementation: thenAfterReveal called with invalid cardsRevealedCount ${cardsRevealedCount}`);
+        if (cardsRevealedCount < 7) {
```
AMMayberry1, line 34: nit: we can reduce the nesting a little and make this easier to read by moving up the final return:

```typescript
private thenAfterReveal(cardsRevealedCount: number, contextWithUnitTarget: AbilityContext): IThenAbilityPropsWithSystems<AbilityContext> {
  Contract.assertTrue(cardsRevealedCount > 0 && cardsRevealedCount < 8, `Error in Don\'t Get Cocky implementation: thenAfterReveal called with invalid cardsRevealedCount ${cardsRevealedCount}`);

  if (cardsRevealedCount === 7) {
      return {
              title: 'Deal damage equal to the chosen unit equal to the total cost of cards revealed, if it is 7 or less',
              immediateEffect: this.afterStopRevealingEffect(7, contextWithUnitTarget)
          };
  }

  // other path continues here
  const then: IThenAbilityPropsWithSystems<AbilityContext> = {
```

server/game/cards/01_SOR/events/DontGetCocky.ts
```diff
@@ -0,0 +1,74 @@
...
+            const then: IThenAbilityPropsWithSystems<AbilityContext> = {
+                title: 'Reveal the next card from your deck or stop revealing cards',
+                targetResolver: {
+                    mode: TargetMode.Select,
+                    choices: {
```
AMMayberry1, line 39: I was thinking it might be good to set activePromptTitle to show the total cost so far. Maybe just something like `Current total cost: ${totalCost}\nSelect one:`. We would want to have a check in the tests for that as well

test/server/cards/01_SOR/events/DontGetCocky.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+                context.player1.clickPrompt('Reveal another card');
+                context.player1.clickPrompt('Reveal another card');
+                context.player1.clickPrompt('Reveal another card');
+                context.player1.clickPrompt('Reveal another card');
+                context.player1.clickPrompt('Reveal another card');
+                context.player1.clickPrompt('Reveal another card');
+                // 7 cards revealed, no choices left
+                expect(context.player2).toBeActivePlayer();
+                expect(context.atatSuppressor.damage).toBe(0);
+                expect(context.player1Object.getTopCardOfDeck()).toBe(context.swoopRacer);
```
AMMayberry1, line 100: These are great! I think we need just one more to test what happens if there aren't enough cards in deck

server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -85,13 +85,16 @@ export class CardTargetResolver extends TargetResolver<ICardTargetResolver<Abili
...
+        // these will appear to have no effect on any target, but should not be skipped
         if (
             !this.dependentTarget &&
+            this.immediateEffect &&
             !legalTargets.some((target) => this.immediateEffect.canAffect(
                 target,
                 this.getContextCopy(target, context),
                 {},
-                GameStateChangeRequired.MustFullyOrPartiallyResolve
+                this.properties.mustChangeGameState
```
Ennis0, line 97: @AMMayberry1 Should I change this line back to how it was and delete the default setting in the `TargetResolver` constructor or is it better this way?

AMMayberry1: I think it should go back to the way it was for now, unless we find a case where it really needs to be configurable

AMMayberry1 (approved):

PR #256, opened 2024-11-23
Add Tarfful and Let The Wookiee Win implementation

Frugghi:
| Tarfful | Let The Wookiee Win |
| ------ | -------- |
| image | image |

### DIFF
--- a/server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
+++ b/server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
@@ -25,7 +25,8 @@ export default class TarffulKashyyykChieftain extends NonLeaderUnitCard {
                 cardTypeFilter: WildcardCardType.Unit,
                 zoneFilter: ZoneName.GroundArena,
                 immediateEffect: AbilityHelper.immediateEffects.damage((context) => ({
-                    amount: context.event.amount
+                    amount: context.event.amount,
+                    source: context.event.card
                 }))
             }
         });
--- a/server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
+++ b/server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
@@ -16,9 +16,8 @@ export default class TarffulKashyyykChieftain extends NonLeaderUnitCard {
             when: {
                 onDamageDealt: (event, context) =>
                     event.type === DamageType.Combat &&
+                    !event.willDefeat &&
                     event.card.hasSomeTrait(Trait.Wookiee) &&
-                    event.isResolved &&
-                    event.card.canBeInPlay() && event.card.isInPlay() && !event.card.pendingDefeat &&
                     event.card.controller === context.source.controller
             },
             targetResolver: {
--- a/server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
+++ b/server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
@@ -1,0 +1,36 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { DamageType, RelativePlayer, Trait, WildcardCardType, ZoneName } from '../../../core/Constants';
+
+export default class TarffulKashyyykChieftain extends NonLeaderUnitCard {
+    protected override getImplementationId () {
+        return {
+            id: '7022736145',
+            internalName: 'tarfful#kashyyyk-chieftain'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addTriggeredAbility({
+            title: 'That unit deals that much damage to an enemy ground unit',
+            when: {
+                onDamageDealt: (event, context) =>
+                    event.type === DamageType.Combat &&
+                    event.card.hasSomeTrait(Trait.Wookiee) &&
+                    event.isResolved &&
+                    event.card.canBeInPlay() && event.card.isInPlay() && !event.card.pendingDefeat &&
+                    event.card.controller === context.source.controller
+            },
+            targetResolver: {
+                controller: RelativePlayer.Opponent,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: ZoneName.GroundArena,
+                immediateEffect: AbilityHelper.immediateEffects.damage((context) => ({
+                    amount: context.event.amount
+                }))
+            }
+        });
+    }
+}
+
+TarffulKashyyykChieftain.implemented = true;
--- a/server/game/gameSystems/DamageSystem.ts
+++ b/server/game/gameSystems/DamageSystem.ts
@@ -25,6 +25,9 @@ export interface ICombatDamageProperties extends IDamagePropertiesBase {
 export interface IAbilityDamageProperties extends IDamagePropertiesBase {
     type?: DamageType.Ability;    // this is optional so it can be the default property type
     amount: number | ((card: UnitCard) => number);
+
+    /** The source of the damage, if different from the card that triggered the ability */
+    source?: Card;
 }
 
 /** Used for abilities that use the excess damage from another instance of damage (currently just Blizzard Assault AT-AT) */
@@ -229,8 +232,8 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext, TPro
     private addAbilityDamagePropertiesToEvent(event: any, card: Card, context: TContext, properties: IAbilityDamageProperties): void {
         const abilityDamageSource: IDamagedOrDefeatedByAbility = {
             type: DamageSourceType.Ability,
-            player: context.player,
-            card: context.source,
+            player: properties.source?.controller ?? context.player,
+            card: properties.source ?? context.source,
             event
         };
 
--- a/server/game/gameSystems/DamageSystem.ts
+++ b/server/game/gameSystems/DamageSystem.ts
@@ -149,6 +149,10 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext, TPro
             default:
                 Contract.fail(`Unexpected damage type: ${properties['type']}`);
         }
+
+        // Check if the damage will defeat the card, this can be used by abilities (e.g. Tarfful) to determine if the card will be defeated or not
+        const damageAmount = event.amount ?? event.sourceEventForExcessDamage?.availableExcessDamage ?? 0;
+        event.willDefeat = card.canBeDamaged() && card.damage + damageAmount >= card.getHp();
     }
 
     private addAttackDamagePropertiesToEvent(event: any, card: Card, context: TContext, properties: ICombatDamageProperties): void {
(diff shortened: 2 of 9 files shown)


server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
```diff
@@ -0,0 +1,36 @@
...
+                    event.card.canBeInPlay() && event.card.isInPlay() && !event.card.pendingDefeat &&
+                    event.card.controller === context.source.controller
+            },
+            targetResolver: {
+                controller: RelativePlayer.Opponent,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: ZoneName.GroundArena,
+                immediateEffect: AbilityHelper.immediateEffects.damage((context) => ({
+                    amount: context.event.amount
+                }))
```
Frugghi, line 30: This will use Tarfful as source of the damage instead of the unit that was attacked, I'm not sure if it's important.

AMMayberry1: Unfortunately I think it probably is, it's not hard to imagine that there could be situations where it's important which specific unit is dealing damage

Frugghi: That's what I thought as well, but there are implemented cards with similar wording (Strike True for example) that don't do anything special. I think those will need to be updated as well!

AMMayberry1: Yeah correct, they will :/ it's on the engine work items list I believe

server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public override setupCardAbilities () {
+        this.addTriggeredAbility({
+            title: 'That unit deals that much damage to an enemy ground unit',
+            when: {
+                onDamageDealt: (event, context) =>
+                    event.type === DamageType.Combat &&
+                    event.card.hasSomeTrait(Trait.Wookiee) &&
+                    event.isResolved &&
```
AMMayberry1, line 20: We shouldn't check `event.isResolved` explicitly like this, it could cause issues. We fire the ability trigger before and after the event resolves partly because of the need to catch triggers on units that are leaving the field during resolution.

An example where this wouldn't work is if the new Maul unit attacks both Tarfful and another wookiee at the same time, and defeats Tarfful but not the other wookiee. Tarfful's ability would not fire afterwards, because he was defeated during the attack and his triggers unregistered immediately after. 

As for the way to actually check this condition, I think it may actually be enough to just check if the event's damage is enough to defeat the unit. There can't be multiple damage events to the same target in the same window due to rule 1.9.8 (linked below), so I'm pretty sure we can safely do that check and assume it will be correct. Lmk if you can think of an exception though.

image

There is one hiccup there which is we need to know if the damage has already been applied or not when we check the unit's `remainingHp` property, because it might be the case that we're seeing the event after it's already resolved. Probably the safest thing to do is to add a new property on the event itself called `willDefeat` or something and just check that.

server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
```diff
@@ -0,0 +1,36 @@
...
+                    event.card.canBeInPlay() && event.card.isInPlay() && !event.card.pendingDefeat &&
```
AMMayberry1, line 21: What was the reason for adding these? Typically these checks are handle by the GameSystem in `immediateEffect`, there's a last-minute check that the target is still legal right before event resolution using the system's `canAffect()`. If it fails, then the event is cancelled. Then, in the triggered ability window, any triggered abilities for events that ended up getting cancelled are also cancelled. So this ability will never fire if the damage didn't actually end up happening, and we shouldn't need to explicitly filter that in the trigger.

test/server/cards/02_SHD/units/TarffulKashyyykChieftain.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+                const { context } = contextRef;
+
+                // Scenario 1: a friendly wookie unit attacks
+                context.player1.clickCard(context.liberatedSlaves);
+                context.player1.clickCard(context.wroshyrTreeTender);
+
+                expect(context.player1).toBeAbleToSelectExactly([context.wampa, context.specforceSoldier, context.wroshyrTreeTender, context.atst]);
+                context.player1.clickCard(context.specforceSoldier);
+
+                expect(context.specforceSoldier).toBeInZone('discard', context.player2);
```
AMMayberry1, line 27: We should do at least one check with a unit that isn't defeated by the ability damage so we can confirm the amount is right

test/server/cards/02_SHD/units/TarffulKashyyykChieftain.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+                        groundArena: ['wampa', 'specforce-soldier', 'wroshyr-tree-tender', 'atst'],
+                        spaceArena: ['cartel-spacer'],
+                    }
+                });
+            });
+
+            it('when a wookie unit is damaged that unit deals that much damage to an enemey ground unit', function () {
+                const { context } = contextRef;
+
+                // Scenario 1: a friendly wookie unit attacks
```
AMMayberry1, line 20: Can we also add a test where a friendly wookiee is attacked

test/server/cards/02_SHD/units/TarffulKashyyykChieftain.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+                expect(context.atst).toBeInZone('discard', context.player2);
+                expect(context.player2).toBeActivePlayer();
+
+                // Scenario 4: a friendly non-wookie unit is attacked
+                context.wroshyrTreeTender.damage = 0;
+                context.player2.clickCard(context.wroshyrTreeTender);
+                context.player2.clickCard(context.isbAgent);
+
+                expect(context.isbAgent).toBeInZone('groundArena', context.player1);
+                expect(context.player1).toBeActivePlayer();
```
AMMayberry1, line 53: Let's add a TODO for using the Maul ability to send combat damage to a friendly underworld wookiee

test/server/cards/02_SHD/units/TarffulKashyyykChieftain.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+            });
```
AMMayberry1, line 54: Can we add a test with Qira since her damage is computed dynamically, just to make sure it doesn't cause an error

Frugghi: Qira doesn't do combat damage so it shouldn't trigger Tarfful. I've added it as a test to check that the ability is not triggered if it's not combat damage.

server/game/cards/02_SHD/units/TarffulKashyyykChieftain.ts
Frugghi: I think I've fixed all the existing cases, Command will be another one but it's implemented yet.

server/game/gameSystems/DamageSystem.ts
```diff
@@ -149,6 +152,10 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext, TPro
             default:
                 Contract.fail(`Unexpected damage type: ${properties['type']}`);
         }
+
+        // Check if the damage will defeat the card, this can be used by abilities (e.g. Tarfful) to determine if the card will be defeated or not
+        const damageAmount = event.amount ?? event.sourceEventForExcessDamage?.availableExcessDamage ?? 0;
```
AMMayberry1, line 157: Would `this.getDamageAmountFromEvent()` work? That also has contract checks to confirm one or the other of these values exists

test/server/cards/02_SHD/units/TarffulKashyyykChieftain.spec.ts
```diff
@@ -51,6 +52,31 @@ describe('Tarfful, Kashyyyk Chieftain', function() {
...
+                context.player1.passAction();
+
+                context.player2.clickCard(context.qira);
+                context.player2.clickPrompt('Deploy Qi\'ra');
+
+                expect(context.player1).toBeActivePlayer();
+                expect(context.player1).toHavePrompt('Choose an action');
+
+                // TODO: Add test with SHD Maul unit redirecting damage to a friendly wookiee underworld unit
+                // TODO: Add test with TWI Maul unit dealing damage to multiple friendly wooikee at the same time
```
AMMayberry1, line 79: Actually, thinking about it let's add a third as well for the case I raised where Tarfful and another wookiee is attacked by Maul2 and only Tarfful is defeated

server/game/gameSystems/DamageSystem.ts
Frugghi: I haven't tried but I don't think so because `getDamageAmountFromEvent` modifies the event to remove available excess damage so the event handler wouldn't work correctly.

AMMayberry1: Gotcha, good catch. I'm pretty sure the 0 case should never happen here, so I definitely want to at least redo the contract checks from that method. So we could either copy them here or maybe add a parameter to the method to disable removing the excess damage. Or maybe split the latter into a separate method, that might make the most sense.

Frugghi: I moved the part the removes the excess damage to the event handler, I think it makes more sense to have it there.

server/game/gameSystems/DamageSystem.ts
```diff
@@ -153,9 +150,13 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext, TPro
                 Contract.fail(`Unexpected damage type: ${properties['type']}`);
         }
 
-        // Check if the damage will defeat the card, this can be used by abilities (e.g. Tarfful) to determine if the card will be defeated or not
-        const damageAmount = event.amount ?? event.sourceEventForExcessDamage?.availableExcessDamage ?? 0;
-        event.willDefeat = card.canBeDamaged() && card.damage + damageAmount >= card.getHp();
+        if (card.canBeDamaged()) {
```
AMMayberry1, line 153: Since this means the card is an illegal target, we don't want to let it ever be the case that we don't go down this if condition. The Contract asserts are typed to allow narrowing to occur, so we can do:

```typescript
Contract.assertTrue(card.canBeDamaged());

const damageAmount = this.getDamageAmountFromEvent(event);
event.availableExcessDamage = damageAmount - Math.min(damageAmount, card.remainingHp);

// Check if the damage will defeat the card, this can be used by abilities (e.g. Tarfful) to determine if the card will be defeated or not
event.willDefeat = damageAmount >= card.remainingHp;
```

Frugghi: Nice! I didn't realise it was possible, that's way better that an if.

AMMayberry1 (approved):

PR #257, opened 2024-11-23
SOR #13 Cassian Andor Leader

brianries:
Added cassian andor leader card / unit.  Had to make a small update to the damage dealt tracker to keep track of amount as well

image
image

### DIFF
--- a/server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
+++ b/server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
@@ -1,0 +1,55 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { LeaderUnitCard } from '../../../core/card/LeaderUnitCard';
+import { StateWatcherRegistrar } from '../../../core/stateWatcher/StateWatcherRegistrar';
+import { attack } from '../../../gameSystems/GameSystemLibrary';
+import { DamageDealtThisPhaseWatcher } from '../../../stateWatchers/DamageDealtThisPhaseWatcher';
+
+export default class CassionAndorDedicatedToTheRebellion extends LeaderUnitCard {
+    private damageDealtThisPhaseWatcher: DamageDealtThisPhaseWatcher;
+
+    protected override getImplementationId() {
+        return {
+            id: '3187874229',
+            internalName: 'cassian-andor#dedicated-to-the-rebellion',
+        };
+    }
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar): void {
+        this.damageDealtThisPhaseWatcher = AbilityHelper.stateWatchers.damageDealtThisPhase(registrar, this);
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addActionAbility({
+            title: 'If you\'ve dealt 3 or more damage to an enemy base this phase, draw a card.',
+            cost: [AbilityHelper.costs.abilityResourceCost(1), AbilityHelper.costs.exhaustSelf()],
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => {
+                    const damageDealtToBase = this.damageDealtThisPhaseWatcher.getDamageDealtByPlayer(
+                        context.source.controller,
+                        (damage) => damage.target.isBase() && damage.target.controller !== context.source.controller
+                    ).reduce((sum, damage) => sum + damage.amount, 0)
+
+                    return damageDealtToBase >= 3;
+                },
+                onTrue: AbilityHelper.immediateEffects.draw((context) => ({ target: context.source.controller })),
+                onFalse: AbilityHelper.immediateEffects.noAction()
+            })
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities() {
+        this.addTriggeredAbility({
+            title: 'When you deal damage to an enemy base: You may draw a card. Use this ability only once each round.',
+            when: {
+                onDamageDealt: (event, context) =>
+                    event.damageSource.attack.target.isBase() &&
+                    event.damageSource.attack.target.controller !== context.source.controller
+            },
+            immediateEffect: AbilityHelper.immediateEffects.draw((context) => ({ target: context.source.controller })),
+            limit: AbilityHelper.limit.perRound(1)
+        });
+    }
+}
+
+
+CassionAndorDedicatedToTheRebellion.implemented = true;
--- a/server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
+++ b/server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
@@ -1,7 +1,6 @@
 import AbilityHelper from '../../../AbilityHelper';
 import { LeaderUnitCard } from '../../../core/card/LeaderUnitCard';
 import { StateWatcherRegistrar } from '../../../core/stateWatcher/StateWatcherRegistrar';
-import { attack } from '../../../gameSystems/GameSystemLibrary';
 import { DamageDealtThisPhaseWatcher } from '../../../stateWatchers/DamageDealtThisPhaseWatcher';
 
 export default class CassionAndorDedicatedToTheRebellion extends LeaderUnitCard {
--- a/server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
+++ b/server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
@@ -26,7 +26,7 @@ export default class CassionAndorDedicatedToTheRebellion extends LeaderUnitCard 
                     const damageDealtToBase = this.damageDealtThisPhaseWatcher.getDamageDealtByPlayer(
                         context.source.controller,
                         (damage) => damage.target.isBase() && damage.target.controller !== context.source.controller
-                    ).reduce((sum, damage) => sum + damage.amount, 0)
+                    ).reduce((sum, damage) => sum + damage.amount, 0);
 
                     return damageDealtToBase >= 3;
                 },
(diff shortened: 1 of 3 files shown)


server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
```diff
@@ -0,0 +1,54 @@
...
+            })
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities() {
+        this.addTriggeredAbility({
+            title: 'When you deal damage to an enemy base: You may draw a card. Use this ability only once each round.',
+            when: {
+                onDamageDealt: (event, context) =>
+                    event.damageSource.attack.target.isBase() &&
```
AMMayberry1, line 44: There are multiple types of damageSource (see IDamageOrDefeatSource.ts), and some of them don't have the `attack` property. If you were to deal damage with an event then this would break - we should add a test case for that.

The default target for a card event can be found as `event.card`, should use that to get the target of the damage and `damageSource.player` to get the responsible player

server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    protected override setupLeaderSideAbilities() {
+        this.addActionAbility({
+            title: 'If you\'ve dealt 3 or more damage to an enemy base this phase, draw a card.',
+            cost: [AbilityHelper.costs.abilityResourceCost(1), AbilityHelper.costs.exhaustSelf()],
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => {
+                    const damageDealtToBase = this.damageDealtThisPhaseWatcher.getDamageDealtByPlayer(
+                        context.source.controller,
+                        (damage) => damage.target.isBase() && damage.target.controller !== context.source.controller
```
AMMayberry1, line 28: Because the target's controller may have changed since the damage happened (due to e.g. Palpatine ability), we need to use the `damageSource` record for determining any information about the damage. In this case `damageSource.player` indicates the responsible player for the damage

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+        describe('Cassian Andor\'s leader ability', function() {
+            beforeEach(function() {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        deck: ['k2so#cassians-counterpart'],
+                        groundArena: ['yoda#old-master'],
+                        spaceArena: ['green-squadron-awing'],
+                        leader: { card: 'cassian-andor#dedicated-to-the-rebellion', deployed: false },
+                        base: { card: 'dagobah-swamp', damage: 5 },
```
AMMayberry1, line 12: What's the reason to declare damage on the bases in these tests? Typically we try not to add any test setup that isn't explicitly needed for the relevant case(s). Generally looks like a lot of stuff in the setup isn't used.

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
```diff
@@ -0,0 +1,95 @@
+describe('Cassian Andor, Dedicated to the Rebellion', function() {
+    integration(function(contextRef) {
+        describe('Cassian Andor\'s leader ability', function() {
+            beforeEach(function() {
+                contextRef.setupTest({
```
AMMayberry1, line 5: If there's only one test case for a given setupTest then you can put the setup directly in the test case instead of doing a beforeEach. See e.g. the Palpatine leader tests for an example.

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+                expect(context.player1.hand).toContain(context.k2so);
+                expect(context.player1.exhaustedResourceCount).toBe(0); // no resources spent now
+
+                context.player2.passAction();
+
+                // Deal more damage to base to see if another event will trigger - it shouldn't
+                context.player1.clickCard(context.greenSquadronAwing);
+                context.player1.clickCard(context.p2Base);
+
+                // No card draw should have happened due to once per round limit -- state should be the same
```
AMMayberry1, line 88: Can we have one more case where use `context.moveToNextActionPhase()` to go to the next phase and confirm that the ability works again, to test the limit

server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
brianries: Looking at the getDamageDealtByPlayer code, am I right that I didn't need this additional filter at all? It was already filtering by player using damageSource.player?  I should just ditch it?

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
brianries: Added this next action phase to both sets of tests (to make sure the phase tracking damage works too)

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
brianries: Cleaned this up

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
brianries: Cleaned this up - removed as much unnecessary setup as possible

server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
brianries: Thanks - much cleaner looking as well

server/game/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.ts
AMMayberry1: Oh yeah actually that's correct now that you say it, missed that

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+                        hand: ['daring-raid', 'daring-raid'],
+                        deck: ['k2so#cassians-counterpart', 'open-fire'],
+                        spaceArena: ['green-squadron-awing'],
+                        leader: { card: 'cassian-andor#dedicated-to-the-rebellion', deployed: false },
+                        resources: 4,
+                    },
+                    player2: {
+                        groundArena: ['wampa'],
+                        spaceArena: ['tieln-fighter'],
+                        base: { card: 'capital-city' }
```
AMMayberry1, line 17: If we're not adding damage or anything this can just be `base: 'capital-city'`. Is there a reason to declare it at all though? There will be a base put in by default, can be referred to by `context.p2Base`

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+                context.player2.passAction();
+
+                // Expect this ability have activated now - so it should draw the top card from the deck
+                expect(context.cassianAndor).toHaveAvailableActionWhenClickedBy(context.player1);
+                expect(context.cassianAndor.exhausted).toBeTrue();
+                expect(context.player1.exhaustedResourceCount).toBe(2);
+                expect(context.player1.hand).toContain(context.k2so);
+
+                context.moveToNextActionPhase();
+                context.player1.readyResources(4);
```
AMMayberry1, line 42: Why is this needed? We're at the start of a new action phase so all resources should be ready

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    integration(function(contextRef) {
+        describe('Cassian Andor\'s leader ability', function() {
+            it('should draw a card aftering dealing 3 damage to an enemy base with an attack', function() {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['daring-raid', 'daring-raid'],
+                        deck: ['k2so#cassians-counterpart', 'open-fire'],
+                        spaceArena: ['green-squadron-awing'],
+                        leader: { card: 'cassian-andor#dedicated-to-the-rebellion', deployed: false },
```
AMMayberry1, line 11: The default state is not deployed so if you want you can just write `leader: 'cassian-andor#dedicated-to-the-rebellion'`. Having it be explicit is also fine though

brianries: I'll remove it to be more consistent with the other tests.

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
brianries: vestigial, forgot to remove.  I thought resources weren't readying correctly, forgot that direct-fire was consuming resources (looked like it hadn't reset -- PEBKAC).  Cleaned up

test/server/cards/01_SOR/leaders/CassianAndorDedicatedToTheRebellion.spec.ts
brianries: good to know about default, cleaned up

PR #260, opened 2024-11-23
Add Capture mechanic

AMMayberry1:
Added the Capture mechanic, available as CaptureSystem. Also added the Rescue mechanic via RescueSystem.

Implemented Take Captive, Discerning Veteran, and Unexpected Escape to test out the systems. Added a test suite for Capture and some tests to ensure that it works for Bounty. Additionally tested cases for Rescue in the same suite.

Fixed some bugs with targeting that came up while writing the tests.


test/server/gameSystems/CaptureSystem.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+            // capture wampa with discerning veteran
+            context.player1.clickCard(context.discerningVeteran);
+
+            expect(context.wampa).toBeCapturedBy(context.discerningVeteran);
+            expect(context.battlefieldMarine).toBeInZone('groundArena');
+            expect(context.battlefieldMarine.exhausted).toBeTrue();
+            expect(context.academyTraining).toBeInZone('discard');
+        });
+    });
+});
```
VJubert, line 111: there are few more tests we can add, 

- [ ] for superlaser blast, captured units should not be defeated 
- [ ] defeat a control-taken unit which guard units should return captured unit to the good controllers

server/game/cards/02_SHD/events/UnexpectedEscape.ts
```diff
@@ -0,0 +1,35 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Exhaust a unit. You may rescue a captured card guarded by that unit.',
+            targetResolvers: {
+                exhaust: {
+                    zoneFilter: WildcardZoneName.AnyArena,
+                    controller: RelativePlayer.Opponent,
```
Frugghi, line 19: I guess technically it could be a unit you control, it doesn't say "an enemy unit".

server/game/cards/02_SHD/events/UnexpectedEscape.ts
```diff
@@ -0,0 +1,35 @@
...
+                    immediateEffect: AbilityHelper.immediateEffects.exhaust()
+                },
+                rescue: {
+                    dependsOn: 'exhaust',
+                    activePromptTitle: 'Rescue a captured card guarded by that unit',
```
Frugghi, line 24: Wouldn't it be better to use the name of the card instead of "that unit"?

server/game/cards/02_SHD/events/UnexpectedEscape.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Exhaust a unit. You may rescue a captured card guarded by that unit.',
+            targetResolvers: {
+                exhaust: {
+                    zoneFilter: WildcardZoneName.AnyArena,
+                    controller: RelativePlayer.Opponent,
+                    immediateEffect: AbilityHelper.immediateEffects.exhaust()
```
Frugghi, line 20: Could you target a unit that is already exhausted just to rescue something?

server/game/cards/02_SHD/units/DiscerningVeteran.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'This unit captures an enemy non-leader ground unit',
+            targetResolver: {
+                controller: RelativePlayer.Opponent,
+                zoneFilter: ZoneName.GroundArena,
+                immediateEffect: AbilityHelper.immediateEffects.capture()
+            }
```
Frugghi, line 20: Do you also need to filter out leader units or it's the default?

test/server/gameSystems/CaptureSystem.spec.ts
AMMayberry1: I added the first test but I don't know what you mean by the second test, could you give an example?

server/game/cards/02_SHD/units/DiscerningVeteran.ts
AMMayberry1: Yeah it's the default since CaptureSystem declares its target type to NonLeaderUnit

test/server/gameSystems/CaptureSystem.spec.ts
VJubert: sure : player1 captures a unit with take captive, player2 take control of this unit (with palpatine for example), player1 use rivals fall on this unit. the captured unit must return in play on player 2 side

VJubert (approved):

PR #263, opened 2024-11-25
Che bato/modal cards

CheBato:
- Implemented playingModalCards system
- Added Vigilance card and tests.

### DIFF
--- a/server/game/core/Constants.ts
+++ b/server/game/core/Constants.ts
@@ -193,7 +193,6 @@ export enum EventName {
     OnCardLeavesPlay = 'onCardLeavesPlay',
     OnCardMoved = 'onCardMoved',
     OnCardPlayed = 'onCardPlayed',
-    OnPlayModalCard = 'onPlayModalCard',
     OnCardReadied = 'onCardReadied',
     OnCardResourced = 'onCardResourced',
     OnCardReturnedToHand = 'onCardReturnedToHand',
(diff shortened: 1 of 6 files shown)


server/game/core/Constants.ts
```diff
@@ -248,7 +247,8 @@ export enum MetaEventName {
     ReplacementEffect = 'replacementEffect',
     SelectCard = 'selectCard',
     Sequential = 'sequential',
-    Simultaneous = 'simultaneous'
+    Simultaneous = 'simultaneous',
+    ChooseModalEffects = 'ChooseModalEffects'
```
AMMayberry1, line 251: nit: can you move this so it aligns alphabetically

test/server/cards/01_SOR/events/Vigilance.spec.ts
```diff
@@ -100,7 +108,7 @@ describe('Vigilance', function() {
                 });
             });
 
-            it('discards 6 cards from an opponent\'s deck dealing 6 damage to his base and heal 0 damage from base.',
+            it('discards 6 cards from an opponent\'s deck dealing 0 damage to his base and heal 0 damage from base.',
```
AMMayberry1, line 111: nit: could we just remove the reference to damage completely? Imo having it say 0 makes it sound like there would normally be an expectation that damage happens, which isn't true

server/game/gameSystems/PlayModalCardSystem.ts
```diff
@@ -0,0 +1,77 @@
...
+        };
+        choiceHandler(context.player, listOfChoices, properties.amountOfChoices);
+    }
+
+    // Helper method for pushing the correct event into the events array.
+    private pushEvent(
+        events: GameEvent[],
+        choice: any,
+        context: TContext,
+        amountOfChoices: number,
```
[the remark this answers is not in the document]
AMMayberry1: Looks like you changed some other naming but not this one, could we change this parameter to be like `amountOfRemainingChoices` or something similar?

AMMayberry1 (approved):

PR #264, opened 2024-11-25
Feat: Implement bib fortuna and home one

jgordijo:
image

image

### DIFF
--- a/test/server/cards/01_SOR/units/BibFortunaJabbasMajordomo.spec.ts
+++ b/test/server/cards/01_SOR/units/BibFortunaJabbasMajordomo.spec.ts
@@ -1,0 +1,81 @@
+describe('Bib Fortuna', function() {
+    integration(function(contextRef) {
+        describe('Bib Fortuna\'s activated ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['bib-fortuna#jabbas-majordomo', 'jawa-scavenger'],
+                        hand: ['repair', 'confiscate', 'swoop-racer', 'surprise-strike'],
+                        base: 'capital-city',
+                        leader: 'grand-inquisitor#hunting-the-jedi',
+                    },
+                });
+            });
+
+            it('should allow the controller to play an event with a discount of 1', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.bibFortuna);
+
+                expect(context.player1).toHaveEnabledPromptButtons(['Attack', 'Play an event from your hand. It costs 1 less.']);
+                context.player1.clickPrompt('Play an event from your hand. It costs 1 less.');
+
+                expect(context.player1).toBeAbleToSelectExactly([context.repair, context.confiscate, context.surpriseStrike]);
+                context.player1.clickCard(context.repair);
+                // selects target for repair
+                context.player1.clickCard(context.capitalCity);
+
+                expect(context.bibFortuna.exhausted).toBe(true);
+                expect(context.player1.exhaustedResourceCount).toBe(0);
+
+
+                context.player2.passAction();
+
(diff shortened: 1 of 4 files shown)


test/server/cards/01_SOR/units/BibFortunaJabbasMajordomo.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+
+
+                context.player2.passAction();
+
+                // cost discount from bib fortuna should be gone
+                context.player1.clickCard(context.confiscate);
+                expect(context.player1.exhaustedResourceCount).toBe(1);
+
+                context.player2.passAction();
+                context.bibFortuna.exhausted = false;
```
AMMayberry1, line 40: What's this for? Shouldn't need any additional actions after the last expect

test/server/cards/01_SOR/units/BibFortunaJabbasMajordomo.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+            });
+
+            it('should not give the next unit played by the controller a discount after the controller declines to play a unit with the ability', function() {
```
AMMayberry1, line 43: These could be combined into one test by unexhausting Bib Fortuna in between, see Rukh tests for an example of the format we tend to use (if I haven't already explained why we do this even though it's bad test practice please ping me and I can walk through it)

test/server/cards/01_SOR/units/BibFortunaJabbasMajordomo.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+                context.player1.clickCard(context.repair);
+                // selects target for repair
+                context.player1.clickCard(context.capitalCity);
+
+                expect(context.player1.exhaustedResourceCount).toBe(1);
+
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should be able to select and play an event that costs exactly 1 more than ready resources', function() {
```
AMMayberry1, line 64: Great tests! Love these cases, can't think of anything else I would want to cover

test/server/cards/01_SOR/units/HomeOneAllianceFlagship.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+                    player2: {
+                        groundArena: ['rugged-survivors', 'cargo-juggernaut']
+                    }
+                });
+            });
+
+            it('should give Restore 1 to friendly units', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.battlefieldMarine);
```
AMMayberry1, line 21: Would also be good to test a friendly space unit and an enemy unit to confirm that the restore targeting is set up correctly

test/server/cards/01_SOR/units/HomeOneAllianceFlagship.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+
+                expect(context.echoBaseDefender).toBeInZone('groundArena');
+
+                // should only exhaust resources up to home one's cost
+                expect(context.player1.exhaustedResourceCount).toBe(8);
+
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should play a heroic unit that costs more than 3 resources from the discard pile with discount', function () {
```
AMMayberry1, line 69: Could you change this test to where there is only one legal target and confirm that it just auto-plays with no user prompt?

test/server/cards/01_SOR/units/BibFortunaJabbasMajordomo.spec.ts
jgordijo: Thanks, but this scenario was stated first on the Alliance Dispatcher, so I just applied the same here :)

AMMayberry1 (approved):

PR #266, opened 2024-11-25
Feature/implement heroic sacrifice

jgordijo:
image

### DIFF
--- a/server/game/cards/01_SOR/events/HeroicSacrifice.ts
+++ b/server/game/cards/01_SOR/events/HeroicSacrifice.ts
@@ -1,0 +1,44 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { AbilityType, DamageType } from '../../../core/Constants';
+
+export default class HeroicSacrifice extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '6954704048',
+            internalName: 'heroic-sacrifice',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Draw a card',
+            immediateEffect: AbilityHelper.immediateEffects.draw(),
+            then: {
+                title: 'Attack with a unit. For this attack, it gets +2/+0 and gains: "When this unit deals combat damage: Defeat it."',
+                targetResolver: {
+                    immediateEffect: AbilityHelper.immediateEffects.attack({
+                        attackerLastingEffects: [
+                            { effect: AbilityHelper.ongoingEffects.modifyStats({ power: 2, hp: 0 }) },
+                            {
+                                effect: AbilityHelper.ongoingEffects.gainAbility({
+                                    type: AbilityType.Triggered,
+                                    title: 'When this unit deals combat damage: Defeat it.',
+                                    when: {
+                                        onDamageDealt: (event, context) => (
+                                            event.type === DamageType.Combat &&
+                                            event.damageSource.attack.attacker === context.source &&
+                                            event.damageSource.damageDealtBy === context.source)
+                                    },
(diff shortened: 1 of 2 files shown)


server/game/cards/01_SOR/events/HeroicSacrifice.ts
```diff
@@ -0,0 +1,44 @@
...
+                                effect: AbilityHelper.ongoingEffects.gainAbility({
+                                    type: AbilityType.Triggered,
+                                    title: 'When this unit deals combat damage: Defeat it.',
+                                    when: {
+                                        onDamageDealt: (event, context) => (
+                                            event.type === DamageType.Combat &&
+                                            event.damageSource.attack.attacker === context.source &&
+                                            event.damageSource.damageDealtBy === context.source)
+                                    },
+                                    immediateEffect: AbilityHelper.immediateEffects.defeat((context) => ({ target: context.source })),
```
AMMayberry1, line 33: nit: `context.source` is the default target when none is provided (except when in a targetResolver, which this is not) so you can just do `AbilityHelper.immediateEffects.defeat()`

test/server/cards/01_SOR/events/HeroicSacrifice.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                        groundArena: ['isb-agent', { card: 'wampa', exhausted: true }],
+                        spaceArena: ['tieln-fighter']
+                    },
+                    player2: {
+                        groundArena: ['sundari-peacekeeper', 'atst', { card: 'crafty-smuggler', upgrades: ['shield'] }],
+                    }
+                });
+            });
+
+            it('should draw a card and attack with a unit giving +2/+0 for this attack and defeat it after dealing combat damage', function () {
```
AMMayberry1, line 18: Could we add one more test where he does damage into a unit just to be safe, since we haven't written many abilities like this one before

test/server/cards/01_SOR/events/HeroicSacrifice.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+
+                context.player1.clickCard(context.craftySmuggler);
+
+                expect(context.craftySmuggler.upgrades.length).toBe(0);
+                expect(context.craftySmuggler).toBeInZone('groundArena');
+
+                expect(context.isbAgent).toBeInZone('groundArena');
+
+                expect(context.player2).toBeActivePlayer();
+            });
```
AMMayberry1, line 57: Here's a fun test we can do to confirm that it is specifically combat damage that works and not damage from e.g. on-attack abilities:

1. Set up Lom Pyke with a Vambrace Flamethrower
2. Have Lom attack into a unit. You'll get a prompt to select which ability to trigger (see TriggeredAbilityWindow.spec.ts for an example of how to manage that)
3. First select Vambrace Flamethrower and deal damage to the defender and other units
4. Then activate Lom Pyke's ability and give a shield to the defender
5. Lom's combat damage won't resolve b/c of the shield but he should stay in play and not be defeated by Heroic Sacrifice b/c the Flamethrower damage wasn't combat damage

test/server/cards/01_SOR/events/HeroicSacrifice.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+        });
```
AMMayberry1, line 58: Let's a TODO for testing with the Maul unit since he redirects the combat damage, but I'm pretty sure it should still count (I may need to ask the judge chat for confirmation)

test/server/cards/01_SOR/events/HeroicSacrifice.spec.ts
jgordijo: Love the suggestion! That's a great scenario

AMMayberry1 (approved):

PR #270, opened 2024-11-26
Bunch of cards 11

VJubert:


server/game/cards/01_SOR/events/Aggression.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Aggression modal ability:',
+            immediateEffect: AbilityHelper.immediateEffects.chooseModalEffects({
+                amountOfChoices: 2,
+                choices: () => ({
+                    ['Draw a card']: AbilityHelper.immediateEffects.draw(),
+                    ['Defeat up to 2 upgrades']: AbilityHelper.immediateEffects.selectCard({
+                        mode: TargetMode.UpTo,
```
AMMayberry1, line 22: This needs `canChooseNoTargets: true`. Actually could you go to CardTargetResolver and change the behavior so that is the default? I feel like that is almost always the expected behavior in swu. Also, we need to test the choose no targets case

test/server/cards/02_SHD/upgrades/TheMandaloriansRifle.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+            it('should capture an exhausted enemy unit because attached unit is The Mandalorian (unit)', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['the-mandalorians-rifle'],
+                        groundArena: ['the-mandalorian#wherever-i-go-he-goes', 'battlefield-marine'],
+                    },
+                    player2: {
+                        groundArena: ['wampa', { card: 'atst', exhausted: true }, { card: 'scout-bike-pursuer', exhausted: true }]
```
AMMayberry1, line 43: Let's add a deployed enemy leader just to confirm the targeting works since capture is new

test/server/cards/02_SHD/upgrades/LegalAuthority.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.legalAuthority);
+
+                // able to choose non-vehicle friendly unit
+                expect(context.player1).toBeAbleToSelectExactly([context.occupierSiegeTank, context.desperadoFreighter]);
+                context.player1.clickCard(context.occupierSiegeTank);
+
+                // wampa is automatically choose
+                expect(context.wampa).toBeCapturedBy(context.occupierSiegeTank);
```
AMMayberry1, line 26: I realized I didn't add any tests for capturing across arenas (e.g. space unit capture ground unit). Could you go to CaptureSystem.spec.ts and add a cross-arena capture test, including confirmation that when the capture unit is defeated the captured unit is returned to the correct arena?

server/game/gameSystems/SelectCardSystem.ts
```diff
@@ -73,6 +74,14 @@ export class SelectCardSystem<TContext extends AbilityContext = AbilityContext>
                 ) && properties.cardCondition(card, context);
             properties.selector = CardSelectorFactory.create(Object.assign({}, properties, { cardCondition }));
         }
+
+        if (properties.mode === TargetMode.UpTo || properties.mode === TargetMode.UpToVariable) {
+            properties.canChooseNoCards = properties.canChooseNoCards ?? true;
+        } else {
+            // should not use this property if mode is not UpTo(Variable)
+            properties.canChooseNoCards = false;
```
AMMayberry1, line 82: If we shouldn't use this property in certain situations then let's instead have a contract check to throw an error if they set it when they shouldn't. The way it is right now, if someone sets canChooseNoCards to true when they shouldn't then it will just be silently changed to false in a way that they wouldn't realize, which is not good.

server/game/gameSystems/SelectCardSystem.ts
```diff
@@ -78,8 +78,9 @@ export class SelectCardSystem<TContext extends AbilityContext = AbilityContext>
         if (properties.mode === TargetMode.UpTo || properties.mode === TargetMode.UpToVariable) {
             properties.canChooseNoCards = properties.canChooseNoCards ?? true;
         } else {
-            // should not use this property if mode is not UpTo(Variable)
-            properties.canChooseNoCards = false;
+            if (properties.canChooseNoCards === true) {
```
AMMayberry1, line 81: Should be `properties.canChooseNoCards != null` to check if they put any value at all

AMMayberry1 (approved):

PR #272, opened 2024-11-27
SOR #143 Fighters for Freedom

brianries:
image

### DIFF
--- a/server/game/cards/01_SOR/units/FightersForFreedom.ts
+++ b/server/game/cards/01_SOR/units/FightersForFreedom.ts
@@ -12,7 +12,7 @@ export default class FightersForFreedom extends NonLeaderUnitCard {
 
     public override setupCardAbilities () {
         this.addTriggeredAbility({
-            title: 'When you play another aggression card: You may deal 1 damage to a base',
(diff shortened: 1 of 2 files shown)


VJubert (approved):

VJubert (dismissed):

server/game/cards/01_SOR/units/FightersForFreedom.ts
```diff
@@ -0,0 +1,31 @@
...
+    protected override getImplementationId () {
+        return {
+            id: '5907868016',
+            internalName: 'fighters-for-freedom'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addTriggeredAbility({
+            title: 'When you play another aggression card: You may deal 1 damage to a base',
```
AMMayberry1, line 15: nit: we don't include the "when" or the "you may" parts in the title, since the game will manage the flow for those. So it should just be `Deal 1 damage to a base`

brianries: Updated

AMMayberry1 (approved):

PR #273, opened 2024-11-27
Feature/force throw and power failure

aruizc:
image
image


server/game/cards/01_SOR/events/ForceThrow.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Choose a player. That player discards a card from their hand',
+            targetResolver: {
+                mode: TargetMode.Player,
+                immediateEffect: AbilityHelper.immediateEffects.discardCardsFromOwnHand({ amount: 1 }),
+            },
+            then: (thenContext) => ({
```
AMMayberry1, line 21: This should probably be an `ifYouDo: (ifYouDoContext)`, which will cause the second part of the ability to not happen if the discard doesn't happen. We should add a test where the target player's hand is empty to confirm that nothing breaks.

server/game/cards/01_SOR/events/ForceThrow.ts
```diff
@@ -0,0 +1,35 @@
...
+                title: 'If you control a Force unit, you may deal damage to a unit equal to the cost of the discarded card',
```
AMMayberry1, line 22: This title will only appear if the condition is met, and we should include the name of the discarded card and the amount for clarity. So we can say maybe `Deal damage to a unit equal to the cost of ${ifYouDoContext.target.title} (${ifYouDoContext.events[0].card.printedCost} damage)`

For example, `Deal damage to a unit equal to the cost of Wampa (4 damage)`

server/game/cards/01_SOR/events/PowerFailure.ts
```diff
@@ -0,0 +1,27 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Defeat any number of upgrades on a unit',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                immediateEffect: AbilityHelper.immediateEffects.defeat((context) => {
+                    const allUpgrades = context.target.upgrades;
```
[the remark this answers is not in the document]
AMMayberry1: +1, look at General Tagge for an example of how to set this up and test it

test/server/cards/01_SOR/events/ForceThrow.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                        groundArena: ['specforce-soldier', 'atst'],
+                        spaceArena: ['tieln-fighter']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                // Damage dealt by opponent discarded card
+                context.player1.clickCard(context.forceThrow);
+                context.player1.clickPrompt('Opponent');
```
AMMayberry1, line 22: Can we add a check that the right set of buttons appears using `.toHaveExactPromptButtons()`

test/server/cards/01_SOR/events/ForceThrow.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                expect(context.karabast).toBeInZone('discard');
```
AMMayberry1, line 23: I think we talked about this in chat but now I actually think it would be good for the player to have multiple cards in hand so we can confirm that exactly one of them gets discarded, apologies for changing that

test/server/cards/01_SOR/events/ForceThrow.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+            it('should select opponent to discard a card but no deal damage as there is not a force unit', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['force-throw'],
+                        groundArena: ['wampa']
+                    },
+                    player2: {
+                        hand: ['karabast'],
+                        groundArena: ['specforce-soldier', 'atst']
```
AMMayberry1, line 39: We should add a Force unit on the opponent side to confirm that doesn't trigger the effect

test/server/cards/01_SOR/events/ForceThrow.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+                // Opponent discards a card no force unit in play
+                context.player1.clickCard(context.forceThrow);
+                context.player1.clickPrompt('Opponent');
+
+                expect(context.karabast).toBeInZone('discard');
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should select myself to discard a card but no deal damage as there is not a force unit', function () {
```
AMMayberry1, line 53: For this case I would actually prefer that we have a Force unit so we can make sure that the whole flow works for both players

aruizc:
Implementing 
image

test/server/cards/01_SOR/events/ForceThrow.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+
+                const { context } = contextRef;
+
+                // Damage dealt by opponent discarded card
+                context.player1.clickCard(context.forceThrow);
+                expect(context.player1).toHaveExactPromptButtons(['You', 'Opponent']);
+
+                // Opponent discards a card
+                context.player1.clickPrompt('Opponent');
+                context.player2.clickCard(context.karabast);
```
AMMayberry1, line 26: Need to add a check about which cards player 2 is allowed to click (`.toBeAbleToSelectExactly`)

test/server/cards/01_SOR/events/ForceThrow.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+
+                // Validate board state and options
+                expect(context.karabast).toBeInZone('discard');
+                expect(context.player1).toHaveEnabledPromptButton('Pass');
```
AMMayberry1, line 30: We should use `expect(context.player1).toHavePassAbilityPrompt('Deal damage to a unit equal to the cost of Karabast (2 damage)');`

test/server/cards/01_SOR/events/ForceThrow.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+                const { context } = contextRef;
+
+                // Discard a card and no force unit in play
+                context.player1.clickCard(context.forceThrow);
+                context.player1.clickPrompt('You');
+                context.player1.clickPrompt('Deal damage to a unit equal to the cost of Karabast (2 damage)');
+                context.player1.clickCard(context.specforceSoldier);
+
+                expect(context.karabast).toBeInZone('discard');
+                expect(context.specforceSoldier).toBeInZone('discard');
```
AMMayberry1, line 108: When we're doing a test that deals damage we should always choose a target that survives the damage, if possible, so that we can validate the amount

test/server/cards/01_SOR/events/ForceThrow.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+
+                const { context } = contextRef;
+
+                // Discard a card and no force unit in play
+                context.player1.clickCard(context.forceThrow);
+                context.player1.clickPrompt('You');
+                context.player1.clickPrompt('Deal damage to a unit equal to the cost of Karabast (2 damage)');
+                context.player1.clickCard(context.specforceSoldier);
+
+                expect(context.karabast).toBeInZone('discard');
```
AMMayberry1, line 107: Could we test with a different event card with a different cost (other than 2) just to confirm that the damage number stays correct

test/server/cards/01_SOR/events/PowerFailure.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+            });
+
+            it('defeats all upgrades on a unit', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.powerFailure);
+                expect(context.player1).toBeAbleToSelectExactly([context.pykeSentinel, context.imperialInterceptor]);
+
+                context.player1.clickCard(context.pykeSentinel);
+                context.player1.clickCard(context.devotion);
```
AMMayberry1, line 25: Need a `toBeAbleToSelectExactly` to confirm that player1 is only able to select the upgrades on Pyke Sentinel

test/server/cards/01_SOR/events/PowerFailure.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+                context.player1.clickCard(context.entrenched);
+                context.player1.clickPrompt('Done');
+
+                expect(context.pykeSentinel.isUpgraded()).toBe(false);
+                expect(context.imperialInterceptor.isUpgraded()).toBe(true);
```
AMMayberry1, line 30: Should check the exact set of upgrades using `.toHaveExactUpgradeNames()`

test/server/cards/01_SOR/events/PowerFailure.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+                expect(context.entrenched).toBeInZone('discard');
+                expect(context.devotion).toBeInZone('discard');
+            });
+
+            it('defeats selected upgrades on a unit', function () {
```
AMMayberry1, line 35: These are good, can we add one more test confirming that we can select an upgraded unit then choose no upgrades?

test/server/cards/01_SOR/events/PowerFailure.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+                        spaceArena: [{ card: 'imperial-interceptor', upgrades: ['academy-training', 'shield'] }]
+                    }
+                });
+            });
+
+            it('defeats all upgrades on a unit', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.powerFailure);
+                expect(context.player1).toBeAbleToSelectExactly([context.pykeSentinel, context.imperialInterceptor]);
```
AMMayberry1, line 22: I think I mentioned this in my comment about setting `optional: true`, but we should add a test here confirming that the pass button appears: `expect(context.player1).toHavePassAbilityPrompt();`

AMMayberry1 (approved):

PR #277, opened 2024-11-29
Implementing Disabling Fang Fighter

supercodepoet:


server/game/cards/01_SOR/units/DisablingFangFighter.ts
```diff
@@ -0,0 +1,25 @@
...
+    protected override getImplementationId () {
+        return {
+            id: '2668056720',
+            internalName: 'disabling-fang-fighter'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addWhenPlayedAbility({
+            title: 'You may defeat an upgrade.',
```
VJubert, line 15: we do not include 'you may' on title, prompt menu handle it for us

supercodepoet: Gotcha, fixed

test/server/cards/01_SOR/units/DisablingFangFighter.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.frozenInCarbonite, context.lukesLightsaber]);
+                expect(context.player1).toHavePassAbilityButton();
+
+                context.player1.clickCard(context.frozenInCarbonite);
+                expect(context.pykeSentinel.isUpgraded()).toBe(false);
+                expect(context.lukeSkywalkerJediKnight.isUpgraded()).toBe(true);
+                expect(context.frozenInCarbonite).toBeInZone('discard');
+            });
+
+            it('can defeat an upgrade on a enemy unit', function () {
```
AMMayberry1, line 30: Thanks for the good test cases! Unfortunately, we have this practical problem where the jasmine tests are very slow and heavy and we have finite Github Actions time in the free tier that we have to conserve. So we try to consolidate our test cases as much as we can.

In this case we generally assume that if a target appears in the target list, we can assume that it will work unless there's a special case. Same for the pass button, so long as we check that it appears we assume it works. So we can remove the bottom two test cases here.

supercodepoet: Understood and fixed

AMMayberry1 (approved):

VJubert (approved):

PR #281, opened 2024-11-29
Grievous Wheel Bike, On Top of Things & Unshakeable Will 

VJubert:


adbronico (dismissed):
lgtm

test/server/cards/03_TWI/upgrades/OnTopOfThings.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+                        hand: ['on-top-of-things'],
+                        groundArena: ['battlefield-marine', 'wampa'],
+                    },
+                    player2: {
+                        groundArena: ['atst']
+                    }
+                });
+            });
+
+            it('should make attached unit untargetable for this phase', function () {
```
AMMayberry1, line 17: Could you go to Sentinel.spec.ts and add a test using On Top of Things and Unshakeable Will to have an explicit test case that the Sentinel overrides the target restriction? I think we have one or two tests for that on random cards but having it in the Sentinel test cases explicitly would be good, and these two cards make writing the case very easy

AMMayberry1 (approved):

PR #283, opened 2024-11-29
Bunch of cards 12

VJubert:


test/server/cards/03_TWI/units/DaughterOfDathomir.spec.ts
```diff
@@ -0,0 +1,35 @@
+describe('Daughter of Dathomir', function () {
+    integration(function (contextRef) {
+        describe('Daughter of Dathomir\'s ability', function () {
+            it('should gain Restore 2 while undamaged', function () {
```
sjmcculloch, line 4: should also have a test for when it is damaged (to make sure it doesn't restore)

sjmcculloch: Oh, my bad. I was expecting separate tests and didn't see you combined into one.

test/server/cards/03_TWI/units/AdmiralYularenAdvisingCaution.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+                    },
+                    player2: {
+                        groundArena: ['battlefield-marine'],
+                        hasInitiative: true,
+                    }
+                });
+
+                const { context } = contextRef;
+
+                // attack echo base defender with battlefield marine
```
adbronico, line 18: We shouldn't need any attacks here. Just check the HP of Echo Base is 4 and that the HP of Battlefield Marine is 3

server/game/cards/03_TWI/units/AdmiralYularenAdvisingCaution.ts
```diff
@@ -0,0 +1,22 @@
...
+        return {
+            id: '0268657344',
+            internalName: 'admiral-yularen#advising-caution',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'Each other friendly Heroism unit gets +0/+1',
+            matchTarget: (card, context) => card !== context.source && card.isUnit() && card.hasSomeAspect(Aspect.Heroism),
```
adbronico, line 16: We should make sure this isn't buffing enemy units

test/server/cards/03_TWI/units/BoKatanKryzeDeathWatchLieutenant.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['bokatan-kryze#death-watch-lieutenant', 'battlefield-marine'],
+                    },
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.bokatanKryze);
+                expect(context.p2Base.damage).toBe(3);
```
adbronico, line 15: Can we have P2 vanquish the Marine and ensure Bo-Katan isn't buffing herself? Ditto with the other effects in the other test

adbronico (changes requested):

test/server/cards/03_TWI/units/ConfederateTriFighter.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+                expect(context.p2Base.damage).toBe(5);
+
+                reset();
+
+                // kill confederate tri-fighter, base healing should work
+                context.player1.clickCard(context.confederateTrifighter);
+                expect(context.confederateTrifighter).toBeInZone('discard');
+
+                // heal 3 damage from base
+                context.player2.clickCard(context.smugglersAid);
```
AMMayberry1, line 50: Could we add a similar test with an event card that has a targeted healing effect to try and heal the base while tri-fighter is still alive? We need to confirm that the base is still targetable / selectable for heal effects but that the effect just does nothing

test/server/cards/03_TWI/units/DevastatingGunship.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+                        groundArena: ['specforce-soldier']
+                    },
+                    player2: {
+                        groundArena: [{ card: 'battlefield-marine', damage: 2 }, 'wampa'],
+                        spaceArena: ['lurking-tie-phantom']
+                    }
+                });
+            });
+
+            it('should not give -1/-1 because no friendly was defeated this phase', function () {
```
AMMayberry1, line 18: nit: wrong test title

server/game/cards/03_TWI/units/EnfysNestChampionOfJustice.ts
```diff
@@ -0,0 +1,31 @@
...
+            title: 'Return an enemy non-leader unit with less power than this unit to its owner\'s hand',
+            when: {
+                onCardPlayed: (event, context) => event.card === context.source,
+                onAttackDeclared: (event, context) => event.attack.attacker === context.source,
+            },
+            optional: true,
+            targetResolver: {
+                controller: RelativePlayer.Opponent,
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                cardCondition: (card, context) => card.isUnit() && card.getPower() < context.source.getPower(),
```
AMMayberry1, line 24: If we have to have a `card.is*()` check anyway we may as well remove the NonLeaderUnit filter and just check `card.isNonLeaderUnit()` here

test/server/cards/03_TWI/units/EnfysNestChampionOfJustice.spec.ts
```diff
@@ -0,0 +1,50 @@
+describe('Enfys Nest, Champion of Justice', function () {
+    integration(function (contextRef) {
+        describe('Enfys Nest\'s ability', function () {
+            it('can return an enemy non-leader unit with less power than this unit to its owner\'s hand', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['enfys-nest#champion-of-justice', 'academy-training'],
```
AMMayberry1, line 8: Can we add a friendly unit as well just to check the filter

server/game/gameSystems/HealSystem.ts
```diff
@@ -37,9 +39,6 @@ export class HealSystem<TContext extends AbilityContext = AbilityContext> extend
         if (properties.isCost && (properties.amount === 0 || card.damage === 0)) {
             return false;
         }
-        if (card.hasRestriction(AbilityRestriction.BeHealed, context)) {
```
AMMayberry1, line 40: Ah yeah okay I was thinking we might hit this issue. So we should keep this check here but make some changes to it.

Basically `canAffect()` for any system always should check if the card is a legal target for the effect and return false if not. By default it doesn't care if the effect will actually change game state or not, but there is a parameter to control that behavior which we need to update this code to respect.

If you look at `ExhaustSystem.canAffect()` you will see that it checks if the target is already exhausted but only if `properties.isCost` is true or the `mustChangeGameState` parameter set to some enum value other than "None". So basically by default it's only checking to see if the card is a legal target for exhaust, but if one of those two is set then it also checks that the effect must change game state. That's what we need here as well (and we should make the same change in the damage system now since we know we'll need it).

AMMayberry1 (approved):

PR #285, opened 2024-11-30
SHD #73 Mandalorian Armor Upgrade Implementation

sjmcculloch:
Mandalorian Armor Upgrade Implementation from SHD

### DIFF
--- a/server/game/cards/02_SHD/upgrades/MandalorianArmor.ts
+++ b/server/game/cards/02_SHD/upgrades/MandalorianArmor.ts
@@ -1,0 +1,27 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { Card } from '../../../core/card/Card';
+import { UpgradeCard } from '../../../core/card/UpgradeCard';
+import { Trait } from '../../../core/Constants';
+
+export default class MandalorianArmor extends UpgradeCard {
+    protected override getImplementationId () {
+        return {
+            id: '3514010297',
+            internalName: 'mandalorian-armor',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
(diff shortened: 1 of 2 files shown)


VJubert (approved):

server/game/cards/02_SHD/upgrades/MandalorianArmor.ts
```diff
@@ -0,0 +1,27 @@
...
+            id: '3514010297',
+            internalName: 'mandalorian-armor',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
+            title: 'Shield attached unit when Mandalorian',
```
adbronico, line 18: nit: I would label it as "If attached unit is a Mandalorian, give it a Shield"

adbronico (changes requested):

AMMayberry1:
In the future can you include the card image in the PR description, makes it easier to verify

server/game/cards/02_SHD/upgrades/MandalorianArmor.ts
```diff
@@ -0,0 +1,27 @@
...
+            targetResolver: {
```
AMMayberry1, line 19: This is incorrect, as currently written it will allow the player to give a shield to _any_ Mandalorian unit on the field.

`targetResolver` should only be used when the player has a choice of targets to select from. In this case the target is known automatically (the attached card), so we can use `immediateEffect` instead and explicitly specify the target. We do that inside of a conditional to check if it's Mandalorian:

```typescript
    condition: (context) => context.source.parentCard?.hasSomeTrait(Trait.Mandalorian),
    onTrue: AbilityHelper.immediateEffects.giveShield((context) => ({ target: context.source.parentCard })),
    onFalse: AbilityHelper.immediateEffects.noAction()
});

PR #61, opened 2024-09-22
Implemented several more cards from the starter decks.

Ennis0:
Obi-Wan Kenobi, Gladiator Star Destroyer, Luke's Lightsaber, Superlaser Technician

PR #289, opened 2024-11-30
SOR #61  Guardian of the Whills

brianries:
Also included a bug fix for Omega - appears it was ignoring aspect penalty on any card so long as a clone wasn't played

image


server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar): void {
+        this.cardsPlayedThisPhaseWatcher = AbilityHelper.stateWatchers.cardsPlayedThisPhase(registrar, this);
+    }
+
+    private isFirstUpgradePlayedOnThis(card: Card) {
+        const playedUpgradesOnThisCardThisPhase = this.cardsPlayedThisPhaseWatcher.getCardsPlayed((playedCardEntry) =>
+            playedCardEntry.card.isUpgrade() &&
+            playedCardEntry.card.parentCard === this &&
```
VJubert, line 25: we never use this, you should use context.source (you have to pass it from caller)

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
```diff
@@ -0,0 +1,55 @@
+describe('Guardian of the Whills', function () {
+    integration(function (contextRef) {
+        describe('Guardian of the Whills\' ability', function () {
+            it('should decrease the cost of the first upgrade played on it by 1 resource, once per round', function () {
```
VJubert, line 4: to check : opponent rules (cf discord)

VJubert (dismissed):

brianries:
Closing this as we need new logic to hand a card becoming a new unit after being played and resetting its state

brianries:
Re-opened as the copy-id logic was added, and tests were updated

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
brianries: Fixed this by using the adjusterSource

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
```diff
@@ -0,0 +1,48 @@
...
+            internalName: 'guardian-of-the-whills'
+        };
+    }
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar): void {
+        this.cardsPlayedThisPhaseWatcher = AbilityHelper.stateWatchers.cardsPlayedThisPhase(registrar, this);
+    }
+
+    private isFirstUpgradePlayedOnThisCopy(card: Card, adjusterSource: Card): boolean {
+        if (!adjusterSource.isUnit()) {
```
AMMayberry1, line 23: Since adjusterSource will always be the host card of the ability we should be okay without this check

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
```diff
@@ -0,0 +1,48 @@
...
+            return false;
+        }
+        const playedUpgradesOnThisCardThisPhase = this.cardsPlayedThisPhaseWatcher.getCardsPlayed((playedCardEntry) =>
```
AMMayberry1, line 26: Can also use `!someCardPlayed()` instead of `getCardsPlayed().length === 0`, slightly simpler

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
```diff
@@ -0,0 +1,48 @@
...
+            playedCardEntry.card.isUpgrade() &&
+            playedCardEntry.card.parentCard === adjusterSource &&
```
AMMayberry1, line 28: We can't directly check `card.parentCard` because the upgrade may have been attached to a new card since it was originally played. We need to add a new property to the watcher which captures the parent card that the card is attached to at the time it is first played. Once implemented, let's also add a test case for that using Survivor's Gauntlet

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
```diff
@@ -0,0 +1,48 @@
...
+            playedCardEntry.card.parentCard.inPlayId === adjusterSource.inPlayId &&
+            playedCardEntry.card !== card
```
AMMayberry1, line 30: We also need to check the play id of the upgrade card itself since it's possible that the same upgrade was returned to hand from play and played again. Test case for this would be good too - could defeat the attached unit and use Razor Crest to return the upgrade to hand

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+                });
+
+                const { context } = contextRef;
+
+                // The first upgrade put on the whills should be 1 cheaper
+                context.player1.clickCard(context.resilient);
+                expect(context.player1).toBeAbleToSelectExactly([context.guardianOfTheWhills, context.systemPatrolCraft]);
+                context.player1.clickCard(context.guardianOfTheWhills);
+                expect(context.guardianOfTheWhills.upgrades).toContain(context.resilient);
+                expect(context.player1.exhaustedResourceCount).toBe(0);
```
AMMayberry1, line 26: Whenever we're testing a numerical increase / decrease I always want to use a higher base number to confirm the exact amount. In this case, can we use an upgrade that costs > 1?

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+                context.player2.passAction();
+
+                // Any further upgrades are full price
+                context.player1.clickCard(context.devotion);
+                expect(context.player1).toBeAbleToSelectExactly([context.guardianOfTheWhills, context.systemPatrolCraft]);
+                context.player1.clickCard(context.guardianOfTheWhills);
+                expect(context.guardianOfTheWhills.upgrades).toContain(context.devotion);
```
AMMayberry1, line 34: I always prefer to do checks for exact sets if possible, it has caught issues for us in the past. So in this case it would be good to use `.toHaveExactUpgradeNames()` throughout instead of `.toContain()`

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
brianries: This was to handle the fact taht the match method has two types of Card passed in, and inPlayId is only available to UnitCards.  This got rid of the warning that adjusterSource was indeed a unit -- is there a better way to do this?

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
brianries: changed this

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
brianries: yes, fixed

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
brianries: Good spot, this will fail for sure since I filtered card === card so I'll skip the first time it was played

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
brianries: Fixed and updated, now tracking parent card and the parent card's playedId at the time of the event as well

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
brianries: Changed all tests to use exactUpgradeNames, agreed thats safer.

server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
brianries: Added a Contract.assertTrue -- works now

server/game/stateWatchers/CardsPlayedThisPhaseWatcher.ts
```diff
@@ -4,11 +4,14 @@ import { StateWatcherRegistrar } from '../core/stateWatcher/StateWatcherRegistra
 import Player from '../core/Player';
 import { TokenOrPlayableCard } from '../core/card/CardTypes';
 import { Card } from '../core/card/Card';
+import { InPlayCard } from '../core/card/baseClasses/InPlayCard';
 
 export interface PlayedCardEntry {
     card: TokenOrPlayableCard;
     inPlayId?: number;
     playedBy: Player;
+    parentCard: InPlayCard;
```
AMMayberry1, line 13: These parent card properties should be optional (`parentCard?`)

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+
+                // Test round ending resets limit
+                context.moveToNextActionPhase();
+                context.player1.clickCard(context.electrostaff);
+                expect(context.player1).toBeAbleToSelectExactly([context.guardianOfTheWhills, context.systemPatrolCraft]);
+                context.player1.clickCard(context.guardianOfTheWhills);
+                expect(context.guardianOfTheWhills).toHaveExactUpgradeNames([context.jetpack.internalName, context.devotion.internalName, context.electrostaff.internalName]);
+                expect(context.player1.exhaustedResourceCount).toBe(1);
+
+                // Test this doesn't accidentally decrease the cost playing on non-Whills units
```
AMMayberry1, line 47: I think this test would be better as the first test in this action phase since the limit is reached already at this point

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+                expect(context.player1).toHavePassAbilityButton();
+                context.player1.clickCard(context.protector);
+                expect(context.player1.exhaustedResourceCount).toBe(6);
+
+                context.player2.passAction();
+
+                context.player1.clickCard(context.protector);
+                expect(context.player1).toBeAbleToSelectExactly([context.guardianOfTheWhills, context.moistureFarmer, context.systemPatrolCraft, context.survivorsGauntlet, context.razorCrestReliableGunship]);
+                context.player1.clickCard(context.guardianOfTheWhills);
+                expect(context.player1.exhaustedResourceCount).toBe(9); // full cost is 3
```
AMMayberry1, line 139: Could we add one more test where the upgrade is only playable with the discount (i.e., holding a 2 cost upgrade but only 1 ready resource) to confirm that the discount is being correctly counted when determining whether a card is legally playable. Can use `context.player1.exhaustResources()` to do it

server/game/stateWatchers/CardsPlayedThisPhaseWatcher.ts
brianries: made optional

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
brianries: Are you talking about the test starting on line 47? we move to next action phase first?

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
brianries: Starting this change in another branch as its far more involved.  We need to handle the fact that payForCost needs to check targets, and checkTarget needs to check payForCost.  Not sure how we'll cleanly do this

test/server/cards/01_SOR/units/GuardianOfTheWhills.spec.ts
AMMayberry1: Oh pfft missed that

AMMayberry1 (approved):

PR #291, opened 2024-11-30
Add docker file to run with ECS w/ dummy env variables

garrettdc2:


Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Use the official Node.js:lst runtime as a base image
+FROM node:lts
```
AMMayberry1, line 2: Is it worth it to use a more specific tag? I'm assuming this will eventually get bumped to another version of node, though not anytime soon I'm sure

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Set the working directory in the container
+WORKDIR /app
+
+# Copy the build over
+COPY ./build/ ./app
+
+# Copy our dependencies
+COPY ./node_modules ./node_modules
```
AMMayberry1, line 11: Do we need to worry about standardizing the way that node_modules is generated in case there are differences depending on the host environment? I was assuming we'd have a build image or something like that, but maybe for node it's not really needed

AMMayberry1 (approved):

PR #296, opened 2024-12-02
Bunch of cards 14

VJubert:
- Calculated Lethality ([SHD 39](https://swudb.com/card/shd/039/calculated-lethality))
- Grenade Strike ([TWI 171](https://swudb.com/card/TWI/171/grenade-strike))
- Soulless One ([TWI 179](https://swudb.com/card/TWI/179/soulless-one))

### DIFF
--- a/server/game/cards/02_SHD/events/CalculatedLethality.ts
+++ b/server/game/cards/02_SHD/events/CalculatedLethality.ts
@@ -26,16 +26,13 @@ export default class CalculatedLethality extends EventCard {
             },
             then: (thenContext) => ({
                 title: 'For each upgrade that was on that unit, give an Experience token to a friendly unit.',
-                immediateEffect: AbilityHelper.immediateEffects.conditional({
-                    condition: thenContext.targets.upgradeAmount != null,
-                    onFalse: AbilityHelper.immediateEffects.noAction(),
-                    onTrue: AbilityHelper.immediateEffects.distributeExperienceAmong({
-                        amountToDistribute: thenContext.targets.upgradeAmount,
-                        cardTypeFilter: WildcardCardType.Unit,
-                        controller: RelativePlayer.Self,
-                        canChooseNoTargets: false,
-                    })
-                }),
+                thenCondition: () => thenContext.targets.upgradeAmount != null,
+                immediateEffect: AbilityHelper.immediateEffects.distributeExperienceAmong({
+                    amountToDistribute: thenContext.targets.upgradeAmount,
+                    cardTypeFilter: WildcardCardType.Unit,
+                    controller: RelativePlayer.Self,
+                    canChooseNoTargets: false,
+                })
             })
         });
     }
--- a/server/game/cards/02_SHD/events/CalculatedLethality.ts
+++ b/server/game/cards/02_SHD/events/CalculatedLethality.ts
@@ -1,0 +1,41 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { RelativePlayer, WildcardCardType } from '../../../core/Constants';
+
+export default class CalculatedLethality extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '0302968596',
+            internalName: 'calculated-lethality',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Defeat a non-leader unit that costs 3 or less',
+            targetResolver: {
+                cardCondition: (card) => card.isNonLeaderUnit() && card.cost <= 3,
+                immediateEffect: AbilityHelper.immediateEffects.sequential([
+                    // TODO: this is a hack to store data before it's defeated.
+                    // once last known state is implemented, we need to read data from that
+                    AbilityHelper.immediateEffects.handler((context) => ({
+                        handler: () => context.targets.upgradeAmount = context.target.upgrades.length
+                    })),
+                    AbilityHelper.immediateEffects.defeat()
+                ])
+            },
+            ifYouDo: (ifYouDoContext) => ({
+                title: 'For each upgrade that was on that unit, give an Experience token to a friendly unit.',
+                immediateEffect: AbilityHelper.immediateEffects.distributeExperienceAmong({
+                    amountToDistribute: ifYouDoContext.targets.upgradeAmount,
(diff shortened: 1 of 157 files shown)


server/game/cards/02_SHD/events/CalculatedLethality.ts
```diff
@@ -0,0 +1,41 @@
...
+                immediateEffect: AbilityHelper.immediateEffects.sequential([
+                    // TODO: this is a hack to store data before it's defeated.
+                    // once last known state is implemented, we need to read data from that
+                    AbilityHelper.immediateEffects.handler((context) => ({
+                        handler: () => context.targets.upgradeAmount = context.target.upgrades.length
+                    })),
+                    AbilityHelper.immediateEffects.defeat()
+                ])
+            },
+            ifYouDo: (ifYouDoContext) => ({
```
AMMayberry1, line 27: This actually should be a "then" not an "ifYouDo" because of this:

image

server/game/cards/02_SHD/events/CalculatedLethality.ts
```diff
@@ -0,0 +1,41 @@
...
+                title: 'For each upgrade that was on that unit, give an Experience token to a friendly unit.',
+                immediateEffect: AbilityHelper.immediateEffects.distributeExperienceAmong({
+                    amountToDistribute: ifYouDoContext.targets.upgradeAmount,
+                    cardTypeFilter: WildcardCardType.Unit,
+                    controller: RelativePlayer.Self,
+                    canChooseNoTargets: false,
```
AMMayberry1, line 33: This is redundant since we have `canDistributeLess: false`

server/game/cards/03_TWI/events/GrenadeStrike.ts
```diff
@@ -0,0 +1,34 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '4042866439',
+            internalName: 'grenade-strike',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Defeat a non-leader unit that costs 3 or less',
```
AMMayberry1, line 15: Wrong title

server/game/gameSystems/DistributeExperienceSystem.ts
```diff
@@ -0,0 +1,27 @@
...
+    public override readonly name = 'distributeExperience';
+
+    public override promptType = StatefulPromptType.DistributeExperience;
+
+    protected override generateEffectSystem(target: Card = null, amount = 1): GiveExperienceSystem {
+        return new GiveExperienceSystem({ target, amount });
+    }
+
+    protected override canDistributeLessDefault(): boolean {
+        return this.properties.canDistributeLess;
```
AMMayberry1, line 25: This should return either true or false since it's what is used as the default value if properties.canDistributeLess is null. I think Calculated Lethality is the only card that uses it so we can just default to false and remove that from the card implementation

test/server/cards/02_SHD/events/CalculatedLethality.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+                reset();
+
+                context.player1.clickCard(context.calculatedLethality);
+
+                // fifth brother is the only legal target and corellian is the only friendly unit, all system should be resolved automatically
+
+                expect(context.player2).toBeActivePlayer();
+                expect(context.fifthBrother).toBeInZone('discard');
+                expect(context.corellianFreighter).toHaveExactUpgradeNames(['experience', 'experience', 'experience']);
+            });
```
AMMayberry1, line 71: Let's add a TODO for testing the TIE Phantom case

test/server/cards/02_SHD/events/CalculatedLethality.spec.ts
```diff
@@ -0,0 +1,74 @@
+describe('Calculated Lethality', function () {
+    integration(function (contextRef) {
+        describe('Calculated Lethality\'s ability', function () {
+            beforeEach(function () {
+                contextRef.setupTest({
```
AMMayberry1, line 5: Might as well move the setup into the test case since there's only one case. Same for other test files

server/game/cards/02_SHD/events/CalculatedLethality.ts
```diff
@@ -24,14 +24,13 @@ export default class CalculatedLethality extends EventCard {
                     AbilityHelper.immediateEffects.defeat()
                 ])
             },
-            ifYouDo: (ifYouDoContext) => ({
+            then: (thenContext) => ({
```
AMMayberry1, line 27: This currently might fail if no target is available for defeat, since `thenContext.targets.upgradeAmount` will not be set. I think we need to add a `thenCondition` checking if no target was selected at all, and we should have a test confirming it works.

server/game/gameSystems/DistributeExperienceSystem.ts
```diff
@@ -22,6 +22,6 @@ export class DistributeExperienceSystem<TContext extends AbilityContext = Abilit
     }
 
     protected override canDistributeLessDefault(): boolean {
-        return this.properties.canDistributeLess;
+        return this.properties.canDistributeLess || false;
```
AMMayberry1, line 25: Still doesn't match the other implementations of this method, and it doesn't really make sense for it to check the properties since it's supposed to be providing the default in case the property isn't present. Also, it's checking `this.properties` instead of using `generatePropertiesFromContext`, so it might be getting a wrong value. Should just return `false`

server/game/cards/02_SHD/events/CalculatedLethality.ts
```diff
@@ -26,11 +26,15 @@ export default class CalculatedLethality extends EventCard {
             },
             then: (thenContext) => ({
                 title: 'For each upgrade that was on that unit, give an Experience token to a friendly unit.',
-                immediateEffect: AbilityHelper.immediateEffects.distributeExperienceAmong({
-                    amountToDistribute: thenContext.targets.upgradeAmount,
-                    cardTypeFilter: WildcardCardType.Unit,
-                    controller: RelativePlayer.Self,
-                    canChooseNoTargets: false,
+                immediateEffect: AbilityHelper.immediateEffects.conditional({
```
AMMayberry1, line 29: Simpler to use `thenCondition` instead of wrapping the whole thing in conditional

AMMayberry1 (approved):

PR #298, opened 2024-12-03
create set code map

danbastin:
Creates a map of SetCode to card id to be used for mapping swudb deck lists to card data.

### DIFF
--- a/bin/fetchdata.js
+++ b/bin/fetchdata.js
@@ -7,6 +7,7 @@ const fs = require('fs/promises');
 const mkdirp = require('mkdirp');
 const path = require('path');
 const cliProgress = require('cli-progress');
+const { set } = require('zod');
 
 const pathToJSON = path.join(__dirname, '../test/json/');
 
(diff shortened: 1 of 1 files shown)


bin/fetchdata.js
```diff
@@ -148,7 +161,7 @@ function getUniqueCards(cards) {
     playableCardTitles.sort();
 
     const uniqueCards = [...uniqueCardsMap].map(([internalName, card]) => card);
-    return { uniqueCards, cardMap, playableCardTitles, duplicatesWithSetCode };
+    return { uniqueCards, cardMap, playableCardTitles, duplicatesWithSetCode, setCodeMap };
```
AMMayberry1, line 164: Do you think it would be worth sorting the map entries by set and number to improve human readability? TS maps keep insertion order so it's doable by converting to an array and then back to a map, just not sure if we think there's much benefit

AMMayberry1 (approved):

PR #299, opened 2024-12-03
Add Choose Sides implementation

Frugghi:
image

### DIFF
--- a/test/server/cards/02_SHD/events/ChooseSides.spec.ts
+++ b/test/server/cards/02_SHD/events/ChooseSides.spec.ts
@@ -1,60 +1,94 @@
 describe('Choose Sides', function() {
     integration(function(contextRef) {
         describe('Choose Sides\'s event ability', function() {
-            beforeEach(function() {
-                contextRef.setupTest({
-                    phase: 'action',
-                    player1: {
-                        hand: ['choose-sides'],
-                        spaceArena: ['cartel-spacer'],
-                        leader: { card: 'boba-fett#daimyo', deployed: true },
-                        resources: 30,
-                    },
-                    player2: {
-                        hand: ['cantina-bouncer'],
-                        leader: { card: 'sabine-wren#galvanized-revolutionary', deployed: true },
-                    }
+            describe('when there are no enemy non-leader units in play', function() {
+                beforeEach(function() {
+                    contextRef.setupTest({
+                        phase: 'action',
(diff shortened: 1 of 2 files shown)


test/server/cards/02_SHD/events/ChooseSides.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+                context.player1.clickCard(context.cartelSpacer);
+                context.player2.passAction();
+
+                context.player1.clickCard(context.chooseSides);
+
+                expect(context.player2).toBeActivePlayer();
+                expect(context.cartelSpacer.controller).toBe(context.player2Object);
+                expect(context.cantinaBouncer.controller).toBe(context.player1Object);
+
+                reset();
```
VJubert, line 57: last reset is not needed

test/server/cards/02_SHD/events/ChooseSides.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+            });
+
+            it('should exchange control of a friendly and enemy units', () => {
+                const { context } = contextRef;
+
+                const reset = () => {
+                    context.player1.moveCard(context.chooseSides, 'hand');
+                };
+
+                // Scenario 1: Only a friendly non-leader unit is in play
```
AMMayberry1, line 27: This is maybe a case where I'd be supportive of having separate test cases just for the sake of readability, since each case requires different board state

test/server/cards/02_SHD/events/ChooseSides.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+                context.player2.clickCard(context.cartelSpacer);
+
+                context.player1.clickCard(context.chooseSides);
+
+                expect(context.player2).toBeActivePlayer();
+                expect(context.cantinaBouncer.controller).toBe(context.player2Object);
+
+                reset();
+
+                // Scenario 3: Exchange control of a friendly and enemy non-leader units
```
AMMayberry1, line 46: An important test for this card is to see what happens if you swap two copies of the same unique unit - we want to make sure that the timing is correct so that the game doesn't think one player controls both copies at some time point and invoke the unique rule to defeat one of them. We could do that as a fourth case or replace this one with it, either is good

AMMayberry1 (approved):

PR #300, opened 2024-12-04
Implementing Valiant Assault Ship

supercodepoet:
image

### DIFF
--- a/server/game/cards/02_SHD/units/ValiantAssaultShip.ts
+++ b/server/game/cards/02_SHD/units/ValiantAssaultShip.ts
@@ -12,7 +12,7 @@ export default class ValiantAssaultShip extends NonLeaderUnitCard {
 
     public override setupCardAbilities () {
         this.addOnAttackAbility({
-            title: 'If the defending player controls more resources than you, this unit gets +2/+0 for this attack.',
+            title: 'If the defending player controls more resources than you, this unit gets +2/+0 for this attack',
             targetResolver: {
                 controller: RelativePlayer.Self,
                 cardTypeFilter: WildcardCardType.Unit,
--- a/server/game/cards/02_SHD/units/ValiantAssaultShip.ts
+++ b/server/game/cards/02_SHD/units/ValiantAssaultShip.ts
@@ -1,6 +1,5 @@
 import AbilityHelper from '../../../AbilityHelper';
 import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
-import { RelativePlayer, WildcardCardType, ZoneName } from '../../../core/Constants';
 
 export default class ValiantAssaultShip extends NonLeaderUnitCard {
     protected override getImplementationId () {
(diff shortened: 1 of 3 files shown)


server/game/cards/02_SHD/units/ValiantAssaultShip.ts
```diff
@@ -0,0 +1,35 @@
...
+    public override setupCardAbilities () {
+        this.addOnAttackAbility({
+            title: 'If the defending player controls more resources than you, this unit gets +2/+0 for this attack',
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: ZoneName.SpaceArena,
+                immediateEffect: AbilityHelper.immediateEffects.conditional({
+                    condition: (context) => {
+                        return context.source.controller.resources.length < context.source.controller.opponent.resources.length;
```
AMMayberry1, line 22: Can do this all in one line as:

```typescript:
condition: (context) => context.source.controller.resources.length < context.source.controller.opponent.resources.length
```

test/server/cards/02_SHD/units/ValiantAssaultShip.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+                    player1: {
+                        spaceArena: ['valiant-assault-ship']
+                    },
+                    player2: {
+                        spaceArena: [{ card: 'system-patrol-craft', upgrades: ['shield'] }]
+                    }
+                });
+            });
+
+            it('should give a unit +2/+0 for this attack while ignoring the Sentinel', function () {
```
AMMayberry1, line 16: nit: let's add something like "when the opponent has more resources" just to make it super explicit. that makes it easier for debugging if someone accidentally breaks this due to a random change and needs to figure out what the test is doing. Same for other test names

test/server/cards/02_SHD/units/ValiantAssaultShip.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+                expect(context.player1).toHaveEnabledPromptButton('If the defending player controls more resources than you, this unit gets +2/+0 for this attack');
+                expect(context.player1).toHaveEnabledPromptButton('Saboteur: defeat all shields');
+                expect(context.valiantAssaultShip.exhausted).toBe(true);
+
+                context.player1.clickPrompt('If the defending player controls more resources than you, this unit gets +2/+0 for this attack');
+                expect(context.systemPatrolCraft.damage).toBe(0);
+                expect(context.p2Base.damage).toBe(5);
+            });
+
+            it('should give a unit +2/+0 for this attack while attacking the Sentinel through the shield', function () {
```
AMMayberry1, line 35: We can omit this test since we've shown with the other that Saboteur is being applied correctly, don't need to check both of its effects

server/game/cards/02_SHD/units/ValiantAssaultShip.ts
```diff
@@ -0,0 +1,35 @@
...
+        return {
+            id: '7922308768',
+            internalName: 'valiant-assault-ship'
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addOnAttackAbility({
+            title: 'If the defending player controls more resources than you, this unit gets +2/+0 for this attack',
+            targetResolver: {
```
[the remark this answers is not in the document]
supercodepoet: ahhh, yes, good call

server/game/cards/02_SHD/units/ValiantAssaultShip.ts
supercodepoet: yep, good catch, that was left over from debugging

test/server/cards/02_SHD/units/ValiantAssaultShip.spec.ts
supercodepoet: 👍

AMMayberry1 (approved):

PR #301, opened 2024-12-04
Coordinate keyword

AMMayberry1:
Added support for the Coordinate keyword, using similar mechanisms to the Bounty keyword. Added several test units to confirm that the timing works and that triggered and constant abilities work.

Two special cases are not covered in this PR and will require additional work:
- Coordinate on leaders (Ahsoka, Padme)
- "Gain coordinate" (For the Republic)

### DIFF
--- a/test/server/core/abilities/keyword/Raid.spec.ts
+++ b/test/server/core/abilities/keyword/Raid.spec.ts
@@ -80,7 +80,5 @@ describe('Raid keyword', function() {
                 expect(context.p2Base.damage).toBe(4);
             });
         });
-
-        // TODO test that a card that attacked and then is bounced back to hand (i.e. Waylay) doesn't receive a second Raid buff
     });
 });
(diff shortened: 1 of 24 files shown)


test/server/core/abilities/keyword/Raid.spec.ts
```diff
@@ -80,7 +80,5 @@ describe('Raid keyword', function() {
                 expect(context.p2Base.damage).toBe(4);
             });
         });
-
-        // TODO test that a card that attacked and then is bounced back to hand (i.e. Waylay) doesn't receive a second Raid buff
```
adbronico, line 84: I'm OK removing this if we're confident it's not going to happen, but do we have a test to cover this and the similar notes in Restore/Shielded?

adbronico (approved):
looks clean

test/server/core/abilities/keyword/Raid.spec.ts
AMMayberry1: Mentioned in chat but Chebato did implement these tests, he just missed the comments in some cases

PR #303, opened 2024-12-05
Implementing Guarding the Way

supercodepoet:
image

I didn't see in the tests for a way to check for Sentinel other than just attacking so that is the direction I went.

### DIFF
--- a/server/game/cards/03_TWI/events/GuardingTheWay.ts
+++ b/server/game/cards/03_TWI/events/GuardingTheWay.ts
@@ -12,7 +12,7 @@ export default class GuardingTheWay extends EventCard {
 
     public override setupCardAbilities() {
         this.setEventAbility({
-            title: 'Give Sentinel for the phase and optionally +2/+2 if you have initiative',
+            title: 'Give Sentinel for the phase. If you have the initiative, also give that unit +2/+2 for this phase.',
             targetResolver: {
                 cardTypeFilter: WildcardCardType.Unit,
                 immediateEffect: AbilityHelper.immediateEffects.simultaneous([
--- a/server/game/cards/03_TWI/events/GuardingTheWay.ts
+++ b/server/game/cards/03_TWI/events/GuardingTheWay.ts
@@ -1,0 +1,35 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { KeywordName, WildcardCardType } from '../../../core/Constants';
+
+export default class GuardingTheWay extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '2359136621',
+            internalName: 'guarding-the-way',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Give Sentinel for the phase and optionally +2/+2 if you have initiative',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                immediateEffect: AbilityHelper.immediateEffects.simultaneous([
+                    AbilityHelper.immediateEffects.forThisPhaseCardEffect({
+                        effect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Sentinel)
+                    }),
+                    AbilityHelper.immediateEffects.conditional({
+                        condition: (context) => context.source.controller.hasInitiative(),
+                        onTrue: AbilityHelper.immediateEffects.forThisPhaseCardEffect({
+                            effect: AbilityHelper.ongoingEffects.modifyStats({ power: 2, hp: 2 })
+                        }),
+                        onFalse: AbilityHelper.immediateEffects.noAction()
+                    })
+                ])
+            }
+        });
+    }
+}
+
+GuardingTheWay.implemented = true;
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/events/GuardingTheWay.ts
```diff
@@ -0,0 +1,35 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '2359136621',
+            internalName: 'guarding-the-way',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Give Sentinel for the phase and optionally +2/+2 if you have initiative',
```
AMMayberry1, line 15: nit, wording: "optional" for us means "you can choose not to," which is not the case here

test/server/cards/03_TWI/events/GuardingTheWay.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+
+                context.player1.clickCard(context.guardingTheWay);
+                expect(context.player1).toBeAbleToSelectExactly([context.lukeSkywalkerJediKnight, context.chewbaccaPykesbane]);
+
+                context.player1.clickCard(context.lukeSkywalkerJediKnight);
+                expect(context.lukeSkywalkerJediKnight.getPower()).toBe(8);
+                expect(context.lukeSkywalkerJediKnight.getHp()).toBe(9);
+                expect(context.player2).toBeActivePlayer();
+
+                context.player2.clickCard(context.chewbaccaPykesbane);
```
AMMayberry1, line 26: We should add a comment explaining that this attack is validating the sentinel keyword

test/server/cards/03_TWI/events/GuardingTheWay.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.chewbaccaPykesbane);
+
+                context.player2.clickCard(context.guardingTheWay);
+                expect(context.player2).toBeAbleToSelectExactly([context.lukeSkywalkerJediKnight, context.chewbaccaPykesbane]);
+
+                context.player2.clickCard(context.lukeSkywalkerJediKnight);
```
AMMayberry1, line 54: Also need to do the Sentinel check in this test to confirm that it was applied correctly

test/server/cards/03_TWI/events/GuardingTheWay.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                    },
+                    player2: {
+                        hand: ['guarding-the-way'],
+                        groundArena: ['luke-skywalker#jedi-knight']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.chewbaccaPykesbane);
```
AMMayberry1, line 49: What's the reason for starting Chewbacca in hand?

test/server/cards/03_TWI/events/GuardingTheWay.spec.ts
```diff
@@ -0,0 +1,60 @@
+describe('Guarding the Way', function () {
+    integration(function (contextRef) {
+        describe('Guarding the Way\'s ability', function () {
+            it('should give a unit Sentinel and +2/+2 for the phase', function () {
```
AMMayberry1, line 4: nit, clarity: should include "when the controller has initiative"

test/server/cards/03_TWI/events/GuardingTheWay.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                context.player2.clickCard(context.chewbaccaPykesbane);
+                expect(context.chewbaccaPykesbane.damage).toBe(8);
+                expect(context.lukeSkywalkerJediKnight.damage).toBe(4);
+
+                context.moveToNextActionPhase();
+                expect(context.lukeSkywalkerJediKnight.getPower()).toBe(6);
+                expect(context.lukeSkywalkerJediKnight.getHp()).toBe(7);
+            });
+
+            it('should give a unit Sentinel for the phase', function () {
```
AMMayberry1, line 35: nit, clarity: should include "when the controller does not have initiative"

server/game/cards/03_TWI/events/GuardingTheWay.ts
supercodepoet: Would it be better to just model the card language? 

`Give Sentinel for the phase. If you have the initiative, also give that unit +2/+2 for this phase.`

test/server/cards/03_TWI/events/GuardingTheWay.spec.ts
supercodepoet: does this work?

`should give a unit Sentinel for the phase and when the controller has initiative also give it +2/+2`

test/server/cards/03_TWI/events/GuardingTheWay.spec.ts
supercodepoet: 👍

test/server/cards/03_TWI/events/GuardingTheWay.spec.ts
supercodepoet: I wanted to simulate play but I can start Chewbacca on the board letting it swing to base then having him swing into Luke when he gains Sentinel.

AMMayberry1 (approved):

PR #304, opened 2024-12-07
Bunch of cards 15

VJubert:
- Ephant Mon ([SHD 088](https://swudb.com/card/SHD/088/ephant-mon))
- Obi-Wan Kenobi ([TWI 003](https://swudb.com/card/twi/003/obi-wan-kenobi))
- Quinlan Vos ([TWI 018](https://swudb.com/card/TWI/018/quinlan-vos))
- Calculating MagnaGuard ([TWI 033](https://swudb.com/card/twi/033/calculating-magnaguard))
- Jyn Erso ([TWI 143](https://swudb.com/card/twi/143/jyn-erso))
- Luminara Unduli ([TWI 050](https://swudb.com/card/TWI/050/luminara-unduli))
- Resolute ([TWI 151](https://swudb.com/card/TWI/151/resolute))
- Steela Gerrera ([TWI 146](https://swudb.com/card/TWI/146/steela-gerrera))
- Foresight ([TWI 068](https://swudb.com/card/TWI/068/foresight))
- General's Blade ([TWI 121](https://swudb.com/card/twi/121/generals-blade))


server/game/cards/03_TWI/leaders/ObiWanKenobiPatientMentor.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    protected override setupLeaderUnitSideAbilities() {
+        this.addOnAttackAbility({
+            title: 'Heal 1 damage from a unit',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                immediateEffect: AbilityHelper.immediateEffects.heal({ amount: 1 })
+            },
+            ifYouDo: (ifYouDoContext) => ({
```
VJubert, line 31: @AMMayberry1 I have no idea how to fix ifyoudo with healsystem, can you help on this one ?

server/game/cards/01_SOR/leaders/GrandAdmiralThrawnPatientAndInsightful.ts
```diff
@@ -0,0 +1,84 @@
...
+    protected override setupLeaderSideAbilities() {
+        this.addLookTrigger();
+
+        this.addActionAbility({
+            title: 'Reveal the top card of any player\'s deck',
+            cost: [AbilityHelper.costs.abilityResourceCost(1), AbilityHelper.costs.exhaustSelf()],
+            targetResolver: {
+                mode: TargetMode.Select,
+                choices: {
+                    ['Reveal your top deck cards']: AbilityHelper.immediateEffects.reveal((context) => ({
```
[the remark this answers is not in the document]
VJubert: You have not to reveal the top cards of deck, opponent should not see them
I can't use TargetMode.Player here, lookAt interface does not fit PlayerTargetSystem

test/server/cards/03_TWI/upgrades/GeneralsBlade.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                phase: 'action',
+                player1: {
+                    hand: ['wampa', 'battlefield-marine', 'smugglers-aid'],
+                    groundArena: [{ card: 'obiwan-kenobi#following-fate', upgrades: ['generals-blade'] }],
+                    base: 'echo-base',
+                    leader: 'cassian-andor#dedicated-to-the-rebellion'
+                },
+            });
+
+            const { context } = contextRef;
```
AMMayberry1, line 14: This came up in chat recently - can we play units with smuggle and from discard (palp's return) to confirm that the discount also applies in those cases?

scripts/fetchdata.js
```diff
@@ -124,7 +124,7 @@ function getUniqueCards(cards) {
     const seenNames = [];
     var duplicatesWithSetCode = {};
     const uniqueCardsMap = new Map();
-    const setNumber = new Map([['SOR', 1], ['SHD', 2], ['TWI', 3]]);
+    const setNumber = new Map([['SOR', 1], ['SHD', 2], ['TWI', 3], ['JTL', 4]]);
```
AMMayberry1, line 127: Let's hold off until at least the meta cards from set 1 - 3 are done :) we're pretty close

server/game/gameSystems/DistributeHealingSystem.ts
```diff
@@ -29,6 +29,6 @@ export class DistributeHealingSystem<TContext extends AbilityContext = AbilityCo
     }
 
     protected override getDistributedAmountFromEvent(event: any): number {
-        return event.damageRemoved;
+        return event.damageRemoved || 0;
```
AMMayberry1, line 32: Just curious, where was this coming up? If it's needed here it's probably needed for damage and experience distribution systems as well.

test/server/cards/01_SOR/events/Vigilance.spec.ts
```diff
@@ -132,7 +132,7 @@ describe('Vigilance', function() {
 
                     // heal 0 damage from base
                     context.player1.clickPrompt('Heal 5 damage from a base.');
-                    context.player1.clickCard(context.p1Base);
+                    // both base are not damaged, no base selectable
```
AMMayberry1, line 135: Hmm I think this isn't the behavior we expect - it should be possible to choose a target for healing even if damage is 0. Does the clickCard fail here?

scripts/fetchdata.js
VJubert: that will not release set 4's json, it's data for next release ^^

test/server/cards/01_SOR/events/Vigilance.spec.ts
VJubert: Yes, update from Obi-wan require to do not do excessive healing, you still can use card

server/game/gameSystems/DistributeHealingSystem.ts
VJubert: redemption do excessive healing with distribute prompt and crash the test

scripts/fetchdata.js
AMMayberry1: Ahh okay, all good then! Thanks

server/game/gameSystems/DistributeHealingSystem.ts
AMMayberry1: Hmm that's interesting, I just did that so there may be a bug. I'll take a look, I think this shouldn't be necessary ideally

test/server/cards/01_SOR/events/Vigilance.spec.ts
AMMayberry1: Ah okay I figured out what's going on here. I forgot that we have a targeting optimization where it will skip targeting if _none_ of the possible targets would have a game state changing effect if they were selected. So like compare these two situations:

- case 1: Wampa and AT-ST are on the field, Wampa has 1 damage and AT-ST has 0. A player plays Midnight Repairs, they are allowed to select either Wampa or AT-ST for healing. They can legally choose to put some heal on AT-ST but nothing will change.

- case 2: same as above except Wampa has 0 damage. In this case, according to SWU rules the player is still legally allowed to "heal" both units but nothing will happen when they do. So, we just skip targeting completely since nothing the player does will matter.

In the future we might choose to turn this optimization off during testing like we did with `autoSingleTarget`, but for now that is the reason we are seeing some tests change now that you have update `HealSystem.canAffect()` to correctly report `false` when there is no healing to be done

test/server/cards/01_SOR/events/Repair.spec.ts
```diff
@@ -39,16 +39,6 @@ describe('Repair', function() {
                 expect(context.p1Base.damage).toBe(0);
             });
 
-            it('can select a target with no damage', function () {
```
AMMayberry1, line 42: In line with my response [here](https://github.com/SWU-Karabast/forceteki/pull/304/files#r1931338009), instead of removing this test can we change it to explicitly check that no prompt happens

server/game/gameSystems/DistributeHealingSystem.ts
AMMayberry1: Okay I think I know what's going on - since the event gets cancelled, `damageRemoved` never gets set.

Instead of doing this check could we instead update `HealSystem.addPropertiesToEvent` so that `event.damageRemoved` is initialized to 0? That way if we need to do this check in other places we don't have to remember to do the `|| 0` every time

server/game/cards/01_SOR/leaders/HeraSyndullaSpectreTwo.ts
```diff
@@ -15,6 +15,7 @@ export default class HeraSyndullaSpectreTwo extends LeaderUnitCard {
             title: 'Ignore the aspect penalty on Spectre cards you play',
             targetController: RelativePlayer.Self,
             ongoingEffect: AbilityHelper.ongoingEffects.ignoreAllAspectPenalties({
+                limit: AbilityHelper.limit.unlimited(),
```
AMMayberry1, line 18: Ah okay, I didn't think about that this change would affect cards like Hera. I will investigate whether it would be majorly disruptive to maybe just remove the default altogether, it feels like people will make mistaken assumptions about what the default is and that will lead to bugs.

server/game/core/cost/CostAdjuster.ts
```diff
@@ -101,7 +103,7 @@ export class CostAdjuster {
             properties.playingTypes &&
             (Array.isArray(properties.playingTypes) ? properties.playingTypes : [properties.playingTypes]);
 
-        this.limit = properties.limit;
+        this.limit = properties.limit || new PerGameAbilityLimit(1);
```
AMMayberry1, line 106: Yeah let's just revert this change then I think we're good, sorry for the back-and-forth

AMMayberry1 (approved):

PR #307, opened 2024-12-07
Implementing Ahsoka's Padawan Lightsaber

supercodepoet:
image

### DIFF
--- a/server/game/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.ts
+++ b/server/game/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.ts
@@ -18,11 +18,11 @@ export default class AhsokasPadawanLightsaber extends UpgradeCard {
             title: 'Attack with a unit',
             optional: true,
             targetResolver: {
-                immediateEffect: AbilityHelper.immediateEffects.conditional((context) => ({
-                    condition: context.source.parentCard?.title === 'Ahsoka Tano',
+                immediateEffect: AbilityHelper.immediateEffects.conditional({
+                    condition: (context) => context.source.parentCard?.title === 'Ahsoka Tano',
                     onTrue: AbilityHelper.immediateEffects.attack(),
                     onFalse: AbilityHelper.immediateEffects.noAction()
-                }))
+                })
             }
         });
     }
--- a/server/game/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.ts
+++ b/server/game/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.ts
@@ -15,14 +15,15 @@ export default class AhsokasPadawanLightsaber extends UpgradeCard {
         this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
 
         this.addWhenPlayedAbility({
-            title: 'Attack with Ahsoka Tano',
+            title: 'Attack with a unit',
             optional: true,
-            immediateEffect: AbilityHelper.immediateEffects.conditional((context) => ({
-                target: context.source.parentCard,
-                condition: context.source.parentCard?.title === 'Ahsoka Tano',
-                onTrue: AbilityHelper.immediateEffects.attack(),
-                onFalse: AbilityHelper.immediateEffects.noAction()
-            }))
+            targetResolver: {
+                immediateEffect: AbilityHelper.immediateEffects.conditional((context) => ({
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.ts
```diff
@@ -0,0 +1,30 @@
...
+            id: '0741296536',
+            internalName: 'ahsokas-padawan-lightsaber'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
+            title: 'Attack with Ahsoka Tano',
```
AMMayberry1, line 18: This implementation is wrong, the card ability is to attack with any unit not necessarily Ahsoka

supercodepoet: Alright, this should be fixed up and test passing. The one question I had: Should the `optional` be on the When Played or should it be on the attack? The attack is technically optional, but seems the user would expect to pass the whole When Played trigger instead of selecting a card then passing on the attack.

server/game/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
+            title: 'Attack with a unit',
+            optional: true,
+            targetResolver: {
+                immediateEffect: AbilityHelper.immediateEffects.conditional((context) => ({
```
AMMayberry1, line 21: Instead of using a context handler for the whole immediateEffect, you could just use one for conditional like `condition: (context) => context.source.parentCard?.title === 'Ahsoka Tano'`

test/server/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+                        groundArena: ['r2d2#ignoring-protocol']
+                    }
+                });
+            });
+
+            it('initiates an attack when played on Ahsoka Tano', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.ahsokasPadawanLightsaber);
+                expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine, context.ahsokaTanoAlwaysReadyForTrouble, context.r2d2IgnoringProtocol]);
```
AMMayberry1, line 22: If you want you can omit the subtitle and just use the title for the unit object - e.g. `context.ahsokaTano`. Unless there are two cards in play with the same title and different subtitles, in which case you have to specify subtitle

test/server/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+                expect(context.ahsokaTanoAlwaysReadyForTrouble).toHaveExactUpgradeNames(['ahsokas-padawan-lightsaber']);
+                expect(context.ahsokaTanoAlwaysReadyForTrouble.exhausted).toBe(false);
+
+                context.player1.clickCard(context.p2Base);
+                expect(context.battlefieldMarine.exhausted).toBe(true);
+                expect(context.p2Base.damage).toBe(3);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('initiates an attack when played on Ahsoka Tano but player selects not to attack', function () {
```
AMMayberry1, line 39: We generally don't bother adding tests that the pass button works, just confirming that it's there with an expect() is good enough

server/game/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.ts
supercodepoet: good deal, fixed up

test/server/cards/03_TWI/upgrades/AhsokasPadawanLightsaber.spec.ts
supercodepoet: gotcha, that is helpful, thanks

AMMayberry1 (approved):

PR #309, opened 2024-12-07
New multiSelectCardCondition, TWI #153 Bold Resistance Card and sameDiscardPile cleanup

brianries:
Three things in here:

- Created a new multiSelectCardCondition property for handling groups of selected cards
- Implemented the Bold Resistance Card  (TWI # 153)
- Removed the sameDiscardPile property and adjusted Restock to use the new code

image

### DIFF
--- a/server/game/cards/03_TWI/events/BoldResistance.ts
+++ b/server/game/cards/03_TWI/events/BoldResistance.ts
@@ -5,15 +5,12 @@ import { CardsPlayedThisPhaseWatcher } from '../../../stateWatchers/CardsPlayedT
 import { Card } from '../../../core/card/Card';
 
 export default class BoldResistance extends EventCard {
-    private cardsPlayedThisPhaseWatcher: CardsPlayedThisPhaseWatcher;
-
     protected override getImplementationId() {
         return {
             id: '8022262805',
             internalName: 'bold-resistance'
         };
     }
-
 
     public override setupCardAbilities() {
         this.setEventAbility({
--- a/server/game/cards/03_TWI/events/BoldResistance.ts
+++ b/server/game/cards/03_TWI/events/BoldResistance.ts
@@ -1,7 +1,6 @@
 import { EventCard } from '../../../core/card/EventCard';
 import AbilityHelper from '../../../AbilityHelper';
 import { TargetMode, WildcardCardType } from '../../../core/Constants';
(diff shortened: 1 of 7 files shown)


server/game/cards/03_TWI/events/BoldResistance.ts
```diff
@@ -0,0 +1,60 @@
+import { EventCard } from '../../../core/card/EventCard';
+import AbilityHelper from '../../../AbilityHelper';
+import { TargetMode, WildcardCardType } from '../../../core/Constants';
+import { CardsPlayedThisPhaseWatcher } from '../../../stateWatchers/CardsPlayedThisPhaseWatcher';
+import { Card } from '../../../core/card/Card';
+
+export default class BoldResistance extends EventCard {
+    private cardsPlayedThisPhaseWatcher: CardsPlayedThisPhaseWatcher;
```
VJubert, line 8: this watcher is not use ?

server/game/cards/03_TWI/events/BoldResistance.ts
```diff
@@ -0,0 +1,60 @@
...
+            // If no traits remain, there is no common trait
+            if (intersectingTraits.size === 0) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    // eslint-disable-next-line @stylistic/keyword-spacing
+    private static IntersectingSet = class<T> extends Set<T> {
```
VJubert, line 49: you can't use this method from prototype : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection ?

server/game/cards/03_TWI/events/BoldResistance.ts
brianries: thanks, cut&paste error missed it

server/game/cards/03_TWI/events/BoldResistance.ts
```diff
@@ -0,0 +1,56 @@
...
```
AMMayberry1, line 45: Now that we know we're keeping this, do you mind moving it to Helpers.ts?

As far as using it after moving, a note in case you haven't seen it yet - our usage convention for stuff in helper files is to import like `import * as Helpers from ...`  and then use it with the name `Helpers.IntersectingSet`

brianries: Thanks -- wasn't sure where we put our custom helper functions, Moved it

AMMayberry1 (approved):

PR #310, opened 2024-12-07
BokatanKyrzeFightingForMandalore (SHD) card implementation 

Twet1:
Implements the logic for BokatanKyrzeFightingForMandalore.

Unit tests include cases for when each player has correct damage on base to trigger when defeated response. 

Please review and provide feedback. First implementation for Karabast!

### DIFF
--- a/test/server/cards/02_SHD/units/BoKatanKryzeFightingForMandalore.spec.ts
+++ b/test/server/cards/02_SHD/units/BoKatanKryzeFightingForMandalore.spec.ts
@@ -10,71 +10,61 @@ describe('Bo-KatanKryze', function () {
                     },
                     player2: {
                         groundArena: ['battlefield-marine'],
+                        hand: ['repair'],
                         base: { card: 'echo-base', damage: 12 }
                     }
                 });
             });
 
-            it('No cards drawn if damage on both bases is less than 15', function () {
+            it('Correct card draw depending on base(s) damage', function() {
                 const { context } = contextRef;
 
+                const reset = () => {
+                    context.player1.moveCard(context.bokatanKryze, 'groundArena');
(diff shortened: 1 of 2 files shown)


test/server/cards/02_SHD/units/BoKatanKryzeFightingForMandalore.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+                context.player1.clickCard(context.bokatanKryze);
+                context.player1.clickCard(context.p2Base);
+                expect(context.p2Base.damage).toBe(15);
+
+                context.player2.clickCard(context.battlefieldMarine);
+                context.player2.clickCard(context.bokatanKryze);
+                expect(context.player1.hand.length).toBe(1);
+            });
+
+            it(' 2 cards drawn if damage on both bases exceeds 15', function () {
```
AMMayberry1, line 42: We have this issue where the Jasmine tests are super heavy and slow, and we have finite Github Actions time in the free tier. So, unfortunately we have to consolidate multiple cases into one where possible.

On-attack abilities are a really good candidate for this because we can reset the attacker by setting exhausted = false in between attacks. See the Mace Windu Party Crasher tests for an example of a typical pattern we use. We could use a card like Repair to bring basr damage back below 15 if needed for a test's setup.

Twet1: No worries, Ill finish what i'm doing on another branch and make the updates to this! 
If I am reading correctly, its simply consolidate all test cases under 1 iteration to save on resources?

Twet1: @AMMayberry1 updated

AMMayberry1 (approved):

PR #311, opened 2024-12-08
add savage opress monster

ssoulier:
Adding Savage Opress Monster


VJubert (approved):

server/game/cards/03_TWI/units/SavageOpressMonster.ts
```diff
@@ -0,0 +1,32 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '8540765053',
+            internalName: 'savage-opress#monster',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'If you control fewer units than an opponent, ready this unit.',
```
AMMayberry1, line 16: nit: since the title will only appear if the condition is already true, we omit the "if" part and just put the effect. So in this case the title would be `Ready this unit`

AMMayberry1 (approved):

PR #322, opened 2024-12-11
add vanguard-droid-bomber

ssoulier:
[Vanguard droid bomber](https://swudb.com/card/twi/160/vanguard-droid-bomber)

Screenshot 2024-12-11 at 18 17 31

### DIFF
--- a/server/game/cards/03_TWI/units/VanguardDroidBomber.ts
+++ b/server/game/cards/03_TWI/units/VanguardDroidBomber.ts
@@ -1,0 +1,25 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { Trait } from '../../../core/Constants';
+
+export default class VanguardDroidBomber extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '3410014206',
+            internalName: 'vanguard-droid-bomber'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'If you control another Separatist unit, deal 2 damage to an enemy base.',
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/units/VanguardDroidBomber.ts
```diff
@@ -0,0 +1,25 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '3410014206',
+            internalName: 'vanguard-droid-bomber'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'If you control another Separatist unit, deal 2 damage to an enemy base.',
```
AMMayberry1, line 15: For our ability titles we don't include "if" clauses, since the ability will only appear at all if the condition is already met. So in this case it can just be `Deal 2 damage to an enemy base`

test/server/cards/03_TWI/units/VanguardDroidBomber.spec.ts
```diff
@@ -0,0 +1,46 @@
+describe('Vanguard Droid Bomber', function () {
+    integration(function (contextRef) {
+        describe('Vanguard Droid Bomber\'s ability', function () {
+            it('should deal 2 domage to the opponent base', function () {
```
AMMayberry1, line 4: nit: just to help other people understand what this test is doing (in case they break it on accident), we should include the reason it's happening to differentiate it from the other test and don't know the card text. So like 'should deal 2 damage to the opponent base because there is a friendly separatist unit'. Same for the other test title

server/game/cards/03_TWI/units/VanguardDroidBomber.ts
ssoulier: I am yet very fluent on what I should put in the title. I learnt from last time that "when played" was not required in the title.
Now I understand it's also the case for the condition.
Will do my best for the next one :)

PR #323, opened 2024-12-11
[+] RoyalGuardAttache: Card Integration

HugoHMZ:
image

### DIFF
--- a/server/game/cards/03_TWI/units/RoyalGuardAttache.ts
+++ b/server/game/cards/03_TWI/units/RoyalGuardAttache.ts
@@ -1,0 +1,24 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+
+export default class RoyalGuardAttache extends NonLeaderUnitCard {
+    protected override getImplementationId () {
+        return {
+            id: '0598115741',
+            internalName: 'royal-guard-attache'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'When Played: Deal 2 damage to this unit.',
+            optional: false,
+            when: {
+                onCardPlayed: (event, context) => event.card.controller === context.source.controller
+            },
+            immediateEffect: AbilityHelper.immediateEffects.damage({ amount: 2 }),
+        });
+    }
+}
+
+RoyalGuardAttache.implemented = true;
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/units/RoyalGuardAttache.ts
```diff
@@ -0,0 +1,24 @@
...
+        return {
+            id: '0598115741',
+            internalName: 'royal-guard-attache'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'When Played: Deal 2 damage to this unit.',
+            optional: false,
```
AMMayberry1, line 15: nit: false is the default so we can omit this

server/game/cards/03_TWI/units/RoyalGuardAttache.ts
```diff
@@ -0,0 +1,24 @@
...
+    protected override getImplementationId () {
+        return {
+            id: '0598115741',
+            internalName: 'royal-guard-attache'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'When Played: Deal 2 damage to this unit.',
```
AMMayberry1, line 14: For our ability titles we don't include the trigger, cost, or any "if" or "may" phrases since those will all be implied by the prompt itself (e.g., this ability title won't even appear if the right conditions aren't met). So we generally just describe the effect - in this case it would be `title: 'Deal 2 damage to this unit'`

test/server/cards/03_TWI/units/RoyalGuardAttache.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+                    player1: {
+                        hand: ['royal-guard-attache'],
+                    },
+                    player2: {
+                        groundArena: ['death-trooper'],
+                    },
+                });
+            });
+
+            it('should take 2 damages when played', function () {
```
AMMayberry1, line 16: nit: if there is only one test case for a setupTest() we can remove the beforeEach and put the setup directly in the test case for readability, see Zillo Beast tests for an example

test/server/cards/03_TWI/units/RoyalGuardAttache.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    integration(function (contextRef) {
+        describe('Royal Guard Attaché\'s ability', function () {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['royal-guard-attache'],
+                    },
+                    player2: {
+                        groundArena: ['death-trooper'],
```
AMMayberry1, line 11: Our practice is to only include things in the test setup that are relevant to the test just for readability, in this case we could remove Death Trooper and the entire `player2` object since there's nothing else in there

AMMayberry1 (approved):

PR #325, opened 2024-12-12
Blood Sport card implementation and tests

sethhelman:


AMMayberry1:
In the future can you include an image of the card in the PR description? Makes it easier to double-check

test/server/cards/03_TWI/events/BloodSport.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+                            spaceArena: ['cartel-spacer']
+                        },
+                        player2: {
+                            groundArena: [],
+                            spaceArena: ['alliance-xwing']
+                        }
+                    });
+                });
+
+                it('does nothing', function () {
```
AMMayberry1, line 54: This is a good test case but since we're trying to keep the numbers down due to long run time we can omit this test, we have a lot of cases checking that an event still plays if it has no targets

AMMayberry1 (approved):

PR #328, opened 2024-12-12
[+] FavorableDelegate: Card Implementation Draft

HugoHMZ:
image


server/game/cards/03_TWI/units/FavorableDelegate.ts
```diff
@@ -0,0 +1,24 @@
...
+            internalName: 'favorable-delegate'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Draw a card.',
+            immediateEffect: AbilityHelper.immediateEffects.draw()
+        });
+        this.addWhenDefeatedAbility({
```
AMMayberry1, line 17: nit: can we have a newline in between the two abilities just for readability

test/server/cards/03_TWI/units/FavorableDelegate.spec.ts
```diff
@@ -0,0 +1,64 @@
+describe('Favorable Delegate', function () {
+    integration(function (contextRef) {
+        describe('Favorable Delegate\'s ability', function () {
+            describe('When played ability', function () {
```
AMMayberry1, line 4: Jasmine will concatenate the successive test case names together, so we try to set it up so it forms a complete sentence as the convention. So for the first test case it would say `Favorable Delegate's ability When played ability should draw a card when played`, whereas we would probably want something like `Favorable Delegate's when played ability should draw a card`

test/server/cards/03_TWI/units/FavorableDelegate.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+                    },
+                    player2: {
+                        leader: { card: 'mace-windu#vaapad-form-master', deployed: true },
+                        hasInitiative: true
+                    },
+                    autoSingleTarget: true
+                });
+            });
+
+            it('should discard a card', function () {
```
AMMayberry1, line 38: This is another instance where we could consolidate tests a bit. We could join this into the previous test case by playing it from hand and then doing an attack with it. The third case could then just have its own separate setup

HugoHMZ: I merged the draw and discard basic cases together, and kept the discard on empty deck. I could remove the beforeEach because of that

AMMayberry1 (approved):

PR #330, opened 2024-12-12
add Independent Senator

ssoulier:
[Independent Senator](https://www.swudb.com/card/TWI/206/independent-senator)

Screenshot 2024-12-12 at 17 11 06


test/server/cards/03_TWI/units/IndependentSenator.spec.ts
```diff
@@ -0,0 +1,39 @@
+describe('Independent Senator', function () {
+    integration(function (contextRef) {
+        describe('Independent Senator\'s ability', function () {
+            it('should heal 2 damage from a unit or a base', function () {
```
AMMayberry1, line 4: Wrong test name. Also, if there's just one case you can ditch the `describe` and just have the entire name in the `it()`

test/server/cards/03_TWI/units/IndependentSenator.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+                    },
+                    player2: {
+                        groundArena: ['wampa', 'atst'],
+                        spaceArena: ['gladiator-star-destroyer', 'green-squadron-awing']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                function reset() {
```
AMMayberry1, line 18: Not needed

test/server/cards/03_TWI/units/IndependentSenator.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+                    context.player1.moveCard(context.echo, 'hand');
+                    context.player2.passAction();
+                }
+
+                // Case 1: A groud areana with less than 4 power can be exhausted
```
AMMayberry1, line 23: nit: we generally only use comments like this if we're doing multiple different scenarios in one case, in this one it's fine since the test case name already describes the one thing we're doing

test/server/cards/03_TWI/units/IndependentSenator.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+                context.player1.clickCard(context.independentSenator);
+                context.player1.clickPrompt('Exhaust a unit with 4 or less power.');
+
+                // can choose any units or base
```
AMMayberry1, line 27: nit: comment is wrong, can't choose base. would be confusing to someone reading the test for the first time

test/server/cards/03_TWI/units/IndependentSenator.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.greenSquadronAwing, context.wampa, context.independentSenator]);
+
+                // Exhaust green squadron A wing
+                expect(context.greenSquadronAwing.exhausted).toBeFalse();
```
AMMayberry1, line 31: Don't need this expect, we know from test setup that the unit is not exhausted

ssoulier:
Sorry for the noise, forgot to push a commit cleaning the test file.

PR #331, opened 2024-12-12
Multiple cards implemented x 4

Twet1:
I am not sure if I have done this correctly, Bokatan was committed and merged prior to this PR - I branched this before the merge was completed. 

This should be 4 cards 
image
image
image
image


server/game/cards/02_SHD/upgrades/ArmedToTheTeeth.ts
```diff
@@ -0,0 +1,27 @@
...
+            internalName: 'armed-to-the-teeth',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addGainOnAttackAbilityTargetingAttached({
+            title: 'Give another friendly unit +2/+0 for this phase',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                controller: RelativePlayer.Self,
```
AMMayberry1, line 18: Need to add the following property so that the ability can't target the attached unit (it says "another unit"):

```typescript
cardCondition: (card, context) => card !== context.source,
```

test/server/cards/02_SHD/upgrades/ArmedToTheTeeth.spec.ts
```diff
@@ -0,0 +1,35 @@
+describe('Armed to the teeth upgrade', function () {
```
AMMayberry1, line 1: nit: our convention would be to have it as `'Armed to the Teeth'`

test/server/cards/02_SHD/upgrades/ArmedToTheTeeth.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    integration(function (contextRef) {
+        describe('Armed to the teeth\'s ability', function () {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: [{ card: 'bokatan-kryze#fighting-for-mandalore', upgrades: ['armed-to-the-teeth'] }],
+                        spaceArena: ['collections-starhopper'],
+                        base: { card: 'echo-base', damage: 10 }
```
AMMayberry1, line 10: Can we remove any test settings that aren't relevant to the tests? Seems like setting the bases isn't really needed

test/server/cards/02_SHD/upgrades/ArmedToTheTeeth.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+            });
+
+            it('ground unit has 5 power and on attack gives other unit +2, which removes at end of phase', function () {
+                const { context } = contextRef;
+                expect(context.bokatanKryze.getPower()).toBe(5);
+                expect(context.collectionsStarhopper.getPower()).toBe(2);
+
+                context.player1.clickCard(context.bokatanKryze);
+                context.player1.clickCard(context.p2Base);
+                context.player1.clickCard(context.collectionsStarhopper);
```
AMMayberry1, line 26: Any time we write a card with a `targetResolver` we always need to have at least one instance of `toBeAbleToSelectExactly` to make sure the filters are set up right

server/game/cards/03_TWI/events/PetitionTheSenate.ts
```diff
@@ -0,0 +1,25 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '3500129784',
+            internalName: 'petition-the-senate',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'If you control 3 or more Official units, draw 3 cards.',
```
AMMayberry1, line 15: We don't include any "if" conditions in the title because if the player is seeing this string, the condition is already met. So we can just say `Draw 3 cards`. Same comment for other titles with "if" in the name in this PR

server/game/cards/03_TWI/events/PetitionTheSenate.ts
```diff
@@ -0,0 +1,25 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => context.source.controller.getOtherUnitsInPlayWithTrait(context.source, Trait.Official).length >= 3,
```
AMMayberry1, line 17: Should use `getUnitsInPlayWithTrait`, the "other" in the name means "other units in play" which isn't relevant since this card is not a unit

test/server/cards/03_TWI/events/PetitionTheSenate.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+                        spaceArena: ['green-squadron-awing']
+                    }
+                });
+            });
+
+            it('If control 3 Official units, draw 3 cards', function () {
+                const { context } = contextRef;
+                context.player1.clickCard(context.petitionTheSenate);
+                expect(context.player1.hand.length).toBe(3);
+            });
```
AMMayberry1, line 24: Need a second test with only two Official units in play to confirm that nothing happens. Ideally there would also be an opponent Official unit in play to confirm that isn't being counted as part of the 3

server/game/cards/03_TWI/units/501stLiberator.ts
```diff
@@ -0,0 +1,25 @@
...
+    protected override getImplementationId () {
+        return {
+            id: '7252148824',
+            internalName: '501st-liberator',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addWhenPlayedAbility({
+            title: 'If you control another Republic unit, you may heal 3 damage from a base.',
```
AMMayberry1, line 15: Since this ability says "may" we need optional: true and need to have a check for a Pass button in the test

server/game/cards/03_TWI/units/501stLiberator.ts
```diff
@@ -0,0 +1,25 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
```
AMMayberry1, line 16: The text says "heal 3 damage from a base" which means the player should be allowed to choose either base. So this should be a target resolver to allow the player to choose. The `immediateEffect` inside the `targetResolver` would be what you currently have here. Tests would need to be updated obviously

server/game/cards/03_TWI/units/_501StLiberator.ts
```diff
@@ -0,0 +1,25 @@
...
+export default class _501stLiberator extends NonLeaderUnitCard {
+    protected override getImplementationId () {
+        return {
+            id: '7252148824',
+            internalName: '501st-liberator',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addWhenPlayedAbility({
```
AMMayberry1, line 14: Can't remember which of these files is correct, can you please remove whichever one is wrong

test/server/cards/03_TWI/units/EchoValiantArcTrooper.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+            const { context } = contextRef;
+            expect(context.echo.getPower()).toBe(2);
+            expect(context.echo.getHp()).toBe(2);
+
+            context.player1.clickCard(context.battlefieldMarine);
+            expect(context.echo.getPower()).toBe(4);
+            expect(context.echo.getHp()).toBe(4);
+
+            context.player2.clickCard(context.cartelSpacer);
+            context.player2.clickCard(context.wingLeader);
```
AMMayberry1, line 26: Great tests!

test/server/cards/03_TWI/events/PetitionTheSenate.spec.ts
Twet1: Added a new test and when an official is defeated and click is done on card I get - Expected player prompt state to change after clicking petition-the-senate but it did not.

Is this correct, or should card go to Discard with no outcome?

Twet1:
Added a new test for Petition the senate and when an official is defeated and click is done on card I get - Expected player prompt state to change after clicking petition-the-senate but it did not. I have added a comment to the test to explain 

Is this correct, or should card go to Discard with no outcome?

Twet1:
@AMMayberry1 Pr updated with items mentioned. Please see other comment for one I was slightly unsure of

test/server/cards/03_TWI/events/PetitionTheSenate.spec.ts
AMMayberry1: Pinged in discord as well, but having trouble understanding the situation you describe - can you switch this to draft and push the broken test so I can see.

Also, if you have open comments that I need to answer please leave them unresolved so I can find them easier. Thanks!

AMMayberry1:
Also quick note that there's still a merge conflict

Twet1:
@AMMayberry1 had a merge conflict between old card - Believe this is now fixed.

The issue I was having with senate card was user error. All resolved now

test/server/cards/03_TWI/units/501stLiberator.spec.ts
```diff
@@ -0,0 +1,44 @@
+describe('501st-liberator', function() {
```
AMMayberry1, line 1: nit: '501st Liberator'

Twet1:
@AMMayberry1 commented items corrected. I have also refactored the tests, based on other PR feedback rather than moving cards around defeating or waylaying!

AMMayberry1 (approved):

PR #336, opened 2024-12-13
[+] Encouraging Leadership: Card Implementation

HugoHMZ:
image

### DIFF
--- a/server/game/cards/03_TWI/units/EncouragingLeadership.ts
+++ b/server/game/cards/03_TWI/units/EncouragingLeadership.ts
@@ -1,0 +1,24 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { WildcardZoneName } from '../../../core/Constants';
+
+export default class EncouragingLeadership extends EventCard {
+    protected override getImplementationId () {
+        return {
+            id: '4916334670',
+            internalName: 'encouraging-leadership'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Give each friendly unit +1/+1 for this phase.',
+            immediateEffect: AbilityHelper.immediateEffects.forThisPhaseCardEffect((context) => ({
+                target: context.source.controller.getUnitsInPlay(WildcardZoneName.AnyArena, (card) => card.isUnit()),
+                effect: AbilityHelper.ongoingEffects.modifyStats({ power: 1, hp: 1 }),
+            })),
+        });
+    }
+}
+
+EncouragingLeadership.implemented = true;
--- a/test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
+++ b/test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
@@ -1,0 +1,40 @@
+describe('EncouragingLeadership', function () {
+    integration(function (contextRef) {
+        describe('EncouragingLeadership\'s ability', function () {
+            it('Gives +1/+1 for this phase to allies', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['encouraging-leadership'],
+                        spaceArena: ['green-squadron-awing'],
+                        groundArena: ['duchesss-champion', 'atst']
+                    },
+                    player2: {
+                        spaceArena: ['star-wing-scout'],
+                        groundArena: ['specforce-soldier'],
+                    },
+                });
+                const { context } = contextRef;
+
+                // Check if effect is applied to both units
+                context.player1.clickCard(context.encouragingLeadership);
+                expect(context.encouragingLeadership).toBeInZone('discard');
+                expect(context.atst.getPower()).toEqual(7);
+                expect(context.atst.getHp()).toEqual(8);
+                expect(context.greenSquadronAwing.getPower()).toEqual(2);
+
+                // Pass the phase
+                context.player2.passAction();
+                context.player1.passAction();
+                context.player1.clickPrompt('Done');
+                context.player2.clickPrompt('Done');
+
+                // Check if units power/hp is back to normal
+                expect(context.greenSquadronAwing.getPower()).toEqual(1);
+                expect(context.atst.getPower()).toEqual(6);
+                expect(context.atst.getHp()).toEqual(7);
+                expect(context.greenSquadronAwing.getPower()).toEqual(1);
+            });
+        });
+    });
+});


server/game/cards/03_TWI/units/EncouragingLeadership.ts
```diff
@@ -0,0 +1,24 @@
...
+            id: '4916334670',
+            internalName: 'encouraging-leadership'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Give each friendly unit +1/+1 for this phase.',
+            immediateEffect: AbilityHelper.immediateEffects.forThisPhaseCardEffect((context) => ({
+                target: context.source.controller.getUnitsInPlay(WildcardZoneName.AnyArena, (card) => card.isUnit()),
```
AMMayberry1, line 17: The card.isUnit() check is redundant, `getUnitsInPlay` already filters out everything that isn't a unit. So you can just remove that argument.

test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
```diff
@@ -0,0 +1,40 @@
+describe('EncouragingLeadership', function () {
+    integration(function (contextRef) {
+        describe('EncouragingLeadership\'s ability', function () {
+            it('Gives +1/+1 for this phase to allies', function () {
```
AMMayberry1, line 4: nit: if there's just one `it` in the `describe` you can just remove the `describe` part and put that text inside of the `it` name (e.g., `it('Encouraging Leadership\'s ability gives +1/+1 for this phase to allies')`)

test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
```diff
@@ -0,0 +1,40 @@
+describe('EncouragingLeadership', function () {
```
AMMayberry1, line 1: nit: our convention would be to write this as `'Encouraging Leadership'` with a space

test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+                const { context } = contextRef;
+
+                // Check if effect is applied to both units
+                context.player1.clickCard(context.encouragingLeadership);
+                expect(context.encouragingLeadership).toBeInZone('discard');
+                expect(context.atst.getPower()).toEqual(7);
+                expect(context.atst.getHp()).toEqual(8);
+                expect(context.greenSquadronAwing.getPower()).toEqual(2);
+
+                // Pass the phase
```
AMMayberry1, line 26: Can use the helper `context.moveToRegroupPhase()` for this

test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+                    },
+                });
+                const { context } = contextRef;
+
+                // Check if effect is applied to both units
+                context.player1.clickCard(context.encouragingLeadership);
+                expect(context.encouragingLeadership).toBeInZone('discard');
+                expect(context.atst.getPower()).toEqual(7);
+                expect(context.atst.getHp()).toEqual(8);
+                expect(context.greenSquadronAwing.getPower()).toEqual(2);
```
AMMayberry1, line 24: Need to check the opponent's units to make sure they aren't affected. Also, there isn't a test for a-wing's hp currently

test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
```diff
@@ -0,0 +1,40 @@
+describe('EncouragingLeadership', function () {
+    integration(function (contextRef) {
+        describe('EncouragingLeadership\'s ability', function () {
+            it('Gives +1/+1 for this phase to allies', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['encouraging-leadership'],
```
AMMayberry1, line 8: Can we add a unit in hand and play it after Encouraging Leadership to confirm that it isn't affected by the buff... if you find that it is affected, please let me know, there might be an engine bug here I need to fix

test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+                player2: {
+                    spaceArena: ['star-wing-scout'],
+                    groundArena: ['specforce-soldier'],
+                },
+            });
+            const { context } = contextRef;
+
+            // Check if effect is applied to both units
+            context.player1.clickCard(context.encouragingLeadership);
+            expect(context.encouragingLeadership).toBeInZone('discard');
```
AMMayberry1, line 20: Don't need this check, we aren't testing event card functions in this test

test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+            context.player2.passAction();
+            context.player1.clickCard(context.wampa);
+            expect(context.wampa.getPower()).toEqual(4);
+            expect(context.wampa.getHp()).toEqual(5);
+
+            // Pass the phase
+            context.moveToRegroupPhase();
+
+            // Check if units power/hp is back to normal
+            expect(context.greenSquadronAwing.getPower()).toEqual(1);
```
AMMayberry1, line 44: This line is repeated below

test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+            expect(context.atst.getPower()).toEqual(6);
+            expect(context.atst.getHp()).toEqual(7);
+            expect(context.greenSquadronAwing.getPower()).toEqual(1);
+            expect(context.greenSquadronAwing.getHp()).toEqual(3);
```
AMMayberry1, line 48: Apologies for the late request but I just thought of one more potential bug it would be great to have a check for - could you add a third unit that receives the buff and is then defeated while the buff is still active?

The targeting for the stat buff effects has been a little glitchy so we just need to make sure it doesn't cause an issue by trying to apply itself you the unit after defeat.

HugoHMZ: I'm not sure to follow. You want me to add another unit on my board that gets defeated ? which check would you expect to have for this unit in particular ?

AMMayberry1: Yep that's it, thanks!

test/server/cards/03_TWI/units/EncouragingLeadership.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            expect(context.imperialInterceptor.getHp()).toEqual(3);
+
+            // Check if effect is not applied to non-allies
+            expect(context.starWingScout.getPower()).toEqual(4);
+            expect(context.starWingScout.getHp()).toEqual(1);
+            expect(context.specforceSoldier.getPower()).toEqual(2);
+            expect(context.specforceSoldier.getHp()).toEqual(2);
+
+
+            context.player2.passAction();
```
AMMayberry1, line 35: Two small requests then we should be done, thanks for taking the care to sort all this out:
1. Can you remove one of the newlines above so there's not two in a row
2. For the benefit of other people who might use this test as a reference, can you add a comment here that says "test that a unit with the buff can be defeated without causing any errors"

HugoHMZ: All my pleasure :)

AMMayberry1 (approved):

PR #338, opened 2024-12-13
2 cards implemented

Twet1:
Clone Dive Trooper -
image

Republic Attack Pod - 
image


server/game/cards/03_TWI/units/CloneDiveTrooper.ts
```diff
@@ -0,0 +1,25 @@
...
+            internalName: 'clone-dive-trooper',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addCoordinateAbility({
+            type: AbilityType.Constant,
+            title: 'While this unit is attacking, the defender gets -2/-0.',
+            targetController: RelativePlayer.Opponent,
+            matchTarget: (card) => card.isUnit() && card.isInPlay() && card.isDefending(),
```
AMMayberry1, line 19: Currently this would active when _any_ friendly unit is attacking, not necessarily this one. Need to add an additional condition `card.activeAttack.attacker === context.source`. Also we should add a test confirming that other friendly units don't get the effect.

test/server/cards/03_TWI/units/CloneDiveTrooper.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+            // Coordinate online
+            context.player1.clickCard(context.cloneDiveTrooper);
+            context.player1.clickCard(context.pykeSentinel);
+            expect(context.cloneDiveTrooper).toBeInZone('groundArena');
+            expect(context.cloneDiveTrooper.damage).toBe(0);
+            expect(context.pykeSentinel.damage).toBe(2);
+            context.player2.passAction();
+
+            context.player1.moveCard('battlefield-marine', 'discard');
```
AMMayberry1, line 24: Try to use `moveCard` as little as possible, it is always better to use a game effect if possible. The reason is that `moveCard` doesn't necessarily trigger game events correctly, so it could cause unexpected game state that could mess with the test results. The most common use of it is for returning events from discard to hand so they could be used multiple times in one test.

In this case, better to defeat the marine with an attack or an event card like vanquish.

test/server/cards/03_TWI/units/RepublicAttackPod.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+                    player2: {
+                        groundArena: ['greedo#slow-on-the-draw', 'rey#keeping-the-past'],
+                    }
+                });
+
+                const { context } = contextRef;
+
+                const reset = () => {
+                    context.player2.passAction();
+                    context.player1.moveCard(context.republicAttackPod, 'hand');
```
AMMayberry1, line 22: Similar note here - better to use Waylay to return the card to hand in between tests. This is also a situation where I'm potentially fine having multiple cases, since it's harder to reset state in between than it is with attacks. It's up to personal preference though.

Twet1:
@AMMayberry1 thank you. Updated

AMMayberry1 (approved):

PR #340, opened 2024-12-14
Add support for token units

AMMayberry1:
Added support for creating Battle Droid and Clone Trooper units. Created a test suite for validating all of the interactions specific to those card types, particularly the rules for defeat / leaving the arena. Slightly cleaned up the token upgrade system as well.

### DIFF
--- a/server/game/cards/03_TWI/events/DroidDeployment.ts
+++ b/server/game/cards/03_TWI/events/DroidDeployment.ts
@@ -12,9 +12,7 @@ export default class DroidDeployment extends EventCard {
     public override setupCardAbilities() {
         this.setEventAbility({
             title: 'Create 2 Battle Droid tokens',
-            immediateEffect: AbilityHelper.immediateEffects.createBattleDroid(
-                (context) => ({ amount: 2, target: context.source.controller })
-            )
+            immediateEffect: AbilityHelper.immediateEffects.createBattleDroid({ amount: 2 })
         });
     }
 }
--- a/server/game/cards/03_TWI/events/DroidDeployment.ts
+++ b/server/game/cards/03_TWI/events/DroidDeployment.ts
@@ -1,0 +1,22 @@
+import { EventCard } from '../../../core/card/EventCard';
+import AbilityHelper from '../../../AbilityHelper';
+
+export default class DroidDeployment extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '6826668370',
+            internalName: 'droid-deployment',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Create 2 Battle Droid tokens',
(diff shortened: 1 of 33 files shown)


server/game/core/gameSystem/PlayerTargetSystem.ts
```diff
@@ -22,6 +25,18 @@ export abstract class PlayerTargetSystem<TContext extends AbilityContext = Abili
         return this.canAffect(event.player, event.context, additionalProperties);
     }
 
+    public override resolve(target: undefined | Player | Player[], context: TContext, triggerHandlingMode?: TriggerHandlingMode) {
+        super.resolve(target, context, triggerHandlingMode);
+    }
+
+    protected override updateEvent(event: GameEvent, player: Player, context: TContext, additionalProperties?: any): void {
```
VJubert, line 32: why override these methods but just call super ?

server/game/cards/03_TWI/events/DroidDeployment.ts
```diff
@@ -0,0 +1,22 @@
...
+            id: '6826668370',
+            internalName: 'droid-deployment',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Create 2 Battle Droid tokens',
+            immediateEffect: AbilityHelper.immediateEffects.createBattleDroid(
+                (context) => ({ amount: 2, target: context.source.controller })
```
VJubert, line 16: target : context.source.controller cannot be in default properties ?

test/server/cards/03_TWI/events/OnTheDoorstep.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+            });
+
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.onTheDoorstep);
+
+            const battleDroids = context.player1.findCardsByName('battle-droid');
+            expect(battleDroids.length).toBe(3);
+            expect(battleDroids).toAllBeInZone('groundArena');
+            expect(battleDroids.every((battleDroid) => !battleDroid.exhausted)).toBeTrue();
```
VJubert, line 18: why adding ! on exhausted and not check tobefalse ?

AMMayberry1: That would be the wrong logic. Checking "every unit is exhausted === false" is just checking if at least one unit is ready. But we need to check if all units are ready

server/game/core/gameSystem/PlayerTargetSystem.ts
AMMayberry1: Oh yeah good question - we're overriding the type of the second parameter to be `Player` instead of `any` so that the child classes are forced to use that same signature. It's definitely unclear though, I'll add a comment

PR #342, opened 2024-12-14
Balkhrog/implementation wolf pack escort

HugoHMZ:
image

### DIFF
--- a/server/game/cards/03_TWI/units/WolfPackEscort.ts
+++ b/server/game/cards/03_TWI/units/WolfPackEscort.ts
@@ -1,6 +1,6 @@
 import AbilityHelper from '../../../AbilityHelper';
 import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
-import { CardType, RelativePlayer, Trait } from '../../../core/Constants';
+import { RelativePlayer, Trait, WildcardCardType } from '../../../core/Constants';
 
 export default class WolfPackEscort extends NonLeaderUnitCard {
     protected override getImplementationId () {
@@ -12,10 +12,10 @@ export default class WolfPackEscort extends NonLeaderUnitCard {
 
     public override setupCardAbilities() {
         this.addWhenPlayedAbility({
-            title: 'You may return a friendly non-leader, non-Vehicle unit to its owner\'s hand.',
+            title: 'Return a friendly non-leader, non-Vehicle unit to its owner\'s hand.',
             optional: true,
             targetResolver: {
-                cardTypeFilter: CardType.BasicUnit,
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
                 controller: RelativePlayer.Self,
                 cardCondition: (card) => !card.hasSomeTrait(Trait.Vehicle),
                 immediateEffect: AbilityHelper.immediateEffects.returnToHand()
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/units/WolfPackEscort.ts
```diff
@@ -0,0 +1,27 @@
...
+    protected override getImplementationId () {
+        return {
+            id: '1882027961',
+            internalName: 'wolf-pack-escort'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'You may return a friendly non-leader, non-Vehicle unit to its owner\'s hand.',
```
AMMayberry1, line 15: We omit the "you may" part for our titles since there will be a "pass" button that appears, indicating that it is optional. So this can just be `Return a friendly non-leader, non-Vehicle unit to its owner\'s hand.`

server/game/cards/03_TWI/units/WolfPackEscort.ts
```diff
@@ -0,0 +1,27 @@
...
+            optional: true,
+            targetResolver: {
+                cardTypeFilter: CardType.BasicUnit,
```
AMMayberry1, line 18: Should be `WildcardCardType.NonLeaderUnit`. `BasicUnit` excludes tokens, it is just "units" and not "leader units" or "token units"

test/server/cards/03_TWI/units/WolfPackEscort.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+        describe('ability', function () {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['wolf-pack-escort'],
+                        groundArena: ['duchesss-champion', 'specforce-soldier'],
+                        spaceArena: ['star-wing-scout'],
+                    },
+                    autoSingleTarget: true
```
AMMayberry1, line 12: We only added this to older code for backwards compatibility, it shouldn't be used for new tests

test/server/cards/03_TWI/units/WolfPackEscort.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+                // Checking we cannot select a vehicle
+                expect(context.player1).toBeAbleToSelectExactly([context.duchesssChampion, context.specforceSoldier]);
+
+                // Returning card to hand and checking it was returned
+                context.player1.clickCard(context.specforceSoldier);
+                expect(context.specforceSoldier).toBeInZone('hand');
+            });
+
+            it('can pass ability', function () {
```
AMMayberry1, line 29: We don't do explicit tests that the pass ability button works, it's enough to just confirm that it's there. So let's not have this test and instead just have the pass button check as noted in the comment above

test/server/cards/03_TWI/units/WolfPackEscort.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+                });
+            });
+
+            it('should return card to hand', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.wolfPackEscort);
+
+                // Checking we cannot select a vehicle
+                expect(context.player1).toBeAbleToSelectExactly([context.duchesssChampion, context.specforceSoldier]);
```
AMMayberry1, line 22: We can also put `expect(context.player1).toHavePassAbilityButton()` to confirm that the optional part is set correctly

AMMayberry1 (approved):

PR #343, opened 2024-12-14
Add Creative Thinking and Manufactured Soldiers

Frugghi:
| Creative Thinking | Manufactured Soldiers |
| --- | ---- |
| image | image |


server/game/cards/03_TWI/events/CreativeThinking.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Exhaust a non-unique unit. Create a Clone Trooper token.',
+            immediateEffect: AbilityHelper.immediateEffects.simultaneous([
+                AbilityHelper.immediateEffects.selectCard({
+                    cardTypeFilter: WildcardCardType.Unit,
+                    zoneFilter: WildcardZoneName.AnyArena,
+                    cardCondition: (card) => !card.unique && card.canBeExhausted() && !card.exhausted,
```
AMMayberry1, line 20: We shouldn't filter out exhausted cards here, per the swu rules they can be legally chosen as the target even though there'd be no effect

AMMayberry1 (approved):

AMMayberry1 (approved):

PR #346, opened 2024-12-15
Che bato/game lobby design

CheBato:
- Added the ability to send deck information to the lobby.
- Added the ability set the deck from the lobby before starting the game (sideboarding) so it is seen in the game.


AMMayberry1 (dismissed):

server/gamenode/Lobby.ts
```diff
@@ -51,7 +52,55 @@ export class Lobby {
...
+
+        // Add this card to the target array
+        // Check if the card already exists in the target
+        const targetIndex = targetArray.findIndex((item) => item.card.id === cardID);
+
+        if (targetIndex === -1) {
+            // Card not in target array, add a new entry
+            targetArray.push({
+                count: 1,
+                card: sourceEntry.card || sourceEntry.card // If we removed from source, we need to ensure we still have the card data
```
AMMayberry1, line 95: I think there's a bug here - looks like both sides of the `||` are the same thing?

server/gamenode/Lobby.ts
```diff
@@ -51,7 +52,55 @@ export class Lobby {
...
+        const [cardID] = args; // ['cardID']
+
+        // Determine the arrays we are moving between
+        const sourceArray = command === 'Deck' ? userDeck.deckCards : userDeck.sideboard;
+        const targetArray = command === 'Deck' ? userDeck.sideboard : userDeck.deckCards;
+
+        // Find the card in the source array
+        const sourceIndex = sourceArray.findIndex((item) => item.card.id === cardID);
+        if (sourceIndex === -1) {
+            console.log(`Card with UUID ${cardID} not found in source ${command}`);
```
AMMayberry1, line 72: Feels like this should be an exception

server/gamenode/Lobby.ts
```diff
@@ -51,7 +52,55 @@ export class Lobby {
...
+        }
+    }
+
+    private updateDeck(socket, command, ...args) {
+        const activeUser = this.users.find((u) => u.id === socket.user.username);
+        const userDeck = activeUser.deck.data;
+        const [cardID] = args; // ['cardID']
+
+        // Determine the arrays we are moving between
+        const sourceArray = command === 'Deck' ? userDeck.deckCards : userDeck.sideboard;
```
AMMayberry1, line 66: Is there anywhere that declares the legal values for `command`? Ideally there'd be an enum or something. Also, feels like we should probably have a check here to validate that it's not something unexpected (or null).

server/gamenode/Lobby.ts
```diff
@@ -51,7 +52,55 @@ export class Lobby {
...
+        }
+
+        // Extract the card entry from the source
+        const sourceEntry = sourceArray[sourceIndex];
+
+        // Decrement the count in the source entry
+        sourceEntry.count -= 1;
+
+        // If count is now zero, remove it from the source array
+        if (sourceEntry.count <= 0) {
```
AMMayberry1, line 83: Is the count allowed to go below 0?

server/gamenode/Lobby.ts
```diff
@@ -51,7 +52,55 @@ export class Lobby {
...
             this.sendGameState(this.game);
+        } else {
+            this.sendDeckInfo();
+        }
+    }
+
+    private updateDeck(socket, command, ...args) {
+        const activeUser = this.users.find((u) => u.id === socket.user.username);
+        const userDeck = activeUser.deck.data;
+        const [cardID] = args; // ['cardID']
```
AMMayberry1, line 63: This might be easier to read as `const cardID = args[0]`;

server/gamenode/Lobby.ts
```diff
@@ -51,7 +53,55 @@ export class Lobby {
...
+        }
+    }
+
+    private updateDeck(socket: Socket, ...args) {
+        const activeUser = this.users.find((u) => u.id === socket.user.username);
+        const userDeck = activeUser.deck.data;
+        const source = args[0]; // [<'Deck'|'Sideboard>'<cardID>]
+        const cardID = args[1];
+
+        Contract.assertTrue(source === 'Deck' || source === 'Sideboard');
```
AMMayberry1, line 67: Since the default error message for this will just say "false condition", we should add an error description (second argument of assertTrue) that prints the actual value of `source` for debugging purposes

server/gamenode/Lobby.ts
```diff
@@ -51,7 +53,55 @@ export class Lobby {
...
+        Contract.assertTrue(sourceIndex !== -1);
+
+        // Extract the card entry from the source
+        const sourceEntry = sourceArray[sourceIndex];
+
+        // Decrement the count in the source entry
+        sourceEntry.count -= 1;
+
+        // If count is now zero, remove it from the source array
+        Contract.assertTrue(sourceEntry.count >= 0);
```
AMMayberry1, line 83: nit: can use `assertNonNegative`

AMMayberry1 (approved):

PR #348, opened 2024-12-15
Implementing In Pursuit

supercodepoet:
image


server/game/cards/03_TWI/events/InPursuit.ts
```diff
@@ -0,0 +1,44 @@
...
+export default class InPursuit extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '6401761275',
+            internalName: 'in-pursuit'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
```
adbronico, line 14: I have a feeling we need to rework this one a bit to use an ifYouDo

test/server/cards/03_TWI/events/InPursuit.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+                        spaceArena: ['green-squadron-awing']
+                    },
+                    player2: {
+                        groundArena: ['pyke-sentinel'],
+                        spaceArena: ['inferno-four#unforgetting']
+                    }
+                });
+            });
+
+            it('should exhaust an enemy unit if you exhaust a friendly unit', function () {
```
adbronico, line 19: Can you add tests for when there is no friendly unit to exhaust? I suspect this will still let you exhaust an enemy unit. I also believe this requires the friendly unit to actually go from not exhausted to to exhausted, so we should add a test with only an exhausted friendly unit

adbronico (dismissed):

server/game/cards/03_TWI/events/InPursuit.ts
supercodepoet: ahhh, yeah, that is a good call. Let me look into that method.

AMMayberry1: Yeah exactly correct. As Chaqa notes below, this version won't work correctly if the friendly unit is already exhausted.

test/server/cards/03_TWI/events/InPursuit.spec.ts
supercodepoet: ok, added a test that verifies if you do not have a unit to exhaust the card does nothing

server/game/cards/03_TWI/events/InPursuit.ts
supercodepoet: I fixed the card to have an `IfYouDo` and ensured you could only exhaust a readied-friendly unit. I have still allowed the exhausting of any enemy unit, regardless of state, as it seems to match the rule, even though if you pick an exhausted enemy unit, the card does nothing.

server/game/cards/03_TWI/events/InPursuit.ts
```diff
@@ -0,0 +1,34 @@
...
+            title: 'Exhaust a friendly unit. If you do, exhaust an enemy unit.',
```
AMMayberry1, line 15: nit: it doesn't yet but the plan is to make things so that the "if you do" part of the title is automatically generated from the title of the `ifYouDo` ability. So it doesn't need to be included here, this can just be `Exhaust a friendly unit`

server/game/cards/03_TWI/events/InPursuit.ts
```diff
@@ -0,0 +1,34 @@
...
+            optional: true,
```
AMMayberry1, line 16: The ability doesn't say "you may" so it is not optional (if there are no targets it will just do nothing and move the event to discard)

server/game/cards/03_TWI/events/InPursuit.ts
```diff
@@ -0,0 +1,34 @@
...
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardCondition: (card) => card.isUnit() && !card.exhausted,
```
AMMayberry1, line 19: We shouldn't filter out exhausted cards since the player can legally choose to exhaust an already-exhausted friendly card, though it means the `ifYouDo` part will not resolve.

server/game/cards/03_TWI/events/InPursuit.ts
```diff
@@ -0,0 +1,34 @@
...
+                immediateEffect: AbilityHelper.immediateEffects.exhaust()
+            },
+            ifYouDo: {
+                title: 'Choose an enemy unit',
```
AMMayberry1, line 23: nit: should say `Exhaust an enemy unit`

test/server/cards/03_TWI/events/InPursuit.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                    player2: {
+                        groundArena: [{ card: 'pyke-sentinel', exhausted: true }],
+                        spaceArena: ['inferno-four#unforgetting']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.inPursuit);
+                expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine]);
```
AMMayberry1, line 21: As noted above, A-Wing is a legal target for this effect

supercodepoet: fixed up

server/game/cards/03_TWI/events/InPursuit.ts
supercodepoet: changed

server/game/cards/03_TWI/events/InPursuit.ts
supercodepoet: removed the filter

server/game/cards/03_TWI/events/InPursuit.ts
supercodepoet: removed

server/game/cards/03_TWI/events/InPursuit.ts
supercodepoet: changed

AMMayberry1 (approved):

PR #350, opened 2024-12-15
Discard from opponents hand logic and SOR # 201 Bodhi Rook

brianries:
Added logic to allow players to discard from an opponents hand and not just their own hand.  Bodhi Rook SOR # 201 uses this

image


server/game/cards/01_SOR/units/BodhiRook.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Look at an opponent\'s hand and discard a non-unit card from it.',
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.lookAt((context) => ({
+                    target: context.player.opponent.hand.sort((a, b) => a.title.localeCompare(b.title)),
+                    sendChatMessage: true
+                })),
+                AbilityHelper.immediateEffects.discardCardsFromHand((context) => ({
```
AMMayberry1, line 21: nit: can we get a space in between these abilities for readability

server/game/cards/01_SOR/units/BodhiRook.ts
```diff
@@ -0,0 +1,33 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Look at an opponent\'s hand and discard a non-unit card from it.',
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.lookAt((context) => ({
+                    target: context.player.opponent.hand.sort((a, b) => a.title.localeCompare(b.title)),
+                    sendChatMessage: true
```
AMMayberry1, line 19: Looking back, I don't know why Viper Probe Droid was formatted with the ability like this. You can drop `sendChatMessage` since the default value is true, and the targets don't need to be sorted.

If you wouldn't mind also updated Probe Droid to match since people will probably refer to it as an example a lot, that'd be really helpful

server/game/core/cardSelector/BaseCardSelector.js
```diff
@@ -135,7 +136,10 @@ class BaseCardSelector {
         if (!EnumHelpers.cardZoneMatches(card.zoneName, this.zoneFilter) && card.zoneName !== ZoneName.Capture) {
             return false;
         }
-        if (card.zoneName === ZoneName.Hand && card.controller !== choosingPlayer) {
+        if ((this.discardingPlayerType == null || this.discardingPlayerType === RelativePlayer.Self) && card.zoneName === ZoneName.Hand && card.controller !== choosingPlayer) {
```
AMMayberry1, line 139: I'm thinking on this some more again - do we actually even need this checking logic at all? I'm not sure why hand would need a specific guard that other zones like resources wouldn't. It seems like this was originally set up to block cards from the opponent's hand even if the card filters were set up to allow it. Would things work if we just remove this altogether and rely on the zone filters to get set up right?

server/game/cards/01_SOR/units/BodhiRook.ts
```diff
@@ -0,0 +1,33 @@
...
+                })),
+                AbilityHelper.immediateEffects.discardCardsFromHand((context) => ({
+                    cardTypeFilter: WildcardCardType.Any,
```
AMMayberry1, line 22: This is the default so we can remove it

server/game/cards/01_SOR/units/BodhiRook.ts
```diff
@@ -0,0 +1,33 @@
...
+                    cardCondition: (card) => !card.isUnit(),
+                    discardingPlayerType: RelativePlayer.Opponent,
```
AMMayberry1, line 24: Can we create a `discardCardsFromOpponentHand()` helper which hides this parameter? I think it is needed but it's probably confusing to understand how to set it correctly

server/game/gameSystems/DiscardCardsFromHandSystem.ts
```diff
@@ -13,10 +13,11 @@ export interface IDiscardCardsFromHandProperties extends IPlayerTargetSystemProp
     amount: Derivable<number, Player>;
     random?: boolean;
 
-    /* TODO: Add reveal system to when card type filter or card condition exists, as this is required to keep the
-    in order to keep the player honest in a in-person game */
+    /* TODO: Add a reveal system to flip over the cards if discarding from an opponent, also in the future
+    this may be necessary for a player discarding from their own ahnds if a card condition or filter exits to keep them honest */
     cardTypeFilter?: CardTypeFilter | CardTypeFilter[];
     cardCondition?: (card: Card, context: AbilityContext) => boolean;
+    discardingPlayerType?: RelativePlayer; // Self for own hand, or opponent chooses
```
AMMayberry1, line 20: This name is a little hard to understand, might be better as just `discardingPlayer`. Also, we should add a jsdoc on this parameter to explain what it means.

server/game/gameSystems/DiscardCardsFromHandSystem.ts
```diff
@@ -35,14 +36,25 @@ export class DiscardCardsFromHandSystem<TContext extends AbilityContext = Abilit
...
     public override canAffect(playerOrPlayers: Player | Player[], context: TContext, additionalProperties = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
         for (const player of Helpers.asArray(playerOrPlayers)) {
             const properties = this.generatePropertiesFromContext(context, additionalProperties);
             const availableHand = player.hand.filter((card) => properties.cardCondition(card, context) && EnumHelpers.cardTypeMatches(card.type, properties.cardTypeFilter));
 
+            // Select this player if its their own hand, or the active player from the context if its the 'opponent' choosing
+            const discardingPlayer = properties.discardingPlayerType === RelativePlayer.Self ? player : context.player;
+
+            // Since autoTarget is off any possible card in hand makes the hand 'canAffect' true
+            if (!discardingPlayer.autoSingleTarget && player.hand.length > 0) {
```
AMMayberry1, line 54: This is not quite right - since the swu rules make it legal to target something for an effect even if there will be no game state change, we need to check the value of `mustChangeGameState` to determine whether the hand length is relevant.

I.e., if `mustChangeGameState === GameStateChangeRequired.None`, we should always return true here because it doesn't matter if any cards are being discarded. If it's `MustFullyOrPartiallyResolve` or `MustFullyResolve`, then we return true if there are any cards to discard (even if there are too few).

That isn't technically correct based on what those values mean but the "partial resolution" concept was only partly implemented, so for now we should match the below code which only checks `ustChangeGameState !== GameStateChangeRequired.None`. We can fix it all in one pass later when we finish implementing support for `MustFullyOrPartiallyResolve`.

If that doesn't make sense feel free to ping me in chat, sorry it's a bit of a mess atm

server/game/gameSystems/DiscardCardsFromHandSystem.ts
```diff
@@ -63,16 +75,19 @@ export class DiscardCardsFromHandSystem<TContext extends AbilityContext = Abilit
...
 
+            // Select this player if its their own hand, or the active player from the context if its the 'opponent' choosing
+            const discardingPlayer = properties.discardingPlayerType === RelativePlayer.Self ? player : context.player;
+
             Contract.assertNonNegative(derive(properties.amount, player));
 
             const amount = Math.min(availableHand.length, derive(properties.amount, player));
 
-            if (amount === 0) {
+            if (amount === 0 && discardingPlayer.autoSingleTarget) {
```
AMMayberry1, line 85: Why does the value of autoSingleTarget matter in this case? Since there are no cards to discard, there's nothing to do in any case right?

server/game/gameSystems/DiscardCardsFromHandSystem.ts
```diff
@@ -63,16 +75,19 @@ export class DiscardCardsFromHandSystem<TContext extends AbilityContext = Abilit
...
                 events.push(this.generateEvent(context, additionalProperties));
                 continue;
             }
 
-            if (amount >= availableHand.length) {
+            if (amount >= availableHand.length && discardingPlayer.autoSingleTarget) {
```
AMMayberry1, line 90: We need to consider the random case here, if it's random then we still shouldn't let them choose

test/server/cards/02_SHD/events/Pillage.spec.ts
```diff
@@ -43,7 +44,30 @@ describe('Pillage', function() {
...
+
+                player1.clickCard(pillage);
+                player1.clickPrompt('Opponent');
+
+                expect(imperialInterceptor).toBeInZone('discard');
+
+                expect(player2).toBeActivePlayer();
+            });
+
+            it('should let the player target the opponent with autoSingleTarget false, so the opponent must select their last remaining card', function() {
```
AMMayberry1, line 62: Instead of adding a new test here could we just remove autoSingleTarget from the above test and update it? Honestly just for testing the feature more broadly it would be good to remove all `autoSingleTarget` from this file and maybe one other, like Force Throw.

server/game/cards/01_SOR/units/BodhiRook.ts
brianries: removed

server/game/core/cardSelector/BaseCardSelector.js
brianries: Agreed. I was able to get rid of this by adding back missing controller: RelativePlayer.Self to various cards to not require this check

server/game/gameSystems/DiscardCardsFromHandSystem.ts
brianries: This has been changed and handled with discardSubEvents that make it very clear who is the discarder and the target

server/game/cards/01_SOR/units/BodhiRook.ts
brianries: I'm not sure where you want the space to go exactly?

server/game/cards/01_SOR/units/BodhiRook.ts
brianries: Updated the setHand function to not put the cards back in reverse.  Removed the sort from Viper and Bodhi.  Updated their unit tests.  Also updated the Chimera's unit test for the now flipped order.

server/game/cards/01_SOR/units/BodhiRook.ts
brianries: Yes created this helper function -- much easier to use

server/game/gameSystems/DiscardCardsFromHandSystem.ts
brianries: Removed my change to this section, it was no longer needed (I believe?)

server/game/gameSystems/DiscardCardsFromHandSystem.ts
brianries: Yeah I guess the question is -- how will the reveal process work here?  Do we want them prompted and blocked while cards are flipped or just depend on the lookAt system printing?

brianries: Reverted this back to the original code, nothing to do here for *either* type of discard if there's nothing to viable to discard

server/game/gameSystems/DiscardCardsFromHandSystem.ts
brianries: I believe this is handled.  If auto single target is disabled, it doesn't generate the events here,  The immediate next section of the code checks for *random* only preventing it from getting to the promptForSelect regardless of the auto setting

test/server/cards/02_SHD/events/Pillage.spec.ts
brianries: Updated this unit test and force throw

server/game/cards/01_SOR/units/BodhiRook.ts
AMMayberry1: Oh sorry I just meant a newline

brianries: added

server/game/cards/01_SOR/units/ViperProbeDroid.ts
```diff
@@ -13,7 +13,7 @@ export default class ViperProbeDroid extends NonLeaderUnitCard {
         this.addWhenPlayedAbility({
             title: 'Look at an opponent\'s hand.',
             immediateEffect: AbilityHelper.immediateEffects.lookAt((context) => ({
-                target: context.player.opponent.hand.sort((a, b) => a.name.localeCompare(b.name)),
+                target: context.player.opponent.hand,
                 sendChatMessage: true
```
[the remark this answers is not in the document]
brianries: removed

server/game/gameSystems/DiscardCardsFromHandSystem.ts
```diff
@@ -9,22 +9,30 @@ import * as Helpers from '../core/utils/Helpers';
...
 }
 
 export class DiscardCardsFromHandSystem<TContext extends AbilityContext = AbilityContext> extends PlayerTargetSystem<TContext, IDiscardCardsFromHandProperties> {
     protected override defaultProperties: IDiscardCardsFromHandProperties = {
         amount: 1,
         random: false,
         cardTypeFilter: WildcardCardType.Any,
-        cardCondition: () => true
+        cardCondition: () => true,
+        discardSubEvents: []
```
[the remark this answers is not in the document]
brianries: Removed the discardSubEvents system and went back to the discardingPlayer (calling it discardingRelativePlayer for now)

server/game/gameSystems/DiscardCardsFromHandSystem.ts
```diff
@@ -60,19 +72,19 @@ export class DiscardCardsFromHandSystem<TContext extends AbilityContext = Abilit
 
     public override queueGenerateEventGameSteps(events: any[], context: TContext, additionalProperties: Record<string, any> = {}): void {
         const properties = this.generatePropertiesFromContext(context, additionalProperties);
-        for (const player of properties.target as Player[]) {
-            const availableHand = player.hand.filter((card) => properties.cardCondition(card, context));
+        for (const discardSubEvent of properties.discardSubEvents) {
```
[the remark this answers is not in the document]
brianries: This is removed

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -290,8 +292,43 @@ export function takeControlOfUnit<TContext extends AbilityContext = AbilityConte
...
+
+    return new DiscardCardsFromHandSystem<TContext>(wrappedPropertyFactory);
+}
+
+export function discardCardsFromOpponentsHand<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IDiscardCardsFromHandProperties, TContext>): DiscardCardsFromHandSystem<TContext> {
+    const wrappedPropertyFactory: PropsFactory<IDiscardCardsFromHandProperties, TContext> = (context: TContext) => {
+        const properties = typeof propertyFactory === 'function' ? propertyFactory(context) : propertyFactory;
+
+        // We might get this from an explicit property, or we might get it via the context with a target resolver etc
+        const targets = Helpers.asArray(properties.target ?? context.target ?? []);
```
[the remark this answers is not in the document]
brianries: This was removed by flipping back to leaning on target and relative player

server/game/gameSystems/DiscardCardsFromHandSystem.ts
```diff
@@ -32,7 +26,7 @@ export class DiscardCardsFromHandSystem<TContext extends AbilityContext = Abilit
         random: false,
         cardTypeFilter: WildcardCardType.Any,
         cardCondition: () => true,
-        discardSubEvents: []
+        discardingRelativePlayer: RelativePlayer.Self
```
AMMayberry1, line 29: Do you think "choosingRelativePlayer" (or alternately "choosingPlayer") is clearer here? There's a little bit of ambiguity around which is the "discarding" player if I'm choosing from my opponent's hand. If you think it's clearer as-is I'm fine with that, just trying to find the least awkward way to name this

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -314,17 +306,9 @@ export function discardCardsFromOwnHand<TContext extends AbilityContext = Abilit
 export function discardCardsFromOpponentsHand<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IDiscardCardsFromHandProperties, TContext>): DiscardCardsFromHandSystem<TContext> {
```
AMMayberry1, line 306: Just noticed this - since we're forcing the value of the `discardingRelativePlayer` property we should use an `Omit<>` type to remove that from the `propertyFactory` type completely so that the caller can't try to specify it. Take a look at `playCardFromHand()` and `playCardFromOutOfPlay()` in this file for example of how to use the `Omit<>`

server/game/gameSystems/DiscardCardsFromHandSystem.ts
brianries: Yes, that would be clearer, I can see the confusion.  choosingRelativePlayer is more obvious

server/game/gameSystems/GameSystemLibrary.ts
brianries: Thanks for this, I was wondering how I could make it non-optional and not require them to pass it on at the top level.   Added this.

server/game/gameSystems/DiscardCardsFromHandSystem.ts
```diff
@@ -17,7 +17,7 @@ export interface IDiscardCardsFromHandProperties extends IPlayerTargetSystemProp
     this may be necessary for a player discarding from their own ahnds if a card condition or filter exits to keep them honest */
     cardTypeFilter?: CardTypeFilter | CardTypeFilter[];
     cardCondition?: (card: Card, context: AbilityContext) => boolean;
-    discardingRelativePlayer?: RelativePlayer;
+    choosingRelativePlayer: RelativePlayer;
```
AMMayberry1, line 20: Message in discord as well but leaving it here. Small favor to ask - seeing this here jogged my memory that we have a similar parameter for target resolvers but it's named choosingPlayer: https://github.com/SWU-Karabast/forceteki/blob/8e330598f15e55c64c1db6526a7f8b97ad9a8e09/server/game/TargetInterfaces.ts#L73

I think it's similar enough in usage to this one that we should make the names the same, I'm fine with either choosingRelativePlayer or choosingPlayer for everything

brianries: Decided to change the new one to choosingPlayer, as the other version is in many many spots.  Didn't seem worth changing that one

AMMayberry1 (approved):

PR #354, opened 2024-12-16
Add Finalizer

Frugghi:
image

### DIFF
--- a/server/game/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.ts
+++ b/server/game/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.ts
@@ -1,0 +1,62 @@
+import AbilityHelper from '../../../AbilityHelper';
+import * as Helpers from '../../../core/utils/Helpers.js';
+import { Card } from '../../../core/card/Card';
+import { TriggeredAbilityContext } from '../../../core/ability/TriggeredAbilityContext';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { RelativePlayer, TargetMode, WildcardCardType, WildcardZoneName, ZoneName } from '../../../core/Constants';
+
+export default class FinalizerMightOfTheFirstOrder extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '9752523457',
+            internalName: 'finalizer#might-of-the-first-order',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Choose any number of friendly units',
+            targetResolver: {
+                activePromptTitle: 'Choose friendly units that will capture enemy units in the same arena',
+                mode: TargetMode.UpToVariable,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: WildcardZoneName.AnyArena,
+                controller: RelativePlayer.Self,
+                canChooseNoCards: true,
+                numCardsFunc: (context) => Math.min(
+                    context.source.controller.getUnitsInPlay().length,
+                    this.countOpponentNonLeaderUnitsInPlay(context, WildcardZoneName.AnyArena)
+                ),
+                multiSelectCardCondition: (card, selectedCards, context) => this.countOpponentNonLeaderUnitsInPlay(context, card.zoneName) > this.countSelectedCardsInZone(selectedCards, card.zoneName),
+            },
+            then: (chosenUnitsContext) => ({
+                title: 'Each of those units captures an enemy non-leader unit in the same arena',
+                immediateEffect: AbilityHelper.immediateEffects.sequential(
+                    Helpers.asArray(chosenUnitsContext.target).map((target) =>
+                        AbilityHelper.immediateEffects.selectCard({
+                            activePromptTitle: `Choose a unit to capture with ${target.title}`,
+                            player: RelativePlayer.Self,
+                            cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                            zoneFilter: target.zoneName,
+                            controller: RelativePlayer.Opponent,
+                            innerSystem: AbilityHelper.immediateEffects.capture({ captor: target })
+                        })
+                    )
+                )
+            })
+        });
+    }
+
+    private countOpponentNonLeaderUnitsInPlay(context: TriggeredAbilityContext, zoneName: ZoneName | WildcardZoneName.AnyArena): number {
+        return context.source.controller.opponent.getUnitsInPlay(
+            zoneName as ZoneName.GroundArena | ZoneName.SpaceArena | WildcardZoneName.AnyArena,
+            (card) => card.isNonLeaderUnit()
+        ).length;
+    }
+
+    private countSelectedCardsInZone(selectedCards: Card[], zoneName: ZoneName): number {
+        return selectedCards.filter((selectedCard) => selectedCard.zoneName === zoneName).length;
+    }
+}
+
+FinalizerMightOfTheFirstOrder.implemented = true;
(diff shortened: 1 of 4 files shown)


server/game/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.ts
```diff
@@ -0,0 +1,62 @@
...
+                mode: TargetMode.UpToVariable,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: WildcardZoneName.AnyArena,
+                controller: RelativePlayer.Self,
+                canChooseNoCards: true,
+                numCardsFunc: (context) => Math.min(
+                    context.source.controller.getUnitsInPlay().length,
+                    this.countOpponentNonLeaderUnitsInPlay(context, WildcardZoneName.AnyArena)
+                ),
+                multiSelectCardCondition: (card, selectedCards, context) => this.countOpponentNonLeaderUnitsInPlay(context, card.zoneName) > this.countSelectedCardsInZone(selectedCards, card.zoneName),
```
Frugghi, line 30: Is this too much complexity? I guess it would work fine using `TargetMode.Unlimited` too.

server/game/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.ts
```diff
@@ -0,0 +1,62 @@
...
+            },
+            then: (chosenUnitsContext) => ({
+                title: 'Each of those units captures an enemy non-leader unit in the same arena',
+                immediateEffect: AbilityHelper.immediateEffects.sequential(
```
Frugghi, line 34: I never understood if for something like this it makes a difference to use `sequential` vs `simultaneous`.

server/game/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.ts
```diff
@@ -0,0 +1,62 @@
...
+                        })
+                    )
+                )
+            })
+        });
+    }
+
+    private countOpponentNonLeaderUnitsInPlay(context: TriggeredAbilityContext, zoneName: ZoneName | WildcardZoneName.AnyArena): number {
+        return context.source.controller.opponent.getUnitsInPlay(
+            zoneName as ZoneName.GroundArena | ZoneName.SpaceArena | WildcardZoneName.AnyArena,
```
AMMayberry1, line 52: For safety and cleaner code we can add a Contract check that `zoneName` is one of these types, then we wouldn't need this cast here

server/game/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.ts
AMMayberry1: Nah I think this is fine, this is a weird card so it's okay if it ends being a bit complicated. I doubt there'll be another ability like this one anytime soon

server/game/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.ts
AMMayberry1: Yeah there's definitely some cases where it doesn't really matter. The difference is mostly about whether constant abilities are given a chance to resolve in between effect - i.e., whether other cards "see" these things happen as separate effects or all at once.

In this case I think it should be all at once from the wording on the card, if that doesn't create issues with it working correctly. Like if there was some relevant Coordinate ability on the opponent side that was keeping a unit alive, with `sequential` the unit could potentially die to seeing Coordinate get shut off but with `simultaneous` it won't since it gets captured in the same moment as all of the others (which I'm pretty sure is the correct behavior here).

server/game/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.ts
```diff
@@ -0,0 +1,62 @@
...
+                numCardsFunc: (context) => Math.min(
+                    context.source.controller.getUnitsInPlay().length,
+                    this.countOpponentNonLeaderUnitsInPlay(context, WildcardZoneName.AnyArena)
+                ),
+                multiSelectCardCondition: (card, selectedCards, context) => this.countOpponentNonLeaderUnitsInPlay(context, card.zoneName) > this.countSelectedCardsInZone(selectedCards, card.zoneName),
+            },
+            then: (chosenUnitsContext) => ({
+                title: 'Each of those units captures an enemy non-leader unit in the same arena',
+                immediateEffect: AbilityHelper.immediateEffects.sequential(
+                    Helpers.asArray(chosenUnitsContext.target).map((target) =>
```
AMMayberry1, line 35: Awesome job with this!

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -73,7 +73,7 @@ export function attachUpgrade<TContext extends AbilityContext = AbilityContext>(
 export function attack<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<IInitiateAttackProperties, TContext> = {}): CardTargetSystem<TContext> {
     return new InitiateAttackSystem<TContext>(propertyFactory);
 }
-export function capture<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<ICaptureProperties, TContext> = {}): GameSystem<TContext> {
+export function capture<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<ICaptureProperties, TContext> = {}): CardTargetSystem<TContext> {
```
AMMayberry1, line 76: Actually you can just remove the return type completely and let it infer, there's no benefit to having it like this and it causes issues like the one you're presumably running into. I have a task to come in and just remove all of them in here but haven't gotten around to it yet, might just knock it out after this PR is merged.

server/game/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.ts
Frugghi: Makes a lot of sense, thanks!`Huyang` could be another example of a card that could kill something if it's captured before the other.

I guess we would need `simultaneous` here but it's not as straight forward as updating this line because we would need to be able to prevent the user to capture the same card from multiple units (e.g. if you take the test as an example, Iden captures the Superlaser Tech, then for the Wampa you shouldn't be able to choose the Superlaser Tech again).

Is there anything similar already? I'll need to have a think for how to do this 🤔

Frugghi: Ok, I think I figured it out. I can use the context's events to see which cards have been captured already and filter them in the `selectCard` step. Let me know if you have a better suggestion!

server/game/core/utils/EnumHelpers.ts
```diff
@@ -20,7 +20,7 @@ export function isEnumValue<T>(value: string, enumObj: T): boolean {
     return Object.values(enumObj).indexOf(value) >= 0;
 }
 
-export const isArena = (zone: ZoneFilter) => {
+export const isArena = (zone: ZoneFilter): zone is ZoneName.GroundArena | ZoneName.SpaceArena | WildcardZoneName.AnyArena => {
```
AMMayberry1, line 23: Ah good idea! Do you mind adding a TODO comment here for adding similar annotations to the other methods?

test/server/cards/02_SHD/units/FinalizerMightOfTheFirstOrder.spec.ts
```diff
@@ -40,16 +40,16 @@ describe('Finalizer, Might of the First Order', function() {
                 context.player1.clickPrompt('Done');
 
                 expect(context.player1).toHavePrompt('Choose a unit to capture with Iden Versio');
-                expect(context.player1).toBeAbleToSelectExactly([context.steadfastBattalion, context.battlefieldMarine, context.isbAgent, context.superlaserTechnician]);
-                context.player1.clickCard(context.superlaserTechnician);
+                expect(context.player1).toBeAbleToSelectExactly([context.steadfastBattalion, context.battlefieldMarine, context.zuckuss, context._4lom]);
```
AMMayberry1, line 43: Just in case we accidentally break this behavior later due to engine changes and cause the Zuckuss / 4-LOM part to fail, do you mind adding a comment above this explaining why they're included and what the expected behavior is (i.e., captured not defeated)

AMMayberry1 (approved):

PR #358, opened 2024-12-17
Implementing Set 3 cards (batch #1)

supercodepoet:
[Knight of the Republic](https://swudb.com/card/TWI/049/knight-of-the-republic)
[General's Guardian](https://swudb.com/card/twi/083/generals-guardian)
[Jesse, Hard-Fighting Patriot](https://swudb.com/card/TWI/145/jesse)
[Tri-Droid Suppressor](https://swudb.com/card/twi/217/tri-droid-suppressor)
[Droid Cohort](https://swudb.com/card/twi/218/droid-cohort)


test/server/cards/03_TWI/units/GeneralsGuardian.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+            it('should create a Battle Droid token when attacked', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.coruscantGuard);
+                context.player1.clickCard(context.generalsGuardian);
+
+                const battleDroids = context.player2.findCardsByName('battle-droid');
+                expect(battleDroids.length).toBe(1);
+                expect(battleDroids).toAllBeInZone('groundArena');
+                expect(battleDroids.every((battleDroid) => battleDroid.exhausted)).toBeTrue();
```
AMMayberry1, line 26: Can we also test with another friendly unit being attacked to confirm it only works for this unit. Same note for Knight of the Republic

test/server/cards/03_TWI/units/JesseHardFightingPatriot.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+            it('should create two Battle Droid tokens for the opponent', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.jesse);
+                expect(context.player2).toBeActivePlayer();
+
+                const battleDroids = context.player2.findCardsByName('battle-droid');
+                expect(battleDroids.length).toBe(2);
+                expect(battleDroids).toAllBeInZone('groundArena');
```
AMMayberry1, line 21: Just to be extra explicit, let's specify the player parameter here: `expect(battleDroids).toAllBeInZone('groundArena', context.player2);`

By default the `toBeInZone` expectations will check that the unit is in the arena of its controller, which is a good default. Since there's some potential issues around controllership here it's good to be explicit though I think.

test/server/cards/03_TWI/units/TriDroidSuppressor.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+                context.player1.clickCard(context.tridroidSuppressor);
+                expect(context.player1).toBeAbleToSelectExactly([context.lomPyke, context.pykeSentinel]);
+
+                context.player1.clickCard(context.lomPyke);
+                expect(context.lomPyke.exhausted).toBe(true);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should not exhaust an enemy ground unit', function () {
```
AMMayberry1, line 27: I think we're okay without this test, since Inferno Four didn't appear in the target list in the test above we can be very confident it won't be selected for targeting even if there are no ground units

server/game/cards/03_TWI/upgrades/DroidCohort.ts
```diff
@@ -0,0 +1,23 @@
...
+            id: '7547538214',
+            internalName: 'droid-cohort',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addGainTriggeredAbilityTargetingAttached({
+            title: 'Create a Battle Droid token.',
+            when: {
+                onCardDefeated: (event, context) => event.card.parentCard === context.source
```
AMMayberry1, line 16: This is slightly wrong - it is checking for the defeat of an upgrade whose `parentCard` is the same as this upgrade's. We're giving a "when defeated" ability to the unit itself, so it needs to check for its own defeat.

Actually, there's a helper that should be in place for this that we missed. There is already `UpgradeCard.addGainOnAttackAbilityTargetingAttached()` which would grant an on-attack ability with the right settings, do you think you'd be up for copying that method and creating a new one `addGainWhenDefeatedAbilityTargetingAttached`? All that would need to be changed is the `when` condition.

test/server/cards/03_TWI/units/TriDroidSuppressor.spec.ts
supercodepoet: sounds good

test/server/cards/03_TWI/units/GeneralsGuardian.spec.ts
supercodepoet: this has been fixed up

test/server/cards/03_TWI/units/JesseHardFightingPatriot.spec.ts
supercodepoet: added

server/game/cards/03_TWI/upgrades/DroidCohort.ts
supercodepoet: added the function to `UpgradeCard`. I think I got it right. I didn't see any core card tests so I am just using the Droid Cohort for the test.

test/helpers/PlayerInteractionWrapper.js
```diff
@@ -407,7 +407,11 @@ class PlayerInteractionWrapper {
     }
 
     findCardByName(name, zones = 'any', side) {
-        return this.filterCardsByName(name, zones, side)[0];
+        var cards = this.filterCardsByName(name, zones, side);
+        if (cards.length === 0) {
```
AMMayberry1, line 411: Can we check that the count is exactly 1 and throw an error if it's either higher or lower (the error should clearly specify the number). If that ends up breaking a lot of tests you can revert and lmk and I'll do it in a separate PR.

test/server/cards/03_TWI/units/GeneralsGuardian.spec.ts
```diff
@@ -22,9 +22,19 @@ describe('General\'s Guardian', function() {
 
                 const battleDroids = context.player2.findCardsByName('battle-droid');
                 expect(battleDroids.length).toBe(1);
-                expect(battleDroids).toAllBeInZone('groundArena');
+                expect(battleDroids).toAllBeInZone('groundArena', context.player2);
                 expect(battleDroids.every((battleDroid) => battleDroid.exhausted)).toBeTrue();
             });
+
+            it('should not create a Battle Droid when a different unit is attacked', function () {
```
AMMayberry1, line 29: Can merge this into the above test pretty easily, same for Knight of the Republic

test/helpers/PlayerInteractionWrapper.js
supercodepoet: @AMMayberry1 I started down that path, checking that anything above 1 throws and error. It broke every test. I can dig in further if you would like.

test/server/cards/03_TWI/units/GeneralsGuardian.spec.ts
supercodepoet: fixed up

test/helpers/PlayerInteractionWrapper.js
AMMayberry1: Ah interesting, probably one of the setup methods is using this when it shouldn't be. That's probably out of scope for this PR, can you just add a TODO here marking it to look into later?

supercodepoet: The TODO is added and everything should be good to go.

AMMayberry1 (approved):

PR #361, opened 2024-12-18
Add "in play copy" concept for correct card history tracking

AMMayberry1:
This PR implements this SWU rule (8.6.4):

```
Whenever a card leaves and later re-enters play, it is considered a “new copy” of that card for the purposes of game rules and does
not regain any modifiers or reapply any effects from when it was previously in play. It continues to be considered a new copy even if it
changes zones.
```

This is most relevant for state watchers since they are most often used to ask the question "did X happen to this card this phase." However, in the past we could only check by matching the Card object itself, meaning that if it had entered, left, and re-entered play that phase, it should count as a different unit but it would not.

Added the concept of an "in-play id" which is basically a counter that gets incremented every time a card enters the arena. Watchers and components that track history in any way will stash this value at the relevant time point, so that later the card's current inPlayId can be compared against the historical one to see if it is still the same unit.

Did a pass across all usages of watchers and updated each one for which this concept was relevant, and added tests for all. Also did a few general card and watcher updates as some cards using watchers had other small issues such as not checking controller.


Frugghi (approved):
Great work 🚀

PR #362, opened 2024-12-18
Implementing Grievous leader and more cards

aruizc:
Implementing Grievous leader, Kit Fisto, Twice the Pride and Old Access Codes.

image
image
image
image
image


server/game/cards/03_TWI/leaders/GeneralGrievousGeneralOfTheDroidArmies.ts
```diff
@@ -0,0 +1,45 @@
...
+        return {
+            id: '2872203891',
+            internalName: 'general-grievous#general-of-the-droid-armies',
+        };
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addActionAbility({
+            title: 'Give a Droid unit Sentinel for this phase',
+            cost: [AbilityHelper.costs.exhaustSelf()],
```
AMMayberry1, line 16: nit: doesn't have to be an array

server/game/cards/03_TWI/leaders/GeneralGrievousGeneralOfTheDroidArmies.ts
```diff
@@ -0,0 +1,45 @@
...
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                cardCondition: (card) => card.isUnit() && card.hasSomeTrait(Trait.Droid),
```
AMMayberry1, line 19: We can remove the `isUnit()` here since we have the `cardTypeFilter` above. Same for the deployed ability

server/game/cards/03_TWI/leaders/GeneralGrievousGeneralOfTheDroidArmies.ts
```diff
@@ -0,0 +1,45 @@
...
+                })
+            }
+        });
+    }
+
+    protected override setupLeaderUnitSideAbilities() {
+        this.addOnAttackAbility({
+            title: 'Give a Droid unit +1/+0 and Sentinel for this phase',
+            optional: true,
+            immediateEffect: AbilityHelper.immediateEffects.selectCard({
```
AMMayberry1, line 31: We should use a `targetResolver` here instead of `selectCard()`. `selectCard()` really only exists for cases where using `targetResolver` is not possible, such inside of another effect like `sequential()`.

Note that this will break your tests currently because the autoSingleTarget feature is not working right at the moment for `selectCard`, there will be at least one place you will need to an add explicit click on the target. I am planning to fix that.

test/server/cards/03_TWI/leaders/GeneralGrievousGeneralOfTheDroidArmies.spec.ts
```diff
@@ -0,0 +1,83 @@
+describe('General Grievous, General Of The Droid Armies', function () {
+    integration(function (contextRef) {
+        describe('General Grievous\'s leader undeployed ability', function () {
+            it('should Sentinel to a Droid unit for this phase', function () {
```
AMMayberry1, line 4: nit: since there's only `it()` in this `describe()` we can just remove the `describe` completely and put the whole name inside of the `it`, same for the other test

test/server/cards/03_TWI/leaders/GeneralGrievousGeneralOfTheDroidArmies.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['battle-droid-legion'],
+                        spaceArena: ['devastating-gunship'],
+                        leader: 'general-grievous#general-of-the-droid-armies'
+                    },
+                    player2: {
+                        groundArena: ['battlefield-marine', 'atst'],
```
AMMayberry1, line 13: We should add an opponent Droid unit to make sure it's targetable, same for the other test case

test/server/cards/03_TWI/leaders/GeneralGrievousGeneralOfTheDroidArmies.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.generalGrievous);
+                context.player1.clickPrompt('Give a Droid unit Sentinel for this phase');
+                expect(context.player1).toBeAbleToSelectExactly([context.devastatingGunship, context.battleDroidLegion]);
+                context.player1.clickCard(context.battleDroidLegion);
+
```
AMMayberry1, line 23: Can you do me a favor - since Battle Droid tokens are kind of new it would be good to test that they work here. Would you mind adding the Droid Deployment card to this test and choose one of the generated droid tokens as the sentinel target? Makes this test a little more complicated but it would be helpful for confirming that the token units work right

AMMayberry1:
Since you have contributor access to the main repo now, do you mind cloning that and pushing branches there instead of using your fork? I needed to investigate something in this PR and it's a little harder to pull down your code if it's in a fork

test/server/cards/03_TWI/units/KitFistoTheSmilingJedi.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            });
+
+            const { context } = contextRef;
+
+            // Coordinate active
+            context.player1.clickCard(context.kitFisto);
+            context.player1.clickCard(context.p2Base);
+
+            expect(context.player1).toHaveEnabledPromptButton('Deal 3 damage to a ground unit');
+            expect(context.player1).toHaveEnabledPromptButton('Saboteur: defeat all shields');
```
AMMayberry1, line 24: Can use `toHaveExactPromptButtons` for these instead

test/server/cards/03_TWI/units/KitFistoTheSmilingJedi.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+            // Coordinate offline
+            context.player2.clickCard(context.vanquish);
+            context.player2.clickCard(context.battlefieldMarine);
+
+            context.kitFisto.exhausted = false;
+            context.player1.clickCard(context.kitFisto);
+            context.player1.clickCard(context.p2Base);
+
+            expect(context.player2).toBeActivePlayer();
```
AMMayberry1, line 40: Great tests!!

test/server/cards/03_TWI/upgrades/OldAccessCodes.spec.ts
```diff
@@ -0,0 +1,50 @@
+describe('Old Access Codes', function() {
+    integration(function(contextRef) {
+        describe('Old Access Codes ability', function() {
+            it('should draw a card', function () {
```
AMMayberry1, line 4: nit: we should include a little more detail in case someone needs to understand this test if they broke it or something. So in this case `should draw a card when the controller has fewer units than the opponent`. Same for the below test case

test/server/cards/03_TWI/upgrades/OldAccessCodes.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+            it('should not draw a card', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['old-access-codes'],
+                        groundArena: ['battlefield-marine']
+                    },
+                    player2: {
+                        groundArena: ['r2d2#ignoring-protocol']
```
AMMayberry1, line 35: Do you mind putting an upgrade on R2 just to make sure that isn't getting counted, our arena methods have had some issues in the past. So like `groundArena: [{ card: 'r2d2#ignoring-protocol', upgrades: ['academy-training'] }]` is good

AMMayberry1 (approved):

PR #363, opened 2024-12-19
Mando Armor SHD

Twet1:
image


server/game/cards/02_SHD/upgrades/MandalorianArmor.ts
```diff
@@ -0,0 +1,28 @@
...
+            internalName: 'mandalorian-armor'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
+            title: 'If attached unit is a Mandalorian, give a Shield token to it.',
+            immediateEffect: AbilityHelper.immediateEffects.conditional((context) => ({ target: context.source.parentCard,
```
AMMayberry1, line 19: I am actually not 100% sure that using `target` in `conditional()` works correctly in all cases, do you mind moving this inside of `giveShield()`?

test/server/cards/02_SHD/upgrades/MandalorianArmor.spec.ts
```diff
@@ -0,0 +1,37 @@
+describe('Mandalorian Armor', function() {
+    integration(function(contextRef) {
+        describe('Mandalorian Armor\'s attached ability', function() {
+            it('If attached unit is a Mandalorian, give a Shield token to it, or dont', function () {
```
AMMayberry1, line 4: These two sentences will be concatenated, so they should say something like `Mandalorian Armor\'s attached ability should give a shield to the attached card if it is a Mandalorian unit`. Also you can remove the `describe` since there's only one `it` case and just put the whole name inside of the `it`

test/server/cards/02_SHD/upgrades/MandalorianArmor.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['ketsu-onyo#old-friend', 'the-mandalorian#wherever-i-go-he-goes', 'battlefield-marine'],
```
AMMayberry1, line 8: Can we have a vehicle unit in play so we can confirm that it doesn't appear in the target list with `toBeAbleToSelectExactly` (which is already set up with the correct list of targets anyway since the vehicle shouldn't be included). That will confirm that the attachment filter is working.

Twet1:
@AMMayberry1 changes made

server/game/cards/02_SHD/upgrades/MandalorianArmor.ts
```diff
@@ -0,0 +1,27 @@
...
+            id: '3514010297',
+            internalName: 'mandalorian-armor'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
+            title: 'Give a Shield token.',
```
AMMayberry1, line 18: nit: this is ambiguous, it could mean "give a shield token to any unit". Needs to indicate that the shield token will go to the attached uit

AMMayberry1 (approved):

PR #367, opened 2024-12-20
Prompt typing

danbastin:
Small cleanup on prompt that distingueshes pass/claim prompts from other prompts and puts prompt information under its own property for gamestate updates.


server/game/core/gameSteps/prompts/ResourcePrompt.ts
```diff
@@ -60,7 +60,8 @@ export class ResourcePrompt extends AllPlayerPrompt {
             menuTitle: promptText,
             buttons: [{ text: 'Done', arg: 'done' }],
             promptTitle: 'Resource Step',
-            promptUuid: this.uuid
+            promptUuid: this.uuid,
+            promptType: 'resource'
```
AMMayberry1, line 64: I recommend we create an enum for tracking these, it could go in Constants.ts with the others

AMMayberry1 (approved):

PR #369, opened 2024-12-20
Add Traitorous and Survivors Gauntlet

Frugghi:
| Traitorous | Survivors Gauntlet |
| --- | --- |
| image | image |

### DIFF
--- a/server/game/cards/01_SOR/upgrades/Traitorous.ts
+++ b/server/game/cards/01_SOR/upgrades/Traitorous.ts
@@ -11,7 +11,7 @@ export default class Traitorous extends UpgradeCard {
 
     public override setupCardAbilities() {
         this.addTriggeredAbility({
-            title: 'Take control of a non-leader unit that costs 3 or less',
+            title: 'Take control of attached non-leader unit if it costs 3 or less',
             when: {
                 onUpgradeAttached: (event, context) => event.upgradeCard === context.source,
             },
@@ -24,6 +24,7 @@ export default class Traitorous extends UpgradeCard {
                 onFalse: AbilityHelper.immediateEffects.noAction({ hasLegalTarget: true }),
             })
         });
+
         this.addTriggeredAbility({
             title: 'That unit’s owner takes control of it',
             when: {
--- a/server/game/cards/01_SOR/upgrades/Traitorous.ts
+++ b/server/game/cards/01_SOR/upgrades/Traitorous.ts
@@ -1,0 +1,40 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { UpgradeCard } from '../../../core/card/UpgradeCard';
+
+export default class Traitorous extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '8055390529',
+            internalName: 'traitorous'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Take control of a non-leader unit that costs 3 or less',
+            when: {
+                onUpgradeAttached: (event, context) => event.upgradeCard === context.source,
+            },
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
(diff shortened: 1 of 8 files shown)


server/game/cards/02_SHD/units/SurvivorsGauntlet.ts
```diff
@@ -0,0 +1,45 @@
...
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Attach an upgrade on a unit to another eligible unit controlled by the same player',
+            when: {
+                onCardPlayed: (event, context) => event.card === context.source,
+                onAttackDeclared: (event, context) => event.attack.attacker === context.source,
+            },
+            targetResolvers: {
+                chooseUpgrade: {
+                    mode: TargetMode.Single,
```
AMMayberry1, line 22: This shouldn't be needed, I believe it's the default (same for the other resolver)

server/game/cards/02_SHD/units/SurvivorsGauntlet.ts
```diff
@@ -0,0 +1,45 @@
...
+                    controller: WildcardRelativePlayer.Any,
+                    optional: true,
+                },
+                chooseUnit: {
+                    dependsOn: 'chooseUpgrade',
+                    mode: TargetMode.Single,
+                    cardTypeFilter: WildcardCardType.Unit,
+                    zoneFilter: WildcardZoneName.AnyArena,
+                    controller: (context) => (context.source.controller === context.targets.chooseUpgrade.controller ? RelativePlayer.Self : RelativePlayer.Opponent),
+                    cardCondition: (card, context) => context.targets.chooseUpgrade.isUpgrade() && context.targets.chooseUpgrade.parentCard !== card,
```
AMMayberry1, line 34: Also needs to check upgrade eligibility, since some upgrades have restrictions like not being attachable to vehicles. I think it should be enough to just call `context.targets.chooseUpgrade.canAttach()`

test/server/cards/01_SOR/upgrades/Traitorous.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.traitorous);
+                expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine, context.superlaserTechnician, context.cartelSpacer, context.wampa, context.lukeSkywalker, context.survivorsGauntlet]);
+                context.player1.clickCard(context.wampa);
+
+                expect(context.wampa).toBeInZone('groundArena', context.player2);
+            });
+
+            it('gives back control of a unit when unattached', function () {
```
AMMayberry1, line 55: This is great! One small ask, could you add a few comments narrating the general flow of what's happening in the test?

server/game/cards/02_SHD/units/SurvivorsGauntlet.ts
Frugghi: That's not really necessary because `AttachUpgradeSystem` will already call it in its implementation. This is already covered in the test, for example when moving `Frozen in Carbonite` the deployed leader is not selectable.

server/game/cards/01_SOR/upgrades/Traitorous.ts
```diff
@@ -0,0 +1,40 @@
...
+            when: {
+                onUpgradeAttached: (event, context) => event.upgradeCard === context.source,
+            },
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => context.event.parentCard.isNonLeaderUnit() && context.event.parentCard.cost <= 3,
+                onTrue: AbilityHelper.immediateEffects.takeControlOfUnit((context) => ({
+                    target: context.event.parentCard,
+                    newController: context.source.owner
+                })),
+                onFalse: AbilityHelper.immediateEffects.noAction({ hasLegalTarget: true }),
```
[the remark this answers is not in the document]
Frugghi: I don't think it's needed 😅

AMMayberry1 (approved):

PR #371, opened 2024-12-21
Implementing Captain Rex, Fighting for his Brothers

supercodepoet:
[Captain Rex, Fighting for his Brothers](https://swudb.com/card/twi/007/captain-rex)

image
image


AMMayberry1:
In the future do you mind including images directly instead of a link, just makes things quicker

test/server/cards/03_TWI/leaders/CaptainRexFightingForHisBrothers.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+                    resources: 5
+                },
+                player2: {
+                    groundArena: ['droid-commando']
+                }
+            });
+        });
+
+        describe('Captain Rex\'s leader undeployed ability', function () {
+            it('should create a Clone Tropper when a friendly unit attacked this phase', function () {
```
AMMayberry1, line 18: nit: can just merge the `it()` and `describe()` together into a single `it()` with one name, same for the below test

test/server/cards/03_TWI/leaders/CaptainRexFightingForHisBrothers.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+                context.player1.clickCard(context.captainRex);
+                context.player1.clickPrompt('If a friendly unit attacked this phase, create a Clone Trooper token.');
+
+                expect(context.captainRex.exhausted).toBeTrue();
+                expect(context.player1.exhaustedResourceCount).toBe(2);
+                expect(context.player1.findCardsByName('clone-trooper').length).toBe(0);
+
+                context.moveToNextActionPhase();
+
+                // Leader ability should be available, but should not create a Clone Trooper since an enemy unit attacked this phase
```
AMMayberry1, line 31: You could remove the above test where nothing attacks at all and just do this one since it covers both cases

test/server/cards/03_TWI/leaders/CaptainRexFightingForHisBrothers.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+        });
+
+        describe('Captain Rex\'s leader deployed ability', function () {
+            it('should create a Clone Tropper on deploy and each friendly Trooper has 0/+1', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.captainRex);
+                context.player1.clickPrompt('Deploy Captain Rex');
+
+                expect(context.captainRex.exhausted).toBeFalse();
```
AMMayberry1, line 69: This expectation isn't needed, we can assume the deployment system works correctly

test/server/cards/03_TWI/leaders/CaptainRexFightingForHisBrothers.spec.ts
supercodepoet: sounds good, I am still feeling out my way on how much and what we want to test with limited GitHub Actions minutes.

test/server/cards/03_TWI/leaders/CaptainRexFightingForHisBrothers.spec.ts
supercodepoet: 👍

test/server/cards/03_TWI/leaders/CaptainRexFightingForHisBrothers.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+                expect(context.advancedReconCommando.getHp()).toBe(4);
+                expect(context.droidCommando.getHp()).toBe(3);
```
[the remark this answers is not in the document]
supercodepoet: 👍

server/game/cards/03_TWI/leaders/CaptainRexFightingForHisBrothers.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addActionAbility({
+            title: 'If a friendly unit attacked this phase, create a Clone Trooper token.',
+            cost: [AbilityHelper.costs.abilityResourceCost(2), AbilityHelper.costs.exhaustSelf()],
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => this.attacksThisPhaseWatcher.someUnitAttackedControlledByPlayer({
+                    controller: context.source.controller,
+                    filter: (attack) => context.source !== attack.attacker
```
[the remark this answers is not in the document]
supercodepoet: yeah, I was getting a `TypeError: filter is not a function` when I initially removed the `filter` parameter. I dug into the code and it appeared that `someUnitAttackedControlledByPlayer` had a `filter?` for the parameter but still fires an exception. I am not as well-versed in Typescript as Javascript. I am digging in, but didn't know if you had ideas. If I use `filter: (entry) => true`, everything works just fine.

test/server/cards/03_TWI/bases/PauCity.spec.ts
```diff
@@ -21,7 +21,7 @@ describe('Pau City', function () {
 
             // unit should not have +0/+1
             expect(context.battlefieldMarine.getPower()).toBe(3);
-            expect(context.battlefieldMarine.getHp()).toBe(3);
+            expect(context.battlefieldMarine.getHp()).toBe(4);
```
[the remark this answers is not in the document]
supercodepoet: 👍

test/server/cards/03_TWI/leaders/CaptainRexFightingForHisBrothers.spec.ts
supercodepoet: 👍

server/game/cards/03_TWI/leaders/CaptainRexFightingForHisBrothers.ts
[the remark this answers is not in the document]
supercodepoet: nm, I think I see the bug

server/game/stateWatchers/AttacksThisPhaseWatcher.ts
```diff
@@ -56,9 +56,11 @@ export class AttacksThisPhaseWatcher extends StateWatcher<IAttacksThisPhase> {
         controller: Player;
         filter?: (event: AttackEntry) => boolean;
     }) {
-        return this.getAttackers(
-            (entry) => entry.attackingPlayer === controller && (filter(entry) ?? true)
-        ).length > 0;
+        return this.getAttackers((entry) => {
+            const additionalFilter = (filter !== undefined) ? filter(entry) : true;
```
AMMayberry1, line 60: When doing checks like this we need to use `==` and `!=` as those will match either null or undefined. It's basically the only time it's right to use those instead of `===` and `!==`

supercodepoet: gotcha

server/game/stateWatchers/AttacksThisPhaseWatcher.ts
```diff
@@ -57,6 +57,7 @@ export class AttacksThisPhaseWatcher extends StateWatcher<IAttacksThisPhase> {
         filter?: (event: AttackEntry) => boolean;
     }) {
         return this.getAttackers((entry) => {
+            // eslint-disable-next-line eqeqeq
```
AMMayberry1, line 60: Huh, really weird that this happened, usually it's okay with this type of check. Could you try doing it like this and see if that lets us remove the eslint comment (we try to avoid doing that if possible):

```ts
const additionalFilter = filter ? filter(entry) : true;
```

supercodepoet: alright, I updated to change this. I also was thinking, would it be better to check `typeof filter === function` since it needs to be a function to not error?

AMMayberry1 (approved):

PR #372, opened 2024-12-21
Now there are two of them

ssoulier:
[Now there are 2 of them](https://swudb.com/card/TWI/225/now-there-are-two-of-them)

Screenshot 2024-12-21 at 09 55 53


server/game/cards/03_TWI/events/NowThereAreTwoOfThem.ts
```diff
@@ -0,0 +1,32 @@
...
+        this.setEventAbility({
+            title: 'If you control exactly one unit, play a non-Vehicle unit from your hand that shares a Trait with the unit you control. It costs 5 less.',
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => context.source.controller.getUnitsInPlay().length === 1,
+                onFalse: AbilityHelper.immediateEffects.noAction(),
+                onTrue: AbilityHelper.immediateEffects.selectCard({
+                    activePromptTitle: 'Play a non-Vehicle unit from your hand that shared a Trait with the unit you control. It costs 5 less.',
+                    cardTypeFilter: CardType.BasicUnit,
+                    zoneFilter: ZoneName.Hand,
+                    cardCondition: (card, context) => (!card.hasSomeTrait(Trait.Vehicle)) && (Array.from(context.source.controller.getUnitsInPlay()[0].traits).some((trait) => card.hasSomeTrait(trait))),
```
AMMayberry1, line 24: nit: unnecessary parens around the boolean cases for `&&`. Also do you mind breaking this into multiple lines, same for the `playCardFromHand` below

test/server/cards/03_TWI/events/NowThereAreTwoOfThem.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                const { context } = contextRef;
+
+                // You can't play a card if you have more than one unit in play.
+                context.player1.clickCard(context.nowThereAreTwoOfThem);
+                expect(context.player2).toBeActivePlayer();
+
+                // You can't play a card if you have 0 unit in play.
+                // Reset setup
+                context.player2.passAction();
+                context.player1.setHand([context.nowThereAreTwoOfThem, context.zuckuss, context.ploKoon]);
```
AMMayberry1, line 61: In this case I would say let's just have another test case instead of completely altering board state, or else do it using game mechanisms like Superlaser Blast or something. Returning the event to hand still would have to be done manually

test/server/cards/03_TWI/events/NowThereAreTwoOfThem.spec.ts
```diff
@@ -0,0 +1,69 @@
+describe('Now There are Two of Them', function() {
+    integration(function(contextRef) {
+        describe('Now There are Two of Them\'s ability -', function() {
+            it('should allow you to play a non-Vehicle unit from your hand that shares a Trait with the unit you control for 5 less, if you have only one unit in play', function() {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['now-there-are-two-of-them', 'zuckuss#bounty-hunter-for-hire', 'cad-bane#hostage-taker', 'plo-koon#kohtoyah', 'fetts-firespray#pursuing-the-bounty'],
```
AMMayberry1, line 8: Could we put in at least one card that is just Bounty Hunter and just Underworld so we can confirm that the trait matching logic works in all cases? Also we should populate the opponent's hand with at least one valid target to make sure it's not getting included accidentally

test/server/cards/03_TWI/events/NowThereAreTwoOfThem.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                        groundArena: ['greedo#slow-on-the-draw'],
+                        base: { card: 'administrators-tower', damage: 0 },
+                        leader: 'qira#i-alone-survived',
+                        resources: 5
+                    }
+                });
+
+                const { context } = contextRef;
+
+                // You can select Zuckuss and Cad Bane. Zuckuss is free to play.
```
AMMayberry1, line 18: Could we play something that costs at least 6 so we can confirm that the discount amount is right

test/server/cards/03_TWI/events/NowThereAreTwoOfThem.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                context.player1.clickCard(context.nowThereAreTwoOfThem);
+                expect(context.player1.exhaustedResourceCount).toBe(3);
+                expect(context.player1).toBeAbleToSelectExactly([context.zuckuss, context.cadBane]);
+                context.player1.clickCard(context.zuckuss);
+                expect(context.zuckuss).toBeInZone('groundArena');
+                expect(context.player1.exhaustedResourceCount).toBe(3);
+
+                // You can't select Cad Bane if do not have enough resources.
```
AMMayberry1, line 26: This check could be done simpler by just including an 8-cost underworld card like Jabba's Rancor or the Chewie unit and confirming that it isn't targetable in the above situation. Then we wouldn't need this part at all

server/game/cards/03_TWI/events/NowThereAreTwoOfThem.ts
AMMayberry1, line 24: Why is `Array.from` needed around the `traits` property? Isn't it an array already?

AMMayberry1 (approved):

PR #381, opened 2024-12-22
Che bato/game card icons

CheBato:
- Server support for GameCard tokens such as shield, upgrades, experience, sentinel, power and hp

### DIFF
--- a/server/game/core/card/UpgradeCard.ts
+++ b/server/game/core/card/UpgradeCard.ts
@@ -54,7 +54,10 @@ export class UpgradeCard extends UpgradeCardParent {
     }
 
     public override getSummary(activePlayer: Player, hideWhenFaceup: boolean) {
-        return { ...super.getSummary(activePlayer, hideWhenFaceup), parentCardId: this._parentCard ? this._parentCard.uuid : null, aspects: this.aspects };
+        return {
+            ...super.getSummary(activePlayer, hideWhenFaceup),
+            parentCardId: this._parentCard ? this._parentCard.uuid : null
+        };
     }
 
     /** The card that this card is underneath */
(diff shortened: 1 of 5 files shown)


server/game/core/card/UpgradeCard.ts
```diff
@@ -53,6 +53,10 @@ export class UpgradeCard extends UpgradeCardParent {
         return actions;
     }
 
+    public override getSummary(activePlayer: Player, hideWhenFaceup: boolean) {
+        return { ...super.getSummary(activePlayer, hideWhenFaceup), parentCardId: this._parentCard ? this._parentCard.uuid : null, aspects: this.aspects };
```
AMMayberry1, line 57: nit: can this split into multiple lines for readability

AMMayberry1, line 57: Why are we handling the aspects here? Shouldn't that be a top-level attribute for every card?

server/gamenode/Lobby.ts
```diff
@@ -143,6 +144,26 @@ export class Lobby {
...
+        try {
+            const response = await axios.get('https://karabast-assets.s3.amazonaws.com/data/cards/' + encodeURIComponent(cardName));
+            return response.data;
+        } catch (error) {
+            console.error(`Error fetching card: ${cardName}`, error);
+            throw error;
+        }
+    }
+
+    public async setTokens(): Promise<void> {
```
AMMayberry1, line 157: Is this an okay time to also implement the initialization for the list of playable card names (e.g., for Regional Governor)? I think Balkhrog is going to work on that really soon so it would be good to go ahead and unblock him. Can be a followup PR if you prefer

danbastin: I can take a look at adding this on my PR since I'm already messing around with game initialization with my current changes.

CheBato: I prefer a follow-up PR for this if that's okay since it'd be cool to have the gameIcons available in the game.

AMMayberry1 (dismissed):

AMMayberry1 (approved):

PR #383, opened 2024-12-23
add holdout blaster

ssoulier:
[Holdout Blaster](https://swudb.com/card/twi/256/hold-out-blaster)

Screenshot 2024-12-23 at 16 12 35

### DIFF
--- a/server/game/cards/03_TWI/upgrades/HoldoutBlaster.ts
+++ b/server/game/cards/03_TWI/upgrades/HoldoutBlaster.ts
@@ -16,13 +16,14 @@ export default class HoldoutBlaster extends UpgradeCard {
 
         this.addWhenPlayedAbility({
             title: 'You may have attached unit deal 1 damage to a ground unit.',
+            optional: true,
             immediateEffect: AbilityHelper.immediateEffects.selectCard({
                 cardTypeFilter: WildcardCardType.Unit,
                 zoneFilter: ZoneName.GroundArena,
-                innerSystem: AbilityHelper.immediateEffects.damage({
-                    optional: true,
-                    amount: 1
-                }),
+                innerSystem: AbilityHelper.immediateEffects.damage((context) => ({
+                    amount: 1,
+                    source: context.source.parentCard
+                })),
             })
         });
     }
--- a/server/game/cards/03_TWI/upgrades/HoldoutBlaster.ts
+++ b/server/game/cards/03_TWI/upgrades/HoldoutBlaster.ts
@@ -17,14 +17,14 @@ export default class HoldoutBlaster extends UpgradeCard {
         this.addWhenPlayedAbility({
             title: 'You may have attached unit deal 1 damage to a ground unit.',
             optional: true,
-            immediateEffect: AbilityHelper.immediateEffects.selectCard({
-                cardTypeFilter: WildcardCardType.Unit,
+            targetResolver: {
                 zoneFilter: ZoneName.GroundArena,
-                innerSystem: AbilityHelper.immediateEffects.damage((context) => ({
+                cardTypeFilter: WildcardCardType.Unit,
+                immediateEffect: AbilityHelper.immediateEffects.damage((context) => ({
                     amount: 1,
                     source: context.source.parentCard
-                })),
-            })
+                }))
+            }
         });
     }
 }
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/upgrades/HoldoutBlaster.ts
```diff
@@ -0,0 +1,31 @@
...
+    public override setupCardAbilities () {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
+            title: 'You may have attached unit deal 1 damage to a ground unit.',
+            immediateEffect: AbilityHelper.immediateEffects.selectCard({
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: ZoneName.GroundArena,
+                innerSystem: AbilityHelper.immediateEffects.damage({
+                    optional: true,
```
AMMayberry1, line 23: Our convention is to always have the `optional` be at the ability level instead of on the effect itself, since the behaviors of the two can be a little different

server/game/cards/03_TWI/upgrades/HoldoutBlaster.ts
```diff
@@ -0,0 +1,31 @@
...
+                    amount: 1
```
AMMayberry1, line 24: Since the _unit_ is technically dealing the damage, not the upgrade, we need to add `source: context.source.parentCard` here. This matters for cards like the Jango leader

test/server/cards/03_TWI/upgrades/HoldoutBlaster.spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+            context.player1.clickCard(context.battlefieldMarine);
+            expect(context.battlefieldMarine.getPower()).toBe(4);
+            expect(context.battlefieldMarine.getHp()).toBe(3);
+            expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine, context.r2d2IgnoringProtocol]);
+            expect(context.player1).toHaveChooseNoTargetButton();
+
+            context.player1.clickCard(context.r2d2IgnoringProtocol);
+            expect(context.r2d2IgnoringProtocol.damage).toBe(1);
+            expect(context.battlefieldMarine.exhausted).toBe(false);
```
AMMayberry1, line 30: Can you add a comment: `TODO: test with Jango leader to confirm damage attribution`

test/server/cards/03_TWI/upgrades/HoldoutBlaster.spec.js
```diff
@@ -0,0 +1,33 @@
...
+                }
+            });
+
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.holdoutBlaster);
+            expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine, context.r2d2IgnoringProtocol]);
+
+            context.player1.clickCard(context.battlefieldMarine);
+            expect(context.battlefieldMarine.getPower()).toBe(4);
```
AMMayberry1, line 23: Don't need to test the unit's stats, we can assume the basic upgrade attachment stuff works

server/game/cards/03_TWI/upgrades/HoldoutBlaster.ts
ssoulier: Thanks!

server/game/cards/03_TWI/upgrades/HoldoutBlaster.ts
```diff
@@ -0,0 +1,32 @@
...
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
+            title: 'You may have attached unit deal 1 damage to a ground unit.',
+            optional: true,
+            immediateEffect: AbilityHelper.immediateEffects.selectCard({
```
AMMayberry1, line 20: Sorry, missed this on the first pass - by default we should use `targetResolver` instead of `selectCard()`. The latter has less functionality and is less easy to read. It's really just intended to be used in cases where we can't have `targetResolver`, like inside of a `sequential` or `conditional`. At some point we plan to migrate to find a way to only use `targetResolver` everywhere.

AMMayberry1 (approved):

PR #386, opened 2024-12-23
Grievous Reassembly, Disruptive Burst, Death by Droids events

aruizc:


AMMayberry1:
Quick note - in the future, do you mind including card images in the PR description? Makes it easier to double check the implementation

server/game/cards/03_TWI/events/DeathByDroids.ts
```diff
@@ -0,0 +1,29 @@
...
+            internalName: 'death-by-droids',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setEventAbility({
+            title: 'Defeat a unit that costs 3 or less. Create 2 Battle Droid tokens',
+            immediateEffect: AbilityHelper.immediateEffects.simultaneous([
+                AbilityHelper.immediateEffects.selectCard({
+                    cardTypeFilter: WildcardCardType.Unit,
```
AMMayberry1, line 18: Can remove this since we `card.isUnit()` in the condition below

server/game/cards/03_TWI/events/DeathByDroids.ts
```diff
@@ -0,0 +1,29 @@
...
+                    zoneFilter: WildcardZoneName.AnyArena,
+                    cardCondition: (card) => card.isUnit() && card.cost <= 3,
+                    innerSystem: AbilityHelper.immediateEffects.defeat(),
+                }),
+                AbilityHelper.immediateEffects.createBattleDroid((context) => ({ target: context.source.controller, amount: 2 })),
```
AMMayberry1, line 23: `context.source.controller` is the default so we can just remove the `(context)` handler and do `{ amount: 2 }`. Same for Grievous Reassembly

test/server/cards/03_TWI/events/DeathByDroids.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+        describe('Death by Droids ability', function () {
+            it('should defeat a unit that cost 3 or less and create 2 Battle Droid tokens', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['death-by-droids'],
+                        groundArena: ['fifth-brother#fear-hunter']
+                    },
+                    player2: {
+                        spaceArena: ['green-squadron-awing'],
```
AMMayberry1, line 12: Need to add a 4-cost unit to confirm that the filter is set up right

server/game/cards/03_TWI/events/DisruptiveBurst.ts
```diff
@@ -0,0 +1,22 @@
...
+export default class DisruptiveBurst extends EventCard {
+    protected override getImplementationId () {
+        return {
+            id: '3596811933',
+            internalName: 'disruptive-burst',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setEventAbility({ title: 'Give each enemy unit -1/-1 for this phase',
```
AMMayberry1, line 13: nit: the spacing for properties and open / close braces and parens in this file is inconsistent with all the others

test/server/cards/03_TWI/events/DisruptiveBurst.spec.ts
```diff
@@ -0,0 +1,45 @@
+describe('Disruptive Burst', function() {
+    integration(function(contextRef) {
+        describe('Disruptive Burst\'s ability', function() {
+            it('should apply -1/-1 to all enemy units for this phase', function () {
```
AMMayberry1, line 4: nit: can merge the `describe` into the `it` here since there's just one case and just have a top-level `it`

test/server/cards/03_TWI/events/DisruptiveBurst.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+                        hand: ['droid-deployment'],
+                        groundArena: ['consular-security-force', 'battle-droid', 'battle-droid'],
+                        spaceArena: ['tieln-fighter']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.passAction();
+                context.player2.clickCard(context.droidDeployment);
```
AMMayberry1, line 21: Looks like this isn't needed now since there are battle droids in ground arena?

test/server/cards/03_TWI/events/GrievousReassembly.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+                context.player1.clickCard(context.grievousReassembly);
+                expect(context.player1).toBeAbleToSelectExactly([context.fifthBrother, context.atst]);
+
+                // Selects heal target
+                context.player1.clickCard(context.fifthBrother);
+                const battleDroids = context.player1.findCardsByName('battle-droid');
+
+                // Validates healing
+                expect(context.fifthBrother.getHp()).toBe(4);
```
AMMayberry1, line 26: `getHp()` always returns max hp, if you want to check hp - damage need to use `remainingHp`

test/server/cards/03_TWI/events/GrievousReassembly.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+                // Validates Battle Droid token creation
+                expect(battleDroids.length).toBe(1);
```
AMMayberry1, line 29: Also need to check zone and exhausted status

test/server/cards/03_TWI/events/GrievousReassembly.spec.ts
```diff
@@ -0,0 +1,54 @@
+describe('Grievous Reassembly', function() {
+    integration(function(contextRef) {
+        describe('Grievous Reassembly\'s ability', function() {
+            it('should heal 3 from a unit and create a Battle Droid token', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['grievous-reassembly'],
+                        groundArena: [{ card: 'fifth-brother#fear-hunter', damage: 3 }]
```
AMMayberry1, line 9: We need to use at least 4 damage so we can confirm the exact amount of healing

aruizc:
Sorry, forgot to add card images:

image
image
image

server/game/cards/03_TWI/events/DeathByDroids.ts
aruizc: I don't know why, but if I don't specify an explicit target, the test cases fail. It seems like this only happens when it is used in an `immediateEffects.simultaneous()` clause, such as in `CreativeThinking`.

AMMayberry1: Ahh sorry, that's my mistake. I forgot that this is not being used inside of `targetResolver` or `selectCard` so you need to provide the target explicitly.

test/server/cards/03_TWI/events/DisruptiveBurst.spec.ts
```diff
@@ -1,45 +1,40 @@
...
-                const { context } = contextRef;
+        it('Disruptive Burst\'s ability should apply -1/-1 to all enemy units for this phase', function () {
+            contextRef.setupTest({
+                phase: 'action',
+                player1: {
+                    hand: ['disruptive-burst'],
+                    groundArena: ['fifth-brother#fear-hunter']
+                },
+                player2: {
+                    hand: ['droid-deployment'],
```
AMMayberry1, line 11: Can we remove this card if it's no longer used

AMMayberry1 (approved):

PR #389, opened 2024-12-24
Sneak attack

ssoulier:
[Sneak Attack](https://swudb.com/card/SOR/219/sneak-attack)

Screenshot 2024-12-24 at 11 01 02


server/game/cards/01_SOR/events/SneakAttack.ts
```diff
@@ -0,0 +1,39 @@
...
+            id: '3426168686',
+            internalName: 'sneak-attack'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Play a unit from your hand. It costs 3 less and enters play ready. At the start of the regroup phase, defeat it.',
+            immediateEffect: AbilityHelper.immediateEffects.selectCard({
+                activePromptTitle: 'Play a unit, it costs 3 less and enters play ready. At the start of the regroup phase, defeat it.',
```
AMMayberry1, line 18: Does the prompt not show up correctly without this? selectCard might be missing some features

server/game/cards/01_SOR/events/SneakAttack.ts
```diff
@@ -0,0 +1,39 @@
...
+                cardTypeFilter: CardType.BasicUnit,
+                zoneFilter: ZoneName.Hand,
```
AMMayberry1, line 20: Let's add `controller: RelativePlayer.Self` for now, we're in the middle of reworking some of the play-from-hand flow and it may cause issues if we don't have it

test/server/cards/01_SOR/events/SneakAttack.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+                },
+            });
+
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.sneakAttack);
+            expect(context.player1).toBeAbleToSelectExactly([context.sabineWren]);
+            context.player1.clickCard(context.sabineWren);
+            expect(context.sabineWren.exhausted).toBeFalse();
+            expect(context.player1.readyResourceCount).toBe(0);
```
AMMayberry1, line 22: Let's use a unit that costs at least 4 so we can confirm the exact amount of the discount

test/server/cards/01_SOR/events/SneakAttack.spec.ts
```diff
@@ -0,0 +1,30 @@
+describe('Sneak Attack', function() {
+    integration(function(contextRef) {
+        it('Sneak Attack ability\'s should play Sabine for 1 resource less and defeat it at the end.', function () {
+            contextRef.setupTest({
+                phase: 'action',
+                player1: {
+                    hand: ['sneak-attack', 'sabine-wren#you-can-count-on-me', 'obiwan-kenobi#following-fate'],
+                    groundArena: ['battlefield-marine'],
+                    spaceArena: ['cartel-spacer'],
+                    base: { card: 'administrators-tower', damage: 0 },
```
AMMayberry1, line 10: Can remove the damage and just do `base: 'administrators-tower'`. Same for leader, undeployed is the default so it can just be `leader: 'luke-skywalker#faithful-friend'`

test/server/cards/01_SOR/events/SneakAttack.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+            expect(context.player1).toBeAbleToSelectExactly([context.sabineWren]);
+            context.player1.clickCard(context.sabineWren);
+            expect(context.sabineWren.exhausted).toBeFalse();
+            expect(context.player1.readyResourceCount).toBe(0);
+
+            // Check that Sabine is defeated at the beginning of the regroup phase
+            context.moveToRegroupPhase();
+            expect(context.sabineWren).toBeInZone('discard');
+            expect(context.player1).toHavePrompt('Select between 0 and 1 cards to resource');
+        });
```
AMMayberry1, line 28: Delayed effects are new so it would be good to have an extra test case to validate something - can we have a test where the unit gets defeated before the end of the phase just to confirm nothing breaks?

Can be a separate test case or can just extend the existing one by moving the card back into hand and playing a different unit. I'm fine either way since moving the card back into hand is a manual action and we try to avoid those, but also reducing test cases is good. So whichever you prefer

server/game/cards/01_SOR/events/SneakAttack.ts
ssoulier: @AMMayberry1 Do you confirm we should use `selectCard` here ? I tried with `targetResolver` as you suggested for holdup blaster but I found it a bit less convenient because it allows to select a card even if we do not have enough resources to play it.
Is it something missing from targetResolver (I looked at alliance dispatcher or Han solo set 2) and they both do not filter on the cards we can play after the discount, or is it me not knowing yet how to achieve it?

test/server/cards/01_SOR/events/SneakAttack.spec.ts
ssoulier: This the case. It's Sabine Set 3.

server/game/cards/01_SOR/events/SneakAttack.ts
AMMayberry1: Yeah that must be some kind of bug, if a target is not legal for an effect then it shouldn't up in the selection list for it. I'll take a look at this when I get a chance.

AMMayberry1: Ah okay, I figured out what's going on. It's a little bit tricky (and `selectCard()` isn't behaving correctly for some reason).

Basically, since swu ability resolution is "do as many effects as possible" (i.e. it's okay if some effects don't happen), the `sequential()` and `simultaneous()` operators will allow a target to be selected if it's legal for at least one of the effects in their list. In this case, any target in hand is technically legal for the DelayedEffect just because that doesn't have any hard restrictions on what cards are legal for it to target.

So, this is like some other cases we've seen where there's really an implicit "if you do" in the card. Even though it doesn't technically say it, since the latter effect (the delayed effect) is meaningless if the first effect didn't happen, logically speaking it's the same as "if you do." Some cards like Greedo have the same issue and we solved them the same way - in Greedo's case, the damage can't resolve if no discard happened since there's no card to check to see if it's a unit.

So in this case let's do the play action as a normal effect inside of a targetResolver (no `sequential`) and then do the delayed effect as an `ifYouDo`. Feel free to ping me in chat if that doesn't make sense.

test/server/cards/01_SOR/events/SneakAttack.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+                expect(context.sabineWren.exhausted).toBeFalse();
+                expect(context.player1.readyResourceCount).toBe(0);
+
+                // Check that Sabine is defeated at the beginning of the regroup phase
+                context.moveToRegroupPhase();
+                expect(context.sabineWren).toBeInZone('discard');
+                expect(context.player1).toHavePrompt('Select between 0 and 1 cards to resource');
+            });
+
+            it('should not bug if Sabine is defeated before the end of the phase', function () {
```
AMMayberry1, line 36: Can you add a test where there are no legal cards in hand just to make sure nothing breaks with the `ifYouDo` check

test/server/cards/01_SOR/events/SneakAttack.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.sneakAttack);
+                context.player1.clickCard(context.sabineWren);
+                context.player2.clickCard(context.atst);
+                context.player2.clickCard(context.sabineWren);
+                expect(context.sabineWren).toBeInZone('discard');
+            });
+        });
```
AMMayberry1, line 45: Apologies, one other test I realized we need to check a potential bug - could you add a test where we play out the Sabine, then waylay her back to hand and play her again? In that situation, per the swu rules, she _shouldn't_ be defeated by the delayed effect since she's technically a new unit.

I'm not 100% sure that is implemented correctly in the engine though. If it doesn't work right then just comment it out with a TODO and I'll add a work item for fixing that

ssoulier: She is indeed defeated at the end. I'll comment the test and add the TODO.

AMMayberry1:
Just a heads-up, looks like a test is failing

ssoulier:
> Just a heads-up, looks like a test is failing

Thanks. I saw that but I do not understand the problem. I do not have the same error when I execute the sneak attack tests. Do you have any pointer to help me solve that?

AMMayberry1:
> > Just a heads-up, looks like a test is failing
> 
> 
> 
> Thanks. I saw that but I do not understand the problem. I do not have the same error when I execute the sneak attack tests. Do you have any pointer to help me solve that?

Hey apologies, didn't have a chance to get to this today. Will take a look first thing tomorrow

test/server/cards/01_SOR/events/SneakAttack.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+                context.player1.clickCard(context.sneakAttack);
+                context.player1.clickCard(context.sabineWren);
+                context.player2.clickCard(context.atst);
+                context.player2.clickCard(context.sabineWren);
+                expect(context.sabineWren).toBeInZone('discard');
+            });
+
+            // TODO: Fix this test cf https://github.com/SWU-Karabast/forceteki/pull/389#discussion_r1898603793
+            it('should not defeat Sabine if she is waylay back in hand and played back the same phase', function () {
```
AMMayberry1, line 77: Looks like this test just didn't get commented out?

ssoulier: Thanks. I am sorry for the noise. I missed it

AMMayberry1 (approved):

PR #390, opened 2024-12-24
add asajj ventress count dooku assassin

ssoulier:
[Asajj Ventress](https://swudb.com/card/twi/134/asajj-ventress)

Screenshot 2024-12-24 at 12 50 59

### DIFF
--- a/server/game/cards/03_TWI/units/AsajjVentressCountDookusAssassin.ts
+++ b/server/game/cards/03_TWI/units/AsajjVentressCountDookusAssassin.ts
@@ -22,7 +22,12 @@ export default class AsajjVentressCountDookusAssassin extends NonLeaderUnitCard 
         this.addOnAttackAbility({
             title: 'If you\'ve attacked with another Separatist unit this phase, this unit gets +3/+0 for this phase.',
             immediateEffect: AbilityHelper.immediateEffects.conditional({
-                condition: (context) => this.attacksThisPhaseWatcher.getAttackersInPlay((attack) => attack.attacker.hasSomeTrait(Trait.Separatist) && attack.attacker !== context.source).length > 0,
+                condition: (context) => this.attacksThisPhaseWatcher.someUnitAttackedControlledByPlayer({
+                    controller: context.source.controller,
+                    filter: (attackEvent) =>
+                        attackEvent.attacker.hasSomeTrait(Trait.Separatist) &&
+                        (attackEvent.attacker !== context.source || attackEvent.attackerInPlayId !== context.source.inPlayId)
+                }),
                 onTrue: AbilityHelper.immediateEffects.forThisPhaseCardEffect({
                     effect: AbilityHelper.ongoingEffects.modifyStats({ power: 3, hp: 0 }),
                 }),
--- a/server/game/cards/03_TWI/units/AsajjVentressCountDookusAssassin.ts
+++ b/server/game/cards/03_TWI/units/AsajjVentressCountDookusAssassin.ts
@@ -1,0 +1,35 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { Trait } from '../../../core/Constants';
+import { StateWatcherRegistrar } from '../../../core/stateWatcher/StateWatcherRegistrar';
+import { AttacksThisPhaseWatcher } from '../../../stateWatchers/AttacksThisPhaseWatcher';
+
+export default class AsajjVentressCountDookusAssassin extends NonLeaderUnitCard {
+    private attacksThisPhaseWatcher: AttacksThisPhaseWatcher;
+
+    protected override getImplementationId () {
+        return {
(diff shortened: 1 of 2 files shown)


test/server/cards/03_TWI/units/AsajjVentressCountDookusAssassin.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+            context.player2.clickCard(context.asajjVentress);
+            expect(context.consularSecurityForce.damage).toBe(5);
+            expect(context.asajjVentress.damage).toBe(3);
+
+            // Asajj Ventress should not get +3/+0 when the next-phase starts
+            context.nextPhase();
+            expect(context.asajjVentress.getPower()).toBe(2);
+            expect(context.asajjVentress.remainingHp).toBe(1);
+
+            // Even if Asaaj Ventress is Separatist, she should not get +3/+0 when she attacks
```
AMMayberry1, line 38: Right before this, could we also attack with an opponent's Separatist unit to confirm it doesn't count for the trigger?

server/game/cards/03_TWI/units/AsajjVentressCountDookusAssassin.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar) {
+        this.attacksThisPhaseWatcher = AbilityHelper.stateWatchers.attacksThisPhase(registrar, this);
+    }
+
+    public override setupCardAbilities () {
+        this.addOnAttackAbility({
+            title: 'If you\'ve attacked with another Separatist unit this phase, this unit gets +3/+0 for this phase.',
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => this.attacksThisPhaseWatcher.getAttackersInPlay((attack) => attack.attacker.hasSomeTrait(Trait.Separatist) && attack.attacker !== context.source).length > 0,
```
AMMayberry1, line 25: This isn't quite right, due to the SWU rules around the idea of a "copy" of a card - i.e., every time a card enters play, it becomes a new "copy" for history purposes. So e.g. if Asajj attacks, is defeated, then played again, she would in that case count her previous attack as "another unit" doing a Separatist attack since she is now a new copy.

The way to implement that correctly should be:

```ts
  condition: (context) => this.attacksThisPhaseWatcher.someUnitAttackedControlledByPlayer({
      controller: context.source.controller,
      filter: (attackEvent) =>
          attackEvent.attacker.hasSomeTrait(Trait.Separatist) &&
          (attackEvent.attacker !== context.source || attackEvent.attackerInPlayId !== context.source.inPlayId)
  }),
```

Also we need to add a test confirming the case works, but I left a separate comment for that

test/server/cards/03_TWI/units/AsajjVentressCountDookusAssassin.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+            context.player1.clickCard(context.shaakTi);
+            context.player1.clickCard(context.p2Base);
+            expect(context.p2Base.damage).toBe(13);
+
+            context.player2.passAction();
+            context.player1.clickCard(context.asajjVentress);
+            context.player1.clickCard(context.p2Base);
+            expect(context.asajjVentress.getPower()).toBe(2);
+            expect(context.p2Base.damage).toBe(15);
+        });
```
AMMayberry1, line 55: Re: the note on how to implement the rule for separate copies of the card, we need to add a case very similar to the last test case in the Lothal Insurgent tests. Basically it's this sequence of steps:

- Asajj attacks
- Asajj is returned to hand and played back out
- Asajj attacks again, this time getting the bonus because her ability counts her previous copy as a separate unit

Since that's a pretty different test setup it can be a separate case, though if you want to do the legwork to keep it as part of this one that's fine as well

server/game/cards/03_TWI/units/AsajjVentressCountDookusAssassin.ts
ssoulier: Thanks for explaining clearly this rule point. I'll try to implement it (not sure yet how to achieve it).

ssoulier: I do not understand why we we need this `attackEvent.attacker !== context.source`?
Is it because before the end of the first attack, there is no `attackerInPlayId`?

AMMayberry1: The inPlayId is just an incrementing counter that's unique to each card. So we can't compare them across two different card objects, only repeated plays of the same card.

Another way to think about it is that the statement `attackEvent.attacker !== context.source || attackEvent.attackerInPlayId !== context.source.inPlayId` is basically saying that a card counts as being a "different" card if either it's a different card than this one altogether; OR if it's the same card, it was a previous instance of it (previous inPlayId)

AMMayberry1 (approved):

PR #391, opened 2024-12-26
Implementing Private Manufacturing

supercodepoet:
image


server/game/core/card/Card.ts
```diff
@@ -310,6 +310,10 @@ export class Card extends OngoingEffectSource {
         return this.type === CardType.TokenUnit || this.type === CardType.TokenUpgrade;
     }
 
+    public isTokenUnit(): this is TokenCard {
```
AMMayberry1, line 313: Should be `this is TokenUnitCard`. If we're adding this we should also add `isTokenUpgrade()` for completeness

test/server/cards/03_TWI/events/PrivateManufacturing.spec.ts
```diff
@@ -0,0 +1,46 @@
+describe('Private Manufacturing', function () {
+    integration(function (contextRef) {
+        describe('Private Manufacturing\'s ability', function () {
+            it('should draw two cards and not ask to move two cards to bottom of the deck', function () {
```
AMMayberry1, line 4: nit: test name should include the condition so it's clearer what's going on. i.e., should end with "because there are no token units". Similar note for the other test name

test/server/cards/03_TWI/events/PrivateManufacturing.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+                context.player1.clickCard(context.privateManufacturing);
+                expect(context.player1.hand.length).toBe(4);
+                expect(context.player1).toHavePrompt('Select 2 cards');
+                expect(context.player1).toBeAbleToSelectExactly(context.player1.hand);
+
+                context.player1.clickCard(context.republicTacticalOfficer);
+                context.player1.clickCard(context.advancedReconCommando);
+                context.player1.clickPrompt('Done');
+
+                const bottomDeckedCards = context.player1.deck.slice(-2);
```
AMMayberry1, line 38: Can use `.toAllBeInBottomOfDeck()` instead of using this approach

test/server/cards/03_TWI/events/PrivateManufacturing.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.privateManufacturing);
+                expect(context.player1.hand.length).toBe(4);
+                expect(context.player1).toHavePrompt('Select 2 cards');
+                expect(context.player1).toBeAbleToSelectExactly(context.player1.hand);
+
+                context.player1.clickCard(context.republicTacticalOfficer);
+                context.player1.clickCard(context.advancedReconCommando);
```
AMMayberry1, line 35: Since we made some changes around the conditions for this, can we have some expectations like `expect(context.player1).not.toHaveEnabledPromptButton('Done');` when not enough targets are selected

server/game/core/card/Card.ts
supercodepoet: yep, good catch. Added the additional method.

test/server/cards/03_TWI/events/PrivateManufacturing.spec.ts
supercodepoet: yeah, thanks for that. I had a hard time coming up with a test sentence that wasn't weird.

test/server/cards/03_TWI/events/PrivateManufacturing.spec.ts
supercodepoet: added the check

test/server/cards/03_TWI/events/PrivateManufacturing.spec.ts
supercodepoet: awesome! thanks!

supercodepoet:
As always, thank you for the great review. It really helps getting up to speed with more the the depths of the system and how you like the tests to function. You do great work for the community.

AMMayberry1 (approved):

PR #392, opened 2024-12-26
Implementing Bazine, Political Pressure and Unmasking The Conspiracy

aruizc:
image
image
image


test/server/cards/02_SHD/units/BazineNetalSpyForTheFirstOrder.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+                    player1: {
+                        hand: ['bazine-netal#spy-for-the-first-order'],
+                    },
+                    player2: {
+                        deck: ['wampa']
+                    }
+                });
+                const { context } = contextRef;
+
+                // Player looks at the opponent's hand and discards a card from it, opponent draws a card
```
AMMayberry1, line 65: nit: these comments are wrong. Fantastic tests though!

server/game/cards/03_TWI/events/PoliticalPressure.ts
```diff
@@ -0,0 +1,28 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Discard a random card from your hand',
+            targetResolver: {
+                choosingPlayer: RelativePlayer.Opponent,
+                mode: TargetMode.Select,
+                choices: () => ({
```
AMMayberry1, line 19: There's a little bit of nuance to this but the short version is that this implementation needs to use the `ifYouDoNot` property instead of being set up as a choice like this. As it is right now, the opponent could choose random discard when they have an empty hand and avoid creating the tokens. But the way "if you do not" works, if their hand is empty they _must_ accept the battle droid tokens. We'll need to add a test case for that.

The `ifYouDoNot` property should handle that correctly, though it's fairly new so it might be buggy. Lmk if you have any problems with it.

test/server/cards/03_TWI/events/PoliticalPressure.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+            expect(context.player2).toHaveEnabledPromptButtons(['Discard a random card from your hand', 'Opponent creates 2 Battle Droid Tokens']);
+            context.player2.clickPrompt('Discard a random card from your hand');
+            expect(context.player2.getCardsInZone('discard').length).toBe(1);
+            expect(context.player1.getCardsInZone('groundArena').length).toBe(0); // No Battle Droids tokens are created
+            expect(context.player2.hand.length).toBe(2);
+
+            // Reset
+            reset();
+
+            // Opponent decides to create Battle Droids tokens
```
AMMayberry1, line 33: As noted above, needs a test case with opponent having an empty hand

test/server/cards/03_TWI/events/PoliticalPressure.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+            context.player1.clickCard(context.politicalPressure);
+            expect(context.player2).toHaveEnabledPromptButtons(['Discard a random card from your hand', 'Opponent creates 2 Battle Droid Tokens']);
+            context.player2.clickPrompt('Opponent creates 2 Battle Droid Tokens');
+            const battleDroids = context.player1.findCardsByName('battle-droid');
```
AMMayberry1, line 37: Also need to check that the random discard didn't happen

server/game/cards/03_TWI/events/UnmaskingTheConspiracy.ts
```diff
@@ -0,0 +1,37 @@
...
+            id: '0959549331',
+            internalName: 'unmasking-the-conspiracy',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Discard a card from your hand',
+            immediateEffect: AbilityHelper.immediateEffects.discardCardsFromOwnHand((context) => ({
+                cardTypeFilter: WildcardCardType.Any,
```
AMMayberry1, line 17: I think this might actually cause problems if it's here, it's the default so we can remove it

server/game/cards/03_TWI/events/UnmaskingTheConspiracy.ts
```diff
@@ -0,0 +1,37 @@
...
+                target: context.source.controller,
+                amount: 1
+            })),
+            ifYouDo: () => ({
```
AMMayberry1, line 21: If no `ifYouDoContext` is needed, can juts remove the handler method and do `ifYouDo: {`

test/server/cards/03_TWI/events/UnmaskingTheConspiracy.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+            // Discards a card from the opponent's hand
+            context.player1.clickCard(context.atst);
+            expect(context.atst).toBeInZone('discard');
+            expect(context.player2.hand.length).toBe(1);
+
+            // Reset
+            reset();
+
+            // Player does not have a card to discard and does not looks at the opponent's hand to discard a card from it
```
AMMayberry1, line 40: Can you add one more case where player1 chooses to discard but player2's hand is empty? Just to confirm that doesn't cause any issues with like empty arrays and stuff not resolving right

AMMayberry1 (approved):

PR #394, opened 2024-12-27
Support for test games

danbastin:
Adds in the workflow elements for starting a testgame. I've added one file to `test/gameSetups` as a template example but changes to this file and other custom files should be ignored.


server/gamenode/GameServer.ts
```diff
@@ -115,6 +125,30 @@ export class GameServer {
         return true;
     }
 
+    private startTestGame(filename: string) {
+        const lobby = new Lobby();
+        this.lobbies.set(lobby.id, lobby);
+        const order66 = { id: 'exe66', username: 'Order66' };
+        const theWay = { id: 'th3w4y', username: 'ThisIsTheWay' };
```
AMMayberry1, line 132: Since these will be referred to as player1 and player2 in the test setups, do you think the fact that those don't map clearly onto these names would cause any issues or confusion?

test/gameSetups/testGameTemplate.json
```diff
@@ -0,0 +1,11 @@
+{
```
AMMayberry1, line 1: Wondering if maybe the example setup should be a little more complex so people know what all the configuration options are. I can send an example one with more stuff going on if that would help

danbastin: Yeah, was thinking the same. An example that touched all options would be good to have as the default.

server/gamenode/GameServer.ts
danbastin: Yeah, lack of clarity has been sacrificed in the name of fun here. Order66 has been the default player 1 for a while now so thats just how I've set to thinking about it. It would probably be good to put in a little extra work and make it so a test game can be created with either account (right now you have to be Order66 then join as ThisIsTheWay). If all of this becomes too cumbersome or confusing for people we can simplify.

AMMayberry1 (approved):

PR #395, opened 2024-12-27
Add Subjugating Starfighter and tests.

nickpelton:
My First Card.

[Subjugating Starfighter](https://swudb.com/card/TWI/112/subjugating-starfighter)


AMMayberry1:
Thanks for the PR! Will review sometime tomorrow. In the meantime, would you mind pinging me (Veld) in the discord so we can get you into the dev conversations?

test/server/cards/03_TWI/units/SubjugatingStarfighter.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+                // Play the card.
+                context.player1.clickCard(context.subjugatingStarfighter);
+
+                // Check if the Battle Droid token was created if the player has initiative.'
+                context.player1.clickPrompt('Create a Battle Droid token.');
+                expect(context.player1.findCardsByName('battle-droid').length).toBe(1);
+                expect(context.player1.findCardsByName('battle-droid')).toAllBeInZone('groundArena', context.player1);
+                expect(context.player1.findCardsByName('battle-droid').every((battleDroid) => battleDroid.exhausted)).toBeTrue();
+
+                // Resolve ambush
```
AMMayberry1, line 26: We don't need to test this since it's not really part of the ability that we had to explicitly implement. Can remove this part and set `context.allowTestToEndWithOpenPrompt = true;` so there isn't an error about the open prompt (should leave a comment explaining it though). Same for other test case below.

nickpelton: Thanks for the tip! Ran into that open prompt issue and wasn't sure how to bypass so I completed the prompt. Removed tests and bypassed open prompt and tests pass.

AMMayberry1 (approved):

PR #396, opened 2024-12-27
fix: hide opposing players cards in the gamestate

farisT:
- Removed the `hideWhenFaceup` element (remnant from jigoku engine)
- Return facedown cards with no details based off active player and zone (e.g. only show opposing cards when they are in play)

### DIFF
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -792,7 +792,7 @@ export class Card extends OngoingEffectSource {
         const selectionState = activePlayer.getCardSelectionState(this);
 
         // If it is not the active player and in opposing hand or deck - return facedown card
-        if (!isActivePlayer && (this._zone.hiddenForPlayers === RelativePlayer.Opponent || this._zone.hiddenForPlayers === WildcardRelativePlayer.Any)) {
+        if (this._zone.hiddenForPlayers === WildcardRelativePlayer.Any || (!isActivePlayer && this._zone.hiddenForPlayers === RelativePlayer.Opponent)) {
             const state = {
                 controller: this.controller.getShortSummary(),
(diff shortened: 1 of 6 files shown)


danbastin:
Condition looks right to me. This should be good to g.

server/game/core/card/Card.ts
```diff
@@ -787,17 +787,12 @@ export class Card extends OngoingEffectSource {
...
 
-        // This is my facedown card, but I'm not allowed to look at it
-        // OR This is not my card, and it's either facedown or hidden from me
-        if (
-            isActivePlayer
-                ? this.facedown
-                : this.facedown || hideWhenFaceup
-        ) {
+        // If it is not the active player and in opposing hand or deck - return facedown card
+        if (!isActivePlayer && (this._zone.hiddenForPlayers === RelativePlayer.Opponent || this._zone.hiddenForPlayers === WildcardRelativePlayer.Any)) {
```
AMMayberry1, line 795: I think this condition is slightly off - `WildcardRelativePlayer.Any` means the zone should be hidden regardless of whether we are the active player. If I'm understanding right, I think it should be:

```ts
if (this._zone.hiddenForPlayers === WildcardRelativePlayer.Any || (!this.activePlayer && this._zone.hiddenForPlayers === RelativePlayer.Opponent))

AMMayberry1:
Just a note - please don't force push if at all possible, it erases the review history and makes it harder to identify the new changes.

server/game/core/card/Card.ts
farisT: Ah good one, we should indeed return facedown if the condition `WildcardRelativePlayer.Any` is returned regardless if it is the active player

AMMayberry1 (approved):

PR #401, opened 2024-12-29
#LobbyFunctions

CheBato:
- Added LobbyObject to use for getting information on the lobby.
- Added chatFunction.
- Added opponent display.
- Added ready function.


server/gamenode/Lobby.ts
```diff
@@ -15,6 +16,7 @@ interface LobbyUser {
     ready: boolean;
     socket: Socket | null;
     deck: Deck | null;
+    owner: boolean;
```
AMMayberry1, line 19: Something like `ownsLobby` or `isLobbyOwner` might be slightly clearer

server/gamenode/Lobby.ts
```diff
@@ -23,23 +25,49 @@ export class Lobby {
...
+            id: this._id,
+            users: this.users.map((u) => ({
+                id: u.id,
+                username: u.username,
+                state: u.state,
+                ready: u.ready,
+                deck: u.deck?.data,
+                owner: u.owner
+            })),
+            gameChat: this.gameChat,
```
AMMayberry1, line 50: Would it be useful to have like the id or username of the owner as top level property?

server/gamenode/Lobby.ts
```diff
@@ -215,6 +263,17 @@ export class Lobby {
         this.sendGameState(game);
     }
 
+    private onLobbyMessage(socket: Socket, command: string, ...args): void {
+        if (!this[command] || typeof this[command] !== 'function') {
+            return;
```
AMMayberry1, line 268: Should this be an exception?

server/gamenode/Lobby.ts
```diff
@@ -26,6 +25,7 @@ export class Lobby {
     private users: LobbyUser[] = [];
     private tokens: { battleDroid: any; cloneTrooper: any; experience: any; shield: any };
     private gameChat: GameChat;
+    private lobbyOwner: string;
```
AMMayberry1, line 28: Since we have multiple ways of identifying a user, probably good to be explicit and call this `lobbyOwnerId`

AMMayberry1 (approved):

PR #405, opened 2025-01-01
Bossk leader

AMMayberry1:
Refactored the bounty resolution process to create an explicit "onBountyCollected" event, and made it possible to store the properties of the bounty keyword in the event. Added the CollectBountySystem which the Bossk ability uses to re-trigger the bounty ability.

Added a big test suite to cover the various types of bounties and ensure that all can double-resolve correctly. Added some specific detailed tests for the Bounty Hunter's Quarry case, including an open TODO on one case while we wait for clarification on some rules.


server/game/abilities/keyword/BountyAbility.ts
```diff
@@ -0,0 +1,68 @@
...
+import { GameEvent } from '../../core/event/GameEvent';
+import type Game from '../../core/Game';
+import * as Contract from '../../core/utils/Contract';
+import type { ITriggeredAbilityBaseProps } from '../../Interfaces';
+
+export type IResolvedBountyProperties = Omit<ITriggeredAbilityBaseProps, 'abilityController'> & {
+    bountySource?: UnitCard;
+};
+
+export class BountyAbility extends TriggeredAbility {
```
adbronico, line 14: Could you explain or leave some notes in here about what this class is used for?

adbronico (dismissed):

PR #406, opened 2025-01-01
#QuickPlay

CheBato:
- Support for Quickplay.
- Added matchmaking queue.
- Changed that the lobby doesn't show if a user disconnects and a game already exists.

### DIFF
--- a/server/gamenode/GameServer.ts
+++ b/server/gamenode/GameServer.ts
@@ -12,12 +12,21 @@ import { Lobby } from './Lobby';
 import Socket from '../socket';
 import * as env from '../env';
 
+/**
+ * Represents a player waiting in the queue.
+ */
+interface QueuedPlayer {
+    user: any;
+    deck: any;
+    socket: socketio.Socket | null;
+}
+
 export class GameServer {
     private lobbies = new Map<string, Lobby>();
     private userLobbyMap = new Map<string, string>();
-
     private protocol = 'https';
     private host = env.gameNodeHost;
+    private queue: QueuedPlayer[] = [];
     private io: socketio.Server;
     private titleCardData: any;
     private shortCardData: any;
@@ -104,6 +113,14 @@ export class GameServer {
         app.post('/api/start-test-game', (req, res) => {
             const { filename } = req.body;
             this.startTestGame(filename);
+            return res.status(200).json({ success: true });
+        });
+        app.post('/api/enter-queue', (req, res) => {
+            const { user, deck } = req.body;
+            const success = this.enterQueue(user, deck, null);
+            if (!success) {
+                return res.status(400).json({ success: false, message: 'Failed to enter queue' });
+            }
             return res.status(200).json({ success: true });
         });
     }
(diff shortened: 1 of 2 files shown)


server/gamenode/GameServer.ts
```diff
@@ -12,12 +12,21 @@ import { Lobby } from './Lobby';
 import Socket from '../socket';
 import * as env from '../env';
 
+/**
+ * Represents a player waiting in the queue.
+ */
+interface QueuedPlayer {
+    user: any;
+    deck: any;
+    socket: socketio.Socket | null;
```
AMMayberry1, line 21: nit: can do `socket?: socketio.Socket` instead of the explicit `| null`

server/gamenode/GameServer.ts
```diff
@@ -206,21 +223,111 @@ export class GameServer {
...
+     * Matchmake two users in a queue
+     */
+    private matchmakeQueuePlayers() {
+        // Simple approach: if at least 2 in queue, pair them up
+        while (this.queue.length >= 2) {
+            const p1 = this.queue.shift();
+            const p2 = this.queue.shift();
+            if (!p1 || !p2) {
+                break;
+            } // Just in case
```
AMMayberry1, line 288: Exception?

server/gamenode/GameServer.ts
```diff
@@ -206,21 +223,111 @@ export class GameServer {
...
+
+            // Create a new Lobby
```
AMMayberry1, line 290: Out of curiosity, how slow is this process? Since the queue might get a little large, if the time to set up each individual game is slow it could cause delays since this is running synchronously.

If it's slow at all then maybe we could hand the lobby creation off to an async worker, but that can get tricky so probably shouldn't try if it doesn't end up being an issue

server/gamenode/GameServer.ts
```diff
@@ -206,21 +223,111 @@ export class GameServer {
...
+    /**
+     * Remove the user from the queue if they disconnect or otherwise.
+     */
+    private removeFromQueue(userId: string): void {
+        this.queue = this.queue.filter((q) => q.user.id !== userId);
     }
 
     public onSocketDisconnected(id: string, reason: string) {
         if (!this.userLobbyMap.has(id)) {
+            this.removeFromQueue(id);
```
AMMayberry1, line 330: I'm pretty positive this isn't thread-safe, if this is trying to modify the queue object at the same time as `matchmakeQueuePlayers` then there will be problems

server/gamenode/Lobby.ts
```diff
@@ -180,7 +180,7 @@ export class Lobby {
     }
 
     public isLobbyFilled(): boolean {
-        return this.users.length === 2;
+        return this.users.length === 2 && !this.game;
```
AMMayberry1, line 183: Just curious, what's the reason for this change?

CheBato: The reason was that when someone left gameBoard when the game was ongoing the lobby becomes visible on the left side of the homepage.

AMMayberry1 (approved):

PR #412, opened 2025-01-03
Maul1 unit

AMMayberry1:
Implemented the Maul1 unit, which required a lot of tweaking of various engine systems. Major changes:

- Added explicit support for gaining replacement effect abilities
- Partial implementation of a separate window type for replacement effects, which better handles the special rules for that case
- Update the DamageSystem and helpers to facilitate explicitly creating new combat damage events and enforcing the right properties for damage events across the board


test/server/cards/02_SHD/units/MaulShadowCollectiveVisionary.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+                    'Defeat shield to prevent attached unit from taking damage',
+                    'Redirect combat damage to another Underworld unit'
+                ]);
+                context.player1.clickPrompt('Defeat shield to prevent attached unit from taking damage');
+
+                expect(context.maul.damage).toBe(0);
+                expect(context.maul.isUpgraded()).toBeFalse();
+                expect(context.luminaraUnduli.damage).toBe(7);
+                expect(context.mercenaryCompany.damage).toBe(0);
+            });
```
brianries, line 224: Other possible test cases.

- Does sentinel have any possible interaction here? Could a sentineled maul send the combat damage to non sentinels? Nothing tries to intercept or block it?
- Out of aspect and random, but the replaced combat damage redirected at chewbecca wouldn't ready him would it?

server/game/core/gameSteps/abilityWindow/ReplacementEffectWindow.ts
```diff
@@ -0,0 +1,118 @@
...
+import type { Card } from '../../card/Card';
+import { AbilityType } from '../../Constants';
+import type { EventWindow } from '../../event/EventWindow';
+import type Game from '../../Game';
+import type Player from '../../Player';
+import { TriggerWindowBase } from './TriggerWindowBase';
+import type Shield from '../../../cards/01_SOR/tokens/Shield';
+
+export class ReplacementEffectWindow extends TriggerWindowBase {
+    private newReplacementEvents = true;
```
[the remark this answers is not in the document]
AMMayberry1: Ah yeah, I struggled with this a bit. It's awkward because I'm still trying to roughly conform to the existing `continue()` function for now and we're going to just write a new one for this class eventually. But basically this needs to start as true so that on the first pass we run all the trigger cleanup and prompt setup stuff. I'll add a comment explaining it.

test/server/cards/02_SHD/units/MaulShadowCollectiveVisionary.spec.ts
AMMayberry1: Sentinel is a good thought. I think though since that only affects the attack targeting stage which is a totally separate system and code path that it's probably not worth considering it for tests here.

Re: the chewbacca question - I'm assuming you're thinking of Chewbacca, Loyal Companion? This is the blue one (Pykesbane). But yeah it shouldn't affect him since he's not the target of an attack, and for similar reasons as above I don't think it's worth adding that as a test.

PR #413, opened 2025-01-03
#PrivateLobbies

CheBato:
- Added private lobbies to gameServer.
- Made it possible to join via links.
- Made it possible for players to create private lobbies without logging in.

### DIFF
--- a/server/gamenode/GameServer.ts
+++ b/server/gamenode/GameServer.ts
@@ -15,12 +15,20 @@ import * as env from '../env';
 import type { Deck } from '../game/Deck';
 
 /**
+ * Represents a user object
+ */
+interface User {
+    id: string;
+    username: string;
+}
+
+/**
  * Represents a player waiting in the queue.
  */
 interface QueuedPlayer {
     deck: Deck;
     socket?: socketio.Socket;
-    user: { id: string; username: string };
+    user: User;
 }
 
 export class GameServer {
@@ -80,8 +88,8 @@ export class GameServer {
 
     private setupAppRoutes(app: express.Application) {
         app.post('/api/create-lobby', (req, res) => {
-            const new_user_id = this.createLobby(req.body.user, req.body.deck, req.body.privacy);
-            return res.status(200).json({ success: true, newUserId: new_user_id });
+            const newUserId = this.createLobby(req.body.user, req.body.deck, req.body.isPrivate);
+            return res.status(200).json({ success: true, newUserId: newUserId });
         });
         app.get('/api/available-lobbies', (_, res) => {
             const availableLobbies = Array.from(this.lobbiesWithOpenSeat().entries()).map(([id, _]) => ({
@@ -127,21 +135,36 @@ export class GameServer {
     private lobbiesWithOpenSeat() {
         return new Map(
             Array.from(this.lobbies.entries()).filter(([_, lobby]) =>
-                !lobby.isLobbyFilled() && lobby.isLobbyPublic && !lobby.hasOngoingGame
+                !lobby.isLobbyFilled() && !lobby.isLobbyPrivate() && !lobby.hasOngoingGame()
             )
         );
     }
 
-    private createLobby(user: any, deck: any, privacy: string) {
-        const lobby = new Lobby(privacy);
+    /**
+     * Creates a new lobby for the given user. If no user is provided and
+     * the lobby is private, a default user is created.
+     *
+     * @param {User | null} user - The user creating the lobby. If null is passed in for a private lobby, a default user is created.
+     * @param {Deck} deck - The deck used by this user.
+     * @param {boolean} isPrivate - Whether or not this lobby is private.
+     * @returns {string} The ID of the user who owns and created the newly created lobby.
+     */
+    private createLobby(user: User | null, deck: Deck, isPrivate: boolean) {
+        if (!isPrivate && !user) {
+            throw new Error('User must be provided for public lobbies');
+        }
+
+        const lobby = new Lobby(isPrivate);
         this.lobbies.set(lobby.id, lobby);
-        // set users for private lobbies
+        // set default user if no user is supplied for private lobbies
         if (!user) {
             user = { id: uuid(), username: 'Player1' };
         }
+
         lobby.createLobbyUser(user, deck);
         lobby.setLobbyOwner(user.id);
         this.userLobbyMap.set(user.id, lobby.id);
+
         lobby.setTokens();
         return user.id;
     }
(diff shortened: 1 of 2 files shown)


server/gamenode/GameServer.ts
```diff
@@ -79,11 +80,8 @@ export class GameServer {
 
     private setupAppRoutes(app: express.Application) {
         app.post('/api/create-lobby', (req, res) => {
-            if (this.createLobby(req.body.user, req.body.deck)) {
-                return res.status(200).json({ success: true });
-            }
-
-            return res.status(400).json({ success: false });
+            const new_user_id = this.createLobby(req.body.user, req.body.deck, req.body.privacy);
```
AMMayberry1, line 83: nit: snake case

server/gamenode/Lobby.ts
```diff
@@ -26,10 +26,14 @@ export class Lobby {
     private tokens: { battleDroid: any; cloneTrooper: any; experience: any; shield: any };
     private gameChat: GameChat;
     private lobbyOwnerId: string;
+    private privacy: string;
+    private connectionLink: string;
 
-    public constructor() {
+    public constructor(privacy: string = 'public') {
```
AMMayberry1, line 32: Feels like the privacy setting should probably be an enum - or maybe even just a boolean `isPrivate`?

server/gamenode/GameServer.ts
```diff
@@ -128,18 +126,24 @@ export class GameServer {
...
         );
     }
 
-    private createLobby(user: any, deck: any) {
-        const lobby = new Lobby();
+    private createLobby(user: any, deck: any, privacy: string) {
+        const lobby = new Lobby(privacy);
         this.lobbies.set(lobby.id, lobby);
+        // set users for private lobbies
+        if (!user) {
```
AMMayberry1, line 139: This doesn't match the type signature - `user` is not an optional parameter currently. Also, if it is the case that we only permit `user` to be null when it's a private game then we should have a contract check confirming that

server/gamenode/GameServer.ts
```diff
@@ -128,18 +126,24 @@ export class GameServer {
...
+            user = { id: uuid(), username: 'Player1' };
+        }
         lobby.createLobbyUser(user, deck);
         lobby.setLobbyOwner(user.id);
         this.userLobbyMap.set(user.id, lobby.id);
         lobby.setTokens();
-        return true;
+        return user.id;
```
AMMayberry1, line 146: Can we add a jsdoc explaining what the return value is, since it's not obvious from the name of the method

server/gamenode/GameServer.ts
```diff
@@ -128,18 +126,24 @@ export class GameServer {
...
-            Array.from(this.lobbies.entries()).filter(([_, lobby]) => !lobby.isLobbyFilled())
+            Array.from(this.lobbies.entries()).filter(([_, lobby]) =>
+                !lobby.isLobbyFilled() && lobby.isLobbyPublic
+            )
         );
     }
 
-    private createLobby(user: any, deck: any) {
-        const lobby = new Lobby();
+    private createLobby(user: any, deck: any, privacy: string) {
```
AMMayberry1, line 135: Do we have types / interfaces for `deck` and `user` that we could use here?

server/gamenode/GameServer.ts
```diff
@@ -215,7 +219,8 @@ export class GameServer {
     }
 
     public onConnection(ioSocket) {
-        const user = JSON.parse(ioSocket.handshake.query.user);
+        let user = JSON.parse(ioSocket.handshake.query.user);
+        const requestedLobbyId = ioSocket.handshake.query.lobbyId || '';
```
AMMayberry1, line 223: We should avoid using `''` as a pseudo-null and just use `null` instead, since `''` still behaves as a string for most purposes it could cause issues

server/gamenode/GameServer.ts
```diff
@@ -232,12 +238,30 @@ export class GameServer {
...
+                this.userLobbyMap.set(newUser.id, lobby.id);
+                socket.send('connectedUser', newUser.id);
+                socket.on('disconnect', (_, reason) => this.onSocketDisconnected(user.id, reason));
+                return;
+            }
+            lobby.addLobbyUser(user, socket);
+            this.userLobbyMap.set(user.id, lobby.id);
+            return;
+        }
         // if they are not in the lobby they could be in a queue
```
AMMayberry1, line 265: Not super related but just caught this - if `queuedPlayer` is false, does that mean we're in an error state? If so, we should throw an exception at the end of the method with some debug info (and pass an error back to the user, once we have a framework in place for that)

server/gamenode/GameServer.ts
```diff
@@ -215,7 +219,8 @@ export class GameServer {
     }
 
     public onConnection(ioSocket) {
```
AMMayberry1, line 221: nit: this method is one giant unbroken code block, could we break it up with newlines a little bit for readability

server/gamenode/Lobby.ts
```diff
@@ -79,7 +85,7 @@ export class Lobby {
         }
     }
 
-    public addLobbyUser(user, socket: Socket, owner = false): void {
+    public addLobbyUser(user, socket: Socket): void {
         const existingUser = this.users.find((u) => u.id === user.id);
```
AMMayberry1, line 89: Not exactly sure where this would go so I'm leaving the comment here - do we have a check somewhere that prevents a third player from trying to join a private lobby?

server/gamenode/GameServer.ts
```diff
@@ -232,12 +238,30 @@ export class GameServer {
...
+            user = lobby.getLobbyUserById(user.id);
             const socket = new Socket(ioSocket);
             lobby.addLobbyUser(user, socket);
+            socket.send('connectedUser', user.id);
             socket.on('disconnect', (_, reason) => this.onSocketDisconnected(user.id, reason));
             return;
         }
-
+        // 2. If user connected to the lobby via a link.
+        if (requestedLobbyId && this.lobbies.has(requestedLobbyId)) {
```
AMMayberry1, line 250: Should we provide some feedback to the user if the requested lobby ID no longer exists?

server/gamenode/Lobby.ts
```diff
@@ -26,10 +26,14 @@ export class Lobby {
...
     private gameChat: GameChat;
     private lobbyOwnerId: string;
+    private privacy: string;
+    private connectionLink: string;
 
-    public constructor() {
+    public constructor(privacy: string = 'public') {
         this._id = uuid();
         this.gameChat = new GameChat();
+        this.connectionLink = `http://localhost:3000/lobby?lobbyId=${this._id}`;
```
AMMayberry1, line 35: I'm thinking we should only generate this if the lobby is private?

server/gamenode/Lobby.ts
CheBato: We could do an OnConnect check for a third player since we will now support getting into the lobby via link. But that will only work if the lobby is already filled.

server/gamenode/GameServer.ts
CheBato: Absolutely agree

server/gamenode/Lobby.ts
```diff
@@ -190,16 +195,12 @@ export class Lobby {
...
-    }
-
     public isLobbyFilled(): boolean {
-        return this.users.length === 2 && !this.game;
+        return this.users.length === 2;
     }
 
-    public isLobbyPublic(): boolean {
-        return this.privacy === 'Public';
+    public isLobbyPrivate(): boolean {
```
AMMayberry1, line 202: nit: just noticed this now, but having `Lobby` in these method names is kind of redundant, e.g. `Lobby.isLobbyPrivate()` vs `Lobby.isPrivate()`

server/gamenode/Lobby.ts
```diff
@@ -190,16 +195,12 @@ export class Lobby {
...
+        return this.isPrivate;
```
AMMayberry1, line 203: An additional thought - instead of using a method to return this value, you could just make it a `public readonly` property of the class. Same for other similar getters in here

server/gamenode/Lobby.ts
```diff
@@ -26,15 +31,19 @@ export class Lobby {
     private tokens: { battleDroid: any; cloneTrooper: any; experience: any; shield: any };
     private gameChat: GameChat;
     private lobbyOwnerId: string;
-    private isPrivate: boolean;
+    public isPrivate: boolean;
```
AMMayberry1, line 34: nit: should be `readonly`, looks like most or all others should as well

server/gamenode/Lobby.ts
```diff
@@ -26,11 +31,19 @@ export class Lobby {
...
     private lobbyOwnerId: string;
+    public isPrivate: boolean;
+    private connectionLink: string | null;
     private playableCardTitles: string[];
 
-    public constructor() {
+    public constructor(lobbyGameType: MatchType) {
+        Contract.assertTrue(
+            [MatchType.Custom, MatchType.Private, MatchType.Quick].includes(lobbyGameType),
+            'Lobby game type doesn\'t match any MatchType values'
```
AMMayberry1, line 41: nit: we should include the value in the error message

server/gamenode/Lobby.ts
```diff
@@ -26,11 +31,19 @@ export class Lobby {
     private tokens: { battleDroid: any; cloneTrooper: any; experience: any; shield: any };
     private gameChat: GameChat;
     private lobbyOwnerId: string;
+    public isPrivate: boolean;
+    private connectionLink: string | null;
```
AMMayberry1, line 35: nit: can instead do `private connectionLink?: string;`

AMMayberry1 (approved):

PR #417, opened 2025-01-05
Tomw/Tactical Droid Commander and Kashyyyk Defender

Twet1:
image
image


server/game/cards/03_TWI/units/KashyyykDefender.ts
```diff
@@ -0,0 +1,36 @@
...
+        this.addTriggeredAbility({
+            title: 'Heal up to 2 damage from another unit',
+            when: {
+                onCardPlayed: (event, context) => event.card.controller === context.source.controller
+            },
+            immediateEffect: AbilityHelper.immediateEffects.distributeHealingAmong({
+                amountToDistribute: 2,
+                cardCondition: (card, context) => card !== context.source,
+                controller: WildcardRelativePlayer.Any,
+                canDistributeLess: true,
```
AMMayberry1, line 23: This is the default value, can remove it

test/server/cards/03_TWI/units/KashyyykDefender.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+            ]));
+            expect(context.tacticalDroidCommander.damage).toBe(2);
+            expect(context.kashyyykDefender.damage).toBe(1);
+
+            // reset card
+            context.player2.moveCard(context.waylay, 'hand');
+            context.player2.clickCard(context.waylay);
+            context.player2.clickCard(context.kashyyykDefender);
+
+            // play card, heal 0 damage from own unit. deals 0 damage to Kashyyyk Defender
```
AMMayberry1, line 45: Can we add one more case where we heal a unit for 2 but it only has 1 damage on it, then confirm that only 1 damage is dealt to KD

server/game/cards/03_TWI/units/TacticalDroidCommander.ts
```diff
@@ -0,0 +1,32 @@
...
+            optional: true,
+            when: {
+                onCardPlayed: (event, context) =>
+                    event.card.isUnit() &&
+                    event.card.controller === context.source.controller &&
+                    event.card.hasSomeTrait(Trait.Separatist)
+            },
+            targetResolver: {
+                cardCondition: (card, context) => card.isUnit() && card.cost <= context.event.card.cost,
+                controller: RelativePlayer.Opponent,
```
AMMayberry1, line 25: The card ability isn't restricted to opponent units

test/server/cards/03_TWI/units/TacticalDroidCommander.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+
+            // play separatist unit, able to exhaust unit of same cost or less.
+            context.player1.clickCard(context.confederateTrifighter);
+            expect(context.player1).toBeAbleToSelectExactly(context.battlefieldMarine);
+            context.player1.clickCard(context.battlefieldMarine);
+            expect(context.battlefieldMarine.exhausted).toBeTrue();
+
+            context.player2.passAction();
+
+            // play non separatist, no trigger.
```
AMMayberry1, line 26: Couple more cases to check:
- Owner plays a separatist _event_ instead of a unit
- Opponent plays a separatist unit

test/server/cards/03_TWI/units/TacticalDroidCommander.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+                player2: {
+                    groundArena: ['battlefield-marine', 'plo-koon#kohtoyah']
+                }
+            });
+
+            const { context } = contextRef;
+
+            // play separatist unit, able to exhaust unit of same cost or less.
+            context.player1.clickCard(context.confederateTrifighter);
+            expect(context.player1).toBeAbleToSelectExactly(context.battlefieldMarine);
```
AMMayberry1, line 20: Two tweaks to this that would be helpful for validating some behaviors:
- Can we test with a card that costs the same as the played separatist card to confirm that "same cost" works
- Can we test with a deployed leader unit and also at least one token unit to confirm that its cost value is interpreted correctly (will probably need to play a more expensive separatist unit)

server/game/cards/03_TWI/units/KashyyykDefender.ts
Twet1: Sorry, which is the default value?

server/game/cards/03_TWI/units/TacticalDroidCommander.ts
```diff
@@ -21,8 +21,7 @@ export default class TacticalDroidCommander extends NonLeaderUnitCard {
                     event.card.hasSomeTrait(Trait.Separatist)
             },
             targetResolver: {
-                cardCondition: (card, context) => card.isUnit() && card.cost <= context.event.card.cost,
-                controller: RelativePlayer.Opponent,
+                cardCondition: (card, context) => card.isUnit() && card.cost <= context.event.card.cost && card !== context.event.card,
```
AMMayberry1, line 24: So this is a really minor technicality but there's actually nothing in the wording of the trigger that says you can't select the played unit to be exhausted. I really can't imagine a case where that would be relevant in the current card pool, but we get caught by surprise a lot with stuff like this so we try to be as faithful to rules-as-written as we possibly can

AMMayberry1 (approved):

PR #418, opened 2025-01-07
Corner the prey

ssoulier:
[Corner the prey](https://swudb.com/card/TWI/139/corner-the-prey)

Screenshot 2025-01-07 at 19 32 22

### DIFF
--- a/server/game/cards/03_TWI/events/CornerThePrey.ts
+++ b/server/game/cards/03_TWI/events/CornerThePrey.ts
@@ -14,15 +14,11 @@ export default class CornerThePrey extends EventCard {
         this.setEventAbility({
             title: 'Attack with a unit. It gets +1/+0 for this attack for each damage on the defender at the start of this attack.',
             initiateAttack: {
-                attackerLastingEffects: [
-                    {
-                        condition: (attack: Attack) => !attack.target.isBase(),
-                        effect: (attack: Attack) => {
-                            const damageOnDefender = attack.target.damage;
-                            return AbilityHelper.ongoingEffects.modifyStats({ power: damageOnDefender, hp: 0 });
-                        }
-                    }
-                ]
+                attackerLastingEffects: (_context, attack) =>
+                    ({
+                        condition: () => !attack.target.isBase(),
+                        effect: AbilityHelper.ongoingEffects.modifyStats({ power: attack.target.damage, hp: 0 })
+                    })
             }
         });
     }
--- a/server/game/cards/03_TWI/events/CornerThePrey.ts
+++ b/server/game/cards/03_TWI/events/CornerThePrey.ts
@@ -14,10 +14,10 @@ export default class CornerThePrey extends EventCard {
             title: 'Attack with a unit. It gets +1/+0 for this attack for each damage on the defender at the start of this attack.',
             initiateAttack: {
                 attackerLastingEffects: (_context, attack) =>
-                    ({
-                        condition: () => !attack.target.isBase(),
-                        effect: AbilityHelper.ongoingEffects.modifyStats({ power: attack.target.damage, hp: 0 })
-                    })
+                ({
+                    condition: () => !attack.target.isBase(),
+                    effect: AbilityHelper.ongoingEffects.modifyStats({ power: attack.target.damage, hp: 0 })
+                })
             }
         });
     }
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/events/CornerThePrey.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Attack with a unit. It gets +1/+0 for this attack for each damage on the defender at the start of this attack.',
+            initiateAttack: {
+                attackerLastingEffects: [
+                    {
+                        condition: (attack: Attack) => !attack.target.isBase(),
+                        effect: (attack: Attack) => {
```
ssoulier, line 20: @AMMayberry1 I am a bit stuck on this card.
This implementation does not work  because effect is `Any` so I do not have access to `attack`. It's probably trivial but with my very limited knowledge ot typescript, I do not see how I can access to defender (or target) damage before the attack starts.
Do you have a few pointers for me?

test/server/cards/03_TWI/events/CornerThePrey.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.cornerThePrey);
+                context.player1.clickCard(context.consularSecurityForce);
+                context.player1.clickCard(context.chewbacca);
+                expect(context.chewbacca.damage).toBe(7);
+                expect(context.consularSecurityForce.getPower()).toBe(3);
+                expect(context.consularSecurityForce.damage).toBe(6);
```
AMMayberry1, line 23: Could we add a couple more cases just to validate the functionality:

1. Since it says "at the _start_ of the attack", we need to make sure that damage which occurs after the attack starts isn't counted. We can do this by adding an upgrade that does damage on attack like Fallen Lightsaber or Vambrace Flamethrower. I think it will work as you currently have it, but if it doesn't please ping me since that may require some engine work.

2. Can we have a basic test where defending unit has no damage just to make sure nothing breaks

test/server/cards/03_TWI/events/CornerThePrey.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+            });
+
+            it('should attack with a unit getting +0/+0 while attacking a base', function() {
```
AMMayberry1, line 26: This test could be merged into the above one

test/server/cards/03_TWI/events/CornerThePrey.spec.ts
ssoulier: Good point. Everything works fine

test/server/cards/03_TWI/events/CornerThePrey.spec.ts
```diff
@@ -1,47 +1,69 @@
...
+                // Attacking a unit with no damage shold not increase the power of the Consular Security Force
+                // and it should not crash
+                const cornerThePreys = context.player1.findCardsByName('corner-the-prey');
+                context.player1.clickCard(cornerThePreys[0]);
                 context.player1.clickCard(context.consularSecurityForce);
+                context.player1.clickCard(context.chewbacca);
+                expect(context.chewbacca.damage).toBe(3);
+                context.player2.passAction();
+
+                // Attacking a base should increase the power of the Battlefield Marine
```
AMMayberry1, line 62: nit: should _not_ increase

test/server/cards/03_TWI/events/CornerThePrey.spec.ts
```diff
@@ -1,47 +1,69 @@
...
+                context.player1.clickCard(cornerThePreys[1]);
+                context.player1.clickCard(context.battlefieldMarine);
```
AMMayberry1, line 64: Apologies for the late request, just thought of this - could we also put some damage on the base to confirm it doesn't trigger the ability? Can do it as like `base: { card: 'echo-base', damage: 2 }`

AMMayberry1:
Looks like there are still lint errors, feel free to ping if it's still not working for you in vscode

test/server/cards/03_TWI/events/CornerThePrey.spec.ts
ssoulier: It's a good point. I should have thought about it.

AMMayberry1 (approved):

PR #421, opened 2025-01-08
[+] Card Implementation: Rush Clovis & Clear the Field Implementation

HugoHMZ:
image
image

### DIFF
--- a/server/game/cards/03_TWI/units/ClearTheField.ts
+++ b/server/game/cards/03_TWI/units/ClearTheField.ts
@@ -1,0 +1,31 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { WildcardCardType, WildcardRelativePlayer, WildcardZoneName } from '../../../core/Constants';
+
+export default class ClearTheField extends EventCard {
+    protected override getImplementationId () {
+        return {
+            id: '9620454519',
+            internalName: 'clear-the-field'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Choose a non-leader that costs 3 or less. Return it and each enemy non-leader unit with the same name as it to their owner\'s hand.',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                zoneFilter: WildcardZoneName.AnyArena,
+                controller: WildcardRelativePlayer.Any,
+                cardCondition: (card) => card.isUnit() && card.cost <= 3,
+                immediateEffect: AbilityHelper.immediateEffects.returnToHand((context) => {
+                    const allOpponentUnits = context.player.opponent.getUnitsInPlay();
+                    const returnedCards = allOpponentUnits.filter((card) => card.title === context.target.title);
+                    return { target: returnedCards };
+                }),
+            }
+        });
+    }
+}
+
+ClearTheField.implemented = true;
--- a/server/game/cards/03_TWI/units/RushClovis.ts
+++ b/server/game/cards/03_TWI/units/RushClovis.ts
@@ -1,0 +1,22 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+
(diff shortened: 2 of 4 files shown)


server/game/cards/03_TWI/units/ClearTheField.ts
```diff
@@ -0,0 +1,31 @@
...
+            id: '9620454519',
+            internalName: 'clear-the-field'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Choose a non-leader that costs 3 or less. Return it and each enemy non-leader unit with the same name as it to their owner\'s hand.',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
```
AMMayberry1, line 17: nit: we can remove this and just do `card.isNonLeaderUnit()` in the condition

server/game/cards/03_TWI/units/ClearTheField.ts
```diff
@@ -0,0 +1,31 @@
...
+                zoneFilter: WildcardZoneName.AnyArena,
+                controller: WildcardRelativePlayer.Any,
```
AMMayberry1, line 19: nit: these are the default values for zoneFilter and controller so we can remove them

test/server/cards/03_TWI/units/ClearTheField.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+            context.player1.clickCard(specforceSoldiersP2[0]);
+            expect(specforceSoldiersP2[0]).toBeInZone('hand', context.player2);
+            expect(specforceSoldiersP2[1]).toBeInZone('hand', context.player2);
+
+            // Check ally card that matches the target name doesn't go back to hand
+            expect(specforceSoldierP1).toBeInZone('groundArena', context.player1);
+
+            // Check card that doesn't match the selected card's name stays in play
+            expect(context.duchesssChampion).toBeInZone('groundArena');
+            expect(context.yoda).toBeInZone('groundArena');
```
AMMayberry1, line 31: Good test! Few more test cases it would be good to have:
- Test with two different versions of a unique unit to confirm that cards will match even if the subtitles are different
- Try with a leader whose name matches to confirm that it doesn't try to return it to hand and break the game
- Apply it to some token units to confirm they are all removed from the game

server/game/cards/03_TWI/units/RushClovis.ts
```diff
@@ -0,0 +1,22 @@
...
+    protected override getImplementationId () {
+        return {
+            id: '9964112400',
+            internalName: 'rush-clovis#banking-clan-scion'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'If the defending player controls no ready resources, create a Battle Droid token.',
```
AMMayberry1, line 14: nit: we don't include the "if" part in the title since this prompt will only appear if the condition is already met

server/game/cards/03_TWI/units/RushClovis.ts
```diff
@@ -0,0 +1,22 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.conditional({ condition: (context) => context.source.controller.opponent.resources.filter((resource) => !resource.exhausted).length === 0,
```
AMMayberry1, line 15: nit: can do `context.source.controller.opponent.readyResourceCount`

test/server/cards/03_TWI/units/RushClovis.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+                player2: {
+                    groundArena: ['rush-clovis#banking-clan-scion'],
+                    resources: 1
+                }
+            });
+            const { context } = contextRef;
+            const rushClovisP1 = context.player1.findCardByName('rush-clovis#banking-clan-scion');
+            const rushClovisP2 = context.player2.findCardByName('rush-clovis#banking-clan-scion');
+
+            // Check player 1 doesn't get a Battle Droid token since player 2 has ready resources.
```
AMMayberry1, line 19: Zero resources is kind of a weird edge case, could we instead test with a nonzero number of exhausted resources? Can use `context.player1.exhaustResources(5);` in the test to accomplish that

PR #422, opened 2025-01-08
[+] Card Implementation: Rush Clovis & Clear the Field Implementation…

HugoHMZ:
image
image


HugoHMZ:
@AMMayberry1 hey, quick question. About this comment "Test with two different versions of a unique unit to confirm that cards will match even if the subtitles are different
Try with a leader whose name matches to confirm that it doesn't try to return it to hand and break the game"

i tried to look up for some card names I could use to test it but i cannot find any (more especially since they need to cost under 4 energy). Do you have any names to give me so i can try more comfortably ?

test/server/cards/03_TWI/units/ClearTheField.spec.ts
```diff
@@ -91,5 +92,30 @@ describe('Clear the Field\'s ability', function () {
             expect(droidToken1).toBeInZone('outsideTheGame');
             expect(droidToken2).toBeInZone('outsideTheGame');
         });
+
+        it('should only make unit return to hand and tokens go out of combat instead', function () {
```
AMMayberry1, line 96: Test name needs to be updated

AMMayberry1 (approved):

PR #424, opened 2025-01-09
Add last played card to game state

danbastin:
Adds a function to add log actions for playing cards and adds the last card played from hand to gameState

### DIFF
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -73,6 +73,7 @@ class Game extends EventEmitter {
         this.stateWatcherRegistrar = new StateWatcherRegistrar(this);
         this.movedCards = [];
         this.randomGenerator = seedrandom();
+        this.lastPlayedCard = null;
 
         this.registerGlobalRulesListeners();
 
@@ -1351,6 +1352,7 @@ class Game extends EventEmitter {
                 players: playerState,
                 phase: this.currentPhase,
                 messages: this.gameChat.messages,
+                lastPlayedCard: this.lastPlayedCard?.cardData.setId,
                 spectators: this.getSpectators().map((spectator) => {
                     return {
                         id: spectator.id,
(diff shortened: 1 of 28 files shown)


server/game/core/ability/PlayCardAction.ts
```diff
@@ -111,6 +111,12 @@ export abstract class PlayCardAction extends PlayerAction {
 
     public abstract clone(overrideProperties: Partial<IPlayCardActionProperties>): PlayCardAction;
 
+    public logPlayCardEvent(context: any): void {
```
AMMayberry1, line 114: A slightly simpler way to do this (that also might catch some corner cases) would be to attach it to the GameEvent in `generateOnPlayEvent`. The `GameEvent` constructor has an option fourth parameter which is a handler, you could do this there.

For context, each play card action generates two events: (1) a specific event type that handles the logic for putting a unit or upgrade into play or playing an event, and (2) a generic "card played" event which is the same regardless of card type. By attaching this logic to the latter we can make it so we don't have to add a call to each play card action type.

The corner case that it could potentially help with that it's still theoretically possible that the event could fail to resolve and the card not end up being played even though `PlayCardAction` was triggered. So if we attach the log emitter to the play card event, it is guaranteed only to happen if the card was actually played.

server/game/core/Game.js
```diff
@@ -73,6 +73,7 @@ class Game extends EventEmitter {
         this.stateWatcherRegistrar = new StateWatcherRegistrar(this);
         this.movedCards = [];
         this.randomGenerator = seedrandom();
+        this.lastPlayedCard = null;
```
AMMayberry1, line 76: Could we instead group all of the "logging" properties together under one interface? I would strongly prefer that we don't start adding a lot of little properties to `Game` if it's avoidable, that makes it hard to track all of the state and make sure things are getting handled correctly.

I think ideally we would have a .ts file that's just the interface definition with all the different logging properties (added as we go along), and then here we can use jsdoc to indicate the type of the property so that the IDE and other TS code can enforce it for us.

AMMayberry1 (approved):

PR #425, opened 2025-01-09
Che bato/manual disconnect

CheBato:
### GameServer.ts
- Changed import "import socketio from 'socket.io' to import type { Socket as IOSocket } from 'socket.io';
import { Server as IOServer } from 'socket.io'; This lets us have more possibilities when extending base types for Socket and Server. 

- extended server and socket to type <ClientToServerEvents, ServerToClientEvents, InterServerEvents, SocketData> [reference document](https://socket.io/docs/v4/typescript/) 

- Added interface SocketData which will be used as additional data for a socket. Currently it only holds manualDisconnect but might be useful later on for more custom data.

- Changed type socketio.Socket throughout the file to IOSocket<any, any, any, SocketData> since we now have an extended Socket type.

- Added a manualDisconnect listener to the socket that sets the manualDisconnect parameter in the socket.data. This will tell us if a user wilfully wishes to leave a game

- Removed throw new Error because players can come to this state when they wish to reconnect to a game that they left but the game has removed them from its mapping (they took longer then 20s).

- rewrote onSocketDisconnected function to check if the players socket has the manualDisconnect property set to true if  it is set to true we remove the player from the userlobbymap and game. We cleanup the lobby if there are no users left.

- Added a new timeout for all other types of disconnect which is 20s.


server/gamenode/GameServer.ts
```diff
@@ -22,12 +23,19 @@ interface User {
...
+    manualDisconnect?: boolean;
+}
+
 /**
  * Represents a player waiting in the queue.
  */
 interface QueuedPlayer {
     deck: Deck;
-    socket?: socketio.Socket;
+    socket?: IOSocket<any, any, any, SocketData>;
```
AMMayberry1, line 38: You know I have to ask - is there a way we could avoid using `any`? :P

server/gamenode/GameServer.ts
```diff
@@ -37,7 +45,7 @@ export class GameServer {
     private protocol = 'https';
     private host = env.gameNodeHost;
     private queue: QueuedPlayer[] = [];
-    private io: socketio.Server;
+    private io: IOServer<any, any, any, SocketData>;
```
AMMayberry1, line 48: Since this is used in many places we could declare it as a new type with `type SomeTypeName = IOServer<any, any, any, SocketData>;`. Same for the `IOSocket` version

server/gamenode/GameServer.ts
```diff
@@ -403,38 +418,42 @@ export class GameServer {
...
         if (!this.userLobbyMap.has(id)) {
             this.removeFromQueue(id);
             return;
         }
         const lobbyId = this.userLobbyMap.get(id);
         const lobby = this.lobbies.get(lobbyId);
-        if (reason === 'client namespace disconnect') {
+
+
+        const wasManualDisconnect = socket?.data?.manualDisconnect === true;
```
AMMayberry1, line 430: nit: could simplify as `!!socket?.data?.manualDisconnect`

server/gamenode/GameServer.ts
```diff
@@ -403,38 +418,42 @@ export class GameServer {
...
+            if (lobby.isEmpty()) {
+                // cleanup process
+                lobby.cleanLobby();
+                this.lobbies.delete(lobbyId);
+            }
+            return;
         }
+
+        lobby.setUserDisconnected(id);
+        setTimeout(() => {
```
AMMayberry1, line 445: This is unrelated, but wondering if at some point we should have a timeout for lobbies where the player(s) are present but not taking any actions. Just to prevent from accidentally accumulating lobbies that have become "dead" somehow. No action for this PR, just a thought

server/gamenode/GameServer.ts
danbastin: I think this should actually change as the socket is not a rew IOSocket, but rather the Socket wrapper class. We could just call this of type `Socket` and then do all of our type definitions in that file.

server/gamenode/GameServer.ts
danbastin: We can skip the typing here honestly. This should not have the SocketData type since we're not setting this on the server. We start the socket server below and it's never referenced again.

server/gamenode/GameServer.ts
CheBato: Absolutely agree its something we should consider.

server/gamenode/GameServer.ts
CheBato: skipped typing for server.

server/gamenode/GameServer.ts
CheBato: simplified.

server/gamenode/GameServer.ts
CheBato: i added a todo so we remeber to do this at one point.

AMMayberry1 (approved):

PR #426, opened 2025-01-10
Implementing Sly Moore Secretive Advisor

supercodepoet:
image

### DIFF
--- a/test/server/cards/03_TWI/units/SlyMooreSecretiveAdvisor.spec.ts
+++ b/test/server/cards/03_TWI/units/SlyMooreSecretiveAdvisor.spec.ts
@@ -1,0 +1,52 @@
+describe('Sly Moore Secretive Advisor', function() {
+    integration(function(contextRef) {
+        describe('Sly Moore\'s ability', function() {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['sly-moore#secretive-advisor'],
+                        groundArena: ['battle-droid'],
+                        spaceArena: ['cartel-spacer']
+                    },
+                    player2: {
+                        groundArena: ['clone-trooper', 'advanced-recon-commando']
+                    }
+                });
+            });
+
+            it('should take control of an enemy token unit and ready it. At the start of the regroup phase, the owner should take control.', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.slyMoore);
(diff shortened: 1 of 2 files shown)


test/server/cards/03_TWI/units/SlyMooreSecretiveAdvisor.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+            it('should take control of an enemy token unit and ready it. At the start of the regroup phase, the owner should take control.', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.slyMoore);
+                expect(context.player1).toBeAbleToSelectExactly([context.cloneTrooper]);
+
+                context.player1.clickCard(context.cloneTrooper);
+                expect(context.cloneTrooper).toBeInZone('groundArena', context.player1);
+                expect(context.cloneTrooper.exhausted).toBeFalse();
```
AMMayberry1, line 26: We should start the clone trooper in an exhausted state to confirm that the ready effect happens - can just use `{ card: 'clone-trooper', exhausted: true }`

test/server/cards/03_TWI/units/SlyMooreSecretiveAdvisor.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                expect(context.player2).toBeActivePlayer();
+
+                context.moveToRegroupPhase();
+                expect(context.cloneTrooper).toBeInZone('groundArena', context.player2);
+            });
+
+            it('should take control of an enemy token unit and ready it. At the start of the regroup phase, the owner should not take control.', function () {
```
AMMayberry1, line 33: nit: can we add a little more context as to why the effect is different in this case - e.g., "should not take control if the unit is defeated"

test/server/cards/03_TWI/units/SlyMooreSecretiveAdvisor.spec.ts
supercodepoet: awesome, good call, updated

test/server/cards/03_TWI/units/SlyMooreSecretiveAdvisor.spec.ts
supercodepoet: changed

AMMayberry1 (approved):

PR #428, opened 2025-01-10
Grim resolve senatorial corvette

danbastin:
https://swudb.com/cards/TWI/172.png
https://swudb.com/cards/TWI/148.png


server/game/cards/03_TWI/units/SenatorialCorvette.ts
```diff
@@ -0,0 +1,23 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '6404471739',
+            internalName: 'senatorial-corvette',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenDefeatedAbility({
+            title: 'Each player discards a card from their hand',
```
AMMayberry1, line 14: nit: Each _opponent_ discards a card from their hand

Actually, we've been adjusting the text to be more literal when it refers to multiple opponents since we don't have Twin Suns right now - so in this case it might be better just to do "The opponent discards ..."

test/server/cards/03_TWI/events/GrimResolve.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+                    hand: ['grim-resolve', 'grim-resolve'],
+                    groundArena: [{ card: 'battlefield-marine', damage: 2 }, 'wampa']
+                }
+            });
+
+            const { context } = contextRef;
+
+            const wampa = context.player1.findCardByName('wampa');
+            const marine = context.player1.findCardByName('battlefield-marine');
+            const p2Base = context.player2.base;
```
AMMayberry1, line 16: nit: can we switch to pulling these from the `context` object like in the other test file (there is a `context.p2Base` built-in as well)

test/server/cards/03_TWI/events/GrimResolve.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+            let grimResolve = context.player1.hand[0];
+            context.player1.clickCard(grimResolve);
+            context.player1.clickCard(marine);
+            context.player1.clickCard(p2Base);
+            expect(p2Base.damage).toBe(5);
+
+            context.player2.passAction();
+
+            grimResolve = context.player1.hand[0];
```
AMMayberry1, line 26: We have a couple of established patterns for this situation that are a little more explicit. Either one works:

1. Use one copy of the card and then `context.player1.moveCard(context.grimResolve, 'hand')` to return it to hand
2. Find both copies of the card up front using `const [grimResolve1, grimResolve2] = player1.findCardsByName('grim-resolve');`

test/server/cards/03_TWI/units/SenatorialCorvette.spec.ts
```diff
@@ -0,0 +1,29 @@
+describe('Senatorial Corvette', function() {
+    integration(function(contextRef) {
+        describe('Senatorial Corvette\'s When Defeated ability', function() {
+            beforeEach(function () {
```
AMMayberry1, line 4: nit: since there's just one case, we can streamline this test a little bit by rolling everything up into one case like in the other test file

test/server/cards/03_TWI/units/SenatorialCorvette.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        spaceArena: ['senatorial-corvette'],
+                    },
+                    player2: {
+                        hand: ['wampa'],
+                        spaceArena: ['ruthless-raider'],
+                    },
+                    autoSingleTarget: true
```
AMMayberry1, line 14: We're disabling autoSingleTarget by default for new tests, this option exists just so we didn't have to rewrite all the legacy tests

test/server/cards/03_TWI/units/SenatorialCorvette.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+            });
+
+            it('should discard a card from opponents hand', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.senatorialCorvette);
+                context.player1.clickCard(context.ruthlessRaider);
+
+                expect(context.player2.handSize).toBe(0);
+                expect(context.wampa).toBeInZone('discard');
```
AMMayberry1, line 25: Can we also put a card in the controller's hand and confirm it isn't discarded

AMMayberry1 (approved):

AMMayberry1 (approved):

PR #431, opened 2025-01-11
add chancellor palpatine wartime chancellor

ssoulier:
[Chancellor Palpatine](https://swudb.com/card/twi/203/chancellor-palpatine)

Screenshot 2025-01-11 at 08 45 19


server/game/cards/03_TWI/units/ChancellorPalpatineWartimeChancellor.ts
```diff
@@ -0,0 +1,45 @@
...
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'Each token unit you create enters play ready.',
+            targetController: RelativePlayer.Self,
+            matchTarget: (card) => card.isTokenUnit(),
+            ongoingEffect: OngoingEffectBuilder.card.static(EffectName.EntersPlayReady)
+        });
+
+        this.addOnAttackAbility({
+            title: 'If a unit left play this phase, create a Clone Trooper token.',
```
AMMayberry1, line 31: nit: we don't include the "if" part of the title

server/game/cards/03_TWI/units/ChancellorPalpatineWartimeChancellor.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar): void {
+        this.cardsLeftPlayThisPhaseWatcher = AbilityHelper.stateWatchers.cardsLeftPlayThisPhase(registrar, this);
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'Each token unit you create enters play ready.',
+            targetController: RelativePlayer.Self,
+            matchTarget: (card) => card.isTokenUnit(),
```
AMMayberry1, line 26: nit: can do `targetCardTypeFilter: CardType.TokenUnit`

server/game/cards/03_TWI/units/ChancellorPalpatineWartimeChancellor.ts
```diff
@@ -0,0 +1,45 @@
...
+            ongoingEffect: OngoingEffectBuilder.card.static(EffectName.EntersPlayReady)
+        });
+
+        this.addOnAttackAbility({
+            title: 'If a unit left play this phase, create a Clone Trooper token.',
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => {
+                    const controllerHasUnitsThatLeftPlayThisPhase = this.cardsLeftPlayThisPhaseWatcher.someCardLeftPlayControlledByPlayer({ controller: context.source.controller, filter: (entry) => entry.card.isUnit() });
```
AMMayberry1, line 34: This is a little convoluted since the watcher could just check both players at the same time. Would you mind adding a `someCardLeftPlay` method to `CardsLeftPlayThisPhaseWatcher`?

test/server/cards/03_TWI/units/ChancellorPalpatineWartimeChancellor.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+            cloneTroopers = context.player1.findCardsByName('clone-trooper');
+            expect(cloneTroopers.length).toBe(4);
+
+            // Opponent token units should not enter play ready
+            context.player2.clickCard(context.droidDeployment);
+            const opponentBattleDroids = context.player2.findCardsByName('battle-droid');
+            expect(opponentBattleDroids.length).toBe(2);
+            expect(opponentBattleDroids.every((battleDroid) => battleDroid.exhausted)).toBeTrue();
+
+            // Even battle droids played by chancellor palpatine controller should enter play ready
```
AMMayberry1, line 70: Can we add one more test that non-token units played by the controller still enter play exhausted

server/game/cards/03_TWI/units/ChancellorPalpatineWartimeChancellor.ts
ssoulier: So I understand correctly, the "if" part is implicit with the condition below?

AMMayberry1: Correct, this prompt will only appear if the "if" condition is already met

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -37,6 +37,25 @@ export class CardsLeftPlayThisPhaseWatcher extends StateWatcher<CardLeftPlayEntr
             .map((entry) => entry.card);
     }
 
+    public getCardsLeftPlayByAllPlayers({ filter }: {
```
AMMayberry1, line 40: What's the benefit of having this in addition to `getCardsLeftPlay`?

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -37,6 +37,25 @@ export class CardsLeftPlayThisPhaseWatcher extends StateWatcher<CardLeftPlayEntr
...
+        filter?: (event: CardLeftPlayEntry) => boolean;
+    }) {
+        return this.getCardsLeftPlay({ filter });
+    }
+
+    public someCardLeftPlayByAnyPlayer({ filter }: {
```
AMMayberry1, line 46: The way you currently have the argument set up it would have to be called like this, requiring an object with the `filter` property:

```ts
const myFilterFn = () => true;
someCardLeftPlayByAnyPlayer({ filter: myFilterFn })
```

Since there's only one parameter it seems simpler to allow it to be used without a wrapper object:

```ts
someCardLeftPlayByAnyPlayer(myFilterFn)
```

Which would be done like this:

```ts
public someCardLeftPlayByAnyPlayer(filter?: (event: CardLeftPlayEntry) => boolean) {
```

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -21,11 +21,11 @@ export class CardsLeftPlayThisPhaseWatcher extends StateWatcher<CardLeftPlayEntr
         return super.getCurrentValue();
     }
 
-    public getCardsLeftPlayControlledByPlayer({ controller, filter }: {
-        controller: Player;
+    protected getCardsLeftPlay({ controller, filter }: {
```
AMMayberry1, line 24: Since an object is being used as the parameter here, it doesn't provide much benefit to refactor the method like this and have other methods that overload the `controller` parameter. For example, with `getCardsLeftPlay` as you have it now, it can be called with or without controller equally easily:

```ts
const myFilterFn = () => true;
getCardsLeftPlay({ controller: player1, filter: myFilterFn })
getCardsLeftPlay({ filter: myFilterFn })
```

So creating new signatures for having `controller` or not having it doesn't seem to add much value. It would be different if we had to explicitly say `controller = null` in some cases but since we're using an object parameter we can just omit it completely.

To summarize, I would recommend we remove `getCardsLeftPlayByAllPlayers` and `getCardsLeftPlayControlledByPlayer` and just use this method (that will require updating the name in a few card implementations).

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
AMMayberry1, line 46: nit: the naming convention for watcher methods assumes that by default we are getting events from all players unless stated otherwise (e.g. `CardsEnteredPlayThisPhaseWatcher.getCardsPlayed()`), so we can remove the `ByAnyPlayer` part of the name to keep things consistent

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
ssoulier: Thanks a lot for taking the time to explain it to me.
I went for 2 methods:
 - getCardsLeftPlay
 - someCardsLeftPlay
 
 and I refactored BobaFett leader set 1 that was the only one using these methods.

AMMayberry1 (approved):

PR #432, opened 2025-01-11
add squad support

ssoulier:
[Squad Support](https://swudb.com/card/twi/122/squad-support)

Screenshot 2025-01-11 at 12 19 50

### DIFF
--- a/test/server/cards/03_TWI/upgrades/SquadSupport.spec.ts
+++ b/test/server/cards/03_TWI/upgrades/SquadSupport.spec.ts
@@ -1,46 +1,105 @@
+
 describe('Squad Support', function() {
     integration(function(contextRef) {
-        it('Squad Support\'s ability should give +1/+1 to the attached unit for each Trooper you control', function () {
-            contextRef.setupTest({
-                phase: 'action',
-                player1: {
-                    hand: ['squad-support'],
-                    groundArena: ['battlefield-marine', 'advanced-recon-commando', 'regional-governor'],
-                    spaceArena: ['alliance-xwing'],
-                    leader: { card: 'finn#this-is-a-rescue', deployed: true }
-                },
-                player2: {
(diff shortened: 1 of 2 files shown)


test/server/cards/03_TWI/upgrades/SquadSupport.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+                    'alliance-xwing',
+                    context.deathStarStormtrooper,
+                    context.tieAdvanced
+                ]
+            ); // Only leader Finn can't be selected
+            context.player1.clickCard(context.advancedReconCommando);
+            expect(context.advancedReconCommando.getPower()).toBe(7);
+            expect(context.advancedReconCommando.getHp()).toBe(6);
+
+            // The unit gains +1/+1 for each Trooper unit you control not the upgrade
```
AMMayberry1, line 35: Good idea! Love this

test/server/cards/03_TWI/upgrades/SquadSupport.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+            expect(context.squadSupport.getPower()).toBe(0);
+            expect(context.squadSupport.getHp()).toBe(0);
+
+            // When a Trooper unit is defeated attached unit should lose +1/+1
+            context.player2.clickCard(context.deathStarStormtrooper);
+            context.player2.clickCard(context.battlefieldMarine);
+            expect(context.advancedReconCommando.getPower()).toBe(6);
+            expect(context.advancedReconCommando.getHp()).toBe(5);
+        });
```
AMMayberry1, line 44: Since we haven't had an upgrade like this before it would be good to add a couple more baseline tests:
- Confirm that it gives 0 bonus when no troopers
- Damage the unit and then defeat a trooper so that its hp goes down and it is defeated

AMMayberry1 (approved):

PR #435, opened 2025-01-14
Add Heroes on Both Sides

Frugghi:
image


server/game/cards/03_TWI/events/HeroesOnBothSides.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Give each chosen unit +2/+2 and Saboteur for this phase',
+            targetResolvers: {
+                republicUnit: {
+                    activePromptTitle: 'Choose a Republic unit',
+                    mode: TargetMode.Exactly,
+                    numCards: 1,
```
AMMayberry1, line 20: iirc, numCards: 1 and TargetMode.Exactly are the defaults so I think we can omit them from the target resolvers. I think `AnyArena` is also unnecessary since type is already restricted to unit, the default is `AnyAttackable` which includes base zone but the unit filter will block that out

Frugghi: The default target mode is `Single` but it doesn't allow to choose no cards so that's why I used `Exactly` and 1. Technically the card says "up to 1" but using `UpTo` forces you to click done after selecting the card so it felt unnecessary.

AMMayberry1 (approved):

PR #438, opened 2025-01-15
Che bato/end game screen

CheBato:
### Game.js
- Changed how we send the winner with sendGameState this is temporary since we wait to see how we will display the endGame popup.

### Lobby.ts
- Added regularRematch function to reset a custom/private game.
- Added this.setTokens(); and this.setPlayableCardTitles(); so I could easily test restarting a game via the testStart method.
- Added two private methods requestRematch and regularRematch which give us the ability to track which user requested the rematch and also to start a rematch game/lobby
- added socket eventListener removal to the addLobbyUser function since when users would go to a new quickMatch the eventListeners would otherwise persist from the previous lobby.

### GameServer.ts
- Added requeueUser which is to send the user back into the queue for a new opponent. It checks if the user who requeued is the last in the lobby and cleans up the lobby if he is.

### socket.js
- Added two methods removeEventsListeners that accepts an array of events from which we wish to remove the listeners and eventContainsListener that check whether the supplied event contains the listener.


AMMayberry1 (dismissed):

server/gamenode/GameServer.ts
```diff
@@ -418,6 +419,33 @@ export class GameServer {
...
+     * requeues the user and removes him from the previous lobby. If the lobby is empty, it cleans it up.
+     */
+    private requeueUser(socket: Socket, user: User, deck: any): void {
+        if (this.userLobbyMap.has(user.id)) {
+            const lobbyId = this.userLobbyMap.get(user.id);
+            const lobby = this.lobbies.get(lobbyId);
+            this.userLobbyMap.delete(user.id);
+            lobby.removeUser(user.id);
+            // check if lobby is empty
+            if (lobby.isEmpty()) {
```
AMMayberry1, line 432: Should we set some kind of timer for the lobby to be cleaned up if the game has already started and one of the users has left? Not sure if there are other mechanisms that cover that case already

server/gamenode/Lobby.ts
```diff
@@ -133,6 +150,37 @@ export class Lobby {
         this.sendLobbyState();
     }
 
+    private requestRematch(socket: Socket, ...args: any[]): void {
+        // Expect the rematch mode to be passed as the first argument: 'reset' or 'regular'
+        Contract.assertTrue(args.length === 1, 'Expected rematch mode argument');
```
AMMayberry1, line 155: nit: would be useful to include the arguments directly in the error message for debugging purposes if this fails. Same note for the assertion on 157

server/gamenode/Lobby.ts
```diff
@@ -133,6 +150,37 @@ export class Lobby {
...
+            this.rematchRequest = {
+                initiator: socket.user.id,
+                mode,
+            };
+            logger.info(`User ${socket.user.id} requested a rematch (${mode}) in lobby ${this._id}`);
+        }
+        this.sendLobbyState();
+    }
+
+    private regularRematch() {
```
AMMayberry1, line 170: This naming is a little confusing - what's the difference between this and "non-regular" rematch?

server/gamenode/GameServer.ts
CheBato: i've been thinking about this kind of scenarios and basically we probably want to keep the lobby until the last user leaves. What I thought about as a scenario was when somehow there are no players left (the last players disconnect socket event didn't trigger) how we should handle if something like that happens. In every other situation I think the onDisconnect socket handles it pretty cleanly.

server/gamenode/Lobby.ts
CheBato: fair point i'll rename it to something else :)

AMMayberry1 (approved):

PR #439, opened 2025-01-15
Implement Aayla Secura, Master of the Blade

moyerr:
-----

### Notes on implementation

I had hoped to land on a solution that would work for other similar damage prevention abilities (like [Boba Fett's Armor](https://swudb.com/card/shd/224/boba-fetts-armor) and [Finn, On the Run](https://swudb.com/card/twi/053/finn)), but ultimately the only `replacementImmediateEffect` that I could get to work was `null`. Interestingly, neither of the following strategies worked for me, but maybe there is someone who knows more about these that could help me understand why:

```typescript
// Did not work
AbilityHelper.immediateEffects.noAction()

// Did not work
AbilityHelper.immediateEffects.combatDamage(
    (damageContext) => ({
        amount: 0,
        sourceAttack: damageContext.event.damageSource.attack
    })
)
```

### DIFF
--- a/server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
+++ b/server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
@@ -1,0 +1,39 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { AbilityType, DamageType } from '../../../core/Constants';
+
+export default class AaylaSecuraMasterOfTheBlade extends NonLeaderUnitCard {
+    protected override getImplementationId () {
+        return {
+            id: '6190335038',
+            internalName: 'aayla-secura#master-of-the-blade',
+        };
+    }
+    
+    public override setupCardAbilities () {
+        this.addCoordinateAbility({
+            type: AbilityType.Triggered,
+            title: "Prevent all combat damage that would be dealt to this unit for this attack.",
+            when: {
+                onAttackDeclared: (event, context) => event.attack.attacker === context.source
+            },
+            immediateEffect: AbilityHelper.immediateEffects.forThisAttackCardEffect((context) => ({
+                target: context.source,
+                effect: AbilityHelper.ongoingEffects.gainAbility({
+                    title: "Prevent all combat damage that would be dealt to this unit",
+                    type: AbilityType.ReplacementEffect,
+                    when: {
+                        onDamageDealt: (event, context) => 
+                            event.card === context.source && event.type === DamageType.Combat
+                    },
+                    replaceWith: {
+                        target: context.source,
+                        replacementImmediateEffect: null
+                    }
+                }),
+            })),
+        });
+    }
+}
+
+AaylaSecuraMasterOfTheBlade.implemented = true;
--- a/server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
+++ b/server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
@@ -26,10 +26,7 @@ export default class AaylaSecuraMasterOfTheBlade extends NonLeaderUnitCard {
                         onDamageDealt: (event, context) =>
                             event.card === context.source && event.type === DamageType.Combat
                     },
-                    replaceWith: {
-                        target: context.source,
-                        replacementImmediateEffect: null
-                    }
+                    replaceWith: { /* replace with nothing */ }
                 }),
             })),
         });
--- a/server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
+++ b/server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
@@ -9,21 +9,21 @@ export default class AaylaSecuraMasterOfTheBlade extends NonLeaderUnitCard {
             internalName: 'aayla-secura#master-of-the-blade',
         };
     }
-    
+
     public override setupCardAbilities () {
         this.addCoordinateAbility({
             type: AbilityType.Triggered,
-            title: "Prevent all combat damage that would be dealt to this unit for this attack.",
+            title: 'Prevent all combat damage that would be dealt to this unit for this attack.',
             when: {
                 onAttackDeclared: (event, context) => event.attack.attacker === context.source
             },
             immediateEffect: AbilityHelper.immediateEffects.forThisAttackCardEffect((context) => ({
                 target: context.source,
                 effect: AbilityHelper.ongoingEffects.gainAbility({
-                    title: "Prevent all combat damage that would be dealt to this unit",
+                    title: 'Prevent all combat damage that would be dealt to this unit',
                     type: AbilityType.ReplacementEffect,
                     when: {
-                        onDamageDealt: (event, context) => 
+                        onDamageDealt: (event, context) =>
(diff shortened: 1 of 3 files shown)


server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
```diff
@@ -0,0 +1,39 @@
...
+                effect: AbilityHelper.ongoingEffects.gainAbility({
+                    title: 'Prevent all combat damage that would be dealt to this unit',
+                    type: AbilityType.ReplacementEffect,
+                    when: {
+                        onDamageDealt: (event, context) =>
+                            event.card === context.source && event.type === DamageType.Combat
+                    },
+                    replaceWith: {
+                        target: context.source,
+                        replacementImmediateEffect: null
```
AMMayberry1, line 31: nit: we can just remove this outright instead of explicit null

server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
```diff
@@ -0,0 +1,39 @@
...
+                target: context.source,
+                effect: AbilityHelper.ongoingEffects.gainAbility({
+                    title: 'Prevent all combat damage that would be dealt to this unit',
+                    type: AbilityType.ReplacementEffect,
+                    when: {
+                        onDamageDealt: (event, context) =>
+                            event.card === context.source && event.type === DamageType.Combat
+                    },
+                    replaceWith: {
+                        target: context.source,
```
AMMayberry1, line 30: Could you try removing this and see if it still works? This is intended to be the target to provide to `replacementImmediateEffect` (kind of a weird API tbh) and since we don't have one this shouldn't be needed.

At that point I guess we can also just try removing `replaceWith` completely

test/server/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.aaylaSecura);
+                context.player1.clickCard(context.consularSecurityForce);
+
+                expect(context.aaylaSecura.damage).toBe(0);
+                expect(context.consularSecurityForce.damage).toBe(6);
+            });
+
+            it('should not prevent combat damage to Aayla when Coordinate condition is not met', function() {
```
AMMayberry1, line 28: So we have this unfortunate issue where the Jasmine tests are pretty slow and heavy on a per-case basis, and we have finite github actions time in the free tier. Our eventual goal is to port to a better test framework but for now we are breaking good test practice a little bit by consolidating tests into fewer bigger cases where it's reasonable.

In this case you could refer to the Maul1 tests, that shows the typical pattern we use for structuring combined cases like that. On-attack triggers are a really easy one to do this with because we can un-exhaust the unit in between attacks so we're doing minimally invasive changes with the test framework and still mostly relying on game effects to do the work.

test/server/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+                        ],
+                    },
+                    player2: {
+                        groundArena: ['consular-security-force']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickPrompt('Pass');
```
AMMayberry1, line 65: nit: can instead do `hasInitiative: true` under `player2` in the setup

test/server/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+                context.player2.clickCard(context.consularSecurityForce);
+                context.player2.clickCard(context.aaylaSecura);
+
+                expect(context.aaylaSecura.damage).toBe(3);
+                expect(context.consularSecurityForce.damage).toBe(6);
+            });
```
AMMayberry1, line 72: Could we add some additional test cases:
- Confirm that her ability only blocks combat damage by using the same setup as with Tarfful in the Maul1 tests (case 3, not 4)
- Check that if she has a shield the player is prompted to choose whether to trigger the shield or trigger her ability, and then the not-chosen option doesn't resolve (so would need to be tested both ways like in the Maul1 test)

test/server/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+                        ],
+                    },
+                    player2: {
+                        groundArena: ['consular-security-force']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.aaylaSecura);
```
AMMayberry1, line 21: Just thought of this - just for stress-testing the coordinate stuff, would you mind starting her in hand for this one and using her ambush ability for the attack? That will let us validate that the timing works right for her turning on her own coordinate when she comes onto the field

server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
moyerr: The `replaceWith` property is not optional, but it does work if I replace with an empty object 👍

test/server/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.spec.ts
moyerr: Ended up covering 7 test cases across two test functions. Lemme know if this is enough! The setup for the shield situation was different enough that it felt like it needed it's own clean slate, similar to the Maul1 implementation.

test/server/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.spec.ts
moyerr: Done!

test/server/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.spec.ts
moyerr: Done!

server/game/cards/03_TWI/units/AaylaSecuraMasterOfTheBlade.ts
```diff
@@ -0,0 +1,36 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.forThisAttackCardEffect((context) => ({
+                target: context.source,
+                effect: AbilityHelper.ongoingEffects.gainAbility({
+                    title: 'Prevent all combat damage that would be dealt to this unit',
+                    type: AbilityType.ReplacementEffect,
+                    when: {
+                        onDamageDealt: (event, context) =>
+                            event.card === context.source && event.type === DamageType.Combat
+                    },
+                    replaceWith: { /* replace with nothing */ }
```
AMMayberry1, line 29: Do you mind changing the interface to make this property optional? Might as well make it fully optional if it's intended to provide an empty object in some cases

AMMayberry1:
Could you do me a small favor and avoid doing force-pushes in the future once reviews have started? It makes it so github can't show the changes between review iterations

AMMayberry1 (approved):

PR #440, opened 2025-01-15
Poe Dameron

ssoulier:
[Poe Dameron](https://swudb.com/card/SHD/153/poe-dameron)

Screenshot 2025-01-15 at 20 58 46

### DIFF
--- a/test/server/cards/01_SOR/units/PoeDameronQuickToImprovise.spec.ts
+++ b/test/server/cards/01_SOR/units/PoeDameronQuickToImprovise.spec.ts
@@ -1,54 +1,107 @@
-
 describe('Poe Demaron Quick to improvise\'s ability', function () {
     integration(function (contextRef) {
-        it('should allow to discard card and to resolve effects', function () {
-            contextRef.setupTest({
-                phase: 'action',
-                player1: {
-                    hand: ['sneak-attack', 'battlefield-marine', 'daring-raid'],
-                    groundArena: ['poe-dameron#quick-to-improvise'],
-                },
-                player2: {
-                    hand: ['open-fire', 'karabast'],
-                    groundArena: [{ card: 'death-star-stormtrooper', upgrades: ['academy-training'] }],
-                    spaceArena: ['tie-advanced']
-                }
+        describe('Poe Demaron Quick to improvise\'s ability', function () {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['sneak-attack', 'battlefield-marine', 'daring-raid', 'specforce-soldier'],
+                        groundArena: ['poe-dameron#quick-to-improvise'],
+                    },
+                    player2: {
+                        hand: ['open-fire', 'karabast'],
+                        groundArena: [{ card: 'death-star-stormtrooper', upgrades: ['academy-training'] }],
+                        spaceArena: ['tie-advanced']
+                    }
+                });
             });
-            const { context } = contextRef;
 
-            context.player1.clickCard(context.poeDameron);
-            context.player1.clickCard(context.p2Base);
-            expect(context.player1).toBeAbleToSelectExactly(['sneak-attack', 'battlefield-marine', 'daring-raid']);
-            context.player1.clickCard(context.sneakAttack);
-            context.player1.clickCard(context.battlefieldMarine);
-            context.player1.clickCard(context.daringRaid);
-            context.player1.clickPrompt('Done');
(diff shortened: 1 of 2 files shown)


test/server/cards/01_SOR/units/PoeDameronQuickToImprovise.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+                // We discard 3 cards
+                expect(context.player1).toBeAbleToSelectExactly([
+                    'sneak-attack',
+                    'battlefield-marine',
+                    'daring-raid',
+                    'specforce-soldier',
+                ]);
+                context.player1.clickCard(context.sneakAttack);
+                context.player1.clickCard(context.battlefieldMarine);
+                context.player1.clickCard(context.daringRaid);
```
AMMayberry1, line 34: We need to test that the max is 3 - can you add a click on the fourth card using `clickCardNonChecking()` and then confirm it stays in the hand

test/server/cards/01_SOR/units/PoeDameronQuickToImprovise.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+                context.player1.clickPrompt('Done');
```
AMMayberry1, line 35: We need to check that the selected cards are discarded

test/server/cards/01_SOR/units/PoeDameronQuickToImprovise.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+                context.player1.clickCard(context.poeDameron);
+                context.player1.clickCard(context.p2Base);
+
+                // We discard 3 cards
+                expect(context.player1).toBeAbleToSelectExactly([
+                    'sneak-attack',
+                    'battlefield-marine',
+                    'daring-raid',
+                    'specforce-soldier',
+                ]);
```
AMMayberry1, line 31: Can we confirm that there is a button available for choosing to not discard any cards - should be `toHaveNoTargetsButton()` or `toHavePassAbilityButton()` depending on which one it is.

Also, I think we don't usually need this but since this ability setup is kind of unique, let's go ahead and add a test case where the player chooses to select no cards and confirm that nothing breaks with the `ifYouDoCondition` (can be part of this same test case)

test/server/cards/01_SOR/units/PoeDameronQuickToImprovise.spec.ts
```diff
@@ -0,0 +1,107 @@
+describe('Poe Demaron Quick to improvise\'s ability', function () {
+    integration(function (contextRef) {
+        describe('Poe Demaron Quick to improvise\'s ability', function () {
+            beforeEach(function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['sneak-attack', 'battlefield-marine', 'daring-raid', 'specforce-soldier'],
```
AMMayberry1, line 8: Can we add a friendly upgrade as well just to confirm it's targetable

test/server/cards/01_SOR/units/PoeDameronQuickToImprovise.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+                expect(context.player1).toHaveEnabledPromptButtons([
+                    'An opponent discards a card from their hand.',
+                ]);
+                context.player1.clickPrompt('An opponent discards a card from their hand.');
+                context.player2.clickCard(context.karabast);
+                expect(context.karabast).toBeInZone('discard');
+            });
+
+            it('should work even if there is not upgrade to defeat or no card discarded', function () {
```
AMMayberry1, line 64: If we can easily extend this case as part of the previous one (I think we can but might be missing something) then let's go ahead and do that

ssoulier: The only pb is:
 - you need to go to next-phase to trigger another Poe attack's
 - but when you go to next action phase you draw 2 cards
 - These cards become discardable but I do not know what are these cards
 - So having another test is less "random" (assuming the cards drawn are not deterministic)

ssoulier: I decided to not check again what are the card we can select and in that case we can regroup everything in one test.

AMMayberry1: Ohh yeah good catch. Either way is good, good call

AMMayberry1 (approved):

PR #449, opened 2025-01-19
Implementing Kuiil, I Have Spoken

supercodepoet:
image

### DIFF
--- a/server/game/cards/02_SHD/units/KuiilIHaveSpoken.ts
+++ b/server/game/cards/02_SHD/units/KuiilIHaveSpoken.ts
@@ -1,0 +1,39 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+
+export default class KuiilIHaveSpoken extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '5511838014',
+            internalName: 'kuiil#i-have-spoken',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'Discard a card from your deck.',
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.discardFromDeck((context) => ({
+                    amount: 1,
+                    target: context.source.controller
+                })),
+                AbilityHelper.immediateEffects.conditional((context) => ({
+                    condition: this.hasMatchingAspects(context),
+                    onTrue: AbilityHelper.immediateEffects.returnToHand({
+                        target: context.events[0].card
+                    }),
+                    onFalse: AbilityHelper.immediateEffects.noAction()
+                }))
+            ])
+        });
+    }
+
+    private hasMatchingAspects(context) {
+        const baseAspects = context.source.controller.base.aspects;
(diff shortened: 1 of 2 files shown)


test/server/cards/02_SHD/units/KuiilIHaveSpoken.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+                // Following prompt is to use Kuiil's Restore 1 ability
+                context.player1.clickCard(context.player1.base);
+
+                expect(context.greenSquadronAwing).toBeInZone('hand');
+            });
+
+            it('should dicarded card that does not share an aspect with their base and not draw the card', function () {
+                const { context } = contextRef;
+
+                context.player1.setDeck([context.restoredArc170, context.greenSquadronAwing]);
```
AMMayberry1, line 35: Would be simpler to just extend this test off of the previous one and have Kuiil do two attacks in a row, also reduces test case count

server/game/cards/02_SHD/units/KuiilIHaveSpoken.ts
```diff
@@ -0,0 +1,39 @@
...
+            title: 'Discard a card from your deck.',
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.discardFromDeck((context) => ({
+                    amount: 1,
+                    target: context.source.controller
+                })),
+                AbilityHelper.immediateEffects.conditional((context) => ({
+                    condition: this.hasMatchingAspects(context),
+                    onTrue: AbilityHelper.immediateEffects.returnToHand({
+                        target: context.events[0].card
```
AMMayberry1, line 23: For stuff that involves acting on a discarded card it's a little safer to use `ifYouDo` since that won't trigger if no card is discarded. See Greedo implementation for an example, and let's go ahead and add a test where the deck is empty just to be safe

test/server/cards/02_SHD/units/KuiilIHaveSpoken.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+            it('should dicarded card that shares aspect with their base and draw the card', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.kuiil);
+                context.player1.clickCard(context.player2.base);
+
+                expect(context.player1).toHaveEnabledPromptButtons(['Discard a card from your deck.', 'Restore 1']);
+
+                context.player1.clickPrompt('Discard a card from your deck.');
+                // Following prompt is to use Kuiil's Restore 1 ability
```
AMMayberry1, line 26: I'm a little confused, can you ping me in chat to explain what's going on with this? The restore shouldn't require any clicks, something might be broken

test/server/cards/02_SHD/units/KuiilIHaveSpoken.spec.ts
supercodepoet: ok, combined all actions into one test. Sorry, my habit is always breaking out the tests into separate contextual tests.

server/game/cards/02_SHD/units/KuiilIHaveSpoken.ts
supercodepoet: updated

PR #451, opened 2025-01-21
Implement Roger Roger (TWI #69)

moyerr:

### DIFF
--- a/server/game/cards/03_TWI/upgrades/RogerRoger.ts
+++ b/server/game/cards/03_TWI/upgrades/RogerRoger.ts
@@ -1,0 +1,28 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { UpgradeCard } from '../../../core/card/UpgradeCard';
+import { RelativePlayer } from '../../../core/Constants';
+
+export default class RogerRoger extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '1555775184',
+            internalName: 'roger-roger',
+        };
+    }
+
+    protected override setupCardAbilities() {
+        this.addWhenDefeatedAbility({
+            title: 'Attach to a friendly Battle Droid token',
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardCondition: (card) => card.isTokenUnit() && card.title === 'Battle Droid',
+                immediateEffect: AbilityHelper.immediateEffects.attachUpgrade((context) => ({
+                    upgrade: this,
+                    target: context.target,
+                }))
+            }
+        });
+    }
+}
+
+RogerRoger.implemented = true;
--- a/server/game/cards/03_TWI/upgrades/RogerRoger.ts
+++ b/server/game/cards/03_TWI/upgrades/RogerRoger.ts
@@ -1,6 +1,6 @@
 import AbilityHelper from '../../../AbilityHelper';
 import { UpgradeCard } from '../../../core/card/UpgradeCard';
-import { RelativePlayer } from '../../../core/Constants';
+import { CardType, RelativePlayer } from '../../../core/Constants';
 
 export default class RogerRoger extends UpgradeCard {
     protected override getImplementationId() {
@@ -15,10 +15,10 @@ export default class RogerRoger extends UpgradeCard {
             title: 'Attach to a friendly Battle Droid token',
             targetResolver: {
                 controller: RelativePlayer.Self,
-                cardCondition: (card) => card.isTokenUnit() && card.title === 'Battle Droid',
+                cardTypeFilter: CardType.TokenUnit,
+                cardCondition: (card) => card.title === 'Battle Droid',
                 immediateEffect: AbilityHelper.immediateEffects.attachUpgrade((context) => ({
-                    upgrade: this,
-                    target: context.target,
+                    upgrade: context.source
                 }))
             }
         });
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/upgrades/RogerRoger.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    protected override setupCardAbilities() {
+        this.addWhenDefeatedAbility({
+            title: 'Attach to a friendly Battle Droid token',
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardCondition: (card) => card.isTokenUnit() && card.title === 'Battle Droid',
+                immediateEffect: AbilityHelper.immediateEffects.attachUpgrade((context) => ({
+                    upgrade: this,
```
AMMayberry1, line 20: We avoid using `this` inside of abilities b/c there are cases where another card may need to activate an ability printed on this card, so we need to be flexible about which card is the source card of the ability. So we always use `context.source` instead of `this`

server/game/cards/03_TWI/upgrades/RogerRoger.ts
```diff
@@ -0,0 +1,28 @@
...
+                    target: context.target,
```
AMMayberry1, line 21: This isn't needed, by default if an effect is inside of a `targetResolver` this assignment will happen automatically

server/game/cards/03_TWI/upgrades/RogerRoger.ts
```diff
@@ -0,0 +1,28 @@
...
+            internalName: 'roger-roger',
+        };
+    }
+
+    protected override setupCardAbilities() {
+        this.addWhenDefeatedAbility({
+            title: 'Attach to a friendly Battle Droid token',
+            targetResolver: {
+                controller: RelativePlayer.Self,
+                cardCondition: (card) => card.isTokenUnit() && card.title === 'Battle Droid',
```
AMMayberry1, line 18: nit: we prefer to use type filters where possible for readability, unless there's a specific need to have the type check inside the condition (e.g. if we're checking a card's cost and need the type inference to happen). So here I would recommend removing the `.isTokenUnit()` check and instead doing `cardTypeFilter: CardType.TokenUnit`

test/server/cards/03_TWI/upgrades/RogerRoger.spec.ts
```diff
@@ -0,0 +1,80 @@
+describe('Roger Roger', function() {
+    integration(function(contextRef) {
+        describe('Roger Roger\'s ability', function() {
```
AMMayberry1, line 3: nit: since there's just one `it()` block we could remove the `describe` and just have the `it()` with all the title text

test/server/cards/03_TWI/upgrades/RogerRoger.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+            it('should be transferred to a Battle Droid token when defeated', function () {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['droid-deployment'],
+                        groundArena: [
+                            'battle-droid',
```
AMMayberry1, line 10: Could we also put a Clone Trooper token in play just to confirm that Battle Droids are the only permissible token type

test/server/cards/03_TWI/upgrades/RogerRoger.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+                // Droid Deployment is in discard
+                expect(context.player1.discard.length).toBe(1);
+
+                // CASE 4: Roger Roger is discard if there are no friendly Battle Droids in play
+
+                context.player1.passAction();
+                context.player2.clickCard(context.superlaserBlast);
+
+                expect(rogerRoger).toBeInZone('discard');
+                expect(context.player1.discard.length).toBe(2); // Droid Deployment & Roger Roger
```
AMMayberry1, line 76: Awesome job on the tests! Couple more edge cases it would be good to have involving controllership changes:

1. Let's confirm that it works right if the attached unit changes control - i.e., p2 takes control of the battle droid (maybe Change of Heart) and it is defeated, Roger Roger should still go back to a droid on p1. This could be combined with one of the other cases above or a new one, whichever you prefer

2. We actually don't have the pieces right now for this but could you add a comment here that says `TODO: test that Roger Roger works correctly if stolen with Evidence of the Crime`

server/game/cards/03_TWI/upgrades/RogerRoger.ts
moyerr: @AMMayberry1 Okay, I was a little unsure about this because of the note on [`context.source` and upgrades](https://github.com/SWU-Karabast/forceteki/wiki/Card-Components-Reference#contextsource-and-upgrades) from the wiki. In this case, `context.source` is still the upgrade, because it's not giving the attached unit an ability, is that correct?

test/server/cards/03_TWI/upgrades/RogerRoger.spec.ts
moyerr: Done!

server/game/cards/03_TWI/upgrades/RogerRoger.ts
AMMayberry1: Oh yeah good question, that's exactly correct

AMMayberry1 (approved):

PR #453, opened 2025-01-21
Add Unlimited Power and MagnaGuard Wing Leader

Frugghi:
| Unlimited Power | MagnaGuard Wing Leader |
| --- | --- |
| image | image |

### DIFF
--- a/test/server/cards/03_TWI/units/MagnaguardWingLeader.spec.ts
+++ b/test/server/cards/03_TWI/units/MagnaguardWingLeader.spec.ts
@@ -1,0 +1,38 @@
+describe('MagnaGuard Wing Leader', function () {
+    integration(function (contextRef) {
+        it('MagnaGuard Wing Leader\'s ability should allow to attack with a droid unit and then another droid unit', function () {
+            contextRef.setupTest({
+                phase: 'action',
+                player1: {
+                    hand: [],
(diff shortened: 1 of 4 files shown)


test/server/cards/03_TWI/units/MagnaguardWingLeader.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+            expect(context.player1).toBeAbleToSelectExactly([context.b1SecurityTeam, context.squadronOfVultures]);
+            context.player1.clickCard(context.b1SecurityTeam);
+            context.player1.clickCard(context.p2Base);
+
+            expect(context.p2Base.damage).toBe(7);
+
+            context.player2.passAction();
+
+            // Player 1 can't choose MagnaGuard Wing Leader action again in the same round
+            expect(context.player1).not.toBeAbleToSelect(context.magnaguardWingLeader);
```
AMMayberry1, line 35: Actions with no cost are a little tricky b/c we have to make sure that they can't be used as a pseudo-pass when they have no game-state-changing effect. Can we add one additional test case where the wing leader is exhausted and there are no droid units available to attack with, and just do a `.not.toHaveAvailableActionWhenClickedBy()` to confirm it can't be used

AMMayberry1 (approved):

PR #459, opened 2025-01-24
Implementing IG-11, I Cannot Be Captured

kleer1:
image

This required a small refactor in the `ReplacementEffectSystem.ts` and `SimultaneousGameSystem.ts` since they didn't support multiple effects replacing a single effect. I also added unit tests to cover multiple defeat effects to make sure that they are all getting resolved correctly with the refactor. The only issue I'm not sure on is the `GameEvent.setReplacementEvent`. Currently the event will just be whichever was last. So far this doesn't seem to be negatively affecting anything. If at some point we do need to track all of the events that did the replacing, the `replacementEvent` could be turned into an array.


AMMayberry1:
Hey thanks a ton for the awesome PR! This is fantastic work. I'll do a review later today, in the meantime would you mind pinging me (Veld) in the karabast discord so we can get you connected to the dev chat?

server/game/cards/02_SHD/units/IG-11ICannotBeCaptured.ts
```diff
@@ -0,0 +1,46 @@
...
+                onCardCaptured: (event, context) => event.card === context.source
+            },
+            replaceWith: {
+                target: this,
+                replacementImmediateEffect: AbilityHelper.immediateEffects.simultaneous([
+                    AbilityHelper.immediateEffects.defeat(),
+                    AbilityHelper.immediateEffects.damage((context) => {
+                        return {
+                            amount: 3,
+                            target: context.game.getOtherPlayer(context.source._controller).getUnitsInPlay(ZoneName.GroundArena)
```
AMMayberry1, line 26: nit: we should use `source.controller` getter instead of the `source._controller` field

server/game/core/event/GameEvent.ts
```diff
@@ -134,7 +134,6 @@ export class GameEvent {
     // TODO: refactor this to allow for "partial" replacement effects like Boba Fett's Armor
     public setReplacementEvent(replacementEvent: any) {
         Contract.assertNotNullLike(replacementEvent, `Attempting to set null replacementEvent for ${this.name}`);
-        Contract.assertIsNullLike(this.replacementEvent, `Attempting to set replacementEvent ${replacementEvent.name} for ${this.name} but it already has a value: ${this.replacementEvent?.name}`);
```
AMMayberry1, line 137: I think here we should go ahead and convert this to an array just so that it behaves in the way someone would expect from a coding / debugging perspective, even if it's not strictly needed right now (should be fairly easy thankfully).

It does bring up a game rules situation that's currently ambiguous - if an effect is replaced, then for the purposes of costs and "if you do" clauses we check that the replacement resolved instead of the original (e.g., defeating a shield counts as paying a damage cost). There's nothing that clearly states what to do if _multiple_ replacement effects happen, which might not all resolve.

There isn't a game case where it matters right now so imo we need to use our best judgment and put something in place, my gut instinct is just to say that if every replacement event resolved then we count the original event as resolved (boolean AND).

server/game/gameSystems/SimultaneousSystem.ts
```diff
@@ -81,7 +81,16 @@ export class SimultaneousGameSystem<TContext extends AbilityContext = AbilityCon
         }
 
         for (const gameSystem of properties.gameSystems) {
-            context.game.queueSimpleStep(queueGenerateEventGameStepsFn(gameSystem), generateStepName(gameSystem));
+            // If it's a replacement effect, just add them to events and let the ReplacementEffectSystem handle them
+            if (properties['replacementEffect'] === true) {
```
AMMayberry1, line 85: nit: our code convention would be to use `properties.replacementEffect` here, unless there's an advantage to doing it this way

server/game/gameSystems/SimultaneousSystem.ts
```diff
@@ -81,7 +81,16 @@ export class SimultaneousGameSystem<TContext extends AbilityContext = AbilityCon
...
+                gameSystem.queueGenerateEventGameSteps(
+                    events,
+                    context,
+                    Object.assign({ replacementEffect: true }, additionalProperties)
```
AMMayberry1, line 89: nit: we're migrating towards preferentially using object expansion format as our convention, i.e., `{ ...additionalProperties, replacementEffect: true }`

test/server/cards/02_SHD/units/IG11ICannotBeCaptured.spec.ts
```diff
@@ -0,0 +1,192 @@
+describe('IG-11, I Cannot Be Captured', function() {
+    integration(function(contextRef) {
+        describe('IG-11\'s on attack ability', function() {
```
AMMayberry1, line 3: nit: for cases where there's only one `it()` inside of a `describe()`, you can roll everything up so that there's just a top-level `it()` with the test setup inside just to make the file read a little cleaner (totally up to you as to which format you prefer). See Zillo Beast tests for an example

test/server/cards/02_SHD/units/IG11ICannotBeCaptured.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['bounty-hunter-crew', 'oomseries-officer'],
+                        spaceArena: ['cartel-spacer'],
+                        hand: ['relentless-pursuit'],
+                        leader: { card: 'iden-versio#inferno-squad-commander', deployed: true }
+                    },
+                    player2: {
+                        groundArena: ['ig11#i-cannot-be-captured', 'wampa'],
+
```
AMMayberry1, line 163: nit: whitespace

test/server/cards/02_SHD/units/IG11ICannotBeCaptured.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+                expect(context.ig11).toBeInZone('discard');
+                expect(context.wampa.damage).toBe(0);
+                expect(context.cartelSpacer.damage).toBe(0);
+                expect(context.rebelPathfinder).toBeInZone('discard');
+                expect(context.bountyHunterCrew.damage).toBe(3);
+                expect(context.p2Base.damage).toBe(0);
+            });
+        });
+
+        describe('IG-11\'s on capture ability still triggers iden\'s ability', function() {
```
AMMayberry1, line 85: I think for the sake of reducing test case count we could omit these middle two, if I'm reading it right they don't provide much unique test coverage beyond what the last test does and I like having the prompt in that one to confirm the exact timing.

Just by way of explanation re: our test practices - we have this unfortunate issue where the jasmine tests are really slow and heavy, and we have finite github actions time in the free tier. We do plan to address that at some point by switching to a better test framework, but for now we have to engage in some less-than-ideal test practices.

The big thing is the consolidation of cases, which you already picked up on (thanks!). So while I'm very pro-generous-test-coverage in most situations I think here we need to try and cut back a little if we can, unless there's some additional value to these middle tests that I'm missing. Overall awesome job on the tests though, you covered all the bases pretty thoroughly

server/game/gameSystems/SimultaneousSystem.ts
kleer1: Unfortunately, the `replacementEffect` field isn't part of the object definition.
image
It is manually assigned to the object in the `ReplacementEffectSystem`.
image
I can add it to the `IGameSystemProperties` if we want to be able to always access it via `properties.replacementEffect`.

AMMayberry1: Yeah I'd prefer that honestly, the original code base was pretty messy JS so there's still a lot of cleaning up happening around type stuff. Thanks!

server/game/core/event/GameEvent.ts
```diff
@@ -134,14 +134,16 @@ export class GameEvent {
...
     public setReplacementEvent(replacementEvent: any) {
         Contract.assertNotNullLike(replacementEvent, `Attempting to set null replacementEvent for ${this.name}`);
+        Contract.assertNotNullLike(this.replacementEvents, 'GameEvent.replacementEvents can not be null');
 
-        this.replacementEvent = replacementEvent;
+        this.replacementEvents.push(replacementEvent);
         this.resolutionStatus = EventResolutionStatus.REPLACED;
     }
 
     public getResolutionEvent() {
```
AMMayberry1, line 143: Looks like this isn't actually used anymore so we can just remove it outright I think

AMMayberry1 (approved):

PR #462, opened 2025-01-25
first draft of generic deck move system

ssoulier:
I am clearly no there yet but at least unit test of Inferno Four and R2D2 are working.

I am not convinced by the value of the generic method to move card from deck vs having a couple of classes dedicated to the different scenarios we may encounter (top/bottom, top/discard, ...) . But this is very softly opinionated. I'll follow your advice on this @AMMayberry1 .
For example, I am wondering if we can't handle the discard destination in MoveCardSystem and then It should be easier to create a specific class for the Top/Discard case.

I have a problem with 

```
export function lookMoveDeckCardsTopOrBottom<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<ILookMoveDeckCardsTopOrBottomProperties, TContext>) {
    return new LookMoveDeckCardsTopOrBottomSystem<TContext>(propertyFactory);
}
```

In gameSystemLibrary. It's not clear how to fix that between the base class and the children class

I have some difficulties creating a generic `getEffectMessage` for all the situations so I decided to not overwrite it in the base class. But maybe I should like in `MoveCardSystem` and implement a complex logic. I was a bit lazzy.

Finally, I am sure to understand how MoveCardSystem ensure we can choose the order of the cards we put on the top or bottom.

### DIFF
--- a/server/game/gameSystems/LookMoveDeckCardsSystem.ts
+++ b/server/game/gameSystems/LookMoveDeckCardsSystem.ts
@@ -1,0 +1,117 @@
+import type { AbilityContext } from '../core/ability/AbilityContext';
+import type { Card } from '../core/card/Card';
+import type { GameEvent } from '../core/event/GameEvent';
+import type { MoveZoneDestination } from '../core/Constants';
+import { GameStateChangeRequired, ZoneName } from '../core/Constants';
+import { EventName, DeckZoneDestination } from '../core/Constants';
+import { MoveCardSystem } from './MoveCardSystem';
+import * as Contract from '../core/utils/Contract';
+import type { IPlayerTargetSystemProperties } from '../core/gameSystem/PlayerTargetSystem';
+import { PlayerTargetSystem } from '../core/gameSystem/PlayerTargetSystem';
+import type Player from '../core/Player';
+import { DiscardFromDeckSystem } from './DiscardFromDeckSystem';
+
+export interface ILookMoveDeckCardsProperties extends IPlayerTargetSystemProperties {
+    amount: number;
+    destinations: MoveZoneDestination[];
+    promptTitle: string;
+}
+
+export class LookMoveDeckCardsSystem<TContext extends AbilityContext = AbilityContext> extends PlayerTargetSystem<TContext, ILookMoveDeckCardsProperties> {
+    public override readonly name = 'lookMoveDeckCardsSystem';
+    protected override readonly eventName = EventName.OnLookMoveDeckCards;
+    public override readonly effectDescription = 'look at the top cards of your deck and move them';
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public override eventHandler(event): void { }
+
+    public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext): void {
+        const player = context.player;
+        const { amount, destinations, promptTitle } = this.generatePropertiesFromContext(context);
+        const deckLength = player.drawDeck.length;
+
+        if (deckLength === 0) {
+            return;
+        }
+
+        if (Array.isArray(destinations)) {
+            if (destinations.length === 0) {
+                throw new Error('destinations must have at least one element');
+            }
+        }
+
+        const perCardButtons = destinations.map((destination) => {
+            switch (destination) {
+                case DeckZoneDestination.DeckTop:
+                    return { text: 'Put on top', arg: 'top' };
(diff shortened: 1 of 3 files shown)


server/game/gameSystems/LookMoveDeckCardsSystem.ts
```diff
@@ -0,0 +1,117 @@
...
+import { MoveCardSystem } from './MoveCardSystem';
+import * as Contract from '../core/utils/Contract';
+import type { IPlayerTargetSystemProperties } from '../core/gameSystem/PlayerTargetSystem';
+import { PlayerTargetSystem } from '../core/gameSystem/PlayerTargetSystem';
+import type Player from '../core/Player';
+import { DiscardFromDeckSystem } from './DiscardFromDeckSystem';
+
+export interface ILookMoveDeckCardsProperties extends IPlayerTargetSystemProperties {
+    amount: number;
+    destinations: MoveZoneDestination[];
```
AMMayberry1, line 16: I think we need to make this a little more complex - unfortunately we can't just pass in 'discard' as the move destination, we need to actually trigger the DiscardSpecificCard system. Two reasons:

1. There might be special rules around discarding that are explicitly handled by that system (or will be in the future)
2. If there are triggered abilities that react to discard, then we need to correctly emit an OnCardDiscarded event instead of a generic OnCardMoved event

In addition, we want to make it possible to provide a string name for the button so we don't have to maintain a list of all the options in this class, that's overall difficult to maintain. Much better to let the caller declare it themselves, except potentially to handle some really common defaults (which is what the subclass idea is about).

What we have used in similar situations such as e.g. the SelectInterface use the `IChoicesInterface` which pairs a string name for the button with the associated action - see Ezra impl for an example.

server/game/gameSystems/LookMoveDeckCardsTopOrBottomSystem.ts
```diff
@@ -1,29 +1,22 @@
...
-    // eslint-disable-next-line @typescript-eslint/no-empty-function
-    public override eventHandler(event): void { }
+export class LookMoveDeckCardsTopOrBottomSystem<TContext extends AbilityContext = AbilityContext> extends LookMoveDeckCardsSystem<TContext> {
+    public constructor(propertiesOrPropertyFactory: ILookMoveDeckCardsTopOrBottomProperties) {
+        const destinations = [DeckZoneDestination.DeckTop, DeckZoneDestination.DeckBottom];
+        const promptTitle = 'Select card to move to the top or bottom of the deck';
+        super({ amount: propertiesOrPropertyFactory.amount, destinations: destinations, promptTitle: promptTitle });
+    }
 
     public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext): void {
```
AMMayberry1, line 18: Probably we should remove this method completely and move it entirely to the parent class. afaik the subclass has no reason to change the behavior of this

AMMayberry1:
Re: `getEffectMessage` yeah just don't worry about it for now, we haven't been consistent about using that in general so we can come do cleanup on it later if we need to (not sure if we'll continue to use it the same way)

AMMayberry1:
For the GameSystemLibrary question - not sure what the specific issue is, but it should be fine to create one function for the parent class and one for the child class. This is what we did for `deckSearch()` (parent) vs `playMultipleCardsFromDeck()` (child)

AMMayberry1:
The way the ordering for top / bottom happens is that when the player clicks a button, we generate a move event for that card and push it onto a list. Then we resolve the move events in that order. So the cards are always moved in the order the player clicked the buttons.

server/game/gameSystems/LookMoveDeckCardsSystem.ts
ssoulier: This is what I am doing line 106, where I use `DiscardFromDeckSystem`. Does it answer your point about using the dedicated system for Discard?

AMMayberry1: I mean for now it's not a huge deal since there's only two different use cases, but in general it's not a great design pattern to try and cover every possible option inside of the system implementation. The reason is that if we have other cards that end up needing different options then we need to keep adding more code inside of the system to cover all of the cases.

It is much easier and more scalable if we just make it so that the cards can provide the button text and the system / handler to use. That way for one-off cases like FACIBI in the future, the code can just be in the card implementation and we don't have to add every case to the system.

AMMayberry1:
Closing for now

PR #463, opened 2025-01-25
Add Stolen Landspeeder implementation

Frugghi:
image

### DIFF
--- a/server/game/cards/02_SHD/units/StolenLandspeeder.ts
+++ b/server/game/cards/02_SHD/units/StolenLandspeeder.ts
@@ -1,0 +1,35 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { PlayType } from '../../../core/Constants';
+import { CostAdjustType } from '../../../core/cost/CostAdjuster';
+
+export default class StolenLandspeeder extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '7642980906',
+            internalName: 'stolen-landspeeder'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'If you played this unit from your hand, an opponent takes control of it',
+            when: {
+                onCardPlayed: (event, context) => event.card === context.source && event.playType === PlayType.PlayFromHand
+            },
+            immediateEffect: AbilityHelper.immediateEffects.takeControlOfUnit((context) => ({
+                newController: context.player.opponent
+            }))
+        });
+
+        this.addBountyAbility({
+            title: 'If you own this unit, play it from your discard pile for free and give an Experience token to it',
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.playCardFromOutOfPlay({
+                    adjustCost: { costAdjustType: CostAdjustType.Free },
+                }),
+                AbilityHelper.immediateEffects.giveExperience(),
+            ]),
+        });
+    }
+}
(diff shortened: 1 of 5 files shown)


server/game/core/ability/TriggeredAbility.ts
```diff
@@ -95,6 +95,35 @@ export default class TriggeredAbility extends CardAbility {
...
+    public override meetsRequirements(context: TriggeredAbilityContext, ignoredRequirements: string[] = [], thisStepOnly: boolean = false): string {
+        let canPlayerTrigger: boolean;
+        let controller = context.source.controller;
+
+        // If the event's card is the source of the ability, use the last known controller of the card instead of the source's controller.
+        // This is because when resolving triggered abilities like "when defeated", the defeated card is in the discard pile already
+        // and it might have changed controller.
+        if (context.event.card === context.source && context.event.lastKnownInformation) {
+            controller = context.event.lastKnownInformation.controller;
+        }
```
Frugghi, line 107: Not sure if there is a smarter way to fix this bug 🤔

AMMayberry1: Ohh good catch, we didn't know about this issue at all.

I think this solution is overall the correct one, it's definitely the type of thing that LKI is intended for. I would prefer though if we could maybe consolidate this code a little more cleanly with `CardAbility` so that there isn't duplicated code - maybe we can add a protected method in `CardAbility` named like `controllerMeetsRequirements` (or anything better you can come up with) which is then overridden in `TriggeredAbility`?

server/game/cards/02_SHD/units/StolenLandspeeder.ts
```diff
@@ -0,0 +1,35 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '7642980906',
+            internalName: 'stolen-landspeeder'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'If you played this unit from your hand, an opponent takes control of it',
```
AMMayberry1, line 16: nit: Can we remove the "if" clauses from these titles

test/server/cards/02_SHD/units/StolenLandspeeder.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+            context.player1.clickCard(context.stolenLandspeeder);
+
+            // and collects the bounty
+            expect(context.player1).toHavePassAbilityPrompt(bountyPrompt);
+            context.player1.clickPrompt(bountyPrompt);
+
+            expect(context.stolenLandspeeder).toBeInZone('groundArena', context.player1);
+            expect(context.stolenLandspeeder).toHaveExactUpgradeNames(['experience']);
+
+            // Player 2 kills Stolen Landspeeder and cannot collect the bounty
```
AMMayberry1, line 37: nit: comments are a little unclear, says "cannot collect the bounty" here but then "collects the bounty" below

server/game/cards/02_SHD/units/StolenLandspeeder.ts
```diff
@@ -0,0 +1,35 @@
...
+            },
+            immediateEffect: AbilityHelper.immediateEffects.takeControlOfUnit((context) => ({
+                newController: context.player.opponent
+            }))
+        });
+
+        this.addBountyAbility({
+            title: 'If you own this unit, play it from your discard pile for free and give an Experience token to it',
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.playCardFromOutOfPlay({
```
AMMayberry1, line 28: I guess the "if you own this unit" clause is implicitly enforced by `playCardFromOutOfPlay` currently, but could we actually add a `conditional()` here just to make it explicit? `playCardFromOutOfPlay` will soon support playing out of opponent's discard which might break this

server/game/cards/02_SHD/units/StolenLandspeeder.ts
Frugghi: Isn't the bounty one important? If the action just said `Collect bounty: Play it from your discard pile for free and give an Experience token to it` could make you think that you'll be able to play it even if you don't own it.

AMMayberry1: Oh yeah good catch!

AMMayberry1 (approved):

PR #465, opened 2025-01-26
Implement Boba Fett's Armor

kleer1:
image


server/game/cards/02_SHD/upgrades/BobaFettsArmor.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addReplacementEffectAbility({
+            title: 'If attached unit is Boba Fett and damage would be dealt to him, prevent 2 of that damage',
+            when: {
+                onDamageDealt: (event, context) => event.card === context.source.parentCard && context.source.parentCard.title === 'Boba Fett' &&
+                  (event.context.event === undefined || event.context.event.card !== context.source.parentCard) // otherwise can replace infinitely
```
kleer1, line 21: This is the bit that stops the infinite replacement. Seems like it caused when the replacement effect: 1) has the same type as the trigger (ie damage -> damage) and 2) it has the same target as the original effect. So far I think this is the only effect in the game that would trigger this recursion, but if there are others or they make more cards like this, it would be worth burying this check somewhere deeper so that we always break this chain.

AMMayberry1: Great job sorting this out - I think it's worth taking some time to solve this one now to prevent confusion if it comes up again later. Thankfully I think the fix is relatively easy.

`TriggeredAbility.ts` is the class for all triggered abilities and manages the logic for determining if an ability should legally trigger or not. It has an array property `eventsTriggeredFor` which is used to prevent an ability from being triggered twice for the same event (we emit an event multiple times for rules timing reasons).

I think we could pretty easily add similar logic in the subclass `ReplacementEffectAbility.ts` to add on an additional check that the "parent" event that was replaced (if any) is also not in `eventsTriggeredFor`. I know we have bookkeeping that lets you find a replacement event from the original, not sure if we have it in the reverse direction as well so we may need to add that.

But yeah generally we probably could just refactor `TriggeredAbility.eventHandler()` a little bit so it has some method for checking `eventsTriggeredFor` which we then override in `ReplacementEffectAbility`.

Let me know if that makes sense or if you have any questions or hit any issues. Looks like `TriggeredAbility.eventHandler()` has a comment saying that method is in the documentation which will need to be updated, it's a little stale - the real docs are in the wiki now. I can manage that update myself once the PR is merged.

test/server/cards/02_SHD/upgrades/BobaFettsArmor.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+                    groundArena: ['battlefield-marine']
+                },
+                player2: {
+                    groundArena: [{ card: 'boba-fett#disintegrator', upgrades: ['boba-fetts-armor'] }]
+                }
+            });
+
+            const { context } = contextRef;
+
+            // initial check
```
AMMayberry1, line 16: nit: I'd be okay without these checks since they're really just validating that the upgrade stat computation works, which isn't really the SUT here

test/server/cards/02_SHD/upgrades/BobaFettsArmor.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+            expect(context.bobaFett.getPower()).toBe(5);
+            expect(context.bobaFett.getHp()).toBe(7);
+
+            // attack base and power up other unit
```
AMMayberry1, line 20: nit: not clear what "power up other unit" means in these comments

test/server/cards/02_SHD/upgrades/BobaFettsArmor.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+            context.player1.clickCard(context.battlefieldMarine);
+            context.player1.clickCard(context.bobaFett);
+            expect(context.bobaFett.damage).toBe(1);
+            expect(context.battlefieldMarine).toBeInZone('discard');
+        });
+
+        it('Boba Fett\'s Armor prevents 2 damage when attacked with 2 power and ping ability', function () {
```
AMMayberry1, line 27: Since most of these cases are just different things attacking Boba I think we could reasonably consolidate them down to one case (again for keeping the test run time down). Could still have a few different Bobas (and the Sabine) wearing different copies of the armor.

Note that in that case you'll need to refer to each Boba by the full name including subtitle since there are multiple with the same title - e.g., `context.bobaFettDaimyo` and `context.bobaFettDisintegrator`.

test/server/cards/02_SHD/upgrades/BobaFettsArmor.spec.ts
kleer1: Sorry. Copy/pasted and didn't clean up.

server/game/cards/02_SHD/upgrades/BobaFettsArmor.ts
kleer1: This makes perfect sense (hopefully). I think I've updated accordingly.

test/server/cards/02_SHD/upgrades/BobaFettsArmor.spec.ts
kleer1: Removed

test/server/cards/02_SHD/upgrades/BobaFettsArmor.spec.ts
kleer1: Consolidated into a single test.

server/game/core/gameSteps/abilityWindow/TriggerWindowBase.ts
```diff
@@ -276,7 +276,7 @@ export abstract class TriggerWindowBase extends BaseStep {
     }
 
     private canAnyAbilitiesResolve(triggeredAbilities: TriggeredAbilityContext[]) {
-        return triggeredAbilities.some((triggeredAbilityContext) => triggeredAbilityContext.ability.hasAnyLegalEffects(triggeredAbilityContext, true));
+        return triggeredAbilities?.some((triggeredAbilityContext) => triggeredAbilityContext.ability.hasAnyLegalEffects(triggeredAbilityContext, true));
```
kleer1, line 279: Had to add this since `triggeredAbilities` was coming in a `undefined` for some of the tests. Its occurring when `let abilitiesToResolve = this.unresolved.get(this.currentlyResolvingPlayer);` in `server\game\core\gameSteps\abilityWindow\TriggerWindowBase.promptUnresolvedAbilities()` doesn't have any unresolved abilities for the currently resolving player. It didn't seem serious to me, but let me know if you disagree.

server/game/core/event/GameEvent.ts
```diff
@@ -140,6 +140,10 @@ export class GameEvent {
         this.resolutionStatus = EventResolutionStatus.REPLACED;
     }
 
+    public hasReplacmentEvent(event: GameEvent): boolean {
```
AMMayberry1, line 143: nit, typo: "Replacment"

server/game/core/ability/TriggeredAbility.ts
```diff
@@ -168,6 +169,11 @@ export default class TriggeredAbility extends CardAbility {
         }
     }
 
+    protected invalidEventInTriggeredFor(event) {
```
AMMayberry1, line 172: Do you think it would be cleaner if we put the `!this.eventsTriggeredFor.includes(event)` in here and then the subclass method does its existing check and also calls this method in the superclass? Totally up to you which way you find cleaner, I was thinking that since both are basically doing the same thing (checking if we're triggering off of an already-triggered event) they could go in the same method but I'm good with whichever you find cleaner. If we do consolidate then probably needs a more specific name like "checkEventAlreadyTriggered" or something.

test/server/cards/02_SHD/upgrades/BobaFettsArmor.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+            expect(context.bobaFettDaimyo.damage).toBe(3);
+
+            resetBobaHealth();
+
+            // CASE 4: Armor prevent 2 damage from an event
+            context.player1.clickCard(context.bombingRun);
+            context.player1.clickPrompt('Ground');
+            context.player1.clickPrompt('Opponent');
+            expect(context.bobaFettDisintegrator.damage).toBe(1);
+            expect(context.bobaFettDaimyo.damage).toBe(1);
```
AMMayberry1, line 63: Apologies for the late comment - one really important test I forgot about is testing when the unit also has a shield. We need to confirm that (a) the player is allowed to choose the order they resolve in, and (b) that they resolve correctly in either order. Actually probably we need three cases:

1. Trigger Armor first with damage > 2, shield still triggers
2. Trigger Armor first with damage < 2, shield doesn't trigger
3. Trigger shield first (with damage > 2 for good measure). No real way to test if Armor triggers so this is more of checking that there's no error.

This should probably a separate test setup since you'll be running the same situation multiple times with different resolution orders. See the Maul1 unit tests for an example of something very similar.

kleer1: Good call-out on this. There is an issue with the shield not procking on damage > 2 that I'll need to work through.

AMMayberry1: That's unfortunate. The `ReplacementEffectWindow` is really only halfway-implemented so the work involved may get a bit high. If it looks like it's going to be a project then feel free to leave a comment here like `TODO BUG: ...` and maybe a github issue and we can do it as a separate PR.

server/game/core/event/GameEvent.ts
kleer1: fixed

server/game/core/ability/TriggeredAbility.ts
kleer1: This makes sense. Moved `this.eventsTriggeredFor.includes(event)` into this method.

test/server/cards/02_SHD/upgrades/BobaFettsArmor.spec.ts
kleer1: Found the issue. The event attached to the shield ability was still the original event that was replaced by the armor. As a result, it was getting removed from the `unresolved` triggers during cleanup. I came up with a solution by adding a `isPartial` prop to replacement effects and then checking that (along with a bunch of other stuff) in the replacement effect system.  I _think_ this fixes partial replacement effects by allowing each card to define what it means for it to be partially resolved. If it is only partially resolved, then it shifts all of the matching unresolved trigger abilities to point at the replacement event instead of the event that has been replaced.

Let me know if anything with it seems wonky.

server/game/cards/02_SHD/upgrades/BobaFettsArmor.ts
```diff
@@ -0,0 +1,35 @@
...
+                    source: context.event.damageSource.damageDealtBy
+                }))
+            },
+            effect: 'Boba Fett\s armor prevents 2 damage to {1}',
+            effectArgs: (context) => [context.source.parentCard],
+        });
+    }
+}
+
+BobaFettsArmor.implemented = true;
```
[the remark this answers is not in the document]
kleer1: fixed

test/server/cards/02_SHD/upgrades/BobaFettsArmor.spec.ts
AMMayberry1: Great find! Yeah the partial replacement definitely causes some weird issues. I like the general direction of the idea but I have a few thoughts on how we go about it.

For the general design, I'd prefer that whatever we do be encapsulated in ReplacementEffectWindow as much as possible since that's the central place where all of the ordering gets managed. What you have currently is actually reading some protected properties of ReplacementEffectWindow such as `unresolved`, but since our event objects are untyped for legacy reasons it isn't getting caught by TS.

As far as specific solutions - I'd like to rearchitect the flow a little bit to see if we can avoid needing to explicitly retarget the triggered events and see if we can preserve the basic flow as much as possible.

The current implementation (before this PR) re-emits any replacement events that get added during replacement resolution. This came up in the case of Maul1 sending the damage to a shielded Underworld unit - it created a new damage event which needed another round of replacement checks to trigger the shield on the Underworld unit.

The same thing should be what we do in this situation - the Armor replacement event generates a new damage event, which the shield then triggers off of. But actually what happens is that the new logic we added prevents that because the shield saw itself be triggered off of the original event, so the `invalidEventInTriggeredFor` check fails thinking that we've already resolved this shield trigger.

My preference would be to find a way to resolve that check failure so that we can keep the basic flow of "when a replacement event emits a new event, check all replacement abilities for that event." That seems pretty elegant and simple if we can get it to work right with this multiple layers issue.

Let me know what you think of the above. There's a bunch of ways we could solve this but one might just be to update the replacement effect check to confirm that we not just triggered off of the event but also actually did resolve on it, though that might require some fiddling so there may also be simpler methods.

kleer1: This does make sense. I _think_ I've accomplished this by moving all of the logic into the `ReplacementEffectWindow` which allowed me to also remove the `invalidEventInTriggeredFor` since it was no longer necessary. Let me know if this is what you had in mind.

server/game/core/gameSteps/abilityWindow/ReplacementEffectWindow.ts
```diff
@@ -42,6 +42,28 @@ export class ReplacementEffectWindow extends TriggerWindowBase {
...
+    public override addTriggeredAbilityToWindow(context: TriggeredAbilityContext) {
+        // We have to do extra checks if the event that triggered this event has been replaced.
+        // The goal is to prevent infinite replacement if a replacement effect's replacment triggers itself.
+        if (context.event?.context?.event?.isReplaced) {
+            const replaced = context.event.context.event.replacementEvents.find((e) => e === context.event);
+            // If the current event is the same as what replaced the previous event, then this is potentially an infinite replacement.
+            if (replaced) {
+                // If the ability that triggered the replacement fully replaced the event (ie all damage was prevented),
+                // then it shouldn't need to be replaced again.
+                if (replaced.context.ability?.properties?.isPartial && !replaced.context.ability?.properties?.isPartial(replaced.context)) {
```
AMMayberry1, line 54: I was a little confused why the `isPartial` logic was needed at all then I realized there's another small bug in the flow.

Basically, if a damage event is emitted with amount = 0, in the normal EventWindow flow, the [DamageSystem will cancel itself](https://github.com/SWU-Karabast/forceteki/blob/27a24c9f1facb0eded98737ffcd640458180f595/server/game/gameSystems/DamageSystem.ts#L112). So I was expecting not to need the `isPartial` property b/c if Boba Fett's Armor negates everything and emits a damage event with amount 0 then it should just naturally fizzle before the shield is triggered.

I realized though that in the flow for generating "secondary" replacement effects, it looks like we never actually evaluate whether the newly added replacement events can resolve. The TriggerWindowBase is heavily based on the flow for standard trigger windows (TriggeredAbilityWindow.ts), so it assumes that events don't normally get added to the window during resolution.

This _should_ be getting caught at the [`cleanUpTriggers()` stage of TriggerWindowBase](https://github.com/SWU-Karabast/forceteki/blob/27a24c9f1facb0eded98737ffcd640458180f595/server/game/core/gameSteps/abilityWindow/TriggerWindowBase.ts#L256), but it looks like it assumes that the events have already been checked and filtered out. Somewhere in that method we need to add a call `context.event.checkCondition()` if-and-only-if we are a replacement window and the event has not yet resolved (calling `checkCondition()` on already-resolved events could cause unexpected issues).

Honestly that method could probably use a refactor into at least partially separated overloads in the child classes for triggered and replacement windows respectively, but this PR has already gone pretty long so feel free to leave that as a TODO comment if you'd rather just get this unblocked. Totally up your preference.

Anyway, the net effect of this change would be that we shouldn't need to explicitly set `isPartial()` in the future since non-partial events (i.e., events that are completely negated) should just weed themselves out automatically. That would make implementations simpler and less error-prone.

server/game/core/gameSteps/abilityWindow/ReplacementEffectWindow.ts
```diff
@@ -42,6 +42,28 @@ export class ReplacementEffectWindow extends TriggerWindowBase {
         this.newReplacementEvents = true;
     }
 
+    public override addTriggeredAbilityToWindow(context: TriggeredAbilityContext) {
+        // We have to do extra checks if the event that triggered this event has been replaced.
+        // The goal is to prevent infinite replacement if a replacement effect's replacment triggers itself.
+        if (context.event?.context?.event?.isReplaced) {
```
AMMayberry1, line 48: nit: this chain of calls is a little difficult to read and understand what exactly is getting grabbed. Could we add a property to GameEvent like `eventReplaced` (or a better name if you have one) and then here we could just check `context.event.eventReplaced == null`?

server/game/core/gameSteps/abilityWindow/ReplacementEffectWindow.ts
kleer1: The rabbit hole continues to go deeper. I found out why the 0 damage wasn't being filtered out. I set the amount up as a function (`amount: () => Math.max(context.event.amount - 2, 0)`) but in the DamageSystem it only considers the amount to be a value (`'amount' in properties && properties.amount === 0`) and since the function definition doesn't equal 0, it is never getting kicked out. This led me back to the original reason that I switched it from a value to a function, which was that the replacement effect wasn't getting triggered when it was just a value. After a fair amount of debugging, it looks like when the replacement ability is evaluated for whether it can affect the game state, the amount of 0 passed to the `DamageSystem.canAffect` results in the replacement effect being ignored since it says it won't have an affect on the game state. Still thinking through what the correct solution is for this, but we somehow need allow preventing down to 0 to be acceptable.

AMMayberry1 (approved):

PR #468, opened 2025-01-26
Implementing Admiral Trench, Holding the Line

supercodepoet:
image


supercodepoet:
@AMMayberry1 thought about adding another test to make sure you can return the unit you Exploit for Trench. Is that overkill?

test/server/cards/03_TWI/units/AdmiralTrenchHoldingTheLine.spec.ts
```diff
@@ -0,0 +1,55 @@
+describe('Admiral Trench, Holding the Line', function () {
+    integration(function (contextRef) {
+        beforeEach(function () {
+            contextRef.setupTest({
+                phase: 'action',
+                player1: {
+                    base: 'maz-kanatas-castle',
```
AMMayberry1, line 7: nit: any reason we need to declare the bases for these?

test/server/cards/03_TWI/units/AdmiralTrenchHoldingTheLine.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+        it('should return up to 3 units defeated this phase to player hand from a discard pile', function () {
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.superlaserBlast);
+            expect(context.player2).toBeActivePlayer();
+
+            context.player2.clickCard(context.admiralTrenchHoldingTheLine);
+            expect(context.player2).toHavePrompt('Select 3 cards');
+            expect(context.player2).toBeAbleToSelectExactly([context.deathStarStormtrooper, context.maul]);
```
AMMayberry1, line 30: Can we defeat more than 3 units to confirm we have limit set correctly. As part of that test, we can click a fourth unit using `.clickCardNonChecking()` and confirm that it stays in discard

test/server/cards/03_TWI/units/AdmiralTrenchHoldingTheLine.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+            context.player2.clickCard(context.deathStarStormtrooper);
+            context.player2.clickCard(context.maul);
+            context.player2.clickPrompt('Done');
+
+            expect(context.player2.hand).toContain(context.deathStarStormtrooper);
+            expect(context.player2.hand).toContain(context.maul);
+        });
+
+        it('should return no units since none where defeated this phase', function () {
```
AMMayberry1, line 41: We can actually kill two birds with one stone if we merge this into the above test and use `.moveToNextActionPhase()` in between, that lets us confirm that Trench doesn't grab units defeated in a previous phase.

I'm less pushy about testing that than I used to be since the watchers have been extremely thoroughly tested at this point, but if you have this test already might as well do it

test/server/cards/03_TWI/units/AdmiralTrenchHoldingTheLine.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+                    base: 'maz-kanatas-castle',
+                    leader: 'qira#i-alone-survived',
+                    hand: ['superlaser-blast'],
+                    groundArena: ['pyke-sentinel'],
+                },
+                player2: {
+                    base: 'lair-of-grievous',
+                    leader: 'general-grievous#general-of-the-droid-armies',
+                    hand: ['admiral-trench#holding-the-line'],
+                    groundArena: ['death-star-stormtrooper', 'maul#shadow-collective-visionary'],
```
AMMayberry1, line 16: Can we put at least one upgrade out to confirm that Trench only allows units

test/server/cards/03_TWI/units/AdmiralTrenchHoldingTheLine.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+            context.player1.clickCard(context.pykeSentinel);
+            context.player1.clickCard(context.p2Base);
+            expect(context.player2).toBeActivePlayer();
+
+            context.player2.clickCard(context.admiralTrenchHoldingTheLine);
+            context.player2.clickPrompt('Play Admiral Trench');
+            expect(context.player1).toBeActivePlayer();
+
+            expect(context.player2.hand).toEqual([]);
+        });
```
AMMayberry1, line 53: There's one kind of subtle edge case we need to check. If the following sequence of events happens:
1. player1 has wampa on the field
2. wampa is defeated
3. player1 returns wampa to hand with e.g. The Emperor's Legion
4. player2 plays Pillage and forces player1 to discard wampa

Then even though the wampa card was technically defeated this phase, it's technically a different copy of the card than the one that is now in discard and therefore isn't eligible to be grabbed with Trench's ability (see SWU rules 8.6.4 for more details). The `wasDefeatedThisPhase()` watcher method _should_ handle that correctly but we still need to check since it's kind of hard to get right

AMMayberry1:
> @AMMayberry1 thought about adding another test to make sure you can return the unit you Exploit for Trench. Is that overkill?

Nah sounds fun to me 😁

test/server/cards/03_TWI/units/AdmiralTrenchHoldingTheLine.spec.ts
supercodepoet: ahhh, was just using the same test setup as I used for test templates to manually test but the manual testing was broken at the time and wouldn't load. I can remove them.

supercodepoet: it has been removed

test/server/cards/03_TWI/units/AdmiralTrenchHoldingTheLine.spec.ts
supercodepoet: I removed this test as it was not needed and replaced it with the Exploit test

test/server/cards/03_TWI/units/AdmiralTrenchHoldingTheLine.spec.ts
supercodepoet: Added to the test

test/server/cards/03_TWI/units/AdmiralTrenchHoldingTheLine.spec.ts
supercodepoet: I added this scenerio but that test fails. It still saw the First Legion Stormtrooper. @AMMayberry1 is my test work or is there something in the system I need to dig into?

supercodepoet: @AMMayberry1 manually testing using the testing template and running through this scenario it looks like the client shows that the First Order Stormtropper is selected but when clicking Done it does not move the card into the hand. So it looks like maybe the selection in the discard zone might need some work?

AMMayberry1: Hey sorry, I totally missed these comments - I wasn't checking the PR since it was still in draft :/ in the future feel free to ping me in discord if I don't respond promptly. Will take a look at this when I have a sec, might have to be tomorrow

AMMayberry1: Hey sorry finally had time to dig into this - so yeah it's really good we added this test, there was a bug in the engine. Basically, when Snowtrooper goes from discard to hand it's supposed to become a new copy of the card for the purposes of stuff like this, but that update wasn't happening.

I just applied a fix to main, if you merge this should work. One note though is that since Trench's ability has no targets at all, the prompt just won't appear. So The end of the test should look like this:

```ts
  context.player2.clickCard(context.admiralTrenchHoldingTheLine);
  context.player2.clickPrompt('Play Admiral Trench');

  // Should not be able to return any cards since First Legion Snowtrooper is a new copy and was not defeated this phase
  expect(context.player1).toBeActivePlayer();
```

supercodepoet: excellent, thanks for all the hard work

AMMayberry1 (approved):

PR #474, opened 2025-01-27
Implement Lurking TIE Phantom

kleer1:
image

### DIFF
--- a/server/game/cards/02_SHD/units/LurkingTIEPhantom.ts
+++ b/server/game/cards/02_SHD/units/LurkingTIEPhantom.ts
@@ -1,0 +1,26 @@
+import { InitiateAttackAction } from '../../../actions/InitiateAttackAction';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+
+export default class LurkingTIEPhantom extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '1810342362',
+            internalName: 'lurking-tie-phantom',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addReplacementEffectAbility({
+            title: 'This unit can\'t be captured, damaged, or defeated by enemy card abilities',
+            when: {
+                onCardCaptured: (event, context) => event.card === context.source && event.context.source.controller !== context.source.controller,
+                onDamageDealt: (event, context) => event.card === context.source && event.context.source.controller !== context.source.controller &&
+                  !(event.context.ability instanceof InitiateAttackAction),
+                onCardDefeated: (event, context) => event.card === context.source && event.defeatSource && !event.defeatSource.attack &&
+                  event.defeatSource.card?.controller !== context.source.controller
+            }
+        });
+    }
+}
+
+LurkingTIEPhantom.implemented = true;
--- a/test/server/cards/02_SHD/units/LurkingTIEPhantom.spec.ts
+++ b/test/server/cards/02_SHD/units/LurkingTIEPhantom.spec.ts
@@ -1,0 +1,110 @@
+describe('Lurking TIE Phantom', function() {
+    integration(function(contextRef) {
+        it('Lurking TIE Phantom\'s ability', function () {
+            contextRef.setupTest({
+                phase: 'action',
+                player1: {
+                    groundArena: ['cassian-andor#rebellions-are-built-on-hope'],
+                    spaceArena: ['punishing-one#dengars-jumpmaster'],
+                    hand: ['relentless-pursuit', 'daring-raid', 'imperial-interceptor', 'takedown', 'make-an-opening', 'devastating-gunship'],
+                },
+                player2: {
+                    spaceArena: ['lurking-tie-phantom'],
+                    hand: ['open-fire', 'rivals-fall', 'devastator#inescapable', 'count-dooku#darth-tyranus'],
+
+                }
+            });
+
+            const { context } = contextRef;
+
+            // Case 1: Cannot be captured by enemeny effects
+            context.player1.clickCard(context.relentlessPursuit);
+            expect(context.player1).toBeAbleToSelectExactly([context.cassianAndor, context.punishingOne]);
+            context.player1.clickCard(context.cassianAndor);
+            expect(context.player1).toBeAbleToSelectExactly([context.lurkingTiePhantom]);
+            context.player1.clickCard(context.lurkingTiePhantom);
+            expect(context.lurkingTiePhantom).toBeInZone('spaceArena');
+
+            context.player2.passAction();
+
+            // Case 2: can be defated/damaged from attack
+            context.player1.clickCard(context.punishingOne);
+            context.player1.clickCard(context.lurkingTiePhantom);
+            expect(context.lurkingTiePhantom).toBeInZone('discard');
+
+            context.player2.moveCard(context.lurkingTiePhantom, 'spaceArena');
+            context.player2.passAction();
+
+            // Case 3: Cannot be damaged by oppenent's event
+            context.player1.clickCard(context.daringRaid);
+            context.player1.clickCard(context.lurkingTiePhantom);
+            expect(context.lurkingTiePhantom).toBeInZone('spaceArena');
+            expect(context.lurkingTiePhantom.damage).toBe(0);
+
+            context.player2.passAction();
+
+            // Case 4: Cannot be damaged by enemy unit ability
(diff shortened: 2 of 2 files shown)


server/game/cards/02_SHD/units/LurkingTIEPhantom.ts
```diff
@@ -0,0 +1,26 @@
...
+export default class LurkingTIEPhantom extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '1810342362',
+            internalName: 'lurking-tie-phantom',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addReplacementEffectAbility({
```
AMMayberry1, line 13: Unfortunately this is a case where we can't use a replacement effect, since they are concretely defined in the SWU rules and there is specific language around what is and isn't one. In this case, it would only be a replacement effect if the wording was more like shield where it said "If this unit would be be captured, damaged, or defeated by enemy card abilities, instead it is not."

The "would / instead" are the key words there. It's important because if this were to happen at the same time as e.g. a shield, it is important that the shield is not allowed to be triggered - but if both of them are replacement effects then they will both go on the stack (like with Boba Fett's Armor).

For cases like this the mechanism we use is the `AbilityRestriction` enum, the values of which can be applied as effects to cards and then the respective GameSystems will check for the presence of those restrictions by name. Some existing examples of this are:

- Sabine1 unit uses `AbilityRestriction.BeAttacked` which is checked in `AttackStepsSystem`
- Confederate Tri-Fighter uses `AbilityRestriction.BeHealed` which is checked in `HealSystem`

We may need to add a couple of restriction types or at least round out some functionality for LTP, it's the most advanced case of ability restrictions so far

test/server/cards/02_SHD/units/LurkingTIEPhantom.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+                phase: 'action',
+                player1: {
+                    groundArena: ['cassian-andor#rebellions-are-built-on-hope'],
+                    spaceArena: ['punishing-one#dengars-jumpmaster'],
+                    hand: ['relentless-pursuit', 'daring-raid', 'imperial-interceptor', 'takedown', 'make-an-opening', 'devastating-gunship'],
+                },
+                player2: {
+                    spaceArena: ['lurking-tie-phantom'],
+                    hand: ['open-fire', 'rivals-fall', 'devastator#inescapable', 'count-dooku#darth-tyranus'],
+
```
AMMayberry1, line 14: nit: whitespace

test/server/cards/02_SHD/units/LurkingTIEPhantom.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+            context.player2.moveCard(context.lurkingTiePhantom, 'spaceArena');
+            context.player1.passAction();
+            context.player2.setResourceCount(14);
+
+            // Case 11: Can be defeated by your own unit that says defeat
+            context.player2.clickCard(context.countDooku);
+            context.player2.clickPrompt('Defeat a unit with 4 or less remaining HP');
+            context.player2.clickCard(context.lurkingTiePhantom);
+            expect(context.lurkingTiePhantom).toBeInZone('discard');
```
AMMayberry1, line 107: I don't think mechanically it's different than any of these others but just to be safe, could we also add a couple tests where the LTP owner chooses where to place an enemy's effect and confirm it still doesn't hit. So like Cad Bane leader ability and Power of the Dark Side.

AMMayberry1, line 107: Not sure if you've seen this document yet, it's the list of rules clarifications from FFG maintained by the judge community: https://nexus.cascadegames.com/resources/Rules_Clarifications/

There's a bunch of references for LTP, most or all of which we should add cases for or add a TODO comment if the relevant card isn't implemented yet (e.g. Force Lightning).

test/server/cards/02_SHD/units/LurkingTIEPhantom.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+                player2: {
+                    spaceArena: ['lurking-tie-phantom'],
+                    hand: ['open-fire', 'rivals-fall', 'devastator#inescapable', 'count-dooku#darth-tyranus'],
+
+                }
+            });
+
+            const { context } = contextRef;
+
+            // Case 1: Cannot be captured by enemeny effects
```
AMMayberry1, line 20: I read through these as carefully as I could but just want to confirm with you in case I missed anything - were there any cases where the LTP wasn't even targetable for the effect? If so that would be an engine bug, in all cases it should selectable but then the effect just doesn't resolve.

If you caught any of that or if it comes up while implementing further tests please ping me.

server/game/cards/02_SHD/units/LurkingTIEPhantom.ts
kleer1: I was basing this being a replacement effect on the designers stating that in the set Lurking TIE would be able to take damage from indirect damage since lurking tie treats it's abilities as prevention.

>Lurking TIE Phantom can be assigned damage from indirect damage. "Can't be damaged" is a prevent effect, as clarified in CR4. We are considering the additional step of errataing LTP to use "prevent" wording but aren’t 100% on that yet.
>
>(Answer shared on 2025-01-20)

https://gist.github.com/macfergusson/256a534a98fc1d45f2595503f7d54953#rules-clarifications-from-ryan-serrano-swu-rules-admiral

This made it seem like it could potentially be a replacement effect (ie "If it would take damage, prevent that damage"). My concern with a more concrete restriction is that it can still be a valid target for all of those actions.

AMMayberry1: Oh interesting - yeah I guess it would depend on how exactly they define "prevent" and whether it will fall under replacement effects or not.

But for now the `AbilityRestriction` stuff is designed so that everything should still be targetable, though there may be some bugs there that need worked out.

AMMayberry1: Followup on this from something we hit on another PR - I forgot there is an optimization that can cause the targeting stage to be passed intentionally. Basically if there are no legal targets for which the effect would change game state - e.g., if you're trying to ready but all legal targets are already readied, or healing and all targets have 0 damage - then we just bypass the targeting phase since the selection will end up being irrelevant to game state: https://github.com/SWU-Karabast/forceteki/blob/fd146c3cc10612b7bfc21472b613fe49c8f993ac/server/game/core/ability/abilityTargets/CardTargetResolver.ts#L111

We have a lot of other targeting optimizations like that used in prod but turned off for unit tests so we can validate targeting a little easier, it might make sense to do the same with this one since it could result in unexpected behaviors while writing tests. But just wanted to give you a heads-up that it's there for now.

kleer1: Was able to reimplement this as constant cannot abilities. Still need to add the extra tests you mentioned elsewhere.

test/server/cards/02_SHD/units/LurkingTIEPhantom.spec.ts
kleer1: added

kleer1: Done

server/game/cards/02_SHD/units/LurkingTIEPhantom.ts
```diff
@@ -10,17 +12,28 @@ export default class LurkingTIEPhantom extends NonLeaderUnitCard {
...
+                restrictedActionCondition: (context) => !(context.ability instanceof InitiateAttackAction) && context.ability.card.controller !== this.controller,
+            })
         });
-    }
-}
 
-LurkingTIEPhantom.implemented = true;
+        this.addConstantAbility({
+            title: 'This unit can\'t be captured, damaged, or defeated by enemy card abilities',
+            ongoingEffect: AbilityHelper.ongoingEffects.cardCannot({
```
AMMayberry1, line 25: nit: `ongoingEffect` can be an array, could we consolidate all of these effects into one ability just for conciseness of the code and the properties on the object when debugging?

kleer1: fixed

AMMayberry1 (approved):

PR #479, opened 2025-01-29
Mister Bones

VJubert:

### DIFF
--- a/server/game/cards/03_TWI/units/MisterBonesIPerformedViolence.ts
+++ b/server/game/cards/03_TWI/units/MisterBonesIPerformedViolence.ts
@@ -1,0 +1,26 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { ZoneName } from '../../../core/Constants';
+
+export default class MisterBonesIPerformedViolence extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '7789777396',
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/units/MisterBonesIPerformedViolence.ts
```diff
@@ -0,0 +1,26 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '7789777396',
+            internalName: 'mister-bones#i-performed-violence',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addOnAttackAbility({
+            title: 'Deal 3 damage to a ground unit',
```
AMMayberry1, line 15: Missing `optional: true`

test/server/cards/03_TWI/units/MisterBonesIPerformedViolence.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+            context.player1.clickCard(context.misterBones);
+            context.player1.clickCard(context.p2Base);
+
+            expect(context.p2Base.damage).toBe(3);
+            expect(context.battlefieldMarine.damage).toBe(0);
+            expect(context.player2).toBeActivePlayer();
+        });
+
+        it('Mister Bones\'s ability should deal 3 damage to a ground unit if there is no card in controller\'s hand', function () {
```
AMMayberry1, line 25: Should be able to join these cases together pretty easily - can just have player1 play out the card and then set misterBones.exhausted = false for second attack

AMMayberry1 (approved):

PR #480, opened 2025-01-29
Implementing Geonosis Patrol Fighter and Pelta Supply Frigate

aruizc:
image
image


test/server/cards/03_TWI/units/GeonosisPatrolFighter.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+                    spaceArena: ['restored-arc170'],
+                    hand: ['waylay']
+                }
+            });
+
+            const { context } = contextRef;
+
+            // Returns Ground unit that costs 3 or less
+            context.player1.clickCard(context.geonosisPatrolFighter);
+            context.player1.clickPrompt('Play Geonosis Patrol Fighter');
```
AMMayberry1, line 21: Could we add another case where we play using Exploit and confirm that the unit we choose for Exploit is not targetable with the ability. If you find that it is targetable, then we need to set `cannotTargetFirst: true` on the ability to prevent targeting from happening until after costs have been paid. You can see Exploit.spec.ts for examples of how to trigger Exploit in tests.

If you don't want to manage multiple Waylays you can also use `context.player1.moveCard()`, though it is _slightly_ riskier that it might break game state. That said, we use it in many places in the tests and it seems fine so far.

test/server/cards/03_TWI/units/GeonosisPatrolFighter.spec.ts
```diff
@@ -0,0 +1,38 @@
+describe('Geonosis Patrol Fighter', function() {
+    integration(function(contextRef) {
+        it('Geonosis Patrol Fighter\'s when played ability should return a non-leader unit that costs 3 or less to its owner\'s hand', function () {
+            contextRef.setupTest({
+                phase: 'action',
+                player1: {
+                    hand: ['geonosis-patrol-fighter'],
+                    groundArena: ['r2d2#ignoring-protocol']
```
AMMayberry1, line 8: We need a set of units that exactly confirms the filter's behavior. So it would be good to have a ground unit that costs 3 and a ground unit that costs 4.

AMMayberry1 (approved):

PR #482, opened 2025-01-29
Darth Vader

VJubert:


test/server/cards/01_SOR/units/DarthVaderCommandingTheFirstLegion.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+            context.player1.clickPrompt('Play cards in selection order');
+
+            // Salacious Crumb ability triggers and heal 1 damage from base
+            expect(context.p1Base.damage).toBe(4);
+
+            // outland tie vanguard should give an experience token to a unit that costs 3 resources or less
+            expect(context.player1).toBeAbleToSelectExactly([context.salaciousCrumb]);
+            context.player1.clickCard(context.salaciousCrumb);
+            expect(context.salaciousCrumb).toHaveExactUpgradeNames(['experience']);
+        });
```
AMMayberry1, line 155: Fantastic! There's one other test I wanted to do but didn't get done with U-Wing. Specifically, we need to confirm that the searched cards go to the bottom of the deck before any of the play actions happen.

The way to test this would be to deploy an Inferno Four as part of Vader's ability and confirm that it sees the top two cards of the deck after all 10 of the search cards have been removed. I have some code in place for this but it is not very well tested so you might need to do some light debugging, happy to give some pointers on where to look if that happens.

test/server/cards/01_SOR/units/DarthVaderCommandingTheFirstLegion.spec.ts
```diff
@@ -153,5 +153,56 @@ describe('Darth Vader, Commanding the First Legion', function () {
...
+
+            context.player1.clickCard(context.darthVader);
+            context.player1.clickPrompt('Ambush');
+            expect(context.player1).toHavePrompt('Choose any units with combined cost 3 or less to play for free');
+
+            // play inferno four
+            context.player1.clickCardInDisplayCardPrompt(context.infernoFour);
+            context.player1.clickPrompt('Play cards in selection order');
+
+            // should show the next 2 cards
```
AMMayberry1, line 191: Could we also add a 1-cost unit such as Tie/LN which selected to be played after Inferno Four. That way when Inferno is resolving we can confirm that Tie/LN is not in the deck at all.

Also would generally be good to confirm the state of the whole deck during this prompt. Specifically should mean:

- The cards which were searched by Vader and returned are in the bottom of the deck in random order
- The cards which were not search by Vader or Inferno Four are now at the top of the deck

AMMayberry1 (approved):

PR #487, opened 2025-01-30
Implement Second Chance (SHD 053)

moyerr:

### DIFF
--- a/server/game/cards/02_SHD/upgrades/SecondChance.ts
+++ b/server/game/cards/02_SHD/upgrades/SecondChance.ts
@@ -10,16 +10,14 @@ export default class SecondChance extends UpgradeCard {
         };
     }
 
-    protected override setupCardAbilities() {
+    public override setupCardAbilities() {
         this.setAttachCondition((card) => card.isNonLeaderUnit());
         this.addGainWhenDefeatedAbilityTargetingAttached({
             title: 'For this phase, this unit\'s owner may play it from their discard pile for free.',
             immediateEffect: AbilityHelper.immediateEffects.forThisPhaseCardEffect((context) => ({
                 effect: [
                     OngoingEffectLibrary.canPlayFromDiscard(),
-                    OngoingEffectLibrary.forFree({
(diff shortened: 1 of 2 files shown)


server/game/cards/02_SHD/upgrades/SecondChance.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card) => card.isNonLeaderUnit());
+        this.addGainWhenDefeatedAbilityTargetingAttached({
+            title: 'For this phase, this unit\'s owner may play it from their discard pile for free.',
+            immediateEffect: AbilityHelper.immediateEffects.forThisPhaseCardEffect((context) => ({
+                effect: [
+                    OngoingEffectLibrary.canPlayFromDiscard(),
+                    OngoingEffectLibrary.forFree({ match: (card) => card === context.source })
```
AMMayberry1, line 20: This is a little bit of an unusual case in that we have to split up the "can play from discard" and "play for free" effects. In the past (e.g. Palpatine's Return) we've been able to combine them since the play effect is an immediate effect instead of an ongoing effect, but for the latter we don't yet have support for directly attaching a cost adjuster.

All that to say, I'm a little worried that since these two effects are disconnected that we might run into weird edge cases where the `forFree` part continues to apply even when it shouldn't. Effectively we have two different things here:

1. "You may play this card from discard"
2. "The next time you play this card, it is played for free"

And I just want to make sure everything is set up correctly so that (2) can't trigger in a situation apart from (1). To check this, could we add a test where the Second Chance'd unit is pulled back to hand with e.g. The Emperor's Legion? Then we need to confirm that if that same unit is played from hand, the `forFree` part has expired and we paying the cost for it as expected.

If we do find that there are some issues please let me know, it may be the case that we either need to add some extra conditions on the `forFree()` part or that we need to expand `canPlayFromDiscard()` to support attaching cost adjusters.

AMMayberry1 (approved):

PR #492, opened 2025-01-31
Implementing San Hill, Chairman of the Banking Clan

supercodepoet:
image

### DIFF
--- a/server/game/cards/03_TWI/units/SanHillChairmanOfTheBankingClan.ts
+++ b/server/game/cards/03_TWI/units/SanHillChairmanOfTheBankingClan.ts
@@ -1,0 +1,29 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import type { StateWatcherRegistrar } from '../../../core/stateWatcher/StateWatcherRegistrar';
+import type { UnitsDefeatedThisPhaseWatcher } from '../../../stateWatchers/UnitsDefeatedThisPhaseWatcher';
+
+export default class SanHillChairmanOfTheBankingClan extends NonLeaderUnitCard {
+    private unitsDefeatedThisPhaseWatcher: UnitsDefeatedThisPhaseWatcher;
+
+    protected override getImplementationId () {
+        return {
+            id: '0021045666',
+            internalName: 'san-hill#chairman-of-the-banking-clan'
+        };
+    }
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar): void {
+        this.unitsDefeatedThisPhaseWatcher = AbilityHelper.stateWatchers.unitsDefeatedThisPhase(registrar, this);
+    }
+
+    public override setupCardAbilities () {
+        this.addOnAttackAbility({
+            title: 'For each friendly unit that was defeated this phase, ready a friendly resource.',
+            immediateEffect: AbilityHelper.immediateEffects.readyResources((context) => ({
+                target: context.source.controller,
+                amount: this.unitsDefeatedThisPhaseWatcher.getDefeatedUnitsControlledByPlayer(context.source.controller).length
+            }))
+        });
+    }
+}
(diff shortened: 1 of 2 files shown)


test/server/cards/03_TWI/units/SanHillChairmanOfTheBankingClan.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+                // Syndicate Lackeys should be defeated by attacking Wampa
+                context.player1.clickCard(context.syndicateLackeys);
+                context.player1.clickPrompt('Ambush');
+                context.player1.clickCard(context.battlefieldMarine);
+
+                expect(context.player1.readyResourceCount).toBe(5);
+                expect(context.player1.exhaustedResourceCount).toBe(5);
+
+                // B1 Attack Platform should be defeated by Fleet Lieutenant
```
AMMayberry1, line 60: nit: comments are incorrect

test/server/cards/03_TWI/units/SanHillChairmanOfTheBankingClan.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+                // San Hill should ready 3 resources
+                context.player1.clickCard(context.sanHillChairmanOfTheBankingClan);
+                context.player1.clickCard(context.p2Base);
+
+                expect(context.player1.readyResourceCount).toBe(8);
+                expect(context.player1.exhaustedResourceCount).toBe(2);
+            });
+
+            it('should not ready resources since no friendly units were defeated this phase', function () {
```
AMMayberry1, line 49: We can combine this into the previous test by using `context.moveToNextActionPhase()`

server/game/cards/03_TWI/units/SanHillChairmanOfTheBankingClan.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar): void {
+        this.unitsDefeatedThisPhaseWatcher = AbilityHelper.stateWatchers.unitsDefeatedThisPhase(registrar, this);
+    }
+
+    public override setupCardAbilities () {
+        this.addOnAttackAbility({
+            title: 'For each friendly unit that was defeated this phase, ready a friendly resource.',
+            immediateEffect: AbilityHelper.immediateEffects.readyResources((context) => ({
+                target: context.source.controller,
```
AMMayberry1, line 24: nit: because of some potential issues we've found in other branches, we need to start using `context.player` instead of `context.source.controller` in all cases unless there's a specific reason to use the latter.

The difference is that `context.player` is the triggering player for the ability, whereas `context.source.controller` is the current controller of the card. It specifically matters when a stolen unit has an on-defeat ability, but I want to start using it generally so everyone is in the habit.

supercodepoet: 👍

test/server/cards/03_TWI/units/SanHillChairmanOfTheBankingClan.spec.ts
supercodepoet: removed this test and combined it in the original test

test/server/cards/03_TWI/units/SanHillChairmanOfTheBankingClan.spec.ts
supercodepoet: done

test/server/cards/03_TWI/units/SanHillChairmanOfTheBankingClan.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+                expect(context.player1.readyResourceCount).toBe(8);
+                expect(context.player1.exhaustedResourceCount).toBe(2);
+
+                // San Hill should not ready and resources since no friendly units were defeated this phase
+                context.moveToNextActionPhase();
+
+                expect(context.player1.readyResourceCount).toBe(10);
+                expect(context.player1.exhaustedResourceCount).toBe(0);
+
+                context.player1.clickCard(context.sanHillChairmanOfTheBankingClan);
```
AMMayberry1, line 54: I think we need to exhaust some resources first for this test to be relevant?

supercodepoet: updated

AMMayberry1 (approved):

PR #506, opened 2025-02-01
Add optional prompt support for "basic" lookat / reveal case

AMMayberry1:
Beginning the process of refactoring for every lookat / reveal effect to use the display prompt instead of relying on text messages. Since there are a lot of cards that need to be updated and many of them are special cases, we are adding an optional property that enables the prompt in the `lookAt()` and `reveal()` functions. This will be removed once all cards are ported over.

Converted VPD and Childsen to use the new prompt system.

### DIFF
--- a/server/Util.ts
+++ b/server/Util.ts
@@ -28,6 +28,27 @@ export function httpRequest(url: string): Promise<string> {
     });
 }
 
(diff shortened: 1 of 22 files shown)


danbastin:
Tried this out on the FE. LGTM

PR #507, opened 2025-02-01
Implementing Unnatural Life

supercodepoet:
image


test/server/cards/03_TWI/events/UnnaturalLife.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+            it('should play Zuckuss for 2 less and enter play ready, since he was defeated this phase', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.fleetLieutenant);
+                context.player1.clickCard(context.zuckuss);
+                expect(context.zuckuss).toBeInZone('discard');
+
+                context.player2.clickCard(context.unnaturalLife);
+                expect(context.player2).toHavePrompt('Choose a unit');
+                expect(context.player2).toBeAbleToSelectExactly([context.zuckuss]);
```
AMMayberry1, line 31: Can we also defeat an opponent unit this turn and confirm it isn't selectable

test/server/cards/03_TWI/events/UnnaturalLife.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+                context.player2.clickCard(context.zuckuss);
+                expect(context.zuckuss.exhausted).toBeFalse();
+                expect(context.player2.exhaustedResourceCount).toBe(6);
+                expect(context.player2.readyResourceCount).toBe(1);
+
+                // Check that Zuckuss is defeated at the beginning of the regroup phase
+                context.moveToNextActionPhase();
```
AMMayberry1, line 39: Since he's defeated at regroup, can we instead do `context.moveToRegroupPhase()` and confirm he's defeated at that point. Then we can just have both players click "Done" to skip resourcing and move to next action phase

test/server/cards/03_TWI/events/UnnaturalLife.spec.ts
supercodepoet: ok, I added a check to make sure Fleet was defeated and in the discard zone.

test/server/cards/03_TWI/events/UnnaturalLife.spec.ts
supercodepoet: yep, updated

AMMayberry1 (approved):

PR #509, opened 2025-02-02
Consolidation of power

ssoulier:
[Consolidation of Power](https://swudb.com/card/TWI/089/consolidation-of-power)

Screenshot 2025-02-02 at 09 55 14


server/game/cards/03_TWI/events/ConsolidationOfPower.ts
```diff
@@ -0,0 +1,47 @@
...
+            id: '4895747419',
+            internalName: 'consolidation-of-power',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Choose any number of friendly units. You may play a unit from your hand if its cost is less than or equal to the combined power of the chosen units for free. Then, defeat the chosen units.',
+            targetResolver: {
+                mode: TargetMode.UpToVariable,
```
AMMayberry1, line 18: We haven't used it much so it might be buggy, but I think we should try to use `TargetMode.Unlimited` here since that's more intuitive than having a limit set equal to all possible targets

server/game/cards/03_TWI/events/ConsolidationOfPower.ts
```diff
@@ -0,0 +1,47 @@
...
+                controller: RelativePlayer.Self,
+            },
+            then: (firstThenContext) => ({
+                title: 'You may play a unit from your hand if its cost is less than or equal to the combined power of the chosen units for free.',
+                targetResolver: {
+                    cardTypeFilter: WildcardCardType.Unit,
+                    optional: true,
+                    mode: TargetMode.Single,
+                    controller: RelativePlayer.Self,
+                    cardCondition: (_, context) => context.target.cost <= firstThenContext.target.reduce((sum, card) => sum + card.getPower(), 0),
```
AMMayberry1, line 32: nit: I'm trying to move us to a convention of, instead of just using `_`, prefixing it to the variable name like `_card`

test/server/cards/03_TWI/events/ConsolidationOfPower.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+            context.player1.clickPrompt('Done');
+            expect(context.player1).toBeAbleToSelectExactly([
+                context.academyDefenseWalker,
+                context.homeOneAllianceFlagship,
+                context.savageOpressMonster,
+            ]);
+            expect(context.player1).toHaveChooseNoTargetButton();
+            context.player1.clickCard(context.homeOneAllianceFlagship);
+            expect(context.homeOneAllianceFlagship).toBeInZone('spaceArena');
+            expect(context.battlefieldMarine).toBeInZone('discard');
```
AMMayberry1, line 76: nit: for checking groups of cards like this, can also do `expect([context.battlefieldMarine, ...]).toAllBeInZone('discard');`

test/server/cards/03_TWI/events/ConsolidationOfPower.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+            expect(context.theDarksaber).toBeInZone('discard');
+            expect(context.allianceXwing).toBeInZone('discard');
+            expect(context.reinforcementWalker).toBeInZone('groundArena');
+            expect(context.bountyHunterCrew).toBeInZone('groundArena');
+            expect(context.finalizerMightOfTheFirstOrder).toBeInZone('hand');
+            expect(context.academyDefenseWalker).toBeInZone('hand');
+            expect(context.savageOpressMonster).toBeInZone('hand');
+            // With Home One we can play a unit for that we discarded with Consolidation of power
```
AMMayberry1, line 84: Oohh good test!

test/server/cards/03_TWI/events/ConsolidationOfPower.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+            expect(context.player1).toBeAbleToSelectExactly([
+                context.reinforcementWalker,
+                context.battlefieldMarine,
+                context.allianceXwing,
+                context.bountyHunterCrew,
+            ]);
+            context.player1.clickCard(context.battlefieldMarine);
+            context.player1.clickCard(context.allianceXwing);
+            context.player1.clickPrompt('Done');
+            expect(context.player1).toBeAbleToSelectExactly([
```
AMMayberry1, line 68: If I'm doing my math right, the total of battlefield marine and alliance x-wing's power is 9. Could we get something that costs 9 and costs 10 in here just to confirm the filter cutoff exactly?

test/server/cards/03_TWI/events/ConsolidationOfPower.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+            expect(context.finalizerMightOfTheFirstOrder).toBeInZone('hand');
+            expect(context.academyDefenseWalker).toBeInZone('hand');
+            expect(context.savageOpressMonster).toBeInZone('hand');
+            // With Home One we can play a unit for that we discarded with Consolidation of power
+            expect(context.player1).toBeAbleToSelectExactly([
+                context.battlefieldMarine,
+                context.allianceXwing,
+            ]);
+            context.player1.clickCard(context.battlefieldMarine);
+            expect(context.player1.readyResourceCount).toBe(1);
```
AMMayberry1, line 90: There's kind of a silly edge case that I'd like to validate nonetheless - if the player doesn't play out a unit, the targets are still defeated.

Could we confirm the following two cases:
1. Player has legal play target(s) and chooses not to play them, target arena units are still defeated
2. Player has no legal play targets, the prompt is skipped completely and the target arena units are defeated

These are sufficiently different from the above setup that I think we'd be justified in having a new case and setup for them, up to your discretion though

server/game/cards/03_TWI/events/ConsolidationOfPower.ts
```diff
@@ -0,0 +1,47 @@
...
+                    immediateEffect: AbilityHelper.immediateEffects.playCardFromHand({
+                        adjustCost: { costAdjustType: CostAdjustType.Free },
+                    }),
+                },
+                then: (secondThenContext) => ({
+                    title: 'Then, defeat the chosen units.',
+                    immediateEffect: AbilityHelper.immediateEffects.defeat({
+                        target: firstThenContext.target,
+                    })
+                })
```
AMMayberry1, line 43: Awesome job putting this all together! Other than the TargetMode note above this looks completely correct

server/game/cards/03_TWI/events/ConsolidationOfPower.ts
ssoulier: I tried first with `Unlimited` but it does not allow to select None and  it does not support `optional: true` or `canChooseNoCard`. I think it shares the same interface with the `Single` mode.

If you think we should make Unlimited support 0 card selected I am happy to try a refactoring.

AMMayberry1: Hmm, so we used `Unlimited` for Power Failure and from the tests it looks like that does present a pass button: https://github.com/SWU-Karabast/forceteki/blob/99479127ec31d729dc0885e374169f299829195b/test/server/cards/01_SOR/events/PowerFailure.spec.ts#L26

Are you seeing different behavior?

ssoulier: There is an optional flag outside the `targetResolver`: https://github.com/SWU-Karabast/forceteki/blob/99479127ec31d729dc0885e374169f299829195b/server/game/cards/01_SOR/events/PowerFailure.ts#L13-L30

I think it makes all the difference with what I tried (I tried to put optional in the target resolver, and in that case unlimited does not allow to select None).
I'll see if I can achieve a similar implementation than the one of PowerFailure

AMMayberry1: Ah I see, yeah that's somewhat unintuitive. Probably at some point we need to add a specific option to `Unlimited` to control whether you have to choose at least one.

ssoulier: It's working following PowerFailure pattern. I'll update the PR

server/game/cards/03_TWI/events/ConsolidationOfPower.ts
ssoulier: It's in fact not working if you have no valid target in the first `then`. You do not enter the second then and the units are not discarded.
One solution can be to do it the other way. We defeat the units first (this is not optional) and then you can select the unit you want to play (assuming there at least one legit target).

It's not exactly the order of the events in the card but in that specific example I am not sure it matters.
If you think It does I am happy to have a pointer on how to achieve things in the right order. I tried ifYouDo,  IfYouDont, but we can use them in the `then` section.

ssoulier: Forget the above. The can't do it in the wrong order.

ssoulier: I had to update the logic. The nested `then` did not work as expected. The new version works well.

AMMayberry1 (approved):

PR #510, opened 2025-02-02
Bunch of cards 18

VJubert:
- Fives ([TWI 216](https://swudb.com/card/TWI/216/fives))
- Republic Tactical Officer ([TWI 91](https://swudb.com/card/twi/091/republic-tactical-officer))
- Shadowed Intentions ([TWI 220](https://swudb.com/card/twi/220/shadowed-intentions))
- Strategic Acumen ([TWI 120](https://swudb.com/card/twi/120/strategic-acumen))


server/game/cards/03_TWI/upgrades/ShadowedIntentions.ts
```diff
@@ -0,0 +1,32 @@
...
+        };
+    }
+
+    protected override setupCardAbilities() {
+        this.addConstantAbilityTargetingAttached({
+            title: 'This unit can\'t be captured, defeated, or returned to its owner\'s hand by enemy card abilities',
+            ongoingEffect: [
+                AbilityHelper.ongoingEffects.cardCannot({
+                    cannot: AbilityRestriction.BeCaptured,
+                    restrictedActionCondition: (context) => context.ability.card.controller !== this.controller,
```
AMMayberry1, line 19: Hmm, I just realized there's a bug in this approach that we used with Lurking TIE Phantom. See this ruling from the judge clarifications:

image

So I think right now LTP _would_ take damage from a friendly Val bounty and it should not. Obviously that exact example isn't relevant for this card since this doesn't prevent damage, but I think we need to be using the right pattern so that it doesn't become a bug later if they add a card with 'Bounty: Defeat a unit.'

Would you mind fixing the issue in LTP, adding the Val test, and then doing the fix here as well? I think we need to check `context.player` instead of `context.ability.card.controller` but not 100% sure.

test/server/cards/03_TWI/upgrades/StrategicAcumen.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+                context.player1.clickPrompt('Play a unit from your hand. It costs 1 less');
+                expect(context.player1).toBeAbleToSelectExactly([context.consortiumStarviper]);
+                expect(context.player1).toHaveChooseNoTargetButton();
+                context.player1.clickCard(context.consortiumStarviper);
+                expect(context.consortiumStarviper).toBeInZone('spaceArena');
+                expect(context.player1.exhaustedResourceCount).toBe(2);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should not give the next unit played by the controller a discount after the controller declines to play a unit with the ability', function() {
```
AMMayberry1, line 50: nit: can we merge this into the above test

server/game/core/ability/PlayerOrCardAbility.js
```diff
@@ -346,6 +346,10 @@ class PlayerOrCardAbility {
     isAttackAction() {
         return false;
     }
+
+    controller() {
```
AMMayberry1, line 350: This is great! Couple small nits.

First is - could we add an explicit return type annotation, that can be done in JS using a jsdoc `@returns` annotation like above. That will help disambiguate it from `abilityController` and make it easier to use in TS code.

The other nit is that our convention would be to name this one of two ways:

option 1: property getter
```ts
get controller() {
    return this.abilityController === RelativePlayer.Self ? this.card.controller : this.card.controller.opponent;
}

// used as ability.controller
```

option 2 - getter method
```ts
getController() {
    return this.abilityController === RelativePlayer.Self ? this.card.controller : this.card.controller.opponent;
}

// used as ability.getController()
```

AMMayberry1 (approved):

PR #519, opened 2025-02-03
Qira

VJubert:


server/game/cards/02_SHD/units/QiraPlayingHerPart.ts
```diff
@@ -0,0 +1,41 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { Duration, TargetMode } from '../../../core/Constants';
+import * as AbilityLimit from '../../../core/ability/AbilityLimit';
+
+export default class QiraPlayingHerPart extends NonLeaderUnitCard {
```
AMMayberry1, line 6: Can you do me a small favor and add this line here:

```ts
```

Since the lookat / reveal prompt is not yet working in the FE, we are manually marking this card not implemented so people who are doing QA know not to expect it to work right now

test/server/cards/02_SHD/units/QiraPlayingHerPart.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+            // steal qira
+            context.player2.clickCard(context.changeOfHeart);
+            context.player2.clickCard(context.qira);
+
+            // player 1 marine should cost only 2
+            context.player1.clickCard(marine1);
+            expect(context.player1.exhaustedResourceCount).toBe(2);
+
+            // player 2 marine should still cost 5
+            context.player2.clickCard(marine2);
```
AMMayberry1, line 70: Oh nice! I had never thought about this case before, glad to see it works right out of the box

AMMayberry1 (approved):

PR #520, opened 2025-02-03
Add Cad Bane unit implementation

Frugghi:
image


server/game/cards/03_TWI/units/CadBaneHostageTaker.ts
```diff
@@ -0,0 +1,45 @@
...
+                controller: RelativePlayer.Opponent,
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                multiSelectCardCondition: (card, selectedCards) => card.isUnit() && selectedCards.reduce((totalHp, selectedCard) => totalHp + (selectedCard as UnitCard).remainingHp, card.remainingHp) <= 8,
+                immediateEffect: AbilityHelper.immediateEffects.capture()
+            }
+        });
+        this.addOnAttackAbility({
+            title: 'Rescue a card you own captured by Cad Bane and the opponent draws 2 cards',
+            optional: true,
+            playerChoosingOptional: RelativePlayer.Opponent,
```
AMMayberry1, line 31: Could let me know if it works without this? I _think_ since the target resolver has `choosingPlayer` set it will work, but I'm actually not totally sure

server/game/cards/03_TWI/units/CadBaneHostageTaker.ts
```diff
@@ -0,0 +1,45 @@
...
+                mode: TargetMode.UpTo,
+                canChooseNoCards: true,
+                numCards: 3,
+                controller: RelativePlayer.Opponent,
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                multiSelectCardCondition: (card, selectedCards) => card.isUnit() && selectedCards.reduce((totalHp, selectedCard) => totalHp + (selectedCard as UnitCard).remainingHp, card.remainingHp) <= 8,
+                immediateEffect: AbilityHelper.immediateEffects.capture()
+            }
+        });
+        this.addOnAttackAbility({
```
AMMayberry1, line 28: nit: can we put a newline between the abilities for readability

test/server/cards/03_TWI/units/CadBaneHostageTaker.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+            expect(context.player1.handSize).toBe(0);
+            expect(context.player2).toHavePassAbilityPrompt('Rescue a card you own captured by Cad Bane and the opponent draws 2 cards');
+            context.player2.clickPrompt('Rescue a card you own captured by Cad Bane and the opponent draws 2 cards');
+
+            // Player 2 can't rescue the Battlefield Marine because they don't own it
+            expect(context.player2).toBeAbleToSelectExactly([context.wampa, context.wingLeader]);
+            context.player2.clickCard(context.wampa);
+
+            expect(context.wampa).toBeInZone('groundArena', context.player2);
+            expect(context.player1.handSize).toBe(2);
```
AMMayberry1, line 69: Can we also do an attack where Cad has no captured units just to confirm nothing breaks. Can be a separate test setup since it would be kind of difficult to fit that in here

server/game/cards/03_TWI/units/CadBaneHostageTaker.ts
```diff
@@ -0,0 +1,45 @@
+import AbilityHelper from '../../../AbilityHelper';
+import type { UnitCard } from '../../../core/card/CardTypes';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { RelativePlayer, TargetMode, WildcardCardType, ZoneName } from '../../../core/Constants';
+
+export default class CadBaneHostageTaker extends NonLeaderUnitCard {
```
AMMayberry1, line 6: Can you do me a small favor and add this line here:

```ts
```

Capture isn't yet rendering in the FE, so we are manually marking this card not implemented so people who are doing QA know not to expect it to work right now

server/game/cards/03_TWI/units/CadBaneHostageTaker.ts
Frugghi: No, it won't work.

AMMayberry1 (approved):

PR #523, opened 2025-02-04
I Have The High Ground

Twet1:
image


test/server/cards/03_TWI/events/IHaveTheHighGround.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+                },
+                player2: {
+                    groundArena: ['wampa', 'wrecker#boom', 'clone-trooper']
+                }
+            });
+
+            const { context } = contextRef;
+
+            // Apply event
+            context.player1.clickCard(context.iHaveTheHighGround);
```
AMMayberry1, line 18: Need to check `.toBeAbleToSelectExactly()`

test/server/cards/03_TWI/events/IHaveTheHighGround.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            expect(context.ploKoon.damage).toBe(3);
+            expect(context.wrecker.damage).toBe(3);
+
+            context.player1.passAction();
+
+            // Another unit attacked, to ensure -4 isnt applied globally
+            context.player2.clickCard(context.cloneTrooper);
+            context.player2.clickCard(context.battlefieldMarine);
+            expect(context.cloneTrooper).toBeInZone('outsideTheGame');
+            expect(context.battlefieldMarine.damage).toBe(2);
```
AMMayberry1, line 43: We should also do an attack with Plo Koon to make sure the effect doesn't apply in that case

test/server/cards/03_TWI/events/IHaveTheHighGround.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            context.player1.passAction();
+
+            // 2nd unit attacked, -4 applied
+            context.player2.clickCard(context.wrecker);
+            context.player2.clickCard(context.ploKoon);
+            expect(context.ploKoon).toBeInZone('groundArena');
+            expect(context.ploKoon.damage).toBe(3);
+            expect(context.wrecker.damage).toBe(3);
+
+            context.player1.passAction();
```
AMMayberry1, line 37: Let's play a unit from the opponent's hand and then attack with (Ambush is fine here) to ensure it still gets the effect, that was a specific rules thing that came up in judge chat

PR #524, opened 2025-02-04
#CaptureZonePile

CheBato:
Added the CapturedUnitsZone to the summary with an array of all capturedUnits + added parentCardId to the unitProperties summary.


server/game/core/Player.js
```diff
@@ -266,6 +266,8 @@ class Player extends GameObject {
                 return this.game.spaceArena.getCards({ controller: this });
             case ZoneName.GroundArena:
                 return this.game.groundArena.getCards({ controller: this });
+            case ZoneName.Capture:
```
AMMayberry1, line 269: To be consistent with the other entries, we would need to filter down to only captured cards owned by this player instead of all captured cards in the game

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -656,10 +656,13 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
...
                     power: this.getPower(),
                     hp: this.getHp(),
-                    sentinel: hasSentinel
+                    sentinel: hasSentinel,
                 };
             }
-            return super.getSummary(activePlayer);
+            return {
+                ...super.getSummary(activePlayer),
+                parentCardId: this.getCaptor() ? this.getCaptor().uuid : null,
```
AMMayberry1, line 664: nit: simpler as `this.getCaptor()?.uuid`

server/game/core/Player.js
CheBato: yeah i just noticed on the FE this exact thing I'll fix it up.

AMMayberry1 (approved):

PR #527, opened 2025-02-04
Implementing Guerilla Insurgency

aruizc:
image

### DIFF
--- a/test/server/cards/03_TWI/events/GuerillaInsurgency.spec.ts
+++ b/test/server/cards/03_TWI/events/GuerillaInsurgency.spec.ts
@@ -1,0 +1,182 @@
+describe('GuerillaI nsurgency', function() {
+    integration(function(contextRef) {
+        describe('Guerilla Insurgency\'s ability -', function() {
+            it('should defeat a friendly and enemy resource, make players to discard 2 cards and deal damage 4 to each ground unit', function() {
+                contextRef.setupTest({
+                    phase: 'action',
+                    player1: {
+                        hand: ['guerilla-insurgency', 'grim-resolve', 'in-pursuit'],
+                        groundArena: ['wampa'],
+                        resources: ['cunning', 'aggression', 'spark-of-rebellion', 'protector', 'atst', 'battlefield-marine', 'tieln-fighter', 'death-trooper', 'blood-sport', 'drop-in', 'hello-there'],
+                    },
+                    player2: {
+                        hand: ['pillage', 'force-choke'],
+                        groundArena: ['seventh-sister#implacable-inquisitor'],
+                        spaceArena: ['corellian-freighter'],
+                        resources: ['confiscate', 'restock', 'breaking-in']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                // Assert initial state
+                expect(context.player1.resources.length).toBe(11);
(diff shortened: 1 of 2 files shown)


test/server/cards/03_TWI/events/GuerillaInsurgency.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+                expect(context.player2).toBeAbleToSelectExactly([context.confiscate, context.restock, context.breakingIn]);
+                expect(context.player1).not.toHaveChooseNoTargetButton();
+                context.player2.clickCard(context.confiscate);
+
+                // Each player discards 2 cards
+                expect(context.player1).toHavePrompt('Choose 2 cards to discard');
+                expect(context.player1).toBeAbleToSelectExactly(['grim-resolve', 'in-pursuit']);
+                context.player1.clickCard('grim-resolve');
+                context.player1.clickCard('in-pursuit');
+                context.player1.clickPrompt('Done');
```
AMMayberry1, line 46: Can we confirm the limit by adding a third card in hand for one player, clicking it with `clickCardNonChecking()`, and confirming that it isn't discarded

test/server/cards/03_TWI/events/GuerillaInsurgency.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+
+                expect(context.player2).toHavePrompt('Defeat a resource you control');
+                expect(context.player2).toBeAbleToSelectExactly([context.confiscate, context.restock, context.breakingIn]);
+                expect(context.player1).not.toHaveChooseNoTargetButton();
+                context.player2.clickCard(context.confiscate);
+
+                // Each player discards 2 cards
+                expect(context.player1).toHavePrompt('Choose 2 cards to discard');
+                expect(context.player1).toBeAbleToSelectExactly(['grim-resolve', 'in-pursuit']);
+                context.player1.clickCard('grim-resolve');
```
AMMayberry1, line 44: Let's check here that there is _not_ a done button available since not enough targets are selected yet

test/server/cards/03_TWI/events/GuerillaInsurgency.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+                expect(context.player2.hand.length).toBe(0);
+                expect(context.player1.hand.length).toBe(0);
+
+                // Assert dealt damage
+                expect(context.wampa.damage).toBe(4);
+                expect(context.seventhSister.damage).toBe(4);
+                expect(context.corellianFreighter.damage).toBe(0); // Space unit should not be affected
+            });
+
+            it('should defeat a friendly and enemy resource, only opponent discards a card and deal damage 4 to each ground unit', function() {
```
AMMayberry1, line 71: Are these tests just checking whether discard works if there aren't enough cards in hand? If so I think we're okay without them, we have a good number of tests of the discard system already. Trying to keep the number of test cases down

aruizc: Yes, basically, it's to validate that the next effect happens even if no card was discarded. I've removed those tests as you mentioned there's a bunch of test for sequential effects.

AMMayberry1 (approved):

PR #528, opened 2025-02-05
Add Chancellor Palpatine and double-sided leader support

adbronico:
image
image


server/game/cards/03_TWI/leaders/ChancellorPalpatinePlayingBothSides.ts
```diff
@@ -0,0 +1,69 @@
...
+        this.addActionAbility({
+            title: 'If a friendly Heroism unit was defeated this phase, draw a card, heal 2 damage from your base, then flip this leader.',
+            cost: AbilityHelper.costs.exhaustSelf(),
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => this.friendlyHeroismCardDefeatedThisPhase(context),
+                onFalse: AbilityHelper.immediateEffects.noAction(),
+                onTrue: AbilityHelper.immediateEffects.sequential((context) => ([
+                    AbilityHelper.immediateEffects.draw(),
+                    AbilityHelper.immediateEffects.heal({ target: context.player.base, amount: 2 }),
+                    AbilityHelper.immediateEffects.flipDoubleSidedLeader({ target: context.source })
```
AMMayberry1, line 33: nit: `context.source` is the default target so don't need to explicitly provide it, same for other side ability

server/game/core/card/Card.ts
```diff
@@ -40,11 +41,13 @@ export type CardConstructor = new (...args: any[]) => Card;
...
-    public readonly subtitle?: string;
-    public readonly title: string;
-    public readonly unique: boolean;
+    protected _aspects: Aspect[] = [];
+    protected _backSideAspects?: Aspect[];
+    protected _internalName: string;
+    protected _subtitle?: string;
+    protected _title: string;
+    protected _backSideTitle?: string;
+    protected _unique: boolean;
```
AMMayberry1, line 50: I think these should all still be readonly

server/game/core/card/Card.ts
```diff
@@ -68,6 +71,34 @@ export class Card extends OngoingEffectSource {
...
+    }
+
+    public get backSideTitle(): string {
+        return this._backSideTitle;
+    }
+
+    public get unique(): boolean {
+        return this._unique;
+    }
+
```
AMMayberry1, line 101: nit: could we keep the property getters in alphabetical order

server/game/core/card/Card.ts
AMMayberry1, line 50: nit: could we keep these in alphabetical order

server/game/core/card/DoubleSidedLeaderCard.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Create card abilities for the second leader side by calling subsequent methods with appropriate properties
+     */
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    protected setupLeaderBackSideAbilities(sourceCard: this) {
+    }
+
+    protected flipLeader() {
```
AMMayberry1, line 41: I think this is supposed to be public, since the flip system calls it (the typing there is just broken so TS didn't catch it)

server/game/core/gameSteps/phases/RegroupPhase.ts
```diff
@@ -38,7 +38,7 @@ export class RegroupPhase extends Phase {
             cardsToReady.push(...player.getUnitsInPlay() as CardWithExhaustProperty[]);
             cardsToReady.push(...player.resources);
 
-            if (!(player.leader.deployed)) {
+            if (!(player.leader.isLeaderUnit() && player.leader.deployed)) {
```
AMMayberry1, line 41: nit: might be simpler and less confusing just to check the zone at this point

server/game/gameSystems/DeployLeaderSystem.ts
```diff
@@ -26,7 +26,7 @@ export class DeployLeaderSystem<TContext extends AbilityContext = AbilityContext
     }
 
     public override canAffect(card: Card, context: TContext): boolean {
-        if (!card.isLeader() || card.deployed) {
+        if (!card.isLeader() || card.isLeaderUnit() && card.deployed) {
```
AMMayberry1, line 29: `isLeaderUnit()` just returns `card.deployed` so that check can be simplified. Theoretically we could just do `card.isLeaderUnit()` by itself but I feel like that's confusing, I wonder if maybe `card.isDeployableLeader() && card.deployed` is the clearest to read? Or else just check zone again

server/game/gameSystems/FlipDoubleSidedLeaderSystem.ts
```diff
@@ -0,0 +1,53 @@
...
+export class FlipDoubleSidedLeaderSystem<TContext extends AbilityContext = AbilityContext> extends CardTargetSystem<TContext, IFlipDoubleSidedLeaderProperties> {
+    public override readonly name = 'flip double-sided leader';
+    public override readonly eventName = EventName.OnLeaderFlipped;
+    public override readonly effectDescription = 'deploy {0}';
+
+    protected override readonly targetTypeFilter = [CardType.Leader];
+
+    public eventHandler(event): void {
+        Contract.assertTrue(event.card.isDoubleSidedLeader());
+        Contract.assertFalse(event.card.isDeployableLeader());
```
AMMayberry1, line 20: nit: can we log the card internalName on these so we have a start for debugging if they get hit somehow. Can just include it in the message parameter

server/game/gameSystems/FlipDoubleSidedLeaderSystem.ts
```diff
@@ -0,0 +1,53 @@
...
+        if (!card.isDoubleSidedLeader()) {
+            return false;
+        }
+        return super.canAffect(card, context);
+    }
+
+    protected override updateEvent(event, card: Card, context: TContext, additionalProperties: any = {}) {
+        super.updateEvent(event, card, context, additionalProperties);
+
+        // TODO: is more needed here? such as below?
```
AMMayberry1, line 39: Actually I think this method isn't needed at all, the default behavior of a GameSystem is to emit an event of type `this.eventName` and it will execute the logic in `eventHandler`. So we should be covered without this, right now it's emitting two separate events named `OnLeaderFlipped`

server/game/stateWatchers/UnitsDefeatedThisPhaseWatcher.ts
```diff
@@ -42,6 +42,11 @@ export class UnitsDefeatedThisPhaseWatcher extends StateWatcher<DefeatedUnitEntr
             .map((entry) => entry.unit);
     }
 
+    /** Get the list of the specified player's units that were defeated */
```
AMMayberry1, line 45: nit: wrong comment, doesn't filter by player

test/server/cards/03_TWI/leaders/ChancellorPalpatinePlayingBothSides.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+                context.player2.clickCard(context.wampa);
+
+                // Ensure no effect besides exhaustion
+                context.player1.clickCard(context.chancellorPalpatine);
+                expect(context.chancellorPalpatine.exhausted).toBe(true);
+                expect(context.p1Base.damage).toBe(10);
+                expect(context.player1.hand.length).toBe(5);
+            });
+
+            it('draws, heals, and flips if a friendly Heroism unit has died', function () {
```
AMMayberry1, line 109: Is this doing anything different than what was covered in the first test case?

test/server/cards/03_TWI/leaders/ChancellorPalpatinePlayingBothSides.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+
+                // Check that Palpatine healed, drew, and flipped
+                context.player1.clickCard(context.chancellorPalpatine);
+                expect(context.chancellorPalpatine.exhausted).toBe(true);
+                expect(context.p1Base.damage).toBe(8);
+                expect(context.player1.hand.length).toBe(1);
+                expect(context.chancellorPalpatine.onStartingSide).toBe(false);
+            });
+
+            it('does not trigger a base with a \'When you deploy a leader\' ability', function () {
```
AMMayberry1, line 136: Ohh good catch

test/server/cards/03_TWI/leaders/ChancellorPalpatinePlayingBothSides.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+                context.player1.clickCard(context.battlefieldMarine);
+
+                context.player1.clickCard(context.chancellorPalpatine);
+                expect(context.chancellorPalpatine.exhausted).toBe(true);
+                expect(context.chancellorPalpatine.onStartingSide).toBe(false);
+                expect(context.p2Base.damage).toBe(0);
+                expect(context.player1.findCardsByName('clone-trooper').length).toBe(0);
+            });
+
+            it('back-side creates a clone, deals 2 damage to enemy base, and flips if a Villainy card has been played', function () {
```
AMMayberry1, line 256: Same question here - was this already covered?

test/server/cards/03_TWI/leaders/ChancellorPalpatinePlayingBothSides.spec.ts
adbronico: Not really. I wrote this and the villainy test before the full cycle test, so I can probably remove this one.

test/server/cards/03_TWI/leaders/ChancellorPalpatinePlayingBothSides.spec.ts
adbronico: The only thing it really tests is that the leader can start flipped in a test context. I can remove it if you want

server/game/gameSystems/DeployLeaderSystem.ts
adbronico: Hmm, this makes me wonder. I changed isLeaderUnit to:
`return this.type === CardType.LeaderUnit;`

I suspect I should change that back since we have isDeployableLeader now as well?

AMMayberry1 (approved):

PR #531, opened 2025-02-05
Add CardDataGetter utility and consolidate testing + lobby flow to use it

AMMayberry1:
Created a CardDataGetter base class that can be used for various implementations of classes that retrieve our card data jsons for us. This will make it much easier to make changes to how we get cards as well as supporting alternate implementations - i.e., we currently have one for downloading from AWS and one for pulling from local.

Consolidated the unit test harness and the GameServer + Lobby flows to use this component. Did a lot of cleanup on the GameServer + Lobby to improve the general flow and to use async correctly.


scripts/run-server.js
```diff
@@ -3,13 +3,15 @@ const childProcess = require('child_process');
...
+function createEnvFile() {
     const defaultEnvFields = {
         GAME_NODE_HOST: 'localhost',
         GAME_NODE_NAME: 'test1',
         GAME_NODE_SOCKET_IO_PORT: '9500',
         ENVIRONMENT: 'development',
-        SECRET: 'verysecret'
+        SECRET: 'verysecret',
+        // set this to "true" to force the local client to pull card data jsons from AWS instead of local folder
+        FORCE_REMOTE_CARD_DATA: 'true',
```
AMMayberry1, line 14: Leaving a note to myself to fix this

CheBato:
This is more of a question regarding decks. When we send the deck data from the FE and want to convert it to our format should we have a utility function that builds the deck using the SynchronousCardDataGetter for each card? And probably a question for Caz at what point should we do this when adding the user to the lobby?

server/gamenode/GameServer.ts
```diff
@@ -39,30 +42,70 @@ interface QueuedPlayer {
...
+    private readonly userLobbyMap = new Map<string, string>();
+    private readonly io: IOServer;
+    private readonly cardDataGetter: CardDataGetter;
+    private readonly testGameBuilder?: any;
+    private readonly tokenCardsData: ITokenCardsData;
+    private readonly playableCardTitles: string[];
+
+    private queue: QueuedPlayer[] = [];
+
+    private constructor(cardDataGetter: CardDataGetter, tokenCardsData: ITokenCardsData, playableCardTitles: string[], testGameBuilder?: any) {
```
CheBato, line 92: I might just be missing something so i'll just ask for clarification just in case where exactly are this.tokenCardsData and this.playableCardTitles set? I see that we'll be setting them once when the gameServer is started I just don't see where they are set.

AMMayberry1: Discussed in chat

AMMayberry1: Nope you were right that was totally missing

PR #532, opened 2025-02-05
Fix bug with Perilous Position

kleer1:
While implementing Bravado, I discovered there was a bug for Perilous Position where if the unit was defeated by the upgrade, an exception would be thrown when the upgrade tried to resolve the exhaust portion of the ability. Since the unit was immediately defeated, when the exhaust resolved the unit and upgrade would both already be in the discard. The simplest solution was to remove the contract checks and let the ability fizzle due to not having a target.

### DIFF
--- a/server/game/core/card/UpgradeCard.ts
+++ b/server/game/core/card/UpgradeCard.ts
@@ -56,9 +56,8 @@ export class UpgradeCard extends UpgradeCardParent {
 
     /** The card that this card is underneath */
     public get parentCard(): UnitCard {
-        if (!this.isInPlay()) {
-            return null;
-        }
+        Contract.assertNotNullLike(this._parentCard);
+        Contract.assertTrue(this.isInPlay());
         return this._parentCard;
     }
 
--- a/server/game/core/card/UpgradeCard.ts
+++ b/server/game/core/card/UpgradeCard.ts
@@ -56,9 +56,9 @@ export class UpgradeCard extends UpgradeCardParent {
 
     /** The card that this card is underneath */
     public get parentCard(): UnitCard {
-        Contract.assertNotNullLike(this._parentCard);
-        Contract.assertTrue(this.isInPlay());
-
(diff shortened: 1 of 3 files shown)


server/game/core/card/UpgradeCard.ts
```diff
@@ -56,9 +56,9 @@ export class UpgradeCard extends UpgradeCardParent {
 
     /** The card that this card is underneath */
     public get parentCard(): UnitCard {
-        Contract.assertNotNullLike(this._parentCard);
-        Contract.assertTrue(this.isInPlay());
-
+        if (!this.isInPlay()) {
```
AMMayberry1, line 59: A big design principle for these properties is to strongly enforce when they are and aren't legal to access so that we can catch problems with filters and targeting early. Then we just need to make sure to set our filters up right so that they don't ask for a property of a card that (from a swu rules perspective) doesn't exist based on the card's current zone.

So in this case I'd much prefer that we change the Perilous Position ability to use 

```ts
target:  context.source.isInPlay() ? context.source.parentCard : null
```

Thankfully this case is very rare (this is the only upgrade that can defeat on attach) so we don't need to think about this check in most cases, if it starts coming up more often then maybe it's worth thinking about a strategy that would let us safely get the parentCard but still catch error cases that are accidentally assuming that the upgrade is in play when it's not

kleer1: Makes sense. That was the other solution I considered. Only concern is that we'll also need to remember to add that check for any other upgrades with negative health modifiers in the future,

AMMayberry1 (approved):

server/game/core/card/UpgradeCard.ts
AMMayberry1: Yeah, if that ends up being the case then we'll need to figure out some way of handling it that doesn't require remembering to do this. But as one-off I'm okay with it

PR #537, opened 2025-02-06
Jango Fett, Concealing the Conspiracy

moyerr:
| Undeployed | Deployed |
| :---: | :---: |
|  |  |

## Developer Notes

The work for implementing this leader uncovered an issue with the `DistributeDamageSystem`. Essentially, this system had no way to identify the source of the damage if it was different from the card that triggered the ability. The only current example is Overwhelming Barrage, where the event has the ability, but a unit deals the distributed damage. I addressed this by adding a `source` property to `IDistributeDamageSystemProperties`. I'm open to feedback on how I implemented this!

### Undeployed

 1. Trigger ability when a friendly unit deals combat damage
 2. Trigger ability when a friendly unit deals damage from its own ability (e.g. Elite P-38)
 3. Trigger ability when a friendly unit deals direct damage from an event (Strike True)
 4. Trigger ability when a friendly unit deals distributed damage from an event (Overwhelming Barrage)
 5. Trigger ability when a friendly unit deals damage from an ability gained from an upgrade (Vambrace Flamethrower)
 6. Trigger ability when a friendly unit deals damage at the start of the regroup phase (Sneak Attack + Ruthless Raider)
 7. Trigger ability when a stolen unit deals combat damage to an enemy unit (Change of Heart)
 8. _Don't_ trigger ability when damage is dealt directly from an event (Open Fire)
 9. _Don't_ trigger ability when damage is dealt directly from an upgrade (Vader's Lightsabeer)
 10. _Don't_ trigger ability when a friendly unit deals 0 damage in combat (Grogu)
 11. _Don't_ trigger ability at the start of the regroup phase if Jango was exhausted in the previous action phase
 12. _Don't_ trigger ability when a friendly unit deals damage to another friendly unit (Doctor Pershing)

### Deployed

 1. Trigger ability when a friendly unit deals combat damage
 2. Trigger ability when a friendly unit deals damage from its own ability (e.g. Elite P-38)
 3. Trigger ability when a friendly unit deals direct damage from an event (Strike True)
 4. Trigger ability when a friendly unit deals distributed damage from an event (Overwhelming Barrage)
 5. Trigger ability when a friendly unit deals damage from an ability gained from an upgrade (Vambrace Flamethrower)
 6. Trigger ability when a stolen unit deals combat damage to an enemy unit (Change of Heart)
 7. _Don't_ trigger ability when damage is dealt directly from an event (Open Fire)
 8. _Don't_ trigger ability when damage is dealt directly from an upgrade (Vader's Lightsabeer)
 9. _Don't_ trigger ability when a friendly unit deals 0 damage in combat (Grogu)
 10. _Don't_ trigger ability when a friendly unit deals damage to another friendly unit (Doctor Pershing)

### Extra

- Added a test case to Hold-Out Blaster, ensuring the damage dealt when played triggers Jango's ability
- Added a test case to Maul - Shadow Collective Visionary, ensuring that the deflected combat damage can still trigger an opposing Jango leader's ability
- Added three test cases to `Bounty.spec.ts`, ensuring damage dealt by collecting a bounty is attributed to the correct unit (including a stolen unit with a bounty)


server/game/gameSystems/DistributeDamageSystem.ts
```diff
@@ -8,20 +8,28 @@ import type { IDistributeAmongTargetsSystemProperties } from './DistributeAmongT
...
+> extends DistributeAmongTargetsSystem<TContext, TProperties> {
     protected override readonly eventName = MetaEventName.DistributeDamage;
     public override readonly name = 'distributeDamage';
 
     public override promptType: DistributePromptType = StatefulPromptType.DistributeDamage;
 
     protected override generateEffectSystem(target: Card = null, amount = 1): DamageSystem | HealSystem {
-        return new DamageSystem({ type: DamageType.Ability, target, amount });
+        const { source } = this.properties;
+        return new DamageSystem({ type: DamageType.Ability, target, amount, source });
```
AMMayberry1, line 32: Looks right! Nice job

server/game/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.ts
```diff
@@ -0,0 +1,74 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.exhaust((context) => {
+                return { target: context.event.card };
+            })
+        });
+    }
+
+    private isEnemyUnitDamagedByFriendlyUnit(event, context): boolean {
+        if (event.card.isUnit() && event.card.controller !== context.source.controller) {
+            console.log('----------------------');
+            console.log(`[Damage Recieved] ${event.card.title}`);
```
AMMayberry1, line 51: nit: can we remove the logs in here

server/game/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.ts
```diff
@@ -0,0 +1,74 @@
...
+                onDamageDealt: (event, context) => this.isEnemyUnitDamagedByFriendlyUnit(event, context)
+            },
+            immediateEffect: AbilityHelper.immediateEffects.exhaust((context) => {
+                return { target: context.event.card };
+            })
+        });
+    }
+
+    private isEnemyUnitDamagedByFriendlyUnit(event, context): boolean {
+        if (event.card.isUnit() && event.card.controller !== context.source.controller) {
```
AMMayberry1, line 49: nit: because of some potential issues we've found, we need to start using `context.player` instead of `context.source.controller` in all cases unless there's a specific reason to use the latter.

To explain quickly - `context.player` is the triggering player for the ability, whereas `context.source.controller` is the current controller of the card. It specifically matters when a stolen unit has an on-defeat ability, but I want to start using it generally so everyone is in the habit.

server/game/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.ts
```diff
@@ -0,0 +1,74 @@
...
+            console.log('----------------------');
+            console.log(`[Damage Recieved] ${event.card.title}`);
+
+            if (
```
AMMayberry1, line 53: Could we rearrange these if blocks just for readability, since some of the conditions are repeated

test/server/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.spec.ts
```diff
@@ -0,0 +1,470 @@
...
+
+                context.player1.clickCard(context.isbAgent);
+                context.player1.clickCard(context.volunteerSoldier);
+                context.player1.clickPrompt('Deal 3 damage divided as you choose among enemy ground units');
+                context.player1.setDistributeDamagePromptState(new Map([
+                    [context.battlefieldMarine, 2],
+                    [context.fleetLieutenant, 1],
+                ]));
+
+                // Choose resolution order (is there a way to know which instant corresponds to which damaged unit?)
```
AMMayberry1, line 127: Yeah this is a really important question we've known about for a bit. I created a [github issue](https://github.com/SWU-Karabast/forceteki/issues/540) and I'll trigger a discussion in the FE chat. Do you mind also changing this comment to be a TODO and maybe link the issue?

test/server/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.spec.ts
```diff
@@ -0,0 +1,470 @@
...
+                // CASE 8: Attacks that deal 0 damage do not trigger Jango's ability
+
+                context.moveToNextActionPhase();
+
+                context.player1.clickCard(context.grogu);
+                context.player1.clickPrompt('Attack');
+                context.player1.clickCard(context.consularSecurityForce);
+
+                expect(context.player1).not.toHavePassAbilityPrompt('Exhaust this leader');
+                expect(context.consularSecurityForce.damage).toBe(0);
```
AMMayberry1, line 466: This is truly fantastic, really appreciate the effort and your thoroughness here. Few more cases it would be good to have and then we're set:

- Val's bounty ability should trigger Jango if the Jango played is the one resolving the bounty
- Let's test with a stolen unit (e.g. with Change of Heart) and confirm that it works
- If we don't already have a case where a friendly unit is damaging another friendly unit (e.g. Pershing), let's also add that

server/game/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.ts
moyerr: @AMMayberry1 For sure, meant to put a comment that these will come out when I take this out of draft.

server/game/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.ts
moyerr: I don't think any of the conditions are repeated, but I'll put it into a switch instead. Let me know if that makes it more clear!

test/server/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.spec.ts
moyerr: Everything addressed but Val - will add that shortly!

moyerr: @AMMayberry1 done!

server/game/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.ts
```diff
@@ -46,26 +46,18 @@ export default class JangoFettConcealingTheConspiracy extends LeaderUnitCard {
...
-        if (event.card.isUnit() && event.card.controller !== context.source.controller) {
-            console.log('----------------------');
-            console.log(`[Damage Recieved] ${event.card.title}`);
+        // If an enemy unit received the damage
+        if (event.card.isUnit() && event.card.controller !== context.player) {
+            switch (event.damageSource.type) {
+                case DamageSourceType.Ability:
+                    // If the damage was dealt by a friendly unit via an ability
+                    return event.damageSource.card.isUnit() &&
+                      event.damageSource.card.controller === context.player;
```
AMMayberry1, line 55: I think this might have a bug - a case we've been hitting a lot is when a stolen unit has a "when defeated" ability, checking `card.controller` ends up giving you a wrong answer because the card has reverted to being under the control of its owner but it was under control of the opponent when the ability was triggered at defeat time.

In this case we do have a record of who the activating player was as `damageSource.player`, however that does seem to have a bug itself: https://github.com/SWU-Karabast/forceteki/blob/41fc9b66505a1d52dbff814f39b015902390b127/server/game/gameSystems/DamageSystem.ts#L257

For the same reason as above, that should just be `player: context.player`. Do you mind fixing that, updating these checks to use `damageSource.player`, and then checking that in the tests. Simplest thing might be to just update the "stolen unit" tests you just added to be on-defeats, that covers all bases (if you want to add it as a separate case feel free as well).

test/server/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.spec.ts
```diff
@@ -0,0 +1,470 @@
...
+                    'Exhaust this leader'
+                ]);
+
+                // HELP WANTED: The prompts in this test case are not what I expected.
+                //   It prompts to resolve another instance after `Exhaust this leader`
+                //   is clicked.
+
+                // These are just here to resolve the prompts and make the test pass
+                context.player1.clickPrompt('Exhaust this leader');
+                context.player1.clickPrompt('Exhaust this leader');
```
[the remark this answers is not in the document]
AMMayberry1: Unresolving this while I take a look at it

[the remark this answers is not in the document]
AMMayberry1: Okay so I figured out what's going on here - it's actually not a bug, it's just we have kind of an awkward UX flow here at the moment.

Basically what's going on is that since the ability is optional, there are two clicks required to trigger it from the triggered ability window. The first selects the ability that we are triggering, which brings us to an inner menu in which we can choose whether to trigger or pass the ability we just selected.

It's not a great flow but until we have a better UX for the trigger window overall, there's not a lot we can do. So in this case what might be good is in between these two lines to have an `expect(context.player1).toHavePassAbilityPrompt('Exhaust this leader')` which will at least communicate what's going on a little more clearly

test/server/core/abilities/keyword/Bounty.spec.ts
```diff
@@ -238,6 +238,69 @@ describe('Bounty', function() {
...
+                // Resolve bounty
+                context.player2.clickCard(context.consularSecurityForce);
+
+                // Jango triggers
+                expect(context.player1).toHavePassAbilityPrompt('Exhaust this leader');
+                context.player1.clickPrompt('Exhaust this leader');
+
+                expect(context.consularSecurityForce.exhausted).toBe(true);
+                expect(context.jangoFett.exhausted).toBe(true);
+            });
```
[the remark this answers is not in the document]
moyerr: Done!

server/game/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.ts
moyerr: This has been updated to consider the case of the stolen unit with a bounty

server/game/cards/03_TWI/leaders/JangoFettConcealingTheConspiracy.ts
```diff
@@ -0,0 +1,67 @@
...
+            })
+        });
+    }
+
+    private isEnemyUnitDamagedByFriendlyUnit(event, context): boolean {
+        // If an enemy unit received the damage
+        if (event.card.isUnit() && event.card.controller !== context.player) {
+            switch (event.damageSource.type) {
+                case DamageSourceType.Ability:
+                    const controller = event.damageSource.controller ?? event.damageSource.card.controller;
```
AMMayberry1, line 53: Let's add a similar TODO comment here - ideally we can get to a place where `event.damageSource.controller` is always just populated correctly and we don't need to have the fallback. That's a later task though

AMMayberry1 (approved):

PR #538, opened 2025-02-06
Fenn rau

ssoulier:
[Fenn Rau](https://swudb.com/card/SHD/067/fenn-rau-protector-of-concord-dawn)

Screenshot 2025-02-06 at 07 31 36


test/server/cards/02_SHD/units/FennRauProtectorOfConcordDawn.spec.ts
```diff
@@ -0,0 +1,77 @@
+describe('Fenn Rau Protector of Concord Dawn\'s ability', function () {
+    integration(function (contextRef) {
+        it('should allow to play an upgrade for 2 less', function () {
+            contextRef.setupTest({
+                phase: 'action',
+                player1: {
+                    hand: ['fenn-rau#protector-of-concord-dawn', 'academy-training', 'jedi-lightsaber'],
+                    groundArena: ['specforce-soldier'],
+                    leader: { card: 'luke-skywalker#faithful-friend' },
```
AMMayberry1, line 9: nit: can do this as `leader: 'luke-skywalker#faithful-friend'`

server/game/cards/02_SHD/units/FennRauProtectorOfConcordDawn.ts
```diff
@@ -0,0 +1,43 @@
...
+                immediateEffect: AbilityHelper.immediateEffects.playCardFromHand({
+                    adjustCost: { costAdjustType: CostAdjustType.Decrease, amount: 2 }
+                })
+            }
+        });
+
+        this.addTriggeredAbility({
+            title: 'Give an enemy unit -2/-2 for this phase.',
+            when: {
+                onUpgradeAttached: (event, context) => event.card === context.source
```
AMMayberry1, line 32: So, there's actually a bit of nuance here. The ability is "when you _play_ an upgrade on this unit," but this trigger is "when you _attach_ an upgrade to this unit." So e.g., if we use Survivors' Gauntlet to move an upgrade onto Fenn Rau (which is a test we should add), this ability shouldn't be triggered.

The Dengar ability is close to correct, it just needs to be able to check the attach target. That actually wasn't possible originally, I just merged a PR that adds `event.attachTarget` so you can use that to check if it's being played on this unit.

server/game/cards/02_SHD/units/FennRauProtectorOfConcordDawn.ts
```diff
@@ -0,0 +1,43 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '3399023235',
+            internalName: 'fenn-rau#protector-of-concord-dawn'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'You may play an upgrade from your hand. It costs 2 less',
```
AMMayberry1, line 16: nit: need to remove the "you may" since this is a triggered ability

server/game/cards/02_SHD/units/FennRauProtectorOfConcordDawn.ts
AMMayberry1: P.S. let me know if that doesn't work for some reason, I didn't get a chance to test that property

ssoulier: It works well. Thanks for the subtlety.

AMMayberry1 (approved):

PR #541, opened 2025-02-06
Add Mercenary Gunship and rename "abilityController" to "abilityActivator"

adbronico:


server/game/cards/02_SHD/events/ANewAdventure.ts
```diff
@@ -21,7 +21,7 @@ export default class ANewAdventure extends EventCard {
             then: (thenContext) => ({
                 title: `Play ${thenContext.target.title} for free`,
                 optional: true,
-                abilityController: thenContext.source.controller !== thenContext.target.controller ? RelativePlayer.Opponent : RelativePlayer.Self,
+                abilityActivator: thenContext.source.controller !== thenContext.target.controller ? RelativePlayer.Opponent : RelativePlayer.Self,
```
AMMayberry1, line 24: So I think we actually need to remove this property from sub-steps such as `then` and `ifYouDo`. The way it's being used here it's not a filter, it's a declaration of which player will trigger this part of the ability. That's different than the intent we're changing it to use.

The specific concrete problem with this situation is - what would it mean to put `WildcardRelativePlayer.Any` here? That would indicate that both players trigger this ability step.

That's why I was saying initially we probably need to change A New Adventure and find a way to have a GameSystem that lets the opponent play a card.

test/server/cards/02_SHD/units/MercenaryGunship.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+                const p1Resources = context.player1.readyResourceCount;
+                const p2Resources = context.player1.readyResourceCount;
+
+                context.player1.passAction();
+
+                // Check that Player 2 can take control
+                context.player2.clickCard(context.mercenaryGunship);
+                expect(context.mercenaryGunship).toBeInZone('spaceArena', context.player2);
+                expect(context.player2.readyResourceCount).toBe(p2Resources - 4);
```
AMMayberry1, line 24: Can we add `toBeAbleToSelectExactly()` check for each player's action to confirm that it is being highlighted when it is their action?

server/game/cards/02_SHD/events/ANewAdventure.ts
adbronico: Do you think we need to knock that part out as part of this PR, or are we OK leaving this as is with a note for now and then reworking the game system later?

AMMayberry1 (approved):

PR #542, opened 2025-02-06
#DeckDataRefactor

CheBato:
- Refactored the way we send data to the backend. Currently we want to send both the reconstructed deck and the swuDB deck. So me and Veld can simultaneously work on the implementation. The user in the lobby also has swuDeck data.

### DIFF
--- a/server/gamenode/Lobby.ts
+++ b/server/gamenode/Lobby.ts
@@ -17,6 +17,7 @@ interface LobbyUser {
     ready: boolean;
     socket: Socket | null;
     deck: Deck | null;
+    swuDeck?: Deck | null;
 }
 export enum MatchType {
     Custom = 'Custom',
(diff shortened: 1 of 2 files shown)


server/gamenode/Lobby.ts
```diff
@@ -197,6 +200,8 @@ export class Lobby {
     private changeDeck(socket: Socket, ...args) {
         const activeUser = this.users.find((u) => u.id === socket.user.id);
         Contract.assertTrue(args[0] !== null);
+        Contract.assertTrue(args[1] !== null);
+        activeUser.swuDeck = new Deck(args[1]);
```
AMMayberry1, line 204: Hmm, how is this working? The `Deck()` constructor shouldn't be able to accept a deck in swudb format

AMMayberry1 (approved):

PR #545, opened 2025-02-06
Implementing Home One, Rose Tico leader and Red Five

aruizc:
image
image
image
image


test/server/cards/04_JTL/leaders/RoseTicoSavingWhatWeLove.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+                expect(context.roseTico.exhausted).toBe(true);
+
+                // Move to next action phase
+                context.moveToNextActionPhase();
+                context.player1.clickCard(context.roseTico);
+                context.player1.clickPrompt('Heal 2 damage from a Vehicle unit that attacked this phase');
+
+                // No Vehicle units attacked this phase so not able to heal
+                expect(context.player2).toBeActivePlayer();
+                expect(context.roseTico.exhausted).toBe(true);
```
AMMayberry1, line 46: Since there's a state watcher involved we need to do the Waylay test - i.e., attack with a vehicle, Waylay it to hand, then put it back in play. Since it's now a new "copy" of the card, it shouldn't be targetable by Rose since the new copy hasn't done an attack.

test/server/cards/04_JTL/leaders/RoseTicoSavingWhatWeLove.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+                context.player2.clickCard(context.p1Base);
+
+                // Heal the Vehicle unit
+                context.player1.clickCard(context.roseTico);
+                context.player1.clickPrompt('Heal 2 damage from a Vehicle unit that attacked this phase');
+                expect(context.player1).toBeAbleToSelectExactly([context.allianceXwing, context.atst]); // Only the Vehicle units that attacked this phase
+                expect(context.player1).not.toHavePassAbilityButton();
+                context.player1.clickCard(context.atst);
+
+                expect(context.atst.damage).toBe(0);
```
AMMayberry1, line 35: Can we put at least 3 damage on the atst to confirm the exact heal amount. Same for heal tests below

test/server/cards/04_JTL/leaders/RoseTicoSavingWhatWeLove.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+                expect(context.player1).toHavePrompt('Choose a unit');
+                expect(context.player1).toHavePassAbilityButton();
+                expect(context.player1).toBeAbleToSelectExactly([context.allianceXwing, context.lowAltitudeGunship, context.atst]); // Only the Vehicle units
+                context.player1.clickCard(context.lowAltitudeGunship);
+
+                expect(context.lowAltitudeGunship.damage).toBe(0);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should not be able to heal a Vehicle unit as there is not Vehicle units', function () {
```
AMMayberry1, line 79: I think we're okay without this test, we have a lot of cases checking that target resolver auto-passes if there are no targets

server/game/cards/04_JTL/units/HomeOneOnMyMark.ts
```diff
@@ -0,0 +1,19 @@
...
+        return {
+            id: '1356826899',
+            internalName: 'home-one#on-my-mark',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addDecreaseCostAbility({
+            title: 'This units costs 3 resources less to play',
+            condition: (context) => context.source.controller.opponent.getUnitsInPlay(ZoneName.SpaceArena).length >= 3,
```
AMMayberry1, line 15: nit: need to use `context.player` instead of `context.source.controller`. This is important if e.g. it was being played from capture zone with Dryden Vos (unfortunately he's not implemented so we can't test it - can you add a TODO for that in the test file)

test/server/cards/04_JTL/units/RedFiveRunningTheTrench.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+            expect(context.player1).toBeAbleToSelectExactly([context.wampa, context.atst]);
+            context.player1.clickCard(context.atst);
+
+            expect(context.atst.damage).toBe(3);
+
+            // Attack with Red Five, Running The Trench but no unit is damaged so ability is not triggered
+
+            // Reset board state
+            context.atst.damage = 0;
+            context.wampa.damage = 0;
```
AMMayberry1, line 32: nit: we should always use `context.setDamage()` instead of directly setting the damage property. It will update game state in case there are any abilities that trigger off of specific damage amount

AMMayberry1 (approved):

PR #549, opened 2025-02-07
Implement Scanning Officer and resource shuffling logic

adbronico:


server/game/cards/02_SHD/units/ScanningOfficer.ts
```diff
@@ -0,0 +1,40 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '0505904136',
+            internalName: 'scanning-officer',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Reveal up to 4 Vigilance cards from your hand',
```
AMMayberry1, line 15: nit: wrong title

server/game/cards/02_SHD/units/ScanningOfficer.ts
```diff
@@ -0,0 +1,40 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.reveal((context) => ({
+                // TODO: should Scanning Officer always choose readied resources?
```
AMMayberry1, line 17: Yeah, it's one of those weird things where it's not clear that it's necessarily optimal to choose ready resources since your opponent is probably hiding the smuggle resources in exhaust. So if we start going down that rabbit hole we probably just end up at letting the defending player choose ready / exhausted for each card, which we obviously want to avoid.

So I think just random on both sides (ready / exhaust and SO selection) is the most straightforward and probably the most fair, and tbh my guess is that the hit rate in actual play also amounts to basically random anyway

server/game/cards/02_SHD/units/ScanningOfficer.ts
```diff
@@ -0,0 +1,40 @@
...
+                target: context.player.opponent.getRandomResources(context, 3),
+                useDisplayPrompt: true,
+                promptedPlayer: RelativePlayer.Self
+            })),
+            then: (thenContext) => ({
+                title: 'Defeat each resource with Smuggle',
+                effect: 'defeat',
+                immediateEffect: AbilityHelper.immediateEffects.sequential(() => {
```
AMMayberry1, line 25: I think this needs to be simultaneous. If there was e.g. a 97th Legion with one point of remaining HP, this would cause it to be defeated since it would briefly see an HP drop (which I don't think is correct), whereas simultaneous would not allow that.

server/game/core/zone/ResourceZone.ts
```diff
@@ -29,4 +31,28 @@ export class ResourceZone extends SimpleZone<TokenOrPlayableCard> {
         this.hiddenForPlayers = RelativePlayer.Opponent;
         this.name = ZoneName.Resource;
     }
+
+    public rearrangeResourceExhaustState(context: AbilityContext, prioritizeSmuggle: boolean = false): void {
```
AMMayberry1, line 35: Looks like the `prioritizeSmuggle` isn't actually getting used

server/game/gameSystems/ResourceCardSystem.ts
```diff
@@ -43,9 +43,9 @@ export class ResourceCardSystem<TContext extends AbilityContext = AbilityContext
     public override generatePropertiesFromContext(context: TContext, additionalProperties = {}): IResourceCardProperties {
         const properties = super.generatePropertiesFromContext(context, additionalProperties);
 
-        if (Array.isArray(properties.target)) {
-            Contract.assertTrue(properties.target.length <= 1, 'Resourcing more than 1 card is not yet supported');
-        }
+        // if (Array.isArray(properties.target)) {
+        //     Contract.assertTrue(properties.target.length <= 1, 'Resourcing more than 1 card is not yet supported');
+        // }
```
AMMayberry1, line 48: Can just remove

AMMayberry1 (approved):

PR #551, opened 2025-02-07
Implement Bravado

kleer1:
image


server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -628,6 +629,9 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
         public unattachUpgrade(upgrade) {
             this.assertPropertyEnabledForZone(this._upgrades, 'upgrades');
             this._upgrades = this._upgrades.filter((card) => card.uuid !== upgrade.uuid);
+            if (upgrade.printedHp !== 0) {
+                this._lastPlayerToModifyHp = upgrade.owner;
```
AMMayberry1, line 633: Hmm, for the unattach case I think it should be whoever controls the effect that removes the upgrade, right? I think that makes it relatively easy to get this case right

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -639,6 +643,10 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
             Contract.assertTrue(this.zone.hasCard(upgrade));
 
             this._upgrades.push(upgrade);
+
+            if (upgrade.printedHp !== 0) {
+                this._lastPlayerToModifyHp = upgrade.owner;
```
AMMayberry1, line 648: Should be `upgrade.controller` in case it's being played out by the opponent with e.g. A Fine Addition. Probably a good idea to add a test case for A Fine Addition + Perilous Position

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -664,5 +672,26 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
                 parentCardId: this.getCaptor()?.uuid,
             };
         }
+
+        public override addOngoingEffect(ongoingEffect: IOngoingCardEffect): void {
+            if (ongoingEffect.type === EffectName.ModifyStats) {
```
AMMayberry1, line 677: Probably we should unpack the stat modified and check to see if it has an hp value or not. If it's just modifying power then I think it shouldn't be counted

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -664,5 +672,26 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
...
+                this._lastPlayerToModifyHp = ongoingEffect.context.source.controller;
+            }
+            super.addOngoingEffect(ongoingEffect);
+        }
+
+        public get lastPlayerToModifyHp() {
+            if (this.isInPlay()) {
+                return this._lastPlayerToModifyHp;
+            }
+            return null;
```
AMMayberry1, line 687: I'd prefer to use the pattern established with other similar properties to keep this from being accessed when the card isn't in the play area. You can reference the `upgrades` getter and `setUpgradesEnabled`. That would also remove the need for the `move()` override below

server/game/gameSystems/FrameworkDefeatCardSystem.ts
```diff
@@ -49,8 +49,15 @@ export class FrameworkDefeatCardSystem<TContext extends AbilityContext = Ability
             case DefeatSourceType.UniqueRule:
                 return { type: DefeatSourceType.UniqueRule, player: card.controller };
             case DefeatSourceType.FrameworkEffect:
-                // TODO: this is a workaround until we get comp rules 3.0
-                return null;
+                let responsiblePlayer = null;
```
AMMayberry1, line 52: Probably good to leave a couple lines of comment here referencing the specific rule number and explaining what this is doing briefly

test/server/cards/02_SHD/events/Bravado.spec.ts
```diff
@@ -0,0 +1,273 @@
...
+            // Case 1: No cost decrease
+            expect(context.battlefieldMarine.exhausted).toBe(true);
+            context.player1.clickCard(context.bravado);
+            context.player1.clickCard(context.battlefieldMarine);
+            expect(context.player1.exhaustedResourceCount).toBe(5);
+            expect(context.battlefieldMarine.exhausted).toBe(false);
+
+            context.player2.passAction();
+
+            // CASE 2: Killing a unit when you attack counts as defeating it.
```
AMMayberry1, line 73: I'm so sorry, I should've thought to say this when you started - since this is really a test for a very fundamental engine component, probably we should leave a couple very basic Bravado cases here and promote the rest of these to be a test suite of their own. Probably like `scenarios/DefeatAttribution.spec.ts`

For tests of engine components (as opposed to specific cards) we're way more lax with spreading out tests to individual cases since they're a small fraction of the overall test suite and they're pretty important to be able to debug easily if they break. So you could remove the `reset()` and spread these out into individual cases using a `beforeEach()` setup.

test/server/cards/02_SHD/events/Bravado.spec.ts
```diff
@@ -0,0 +1,273 @@
...
+            context.player1.moveCard(context.takedown, 'hand');
+            context.player1.clickCard(context.takedown);
+            context.player1.clickCard(context.rebelPathfinder);
+
+            context.player2.passAction();
+            context.player1.readyResources(10);
+
+            // Currently needed due to bug where smuggling with empty deck
+            // throws error when trying to replace the smuggle card.
+            context.player1.moveCard(context.perilousPosition, 'deck');
```
AMMayberry1, line 264: Ahh okay, I'll take a look at this. Looks like we somehow forgot to add the empty deck case to the smuggle tests...

test/server/cards/02_SHD/events/Bravado.spec.ts
```diff
@@ -0,0 +1,273 @@
...
+
+            expect(context.battlefieldMarine.exhausted).toBe(true);
+            context.player1.clickCard(context.bravado);
+            context.player1.clickCard(context.battlefieldMarine);
+            expect(context.player1.exhaustedResourceCount).toBe(5);
+            expect(context.battlefieldMarine.exhausted).toBe(false);
+        });
```
AMMayberry1, line 271: I would take a look at some of the Jango damage cases and see if you think any are worth adding here. Specifically we've found some issues with Bounty. Also, since we've had some issues with when defeated on stolen units being attributed correctly, definitely worth explicitly testing those.

https://github.com/SWU-Karabast/forceteki/pull/537

Also just linking for general awareness that the Jango PR is also making some changes to attribution potentially, might need to coordinate with ThatSecondMouse if there are a collisions

test/server/cards/02_SHD/events/Bravado.spec.ts
```diff
@@ -0,0 +1,273 @@
...
+
+            context.player2.passAction();
+            context.player1.readyResources(10);
+
+            context.player1.moveCard(context.takedown, 'hand');
+            context.player1.clickCard(context.takedown);
+            context.player1.clickCard(context.rebelPathfinder);
+
+            context.player2.passAction();
+            context.player1.readyResources(10);
```
AMMayberry1, line 260: Another interesting case that I'm not sure if we can currently get right is Huyang. I _think_ that since Huyang's +hp effect is a lasting effect (it has an explicit expiration), then if a unit dies due to it going away that it falls under third case and no one should be credited with the defeat.

test/server/cards/02_SHD/events/Bravado.spec.ts
AMMayberry1: PS if you do find issues with Bounty or the stolen unit cases then definitely do ping ThatSecondMouse, he's actively debugging those right now

server/game/core/card/propertyMixins/UnitProperties.ts
kleer1: fixed

server/game/core/card/propertyMixins/UnitProperties.ts
kleer1: fixed

server/game/core/card/propertyMixins/UnitProperties.ts
kleer1: fixed

server/game/gameSystems/FrameworkDefeatCardSystem.ts
kleer1: added

AMMayberry1:
The PR with the cut for the beta is here, solved a lot of the merge issues: https://github.com/SWU-Karabast/forceteki/pull/551

PR #555, opened 2025-02-08
Dryden Vos & Play Unit under control engine change

brianries:
image


server/game/actions/PlayUnitAction.ts
```diff
@@ -29,7 +29,6 @@ export class PlayUnitAction extends PlayCardAction {
             context.source,
         );
 
-        // TODO TAKE CONTROL
         const playForOpponentEffect = context.source.getOngoingEffectValues(EffectName.EntersPlayForOpponent);
         const player = playForOpponentEffect.length > 0 ? RelativePlayer.Opponent : RelativePlayer.Self;
```
AMMayberry1, line 33: Let's remove this logic from here and the `EntersPlayForOpponent` effect from the engine entirely since we're taking a different to this situation than ringteki did

server/game/cards/02_SHD/units/DrydenVosOfferingNoEscape.ts
```diff
@@ -0,0 +1,31 @@
...
+export default class DrydenVos extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '0598830553',
+            internalName: 'dryden-vos#offering-no-escape',
+        };
+    }
+
+    private checkIfTargetIsGuardedByControlledUnit(card, context) {
+        return card.zoneName === ZoneName.Capture && card.zone.captor.controller === context.source.controller;
```
AMMayberry1, line 15: Can't remember if I mentioned this to you already or not, but we need to switch to preferentially using `context.player` instead of `context.source.controller` because the former behaves more correctly in certain edge cases. The former is the player activating the ability, the latter is the current controller of the card.

 It usually only comes up around stolen units being defeated since their controller reverts back to the owner before the trigger happens, but I think it's a good habit for everyone to start defaulting to `context.player` in all cases

server/game/gameSystems/RescueSystem.ts
```diff
@@ -35,7 +35,10 @@ export class RescueSystem<TContext extends AbilityContext = AbilityContext, TPro
     protected override updateEvent(event, card: Card, context: TContext, additionalProperties): void {
         super.updateEvent(event, card, context, additionalProperties);
 
+        // The card is being rescued, so it's controller should reset to the owner, but in relative terms to the player.  Null player is assumed to be self
+        const newController = (event.player == null) ? RelativePlayer.Self : event.player === card.owner ? RelativePlayer.Self : RelativePlayer.Opponent;
```
AMMayberry1, line 39: Yeah I see what you mean about the awkwardness it creates to be forced to use the enum in all cases. If you want, you could switch over to using `Player | RelativePlayer` as the type for the `newController` property and then add the `getConcretePlayer()` enum helper we discussed. That would help us get started on making similar properties throughout the codebase more flexible.

test/server/cards/02_SHD/units/DrydenVosOfferingNoEscape.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+                context.player1.clickCard(context.drydenVosOfferingNoEscape);
+                context.player1.clickPrompt('Shielded');
+                expect(context.drydenVosOfferingNoEscape).toHaveExactUpgradeNames(['shield']);
+
+                // Now selecting the captured card -- ensure it doesn't see an opponent's captured card
+                expect(context.player1).toBeAbleToSelectExactly([context.recklessGunslinger]);
+                context.player1.clickCard(context.recklessGunslinger);
+
+                expect(context.player1.exhaustedResourceCount).toBe(10); // the captured card should play for free (only +7 for dryden)
```
AMMayberry1, line 48: nit: prefer to use the pattern where we first put the current exhausted resource count in a variable and then check `exhaustedResourceCount + 7`. that keeps things flexible in case we change something earlier in the test down the road

test/server/cards/02_SHD/units/DrydenVosOfferingNoEscape.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+                expect(context.p2Base.damage).toBe(1);
+
+                // Verify the reckless gunslinger is now under player 1's control
+                expect(context.recklessGunslinger.controller).toEqual(context.player1.player);
+
+                context.player2.passAction();
+
+                context.player1.moveCard(takeCaptive1, 'hand');
+
+                // Use take-captive to set up a captured card for player 1
```
AMMayberry1, line 61: I think this might need to be an instance where we relax the restriction on consolidating test cases, just b/c the setup is so hard to follow with the various iterations of capturing stuff. Or else have comments at each stage clearly indicating what the current board state is.

Eventually we'll add support for capture in the test setup and it'll be less awkward

test/server/cards/02_SHD/units/DrydenVosOfferingNoEscape.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+                context.player1.clickCard(context.recklessGunslinger);
+
+                expect(context.player1.exhaustedResourceCount).toBe(10); // the captured card should play for free (only +7 for dryden)
+
+                // Confirm that reckless-gunslinger triggered its when played ability
+                expect(context.p1Base.damage).toBe(1);
+                expect(context.p2Base.damage).toBe(1);
+
+                // Verify the reckless gunslinger is now under player 1's control
+                expect(context.recklessGunslinger.controller).toEqual(context.player1.player);
```
AMMayberry1, line 55: Any way we could test this with a card effect? Since the controller logic can get kind of complex I prefer to test it through game logic if at all possible.

Additionally, since this is first case of playing a unit under opponent's control, would be good to specifically try out a when-played ability, when-defeated ability, and constant ability (these can be separate test setups)

server/game/cards/02_SHD/units/DrydenVosOfferingNoEscape.ts
brianries: thanks, fixed

test/server/cards/02_SHD/units/DrydenVosOfferingNoEscape.spec.ts
brianries: fixed

test/server/cards/02_SHD/units/DrydenVosOfferingNoEscape.spec.ts
brianries: splitting out the tests, thanks!

test/server/cards/02_SHD/units/DrydenVosOfferingNoEscape.spec.ts
brianries: Added a high-ground test to determine if its friendly, and then added test for when-defeated and constant

server/game/actions/PlayUnitAction.ts
brianries: Removed the logic here, and under Constants.ts and OngoingEffectLibrary.ts

server/game/gameSystems/RescueSystem.ts
brianries: I made it take both RelativePlayer -- which will be used downstream against context.player, and an explicit Player.  Adjusted all functions to use explicit player (was available).  Currently nothing using relative player, but its available if need be?

server/game/core/utils/EnumHelpers.ts
```diff
@@ -196,3 +197,13 @@ export const getCardTypesForFilter = (cardTypeFilter: CardTypeFilter): CardType[
             return [cardTypeFilter];
     }
 };
+
+export const asConcretePlayer = (contextPlayer: Player, player: Player | RelativePlayer): Player => {
```
AMMayberry1, line 201: nit: could we swap the order of these parameters? in my mind, `player` is the thing being acted on and `contextPlayer` is the supporting data

server/game/gameSystems/PutIntoPlaySystem.ts
```diff
@@ -82,14 +83,7 @@ export class PutIntoPlaySystem<TContext extends AbilityContext = AbilityContext>
...
-    private getFinalController(controller: RelativePlayer, context: TContext) {
-        switch (controller) {
-            case RelativePlayer.Self:
-                return context.player;
-            case RelativePlayer.Opponent:
-                return context.player.opponent;
-            default:
-                Contract.fail(`Unknown value of RelativePlayer: ${controller}`);
-        }
+    private getFinalController(controller: Player | RelativePlayer, context: TContext) {
```
AMMayberry1, line 86: nit: now that we have `asConcretePlayer` we can probably just remove this method

brianries: removed method

server/game/core/utils/EnumHelpers.ts
brianries: Rearranged

AMMayberry1 (approved):

PR #556, opened 2025-02-08
Osi Sobeck

VJubert:


server/game/cards/03_TWI/units/FinnOnTheRun.ts
```diff
@@ -0,0 +1,43 @@
...
+export default class FinnOnTheRun extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '7244268162',
+            internalName: 'finn#on-the-run',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
```
AMMayberry1, line 14: nit: needs to use `addReplacementEffectAbility`, I think technically this setup works but I can't guarantee it. `addReplacementEffectAbility` is pretty well-established so we should have higher confidence that it works in all cases

test/server/cards/03_TWI/units/FinnOnTheRun.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+                },
+                player2: {
+                    hand: ['daring-raid'],
+                    groundArena: ['battlefield-marine', 'maul#shadow-collective-visionary'],
+                }
+            });
+
+            const { context } = contextRef;
+
+            // trigger finn ability
```
AMMayberry1, line 17: Since replacement effects are pretty tricky and can fail in a lot of ways, we need to test pretty thoroughly when we write them. A few categories of tests we need to add:

- Can you take a look at the Boba Fett's Armor test cases and make sure you have all the similar cases covered?

- It would be good to bring in case 4 from the Maul1 tests so we can confirm that the damage still counts as combat damage correctly. We would also need a similar case where we test to confirm that event damage with e.g. Daring Raid does _not_ trigger the Tarfful ability.

- The Jango PR is almost done, if it gets finished in time it would be good to use Jango here to confirm that the source is correctly propagated. E.g., if Jango's unit attacks into the Finn target, Jango should still get his ability trigger.

- One really interesting one we can do now would be to chain the Finn effect -> Maul1 effect -> shield effect. I.e., have Finn target Maul1, then Maul1 ability redirects damage onto a unit with a shield

- It would also be good to confirm Finn -> Maul1 reduces damage right, meaning that the target Underworld unit receives 1 less damage due to the Finn effect

Let me know if any of the above don't work, the replacement effect system is still a little temperamental and very difficult to debug unfortunately

test/server/cards/03_TWI/units/OsiSobeckWardenOfTheCitadel.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+            context.player2.moveCard(context.wampa, 'groundArena');
+
+            context.player2.passAction();
+
+            // play osi sobeck free with palpatine's return
+            context.player1.clickCard(context.palpatinesReturn);
+            context.player1.clickCard(context.osiSobeck);
+
+            // cannot capture anybody
+            expect(context.player2).toBeActivePlayer();
```
AMMayberry1, line 58: Could we add a token ground unit? I think he actually should be able to capture that since its cost is technically 0

test/server/cards/03_TWI/units/OsiSobeckWardenOfTheCitadel.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+            });
+
+            const { context } = contextRef;
+
+            // play osi sobeck using 6 resources
+            context.player1.clickCard(context.osiSobeck);
+            context.player1.clickPrompt('Play Osi Sobeck');
+
+            // can capture any enemy ground unit which cost 6 or less
+            expect(context.player1).toBeAbleToSelectExactly([context.scoutBikePursuer, context.wampa, context.atst]);
```
AMMayberry1, line 25: For these checks, can we add a unit that costs exactly one more in each case to confirm the filter? E.g. a 7 cost, 5 cost, and 1 cost unit

server/game/cards/03_TWI/units/FinnOnTheRun.ts
VJubert: Finn ability must target unit before adding his replacement and should only trigger when attack completes, I can't use addReplacementAbility here

test/server/cards/03_TWI/units/FinnOnTheRun.spec.ts
VJubert: I add few more tests as you want, I don't understand what you wannna test in your second point and I'm waiting for Jango but I think tarfful test should cover the same case

AMMayberry1 (approved):

PR #559, opened 2025-02-10
No Disintegrations and Hound's Tooth

VJubert:

### DIFF
--- a/test/server/cards/04_JTL/events/NoDisintegrations.spec.ts
+++ b/test/server/cards/04_JTL/events/NoDisintegrations.spec.ts
@@ -35,6 +35,12 @@ describe('No Disintegrations', function() {
             context.player1.clickCard(context.noDisintegrations);
             context.player1.clickCard(context.battlefieldMarine);
             expect(context.battlefieldMarine.damage).toBe(3);
+
+            context.player2.passAction();
+            context.player1.moveCard(context.noDisintegrations, 'hand');
+            context.player1.clickCard(context.noDisintegrations);
+            context.player1.clickCard(context.wampa);
+            expect(context.wampa.damage).toBe(4);
         });
     });
 });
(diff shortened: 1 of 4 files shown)


test/server/cards/04_JTL/units/HoundsToothReliableAndDeadly.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+            context.player2.clickCard(context.strafingGunship);
+            context.player2.clickCard(context.p1Base);
+
+            context.houndsTooth.exhausted = false;
+
+            // attack an exhausted unit which didn't enters play this phase, dealing damage before defender
+            context.player1.clickCard(context.houndsTooth);
+            context.player1.clickCard(context.strafingGunship);
+            expect(context.houndsTooth.damage).toBe(0);
+            expect(context.strafingGunship).toBeInZone('discard');
```
AMMayberry1, line 49: I don't think we've tested return from capture with this state watcher yet, since that technically counts as "enters play."

Let's set up a test where a space unit is captured in one phase abs then rescued the next phase, and confirm that the Hound's Tooth ability _doesn't_ trigger

test/server/cards/04_JTL/events/NoDisintegrations.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+        it('No Disintegrations\'s ability should deal damage to a non leader unit equal to its remaining hp minus 1', function() {
+            contextRef.setupTest({
+                phase: 'action',
+                player1: {
+                    hand: ['no-disintegrations'],
+                    groundArena: ['atst']
+                },
+                player2: {
+                    groundArena: [{ card: 'wampa', damage: 3 }, { card: 'battlefield-marine', upgrades: ['experience'] }],
+                    spaceArena: ['green-squadron-awing'],
```
AMMayberry1, line 12: Can we add a unit that has one remaining hp and confirm that it's still selectable with the ability

AMMayberry1 (approved):

PR #560, opened 2025-02-10
Add Endless Legions implementation

Frugghi:
image

### DIFF
--- a/server/game/cards/02_SHD/events/EndlessLegions.ts
+++ b/server/game/cards/02_SHD/events/EndlessLegions.ts
@@ -36,7 +36,7 @@ export default class EndlessLegions extends EventCard {
                 cardTypeFilter: WildcardCardType.Unit,
                 zoneFilter: ZoneName.Resource,
                 controller: RelativePlayer.Self,
-                ignoreHiddenZoneRule: false,
+                ignoreHiddenZoneRule: true,
                 cardCondition: (card: Card) => Helpers.asArray(revealedCardsContext.target).includes(card) && !playedCards.includes(card),
                 immediateEffect: AbilityHelper.immediateEffects.playCardFromOutOfPlay({
                     adjustCost: { costAdjustType: CostAdjustType.Free },
--- a/server/game/cards/02_SHD/events/EndlessLegions.ts
+++ b/server/game/cards/02_SHD/events/EndlessLegions.ts
@@ -1,0 +1,50 @@
+import AbilityHelper from '../../../AbilityHelper';
+import * as Helpers from '../../../core/utils/Helpers';
+import { EventCard } from '../../../core/card/EventCard';
+import { RelativePlayer, TargetMode, WildcardCardType, ZoneName } from '../../../core/Constants';
+import type { Card } from '../../../core/card/Card';
+import { CostAdjustType } from '../../../core/cost/CostAdjuster';
+import type { AbilityContext } from '../../../core/ability/AbilityContext';
+import type { IThenAbilityPropsWithSystems } from '../../../Interfaces';
+
+export default class EndlessLegions extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '5576996578',
+            internalName: 'endless-legions',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Reveal any number of resources you control',
+            targetResolver: {
+                mode: TargetMode.Unlimited,
+                zoneFilter: ZoneName.Resource,
+                controller: RelativePlayer.Self,
+                immediateEffect: AbilityHelper.immediateEffects.reveal(),
+            },
+            then: (context) => this.playRevealedCard([], context),
+        });
+    }
+
+    private playRevealedCard(playedCards: Card[], revealedCardsContext: AbilityContext): IThenAbilityPropsWithSystems<AbilityContext> {
+        return {
+            title: 'Play a revelead unit for free',
+            targetResolver: {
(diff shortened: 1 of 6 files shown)


server/game/cards/02_SHD/events/EndlessLegions.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    private playRevealedCard(playedCards: Card[], revealedCardsContext: AbilityContext): IThenAbilityPropsWithSystems<AbilityContext> {
+        return {
+            title: 'Play a revelead unit for free',
+            targetResolver: {
+                activePromptTitle: 'Choose a unit to play for free',
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: ZoneName.Resource,
+                controller: RelativePlayer.Self,
+                choosingFromHiddenOverride: false,
```
AMMayberry1, line 39: nit: the naming here feels a little counterintuitive to me, I would assume that the boolean indicates whether we are engaging the override, not setting the value of the override.

Maybe inverting the logic and naming it something like `ignoreHiddenZoneRule` or would be clearer?

server/game/gameSystems/PutIntoPlaySystem.ts
```diff
@@ -56,7 +57,7 @@ export class PutIntoPlaySystem<TContext extends AbilityContext = AbilityContext>
             return false;
         } else if (!card.canBeInPlay() || card.isInPlay()) {
             return false;
-        } else if (card.zoneName === ZoneName.Resource && !card.hasSomeKeyword(KeywordName.Smuggle)) {
+        } else if (card.zoneName === ZoneName.Resource && !card.hasSomeKeyword(KeywordName.Smuggle) && !(context.playType === PlayType.Smuggle || context.playType === PlayType.PlayFromOutOfPlay)) {
```
AMMayberry1, line 60: I _think_ we can remove the hasSomeKeyword check if we're checking `context.playType`, since that will always return PlayType.Smuggle if we're trying to smuggle.

nit: if we do need all of these checks, can we break this into multiple lines just readability

test/server/cards/02_SHD/events/EndlessLegions.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+            context.player1.clickCard(context.resupply);
+            context.player1.clickCard(context.relentless);
+            context.player1.clickCard(context.arquitensAssaultCruiser);
+            context.player1.clickCard(context.cloneCommanderCody);
+            context.player1.clickCard(context.admiralPiett);
+            context.player1.clickCard(context.frozenInCarbonite);
+            context.player1.clickCard(context.wrecker);
+            context.player1.clickPrompt('Done');
+
+            expect(context.getChatLogs(1)).toContain('player1 reveals Resupply, Relentless, Arquitens Assault Cruiser, Clone Commander Cody, Admiral Piett, Frozen in Carbonite, Wrecker due to Endless Legions');
```
AMMayberry1, line 65: Could we switch over to using the version of this in the Childsen tests. That's a little less brittle if we change the exact wording of the reveal log

test/server/cards/02_SHD/events/EndlessLegions.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+                context.relentless,
+            ]);
+
+            // Player 1 plays Wrecker for free
+            context.player1.clickCard(context.wrecker);
+            context.player1.clickPrompt('Defeat a friendly resource. If you do, deal 5 damage to a ground unit');
+
+            // Player 1 defeats Relentless to deal 5 damage to Wampa
+            context.player1.clickCard(context.relentless);
+            context.player1.clickCard(context.wampa);
```
AMMayberry1, line 100: Can we use this to defeat something that has a "when defeated" ability to confirm the timing there

test/server/cards/02_SHD/events/EndlessLegions.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+                context.wrecker,
+                context.arquitensAssaultCruiser,
+                context.cloneCommanderCody,
+                context.relentless,
+            ]);
+
+            // Player 1 plays Clone Commander Cody for free
+            context.player1.clickCard(context.cloneCommanderCody);
+
+            expect(context.player1).toBeAbleToSelectExactly([
```
AMMayberry1, line 88: A neat test we can add would be to put out a 97th Legion with some damage on it and confirm that it gets defeated at the right time point as cards get played out

server/game/cards/02_SHD/events/EndlessLegions.ts
```diff
@@ -0,0 +1,50 @@
+import AbilityHelper from '../../../AbilityHelper';
+import * as Helpers from '../../../core/utils/Helpers';
+import { EventCard } from '../../../core/card/EventCard';
+import { RelativePlayer, TargetMode, WildcardCardType, ZoneName } from '../../../core/Constants';
+import type { Card } from '../../../core/card/Card';
+import { CostAdjustType } from '../../../core/cost/CostAdjuster';
+import type { AbilityContext } from '../../../core/ability/AbilityContext';
+import type { IThenAbilityPropsWithSystems } from '../../../Interfaces';
+
+export default class EndlessLegions extends EventCard {
```
AMMayberry1, line 10: One last thing I forgot sorry - can you add the unimplemented override for this one as well, we're migrating `reveal()` to use a real prompt so until it's in place this won't work in the FE. I'm doing the update pass on all the cards soon once the system is finished.

```ts
```

AMMayberry1 (approved):

PR #565, opened 2025-02-11
transform data from FE for stateful prompt

danbastin:
Transfor card uuids to the card objects to be consumed by prompt

### DIFF
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -767,9 +767,21 @@ class Game extends EventEmitter {
      */
     statefulPromptResults(playerId, result, uuid) {
         var player = this.getPlayerById(playerId);
-
+        result = this.transformStatefulResult(result);
         // check to see if the current step in the pipeline is waiting for input
         return this.pipeline.handleStatefulPromptResults(player, result, uuid);
+    }
+
+    transformStatefulResult(result) {
+        const data = new Map(
+            result.targets
+                .map((target) => {
+                    const card = this.allCards.find((card) => target.uuid === card.uuid);
+                    return card ? [card, target.amount] : null;
+                })
+                .filter(Boolean)
+        );
+        return { type: result.type, valueDistribution: data };
     }
 
     /**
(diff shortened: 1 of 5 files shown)


server/game/core/Game.js
```diff
@@ -767,11 +767,23 @@ class Game extends EventEmitter {
      */
     statefulPromptResults(playerId, result, uuid) {
         var player = this.getPlayerById(playerId);
-
+        result = this.transformStatefulResult(result);
         // check to see if the current step in the pipeline is waiting for input
         return this.pipeline.handleStatefulPromptResults(player, result, uuid);
     }
 
+    transformStatefulResult(result) {
```
AMMayberry1, line 775: Long term I'd much rather pass the raw data all the way down to the prompt and let the prompt do its own checking, but I'm fine with this for now to unblock the FE work

AMMayberry1 (dismissed):

server/game/core/gameSteps/PromptInterfaces.ts
```diff
@@ -41,7 +41,13 @@ export type DistributePromptType =
   | StatefulPromptType.DistributeExperience
   | StatefulPromptType.DistributeHealing;
 
-export type IStatefulPromptResults = IDistributeAmongTargetsPromptResults;
+export interface IStatefulPromptResults {
```
AMMayberry1, line 44: Can we make this change to the distribute prompt results type instead. This was set up in case we end up with other stateful prompt types besides the distribute one

server/game/core/gameSteps/prompts/DistributeAmongTargetsPrompt.ts
```diff
@@ -103,14 +103,28 @@ export class DistributeAmongTargetsPrompt extends UiPrompt {
...
 
         return true;
     }
 
-    private assertPromptResultsValid(player: Player, results: IStatefulPromptResults): asserts results is IDistributeAmongTargetsPromptResults {
+    private formatPromptResults(results: IStatefulPromptResults): IDistributeAmongTargetsPromptResults {
+        const targets = new Map(
+            results.valueDistribution
+                .map((target) => {
+                    const card = this.game.allCards.find((card) => target.uuid === card.uuid);
```
AMMayberry1, line 118: Can we check this against the list in `this.properties.legalTargets`

danbastin: This check is covered in `assertPromptResultsValid`

AMMayberry1 (approved):

PR #568, opened 2025-02-12
SOR Display Prompt Updates: ISB Agent, Chimaera, Bodhi Rook

moyerr:


test/server/cards/01_SOR/units/BodhiRook.spec.ts
```diff
@@ -20,31 +20,36 @@ describe('Bodhi Rook', function () {
...
+                });
+                context.player1.clickCardInDisplayCardPrompt(context.protector);
                 expect(context.protector).toBeInZone('discard');
 
                 context.player2.passAction();
                 context.player1.moveCard(context.bodhiRook, 'hand');
 
                 // No choice here so no prompt, but the player still sees an opponent hand reveal
                 context.player1.clickCard(context.bodhiRook);
-                expect(context.getChatLogs(1)).toContain('Bodhi Rook sees Sabine Wren, Battlefield Marine, and Inferno Four');
```
AMMayberry1, line 47: Can we add at least one or two checks that the named cards are _not_ emitted in the logs. Can use the same pattern as the reveal effect cards, just with `expect().not.toContain()`

test/server/cards/01_SOR/units/ISBAgent.spec.ts
```diff
@@ -34,6 +34,9 @@ describe('ISB Agent', function() {
                         'player1 reveals Confiscate due to ISB Agent',
                     ]);
 
+                    expect(context.player2).toHaveExactViewableDisplayPromptCards([context.confiscate]);
+                    context.player2.clickPrompt('Done');
```
AMMayberry1, line 38: Let's also check the log here since this is reveal

AMMayberry1 (approved):

PR #574, opened 2025-02-13
Indirect damage, Planetary Bombardment, Devastator, and First Order Stormtrooper

Frugghi:
| Planetary Bombardment | Devastator | First Order Stormtrooper |
| --- | --- | --- |
| image | image | image |


server/game/core/gameSteps/prompts/DistributeAmongTargetsPrompt.ts
```diff
@@ -166,5 +169,15 @@ export class DistributeAmongTargetsPrompt extends UiPrompt {
...
+
+        if (results.type === StatefulPromptType.DistributeIndirectDamage) {
+            const cardsWithOverdistributedDamage = Array.from(results.valueDistribution.entries())
+                .filter((entry) => entry[0].isUnit() && entry[0].remainingHp < entry[1])
+                .map((entry) => entry[0]);
+            Contract.assertFalse(
+                cardsWithOverdistributedDamage.length > 0,
+                `Illegal prompt results '${this._activePrompt.menuTitle}', distributed ${this.distributeType} should not exceed the remaining HP of the target units: ${cardsWithOverdistributedDamage.map((card) => card.internalName).join(', ')}`
+            );
+        }
```
Frugghi, line 181: I guess this needs to be enforced on the frontend as well?

server/game/TargetInterfaces.ts
```diff
@@ -82,7 +81,7 @@ export interface ITargetResolverBase<TContext extends AbilityContext> {
 // TODO: add functionality to PlayerTargetResolver to autodetect any invalid target players.
 export interface IPlayerTargetResolver<TContext extends AbilityContext> extends ITargetResolverBase<TContext> {
     mode: TargetMode.Player | TargetMode.MultiplePlayers;
-    immediateEffect?: PlayerTargetSystem<TContext>;
+    immediateEffect?: GameSystem<TContext>;
```
Frugghi, line 84: Had to change this to be able to use a conditional effect with a player target resolver.

server/game/gameSystems/DamageSystem.ts
```diff
@@ -76,7 +79,8 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext, TPro
 
     protected override defaultProperties: IAbilityDamageProperties = {
         amount: null,
-        type: DamageType.Ability
+        type: DamageType.Ability,
+        indirect: false
```
AMMayberry1, line 83: nit: could we call this and the property on the event `isIndirect` for clarity + consistency

server/game/ongoingEffects/OngoingEffectLibrary.ts
```diff
@@ -62,6 +62,7 @@ export = {
     //    }),
 
     canPlayFromDiscard: () => OngoingEffectBuilder.card.static(EffectName.CanPlayFromDiscard),
+    canAssignIndirectDamageDealtToOpponents: () => OngoingEffectBuilder.player.static(EffectName.CanAssignIndirectDamageDealtToOpponents),
```
AMMayberry1, line 65: "Can" implies a choice to me, can we say "assignsIndirectDamageDealtToOpponents" or something similar

server/game/core/gameSteps/prompts/DistributeAmongTargetsPrompt.ts
AMMayberry1: Yeah this is a case where we have to count on the FE to get it right

test/server/cards/04_JTL/units/DevastatorHuntingTheRebellion.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+            ]));
+
+            expect(context.cartelSpacer.damage).toBe(2);
+            expect(context.p2Base.damage).toBe(2);
+
+            // Player 2 defeats Devastator
+            context.player2.clickCard(context.rivalsFall);
+            context.player2.clickCard(context.devastator);
+
+            // Plauer 1 plays Planetary Bombardment and they don't assign indirect damage anymore
```
AMMayberry1, line 36: Can we also do a Bombardment before Devastator is defeated to confirm its effect works on more than just itself. Also wouldn't hurt to test that the opponent doesn't get the effect

AMMayberry1:
Can we also add a test suite for this under `gameSystems/` to validate the basics and a few edge cases. Would be good to test with Shield, Boba Fett's Armor, and LTP.

server/game/cards/01_SOR/tokens/Shield.ts
```diff
@@ -28,7 +28,7 @@ export default class Shield extends TokenUpgradeCard {
         this.addReplacementEffectAbility({
             title: 'Defeat shield to prevent attached unit from taking damage',
             when: {
-                onDamageDealt: (event, context) => event.card === context.source.parentCard
+                onDamageDealt: (event, context) => event.card === context.source.parentCard && !(event.indirect ?? false)
```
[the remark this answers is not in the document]
Frugghi: I think it needs to be `!!event.isIndirect` to evaluate correctly.

server/game/gameSystems/DistributeIndirectDamageSystem.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public override generatePropertiesFromContext(context: TContext, additionalProperties: any = {}): IDistributeAmongTargetsSystemProperties<AbilityContext<Card>> {
+        const properties = super.generatePropertiesFromContext(context, {
+            ...additionalProperties,
+            player: RelativePlayer.Opponent,
+            controller: this.properties.player ?? RelativePlayer.Opponent,
+        });
+
+        if (context.player.canAssignIndirectDamageDealtToOpponents()) {
```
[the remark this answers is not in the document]
Frugghi: Not sure I understand this, I think logically it makes more sense here because it's the system that handles distributing the damage. If we move it out and then in the future we add more systems that call `DistributeIndirectDamageSystem`, we would need to repeat the logic everywhere. IMO, `IndirectDamageSystem` should only care about the amount of damages and who is the target.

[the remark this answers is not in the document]
AMMayberry1: Oh gosh yeah my comment is totally wrong, I was mulling over Hunting Aggressor and just got some wires crossed (I was pretty tired when I did the review). Apologies for that, thanks for taking the time to reply.

server/game/gameSystems/IndirectDamageToPlayerSystem.ts
```diff
@@ -0,0 +1,63 @@
...
+export interface IIndirectDamageToPlayerProperties extends IPlayerTargetSystemProperties {
+    amount: number;
+}
+
+export class IndirectDamageToPlayerSystem<TContext extends AbilityContext = AbilityContext> extends PlayerTargetSystem<TContext, IIndirectDamageToPlayerProperties> {
+    public override readonly name = 'indirectDamage';
+    public override readonly eventName = EventName.OnIndirectDamageDealt;
+
+    protected override defaultProperties: IIndirectDamageToPlayerProperties = {
+        amount: 1
```
AMMayberry1, line 19: Not sure it makes sense to have a default amount here - there's not an abundance of cards that only do 1 indirect damage, I think this would just facilitate people forgetting to put the value (`DamageSystem` has this as `null` fwiw)

server/game/gameSystems/DistributeIndirectDamageSystem.ts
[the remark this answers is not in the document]
Frugghi: > One thing I will note is that since it's a bit unusual that we have to split a system into two pieces like this, I would advocate for never invoking this system except from IndirectDamageSystem. Ideally we could join them into one in a straightforward way, but I don't think that's really doable

100% agree. Ideally we should never call this outside `IndirectDamageSystem` (but you never know what the next card will say 😂 ) and it'd be great to combine them into one but I don't see it possible atm since indirect damage targets players and distributing damage targets cards 😞

test/server/gameSystems/IndirectDamageToPlayerSystem.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+                        [context.chirrutImwe, 1],
+                    ]))
+                ).toThrowError();
+
+                context.player2.setDistributeIndirectDamagePromptState(new Map([
+                    [context.p2Base, 1],
+                ]));
+            });
+
+            // TODO: Add test for indirect damage with Finn TWI unit
```
AMMayberry1, line 101: I think we're okay without this actually, the test here is more to confirm that generic reduction effects are triggered correctly. If we do decide to have specific tests for other units we can do them in the test suites for those cards.

server/game/gameSystems/DistributeIndirectDamageToCardsSystem.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    public override generatePropertiesFromContext(context: TContext, additionalProperties: any = {}): IDistributeAmongTargetsSystemProperties<AbilityContext<Card>> {
+        const properties = super.generatePropertiesFromContext(context, {
+            ...additionalProperties,
+            player: RelativePlayer.Opponent,
+            controller: this.properties.player ?? RelativePlayer.Opponent,
+        });
+
+        if (context.player.assignIndirectDamageDealtToOpponents()) {
+            properties.player = RelativePlayer.Self;
```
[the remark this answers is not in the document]
Frugghi: Can't do it there because the additional properties don't override the properties that you passed when you initialise the system.

AMMayberry1 (approved):

PR #581, opened 2025-02-13
add the clone wars

ssoulier:
[The Clone Wars](https://swudb.com/card/TWI/125/the-clone-wars)

Screenshot 2025-02-13 at 21 06 30


server/game/cards/03_TWI/events/TheCloneWars.ts
```diff
@@ -0,0 +1,44 @@
...
+    private _readyResourcesCount(context): number {
+        return context.source.controller.resources.filter((card) => !card.exhausted).length;
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Pay any number of resources.',
+            targetResolver: {
+                mode: TargetMode.DropdownList,
+                options: (context) => Array.from({ length: this._readyResourcesCount(context) + 1 }, (x, i) => `${i}`),
```
ssoulier, line 22: I added the `context` here. Let me know if it's not the right approach.

server/game/cards/03_TWI/events/TheCloneWars.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export default class TheCloneWars extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '2267524398',
+            internalName: 'the-clone-wars'
+        };
+    }
+
+    private _readyResourcesCount(context): number {
```
AMMayberry1, line 13: nit: our convention is to only use the leading `_` on private fields that back properties. you can see some examples in `Card.ts`

server/game/cards/03_TWI/events/TheCloneWars.ts
```diff
@@ -0,0 +1,44 @@
...
+        return context.source.controller.resources.filter((card) => !card.exhausted).length;
```
AMMayberry1, line 14: nit: we need to start using `context.player` everywhere instead of `context.source.controller`, the former works better in some edge cases we've been running into

server/game/cards/03_TWI/events/TheCloneWars.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Pay any number of resources.',
+            targetResolver: {
+                mode: TargetMode.DropdownList,
+                options: (context) => Array.from({ length: this._readyResourcesCount(context) + 1 }, (x, i) => `${i}`),
+                condition: (context) => this._readyResourcesCount(context) > 0   // skip ability if there is no resources available
```
AMMayberry1, line 23: Let's also include this as a `thenCondition` so resolution of the `then` can just be shortcut

server/game/cards/03_TWI/events/TheCloneWars.ts
```diff
@@ -0,0 +1,44 @@
...
+            },
+            then: (thenContext) => ({
+                title: 'Create that many Clone Trooper tokens. Each opponent creates that many Battle Droid tokens.',
+                immediateEffect: AbilityHelper.immediateEffects.sequential([
+                    AbilityHelper.immediateEffects.payResourceCost({
+                        amount: parseInt(thenContext.select),
+                        target: thenContext.source.controller,
```
AMMayberry1, line 30: Same comment on these, all need to be `thenContext.player`

server/game/cards/03_TWI/events/TheCloneWars.ts
AMMayberry1: It is! Nice job

server/game/cards/03_TWI/events/TheCloneWars.ts
ssoulier: I'll add it but It's not clear to me what it will change.

AMMayberry1 (approved):

PR #587, opened 2025-02-14
Implement For a Cause I Believe In

moyerr:
This card uses the new display prompt:

### DIFF
--- a/test/server/cards/01_SOR/events/ForACauseIBelieveIn.spec.ts
+++ b/test/server/cards/01_SOR/events/ForACauseIBelieveIn.spec.ts
@@ -18,16 +18,11 @@ describe('For A Cause I Believe In', function() {
 
             context.player1.clickCard(context.forACauseIBelieveIn);
 
-            // P2 gets to see the top 4 cards of P1's deck
-            expect(context.player2).toHaveEnabledPromptButton('Done');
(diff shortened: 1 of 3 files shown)


test/server/cards/01_SOR/events/ForACauseIBelieveIn.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+
+            // SpecForce Soldier should be on the top of deck
+            expect(context.player1.deck[0]).toBe(context.specforceSoldier);
+            expect(context.player1.deck[1]).toBe(context.battlefieldMarine);
+
+            expect(context.echoBaseDefender).toBeInZone('discard');
+            expect(context.regionalSympathizers).toBeInZone('discard');
+        });
+
+        it('deals 0 damage to opponents base when 0 heroism cards are revealed', function() {
```
moyerr, line 53: @AMMayberry1 I split the different cases into different specs for this because the setup is different for each. Test suite still runs in less than a minute, but let me know if you want me to consolidate any of these.

AMMayberry1: That's totally fine - I'm actually going to make an announcement soon, we're going to relax about consolidating test cases due to some new information

AMMayberry1 (approved):

PR #588, opened 2025-02-14
Switch to using interfaces for card types

AMMayberry1:
With Piloting around the corner, we need to have more loose coupling between the specific card class implementations and how they present themselves to the various systems. This will allow a unit or leader to present itself as an upgrade when necessary.

The problem was intractable with our previous design, so we have switched to primarily using interfaces for reasoning about card types.

This is mostly represented in our type assertions such as `isUnit()`, `isUpgrade()` etc. Whereas before they would be of type `this is Unit`, now it is `this is IUnit`. Unfortunately this change is _extremely_ noisy.

While moving things around, found a lot of opportunities for refactoring to clean up the type hierarchy and make it so that there are fewer instances of methods reproduced in multiple places in the type hierarchy or classes having access to methods that they shouldn't based on their card type.


adbronico (dismissed):
Tests pass, we happy

Issue #575, opened 2025-02-13
Ketsu Onyo, Old Friend | Ability triggers every time any base is damaged


moyerr:
Steps to reproduce:

1. Play Ketsu Onyo
2. Attack any base with any unit
3. Result:

<img width=400 src=https://github.com/user-attachments/assets/eec58d88-d1ac-4c72-8cf6-b5b1937d714a>

PR #591, opened 2025-02-14
#575 + moff gideon

VJubert:


test/server/cards/04_JTL/units/MoffGideonIKnowEverything.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+            context.player2.clickCard(context.jawaScavenger);
+            expect(context.player2.exhaustedResourceCount).toBe(exhaustedResourceCount + 2);
+            exhaustedResourceCount = context.player2.exhaustedResourceCount;
+
+            context.player1.passAction();
+
+            // event does not cost more
+            context.player2.clickCard(context.droidDeployment);
+            expect(context.player2.exhaustedResourceCount).toBe(exhaustedResourceCount + 2);
+        });
```
AMMayberry1, line 107: Can we attack a unit with Gideon (no overwhelm damage) and confirm it doesn't trigger. Would be good to do for the other cards as well if they don't have that test already

AMMayberry1 (approved):

PR #596, opened 2025-02-15
Barris offee

ssoulier:
[Barriss Offee](https://swudb.com/card/TWI/042/barriss-offee-unassuming-apprentice)

Screenshot 2025-02-15 at 12 15 54


server/game/stateWatchers/UnitsHealedThisPhaseWatcher.ts
```diff
@@ -0,0 +1,84 @@
...
+    public wasHealedThisPhase(card: IUnitCard, inPlayId?: number): boolean {
+        const inPlayIdToCheck = inPlayId ?? (card.isInPlay() ? card.inPlayId : card.mostRecentInPlayId);
+
+        return this.getCurrentValue().some(
+            (entry) => entry.unit === card && entry.inPlayId === inPlayIdToCheck
+        );
+    }
+
+    /** Check if there is some units controlled by player that was healed this phase */
+    public someHealedUnitControlledByPlayer(controller: Player): boolean {
```
AMMayberry1, line 66: Sorry to have to ask this - since this is the only helper method wrapping `getCurrentValue()` that's used, let's remove the others for now since they wouldn't be getting checked in with any test coverage and it's easy to get these wrong due to issues around `inPlayId`. It could cause a lot of headache if we accidentally check one in that has a bug and someone uses it without realizing it doesn't have test coverage

test/server/cards/03_TWI/units/BarrissOffeeUnassumingApprentice.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+            context.player1.clickCard(firstBattlefieldMarine);
+            expect(firstBattlefieldMarine.getPower()).toBe(4);
+            expect(firstBattlefieldMarine.getHp()).toBe(3);
+
+            // We defeat the first copy of Battlefield Marine
+            context.player2.clickCard(context.atst);
+            context.player2.clickCard(firstBattlefieldMarine);
+            expect(firstBattlefieldMarine).toBeInZone('discard');
+
+            // We play a new copy of Battlefield Marine, it should not have +1/+0
```
AMMayberry1, line 125: Tests look great! Didn't miss anything that I can see, nice job

server/game/stateWatchers/UnitsHealedThisPhaseWatcher.ts
ssoulier: If I understand correctly you suggest keeping only the method `wasHealedThisPhase` that is used in Barriss Offee implementation and to remove all the others?
No problem. I think it makes sense to add them only when required for some future cards.

PR #598, opened 2025-02-15
Reprocess

ssoulier:
[Reprocess](https://swudb.com/card/TWI/088/reprocess)

Screenshot 2025-02-15 at 18 33 34


test/server/cards/03_TWI/events/Reprocess.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+            expect(context.atst).toBeInZone('discard');
+            expect(context.wampa).toBeInZone('discard');
+            expect(context.restock).toBeInZone('discard');
+            expect(context.resupply).toBeInZone('discard');
+
+            const battleDroids = context.player1.findCardsByName('battle-droid');
+            expect(battleDroids.length).toBe(4);
+
+            const opponentBattleDroids = context.player2.findCardsByName('battle-droid');
+            expect(opponentBattleDroids.length).toBe(0);
```
AMMayberry1, line 61: Can we do one more test where we select less than 4 targets.

I'll announce this soon but we are relaxing the restriction on creating multiple test cases, it turns out the github actions time is not a problem like we thought. So you could create a `beforeEach()` here and just select a different number of cards in the second case.

ssoulier: Thanks. I did 3 tests without `beforeEach` because of the third tests that requires no units in the discard pile.

AMMayberry1 (approved):

PR #601, opened 2025-02-15
Add No Glory, No Results and Profundity implementation

Frugghi:
| No Glory, No Results | Profundity |
| --- | --- |
| image | image |

### DIFF
--- a/server/game/cards/01_SOR/units/GreedoSlowOnTheDraw.ts
+++ b/server/game/cards/01_SOR/units/GreedoSlowOnTheDraw.ts
@@ -16,7 +16,7 @@ export default class GreedoSlowOnTheDraw extends NonLeaderUnitCard {
             optional: true,
             immediateEffect: AbilityHelper.immediateEffects.discardFromDeck((context) => ({
                 amount: 1,
-                target: context.source.controller
+                target: context.player
             })),
             ifYouDo: (context) => ({
                 title: 'Deal 2 damage to a ground unit',
--- a/server/game/cards/04_JTL/events/NoGloryOnlyResults.ts
+++ b/server/game/cards/04_JTL/events/NoGloryOnlyResults.ts
@@ -16,13 +16,13 @@ export default class NoGloryOnlyResults extends EventCard {
             targetResolver: {
                 cardTypeFilter: WildcardCardType.NonLeaderUnit,
                 immediateEffect: AbilityHelper.immediateEffects.takeControlOfUnit((context) => ({
-                    newController: context.source.controller,
+                    newController: context.player,
                 }))
             },
-            then: (context) => ({
-                title: `Defeat ${context.target.title}`,
+            then: (thenContext) => ({
+                title: `Defeat ${thenContext.target.title}`,
                 immediateEffect: AbilityHelper.immediateEffects.defeat({
-                    target: context.target,
+                    target: thenContext.target,
                 })
             }),
         });
--- a/server/game/cards/04_JTL/events/NoGloryOnlyResults.ts
+++ b/server/game/cards/04_JTL/events/NoGloryOnlyResults.ts
@@ -1,0 +1,30 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { WildcardCardType } from '../../../core/Constants';
+
+export default class NoGloryOnlyResults extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '9810057689',
+            internalName: 'no-glory-only-results',
+        };
+    }
+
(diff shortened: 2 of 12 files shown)


server/game/cards/01_SOR/units/GreedoSlowOnTheDraw.ts
```diff
@@ -16,7 +16,7 @@ export default class GreedoSlowOnTheDraw extends NonLeaderUnitCard {
             optional: true,
             immediateEffect: AbilityHelper.immediateEffects.discardFromDeck((context) => ({
                 amount: 1,
-                target: context.source.controller
+                target: context.player
```
Frugghi, line 19: In "when defeated" abilities we should always use `context.player` vs `context.source.controller`, correct? I fixed all usages I could find.

AMMayberry1: Thanks! Yeah "when defeated" abilities were causing the problem, so those definitely needed converted. My goal is actually to convert everything in the backlog to using `context.player` so that everyone will use that by default going forward, afaik there's almost never a case where it's incorrect to do so and if we just default to that then we'll reduce the frequency of accidentally forgetting to use it at the right time.

I definitely wouldn't tackle that in this PR, I have a github issue up and we'll get around to it eventually. Now that the when-defeated's are covered we're probably safe from the actual bugs, so it's more of a hygiene thing at this point. Thanks for taking care of that!

server/game/cards/04_JTL/events/NoGloryOnlyResults.ts
```diff
@@ -0,0 +1,30 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Take control of a non-leader unit',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                immediateEffect: AbilityHelper.immediateEffects.takeControlOfUnit((context) => ({
+                    newController: context.source.controller,
```
AMMayberry1, line 19: nit: let's use `context.player`

server/game/cards/04_JTL/events/NoGloryOnlyResults.ts
```diff
@@ -0,0 +1,30 @@
...
+                }))
+            },
+            then: (context) => ({
```
AMMayberry1, line 22: nit: can we use `thenContext` here just for consistency

server/game/cards/04_JTL/units/ProfundityWeFight.ts
```diff
@@ -0,0 +1,35 @@
...
+                onCardDefeated: (event, context) => event.card === context.source,
+            },
+            targetResolver: {
+                activePromptTitle: 'Choose a player to discard a card from their hand',
+                mode: TargetMode.Player,
+                immediateEffect: AbilityHelper.immediateEffects.discardCardsFromOwnHand({ amount: 1 })
+            },
+            then: (thenContext) => ({
+                title: `${thenContext.target} discards a card from their hand`,
+                immediateEffect: AbilityHelper.immediateEffects.conditional({
```
AMMayberry1, line 27: nit: to make this a little cleaner, could use `thenCondition` instead of wrapping the effect with a conditional

test/server/cards/04_JTL/events/NoGloryOnlyResults.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+            context.player1.moveCard(context.noGloryOnlyResults, 'hand');
+            context.player1.clickCard(context.noGloryOnlyResults);
+            expect(context.player1).toBeAbleToSelectExactly([context.senatorialCorvette]);
+
+            // Choose Senatorial Corvette and defeat it
+            context.player1.clickCard(context.senatorialCorvette);
+            expect(context.senatorialCorvette).toBeInZone('discard');
+
+            context.player2.clickCard(context.cartelSpacer);
+            expect(context.cartelSpacer).toBeInZone('discard');
```
AMMayberry1, line 55: Can we add a test with Snoke, there was an FFG clarification explicitly confirming that his ability will activate briefly under the control of the NGOR player

server/game/cards/04_JTL/units/ProfundityWeFight.ts
Frugghi: TIL, I didn't know `thenCondition` was a thing!

AMMayberry1: Yup! We just added `ifYouDoCondition` as well to help keep things clean

AMMayberry1 (approved):

PR #602, opened 2025-02-15
droid missile platform + rampart

VJubert:


server/game/cards/04_JTL/units/RampartEnjoyTheExit.ts
```diff
@@ -0,0 +1,20 @@
...
+            id: '9720757803',
+            internalName: 'rampart#enjoy-the-exit',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'This unit doesn\'t ready during the regroup phase unless its power is 4 or more',
+            condition: (context) => context.source.getPower() < 4 && context.game.currentPhase === PhaseName.Regroup,
+            ongoingEffect: AbilityHelper.ongoingEffects.cardCannot(AbilityRestriction.Ready),
```
AMMayberry1, line 17: So I think for long-term safety we might want to do this slightly differently. What's coded here is "this cannot ready during the regroup phase at all," but I _think_ the intent is that "it doesn't ready during the ready step of the regroup phase."

It would matter if there was ever some effect that messed with ready / exhausted of units during the regroup phase. Currently doesn't exist but better safe imo.

So I would recommend we add a new effect like "DoesNotReadyDuringRegroup" and then check for that effect in `RegroupPhase.readyAllCards()`

test/server/cards/04_JTL/units/RampartEnjoyTheExit.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+            expect(context.greenSquadronAwing.exhausted).toBeFalse();
+
+            // give an upgrade to rampart
+            context.player1.clickCard(context.smugglingCompartment);
+            context.player1.clickCard(context.rampart);
+
+            context.moveToNextActionPhase();
+
+            // rampart should be ready
+            expect(context.rampart.exhausted).toBeFalse();
```
AMMayberry1, line 47: Can we also do a couple timing tests involving lasting effects. Thinking specifically:

- Huyang effect that is active during the regroup phase
- Tactical Advantage test that ends with the action phase and therefore shouldn't count for the Rampart ability

These can be separate test cases now 🥳🥳

AMMayberry1 (approved):

PR #603, opened 2025-02-15
Added capturedUnits property to unit test suite

brianries:
Added support for capturedUnits in the player entries under the unit test suite.  Updated CaptureSystem.spec.ts to use it first.


test/helpers/PlayerInteractionWrapper.js
```diff
@@ -269,6 +277,19 @@ class PlayerInteractionWrapper {
         }
     }
 
+    setCapturedUnits(card, capturedUnits, prevZones = 'any') {
+        for (const capturedUnit of capturedUnits) {
+            const capturedUnitName = (typeof capturedUnit === 'string') ? capturedUnit : capturedUnit.card;
```
AMMayberry1, line 282: We should force the captured unit name to always be a string and throw an error if it's an object, since captured cards can't have any state (damage, upgrades, etc.)

test/helpers/DeckBuilder.js
```diff
@@ -258,6 +268,22 @@ class DeckBuilder {
...
     }
 
+    getCapturedUnitsFromArena(arenaList) {
+        if (!arenaList) {
+            return [];
+        }
+        let capturedUnits = [];
+        for (const card of arenaList) {
+            if (typeof card !== 'string' && card.capturedUnits) {
+                for (const capturedUnit of card.capturedUnits) {
```
AMMayberry1, line 278: We should add this to the `SwuPlayerSetupOptions` interface in IntegrationHelper.d.ts

test/server/gameSystems/CaptureSystem.spec.ts
```diff
@@ -175,36 +175,16 @@ describe('Capture system', function() {
                 contextRef.setupTest({
                     phase: 'action',
                     player1: {
-                        hand: ['take-captive', 'evacuate', 'superlaser-blast'],
-                        groundArena: ['wampa', 'atst'],
-                        spaceArena: ['wing-leader']
+                        hand: ['evacuate', 'superlaser-blast'],
+                        spaceArena: [{ card: 'wing-leader', capturedUnits: ['tieln-fighter'] }]
```
AMMayberry1, line 179: This is one of the few cases where I think the name might be overkill - I think just `captured` is probably fine and saves a little space, if you prefer `capturedUnits` though we can leave it

test/helpers/PlayerInteractionWrapper.js
AMMayberry1: Actually wait I take that back, we might need to be able to specify `owner` for certain tests. We probably should add support for that in this PR.

test/server/gameSystems/CaptureSystem.spec.ts
brianries: Rare case where I'm going to push back as the name matches the getter "capturedUnits" in the code underneath, and the 1:1 matching (even though its just a player entry prop file) could help in future reading/writing

test/helpers/DeckBuilder.js
brianries: We don't have interfaces at the card level yet, can add that when we do?

test/helpers/PlayerInteractionWrapper.js
brianries: Added support for owner on the captured cards and extended the unit tests under the CaptureSystem to test it

test/helpers/DeckBuilder.js
```diff
@@ -278,15 +282,18 @@ class DeckBuilder {
...
+    getCapturedUnitsFromArena(arenaList, ownerFilter = () => true) {
         if (!arenaList) {
             return [];
         }
         let capturedUnits = [];
         for (const card of arenaList) {
             if (typeof card !== 'string' && card.capturedUnits) {
                 for (const capturedUnit of card.capturedUnits) {
-                    capturedUnits.push(capturedUnit);
+                    let capturedUnitName = (typeof capturedUnit === 'string') ? capturedUnit : capturedUnit.card;
```
AMMayberry1, line 293: Can we go ahead and add a check here that the only permissible properties if it's an object are `card` and `owner`. This is mostly useful for catching typos (like if someone typed `onwer` it would be silently ignored)

brianries: Added this test

brianries: Example output:     TestSetupError: Invalid property in capturedUnit: card, onwer

AMMayberry1 (approved):

PR #605, opened 2025-02-16
invincible + fett firespray + radiant VII

VJubert:


server/game/cards/04_JTL/units/FettsFiresprayFearedSilhouette.ts
```diff
@@ -0,0 +1,34 @@
...
+            title: 'Deal 1 indirect damage to a player. If you control Boba Fett, deal 2 indirect damage instead',
+            when: {
+                onAttackDeclared: (event, context) => event.attack.attacker === context.source,
+                onCardPlayed: (event, context) => event.card === context.source
+            },
+            targetResolver: {
+                mode: TargetMode.Player,
+                immediateEffect: AbilityHelper.immediateEffects.conditional({
+                    condition: (context) =>
+                        context.player.leader.title === 'Boba Fett' ||
```
AMMayberry1, line 24: We have `Player.controlsLeaderOrUnitWithTitle()`, which we will extend to cover upgrades as well once Piloting is implemented. Can we use that here instead?

server/game/cards/04_JTL/units/InvincibleNavalAdversary.ts
```diff
@@ -0,0 +1,34 @@
...
+        return {
+            id: '9958088138',
+            internalName: 'invincible#naval-adversary',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addDecreaseCostAbility({
+            title: 'If you control a unique Separatist card, this unit costs 1 resource less to play',
+            condition: (context) => context.player.isTraitInPlay(Trait.Separatist),
```
AMMayberry1, line 17: Need to check for `card.unique`, also we should add a test to cover that

server/game/cards/04_JTL/units/InvincibleNavalAdversary.ts
```diff
@@ -0,0 +1,34 @@
...
+            amount: 1
+        });
+
+        this.addTriggeredAbility({
+            title: 'Return a non-leader unit that costs 3 or less to its owner\'s hand',
+            when: {
+                onLeaderDeployed: (event, context) => event.card.controller === context.source.controller,
```
AMMayberry1, line 24: I really can't imagine it ever being an issue but for consistency, let's use `event.context.player` and `context.player` here

server/game/cards/04_JTL/units/InvincibleNavalAdversary.ts
```diff
@@ -0,0 +1,34 @@
...
+            },
+            optional: true,
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                cardCondition: (card: IUnitCard) => card.cost <= 3,
```
AMMayberry1, line 29: nit: we now have a `card.hasCost()` type check we can use here instead of casting to `IUnitCard` or using `isUnit()`. We always want to do a type assertion over a cast b/c it's safer

test/server/cards/04_JTL/units/InvincibleNavalAdversary.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+            expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine, context.lurkingTiePhantom]);
+            context.player1.clickCard(context.lurkingTiePhantom);
+
+            expect(context.player2).toBeActivePlayer();
+            expect(context.lurkingTiePhantom).toBeInZone('hand');
+        });
+
+        it('Invincible\'s ability should not trigger when opponent leader deploy', function () {
+            contextRef.setupTest({
```
AMMayberry1, line 84: No need to update this time, but going forward just for readability it would be good if we can still try and consolidate `setupTest` instances so that we could use a `beforeEach` and then have multiple test cases that use the same setup. Not a performance issue, just makes it easier to read through the cases. And not a hard rule, just in cases where it makes sense (e.g. I think these two leader deploy cases could have been used with one setupTest)

AMMayberry1 (approved):

PR #606, opened 2025-02-16
Che bato/deck data refactor

CheBato:
- Removed redundant swudbDeck parameter from gameServer.ts
- Changed the moveToSideboard, moveToDeck and convertCardListToMap functions to save and move cards as it worked before.


AMMayberry1:
Merged into mine

PR #615, opened 2025-02-17
Add Guerilla Soldier and Tactical Heavy Bomber

Frugghi:
| Guerilla Soldier | Tactical Heavy Bomber |
| --- | --- |
| image | image |


server/game/cards/04_JTL/units/GuerillaSoldier.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Deal 3 indirect damage to a player. If a base is damaged this way, ready this unit',
+            targetResolver: {
+                activePromptTitle: 'Choose a player to deal 3 indirect damage to',
+                mode: TargetMode.Player,
+                immediateEffect: AbilityHelper.immediateEffects.indirectDamageToPlayer({ amount: 3 })
+            },
+            then: {
```
AMMayberry1, line 23: Right now there's not a case where it would matter but for the future it might be safer to make this an `ifYouDo`. That way if there was some effect that could prevent indirect damage, this part won't fire. Same note for Heavy Bomber

server/game/cards/04_JTL/units/GuerillaSoldier.ts
```diff
@@ -0,0 +1,38 @@
...
+                title: 'Ready this unit',
+                thenCondition: (context) => context.events.some((event) =>
+                    event.name === EventName.OnDamageDealt &&
+                    (event.card as Card).isBase() &&
+                    event.damageSource.type === DamageSourceType.Ability &&
+                    event.damageSource.card === context.source &&
+                    event.isIndirect
+                ),
+                immediateEffect: AbilityHelper.immediateEffects.ready((context) => ({
```
AMMayberry1, line 32: nit: `context.source` is the default target so can omit the parameter completely

test/server/cards/04_JTL/units/GuerillaSoldier.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+            context.player1.setDistributeIndirectDamagePromptState(new Map([
+                [context.guerillaSoldier, 2],
+                [context.p1Base, 1],
+            ]));
+
+            expect(context.guerillaSoldier.exhausted).toBeFalse();
+        });
+
+        it('Guerilla Soldier\'s ability should deal 3 indirect damage to a player and not ready the unit if a base was not damaged', async function () {
+            await contextRef.setupTestAsync({
```
AMMayberry1, line 34: nit: we no longer need it for performance reasons, but just for readability could we converge these to one test setup with a beforeEach()? Since the setup is the same between the two

AMMayberry1 (approved):

PR #617, opened 2025-02-17
Implementing Black One, Straight at Them

supercodepoet:
image


AMMayberry1:
@supercodepoet is this ready for review again? not sure since it's still in draft

supercodepoet:
@AMMayberry1 it can be reviewed but I figured we wanted the Poe leader done to add those tests.

server/game/cards/04_JTL/units/BlackOneStraightAtThem.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    public override setupCardAbilities () {
+        this.addConstantAbility({
+            title: 'While this unit is upgraded, it gets +1/+0.',
+            condition: (context) => context.source.isUpgraded(),
+            ongoingEffect: AbilityHelper.ongoingEffects.modifyStats({ power: 1, hp: 0 })
+        });
+
+        this.addOnAttackAbility({
+            title: 'If you control Poe Dameron (as a unit, upgrade, or leader), you may deal 1 damage to a unit.',
```
AMMayberry1, line 21: nit: shouldn't include the "if" condition or the "you may"

server/game/cards/04_JTL/units/BlackOneStraightAtThem.ts
```diff
@@ -0,0 +1,36 @@
...
+            optional: true,
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                cardCondition: (card, context) => this.hasControlOfPoeDameron(context),
+                immediateEffect: AbilityHelper.immediateEffects.damage({ amount: 1 })
+            }
+        });
+    }
+
+    private hasControlOfPoeDameron (context): boolean {
```
AMMayberry1, line 31: We have a helper method named `Player.controlsLeaderOrUnitWithTitle()` which we will expand to include upgrades once piloting comes in, so let's use that here instead

test/server/cards/04_JTL/units/BlackOneStraightAtThem.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+            it('should get +1/0 because the unit is upgraded', function() {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.topTarget);
+                context.player1.clickCard(context.blackOne);
+
+                expect(context.blackOne).toHaveExactUpgradeNames(['top-target']);
+                expect(context.blackOne.getPower()).toBe(3);
+
+                context.player2.clickCard(context.blackOne);
```
AMMayberry1, line 37: nit: we don't need to do an attack to check, just checking the getPower() is enough. We do need to also check getHp() though to check for any typo or issue in the `hp` part of the stats modifier

test/server/cards/04_JTL/units/BlackOneStraightAtThem.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+                expect(context.player1).toBeActivePlayer();
+            });
+
+            it('should get +1/0 because the unit is upgraded', function() {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.topTarget);
+                context.player1.clickCard(context.blackOne);
+
+                expect(context.blackOne).toHaveExactUpgradeNames(['top-target']);
```
AMMayberry1, line 34: Can we have a test where it gets two upgrades to confirm that it still only gets +1

test/server/cards/04_JTL/units/BlackOneStraightAtThem.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+                context.player1.clickCard(context.blackOne);
+                context.player1.clickCard(context.p2Base);
+
+                expect(context.player1).toHavePrompt('Choose a unit');
+                expect(context.player1).toHavePassAbilityButton();
+                expect(context.player1).toBeAbleToSelectExactly([context.blackOne, context.deathStarStormtrooper, context.infernoFour]);
+
+                context.player1.clickCard(context.deathStarStormtrooper);
+
+                expect(context.deathStarStormtrooper).toBeInZone('discard');
```
AMMayberry1, line 70: When dealing an amount of damage, we prefer not to defeat the unit so that we can confirm we dealt the exact amount expected and no more

server/game/cards/04_JTL/units/BlackOneStraightAtThem.ts
supercodepoet: Updated

server/game/cards/04_JTL/units/BlackOneStraightAtThem.ts
supercodepoet: I have used the method. I believe I have used `context.source.controller` correctly here because if the unit is stolen then it needs to check if the controller has a Poe Dameron. Is this correct?

test/server/cards/04_JTL/units/BlackOneStraightAtThem.spec.ts
supercodepoet: Added the check for HP

test/server/cards/04_JTL/units/BlackOneStraightAtThem.spec.ts
supercodepoet: Added and additional upgrade

test/server/cards/04_JTL/units/BlackOneStraightAtThem.spec.ts
supercodepoet: Changed the unit who received the damage

server/game/cards/04_JTL/units/BlackOneStraightAtThem.ts
AMMayberry1: In this case we would still want to use `context.player`. The "you" always refers to the player who is resolving the ability.

In this case it's extremely unlikely that it would make a difference so the only time we've run into issues with it is on when-defeated abilities, but we still want to be using `context.player` everywhere by default

supercodepoet: ok, updated

AMMayberry1 (approved):

PR #620, opened 2025-02-17
Deck validation logic

AMMayberry1:


server/utils/deck/DeckInterfaces.ts
```diff
@@ -33,3 +33,35 @@ export interface ICardIdAndName {
...
+    [DeckValidationFailureReason.MaxMainboardSizeExceeded]?: { maxBoardedSize: number; actualBoardedSize: number };
+    [DeckValidationFailureReason.MinMainboardSizeNotMet]?: { minBoardedSize: number; actualBoardedSize: number };
+    [DeckValidationFailureReason.TooManyCopiesOfCard]?: { card: ICardIdAndName; maxCopies: number; actualCopies: number }[];
+    [DeckValidationFailureReason.UnknownCardId]?: { id: string }[];
+}
+
+export interface ICardValidationFailures {
+    notImplemented?: boolean;
+    illegalInFormat?: boolean;
+}
```
CheBato, line 67: Since we have NotImplemented and illegalInFormat is this necesary? I didn't see it being used anywhere.

server/utils/deck/DeckValidator.ts
```diff
@@ -0,0 +1,218 @@
...
+
+        unimplementedCards.sort((a, b) => a.set.localeCompare(b.set) || a.titleAndSubtitle.localeCompare(b.titleAndSubtitle));
+
+        return unimplementedCards;
+    }
+
+    // TODO: account for new bases that modify these values
+    public getMinimumSideboardedDeckSize(_deck: ISwuDbDecklist): number {
+        return 50;
+    }
```
CheBato, line 90: i think the naming of this might be a bit off would it make sense to just have getMinimumDeckSize or is there perhaps gonna be anything that might change the sideboard size?

server/utils/deck/DeckValidator.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    public getMaximumSideboardedDeckSize(_deck: ISwuDbDecklist): number {
+        return this.getMinimumSideboardedDeckSize(_deck) + DeckValidator.MaxSideboardSize;
+    }
```
CheBato, line 94: This goes to max 60? Is that correct since the Deck can be any size above 50?

server/utils/deck/DeckValidator.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    public validateDeck(deck: ISwuDbDecklist, format: SwuGameFormat): IDeckValidationFailures {
+        try {
+            if (!deck.leader || !deck.base || !deck.deck || deck.deck.length === 0) {
+                return { [DeckValidationFailureReason.InvalidDeckData]: true };
+            }
+
```
CheBato, line 101: i don't know if this makes sense or if we check this anywhere else. But should a we check if a deck also has a sideboard parameter even if its empty []?

server/utils/deck/DeckValidator.ts
```diff
@@ -0,0 +1,218 @@
...
+                }
+
+                if (card.count > 3) {
+                    failures[DeckValidationFailureReason.TooManyCopiesOfCard].push({
+                        card: { id: card.id, name: cardData.titleAndSubtitle },
+                        maxCopies: 3,
+                        actualCopies: card.count
+                    });
+                }
+            }
```
CheBato, line 181: Just a thought i think we'll need to make this into a function since the new set introduces cards that can have up to 15 copies in a deck.

server/utils/deck/DeckValidator.ts
AMMayberry1: So this number is primarily used to tell us if the user has sideboarded out too many cards and their deck is now below 50. If their decklist doesn't even have 50 cards then we return a slightly different error

server/utils/deck/DeckInterfaces.ts
AMMayberry1: Oh yeah good catch, this is from an earlier iteration

server/utils/deck/DeckValidator.ts
AMMayberry1: What do you mean? The max sideboard count is 10 in premier

server/utils/deck/DeckValidator.ts
AMMayberry1: Not sure I follow - can you explain further? I think it's okay if they don't provide any cards in the sideboard part

server/utils/deck/DeckValidator.ts
AMMayberry1: Oh yep good call

server/utils/deck/DeckValidator.ts
CheBato: yeah my bad a little more context is needed here. Basically i'm referencing this code where the getMaximumSideBoardedDeckSize is used:

```python
const maxBoardedSize = this.getMaximumSideboardedDeckSize(deck); # we get 60 here since 50 from getMinimumSideboardedDeckSize and 10 from DeckValidator.MaxSideboardSize

const decklistCardsCount = this.getTotalCardCount(deckCards); # we get the total amount of cards
const boardedCardsCount = this.getTotalCardCount(deck.deck);
const sideboardCardsCount = deck.sideboard ? this.getTotalCardCount(deck.sideboard) : 0;

if (decklistCardsCount > maxBoardedSize) { # here we check if the decklistCardsCount is higher then 60 and if it is we throw an error 
    failures[DeckValidationFailureReason.MaxDecklistSizeExceeded] = {
        maxDecklistSize: maxBoardedSize,
        actualDecklistSize: decklistCardsCount
    };
```


i hope i was able to present my thoughts more accurately

server/utils/deck/DeckValidator.ts
CheBato: i agree that its okay if the user provides an empty list for sideboarded cards i was just wondering if we should check if the sideboard exists just like we check if the deck exists. For instance !deck.sideboard . I don't know how swudb sends sideboards whether they are an empty array or they don't add a sideboard parameter to the json in that case.

server/utils/deck/DeckValidator.ts
CheBato: yes that makes sense i noticed its used in
```ts
const minBoardedSize = this.getMinimumSideboardedDeckSize(deck);
```

server/utils/deck/DeckValidator.ts
AMMayberry1: I forgot to check, but I think it shouldn't matter right? They can both indicate an empty sideboard

CheBato: indeed it might not matter. And its an easy fix if it will.

PR #633, opened 2025-02-19
Improve TS interfaces for Game and Player

### LINKED ISSUE #609: Interfaces for Game and Player config properties and general cleanup
Currently, the Game and Player classes accept configuration objects of type `any` and then pull values out of them. We need to establish TS interfaces for these and generally do some cleaning to remove unused fields. In addition, we will want to add explicit checks that all required values are populated in the constructor of each class.

This may also mean refactoring some of the logic in the Lobby class and the test game setup infrastructure.

Frugghi:
Fix #609

### DIFF
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -21,9 +21,6 @@ const AbilityResolver = require('./gameSteps/AbilityResolver.js');
 const { AbilityContext } = require('./ability/AbilityContext.js');
 const Contract = require('./utils/Contract.js');
 const { cards } = require('../cards/Index.js');
-// const { Conflict } = require('./conflict');
-// const ConflictFlow = require('./gamesteps/conflict/conflictflow');
-// const MenuCommands = require('./MenuCommands');
 
 const { EventName, ZoneName, Trait, WildcardZoneName, TokenUpgradeName, TokenUnitName } = require('./Constants.js');
 const { StateWatcherRegistrar } = require('./stateWatcher/StateWatcherRegistrar.js');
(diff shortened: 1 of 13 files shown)


server/game/core/Game.js
```diff
@@ -42,31 +39,45 @@ const { DisplayCardsForSelectionPrompt } = require('./gameSteps/prompts/DisplayC
...
+     */
+    constructor(details, options) {
         super();
 
+        Contract.assertNotNullLike(details);
+        Contract.assertNotNullLike(details.id);
+        Contract.assertNotNullLike(details.name);
+        Contract.assertNotNullLike(details.owner);
+        Contract.assertNotNullLike(details.gameMode);
+        Contract.assertNotNullLike(details.cardDataGetter);
```
AMMayberry1, line 54: To improve maintainability, could we wrap these checks in a function that lives next to the interface definition in `GameInterfaces.ts`? That way people will see it when they change the interface and hopefully remember to update the checks.

The ideal would be that they get a build error if they forget to update the checks, but I'm not sure if there's a way to do that that's simple enough to be worth the effort

server/game/core/Player.js
```diff
@@ -144,7 +144,7 @@ class Player extends GameObject {
      * @param { WildcardZoneName.AnyArena | ZoneName.GroundArena | ZoneName.SpaceArena } arena Arena to select units from
      * @param {(card: import('./card/propertyMixins/UnitProperties').IUnitCard) => boolean} cardCondition Condition to filter cards
      */
-    getUnitsInPlay(arena = WildcardZoneName.AnyArena, cardCondition = (card) => true) {
+    getUnitsInPlay(arena = WildcardZoneName.AnyArena, cardCondition = () => true) {
```
AMMayberry1, line 147: Why did we remove the card parameter? `cardCondition` isn't really useful without it

server/game/core/Game.js
```diff
@@ -42,31 +39,45 @@ const { DisplayCardsForSelectionPrompt } = require('./gameSteps/prompts/DisplayC
 const { DisplayCardsBasicPrompt } = require('./gameSteps/prompts/DisplayCardsBasicPrompt.js');
 
 class Game extends EventEmitter {
-    constructor(details, options = {}) {
+    /**
+     * @param {import('./GameInterfaces.js').GameConfiguration} details
+     * @param {import('./GameInterfaces.js').GameOptions} options
+     */
+    constructor(details, options) {
```
AMMayberry1, line 46: Fantastic work on all this!! Thanks a ton for cleaning it up, everything looks way better now.

Not sure if you did this already, so just in case - any time we mess with Lobby stuff or Game initialization we need to also manually check that all methods for starting a game work in local dev, since we don't have e2e tests for that yet. If you haven't already, can you please validate that all three of the following work (lmk in discord if you're not sure how to set any of them up):

1. Test game using a prepared test setup json
2. Private lobby game
3. Matchmade game using the queue

server/game/core/Player.js
Frugghi: VSCode was "complaining" that `card` was unused. I think the type annotation above is enough to make Typescript know that `cardCondition` accepts a `card` parameter and removing it here affects only the default value.

server/gamenode/Lobby.ts
```diff
@@ -295,34 +292,32 @@ export class Lobby {
             }
         });
 
-        game.initialiseTokens(this.tokenCardsData);
+        game.initialiseTokens(this.cardDataGetter.tokenData);
```
CheBato, line 295: Since we already moved playableCardTitles would it make sense to also initilialiseTokens in the actualy game initialization similiarly to what was done with palyableCardTitles?

AMMayberry1 (approved):

PR #637, opened 2025-02-20
Add initial support for Piloting

adbronico:


server/game/Interfaces.ts
```diff
@@ -24,12 +24,26 @@ import type { GroundArenaZone } from './core/zone/GroundArenaZone';
...
+import type { ICardCanChangeControllers } from './core/card/CardInterfaces';
+import type { ICardWithPrintedHpProperty } from './core/card/propertyMixins/PrintedHp';
 
 // allow block comments without spaces so we can have compact jsdoc descriptions in this file
 /* eslint @stylistic/lines-around-comment: off */
 
 // ********************************************** EXPORTED TYPES **********************************************
 
+/** IUpgradeCard definition (exists here to prevent import loops) */
+export interface IUpgradeCard extends IInPlayCard, ICardWithPrintedHpProperty, ICardWithPrintedPowerProperty, ICardWithCostProperty, ICardCanChangeControllers {
```
AMMayberry1, line 39: nit: there's a `CardInterfaces.ts` we could put this in

server/game/Interfaces.ts
```diff
@@ -326,6 +341,12 @@ interface IOverwhelmKeywordProperties extends IKeywordPropertiesBase {
     keyword: KeywordName.Overwhelm;
 }
 
+interface IPilotingKeywordProperties extends IKeywordPropertiesBase {
```
AMMayberry1, line 344: Since this has the same properties as Smuggle we can create a base interface similar to `INumericKeywordPropertiesBase`

server/game/actions/PlayUpgradeAction.ts
```diff
@@ -23,11 +24,14 @@ export class PlayUpgradeAction extends PlayCardAction {
...
-        Contract.assertTrue(context.source.isUpgrade());
+        const isUpgrade = context.source.isUpgrade();
+        const isPilot = !isUpgrade && (context.source.isUnit() && context.source.hasSomeKeyword(KeywordName.Piloting));
+
+        Contract.assertTrue(isUpgrade || isPilot);
 
         const events = [
             new AttachUpgradeSystem({
-                upgrade: context.source,
+                upgrade: context.source as IUpgradeCard,
```
AMMayberry1, line 34: Can we change it so that we don't have to cast this? At this stage the card might not actually be an upgrade yet, the AttachSystem should be able to handle that

server/game/cards/01_SOR/events/Bamboozle.ts
```diff
@@ -15,7 +15,7 @@ export default class Bamboozle extends EventCard {
     }
 
     protected override buildPlayCardActions(playType: PlayType = PlayType.PlayFromHand, propertyOverrides: IPlayCardActionOverrides = null) {
-        const bamboozleAction = playType === PlayType.Smuggle
+        const bamboozleAction = playType === PlayType.Smuggle || playType === PlayType.Piloting // This card can't be piloted - can we do something with an Omit here?
```
AMMayberry1, line 18: It's fine as-is, the design is that if the card doesn't support the provided play mode then it just returns no actions

server/game/core/ability/PlayCardAction.ts
```diff
@@ -31,14 +31,22 @@ interface IStandardPlayActionProperties extends IPlayCardActionPropertiesBase {
     playType: PlayType.PlayFromHand | PlayType.PlayFromOutOfPlay;
 }
 
+// TODO - can we refacor these two to be shared?
+export interface IPilotingCardActionProperties extends IPlayCardActionPropertiesBase {
```
AMMayberry1, line 35: Similar to the keyword interfaces, could just make an `IAlternatePlayModeProperties` and name the properties like `alternatePlayModeResourceCost` or something potentially less verbose. Then have the `playType` property on the specific Smuggle / Piloting interfaces for the discriminated union.

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -69,6 +73,15 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
         this._exhausted = val;
     }
 
+    /** The card that this card is underneath */
+    public get parentCard(): IUnitCard {
+        Contract.assertNotNullLike(this._parentCard);
+        // TODO: move IsInPlay to be usable here
+        // Contract.assertTrue(this.isInPlay());
```
AMMayberry1, line 80: I can't remember, did we have a specific reason for putting the upgrade logic in `PlayableOrDeployableCard` instead of `InPlayCard`? Feels like the latter would make more sense for a handful of reasons

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -90,7 +103,11 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
      */
     public getPlayCardActions(propertyOverrides: IPlayCardActionOverrides = null): PlayCardAction[] {
         if (this.zoneName === ZoneName.Hand) {
-            return this.buildPlayCardActions(PlayType.PlayFromHand, propertyOverrides);
+            let playActions = this.buildPlayCardActions(PlayType.PlayFromHand, propertyOverrides);
+            if (this.hasSomeKeyword(KeywordName.Piloting)) {
```
AMMayberry1, line 107: Can we put a TODO here to update this when we're ready to handle playing with Piloting from discard

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -145,12 +166,77 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
...
+    public canAttach(targetCard: Card, controller: Player = this.controller): boolean {
+        if (!targetCard.isUnit() || (this.attachCondition && !this.attachCondition(targetCard))) {
+            return false;
+        }
+
+        return true;
+    }
+
+    // TODO: Simplify this if it turns out there are no alternative ways to gain a Piloting cost
+    protected buildCheapestPilotingAction(propertyOverrides: IPlayCardActionOverrides = null) {
```
AMMayberry1, line 213: If we do keep this, we definitely could find a way to merge it with `buildCheapestSmuggleAction` since the logic is the same

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -145,12 +166,77 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
...
+            this.takeControl(newController, newParentCard.zoneName);
+        } else {
+            this.moveTo(newParentCard.zoneName);
+        }
+
+        newParentCard.attachUpgrade(this);
+        this._parentCard = newParentCard;
+    }
+
+    public isAttached(): boolean {
```
AMMayberry1, line 189: Since a lot of these methods are meaningless or even potentially dangerous if the card isn't an upgrade, it would be good to follow the practice we have with other similar properties and have a guard that will throw if they're being called at the wrong time.

Could just create a method called `assertIsUpgrade()` that's just a contract check or two and use it at the top of these

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -69,6 +73,15 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
...
+        Contract.assertNotNullLike(this._parentCard);
+        // TODO: move IsInPlay to be usable here
+        // Contract.assertTrue(this.isInPlay());
+
+        return this._parentCard;
+    }
+
     // see Card constructor for list of expected args
     public constructor(owner: Player, cardData: any) {
         super(owner, cardData);
```
AMMayberry1, line 87: A useful guard we could add would be to check if the cardData has the `upgradeHp` and `upgradePower` properties and set a property that indicates whether this card can be an IUpgradeCard off of that. That might make it make more sense to just use those properties for storing upgrade stats instead of copying them over to `hp` and `power` in the fetchdata script

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -145,12 +166,77 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
...
+
+    public unattach() {
+        Contract.assertNotNullLike(this._parentCard, 'Attempting to unattach upgrade when already unattached');
+
+        this.parentCard.unattachUpgrade(this);
+        this._parentCard = null;
+    }
+
+    /**
+         * Checks whether the passed card meets any attachment restrictions for this card. Upgrade
```
AMMayberry1, line 201: nit: indent on the comments

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -159,6 +159,10 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
             return this.upgrades.some((card) => card.isShield());
         }
 
+        public override isUpgrade(): this is IUpgradeCard {
+            return this._parentCard !== null; // TODO: is there a better check?
```
AMMayberry1, line 163: Nah I think this is good actually

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -649,10 +653,42 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
          * Checks whether an attachment can be played on a given card.  Intended to be
          * used by cards inheriting this class
          */
+        // TODO - is this used at all anywhere?
```
AMMayberry1, line 656: Don't think so, can remove

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -649,10 +653,42 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
...
         public canPlayOn(card) {
             return true;
         }
 
+        /**
+         *  This should only be called if a unit is a Pilot or has some other ability that lets it attach as an upgrade
```
AMMayberry1, line 662: This should be enforced with a contract check. We could later add some way for cards with non-piloting attach abilities to have a special override that we could check here

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -649,10 +653,42 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
...
+         */
+        public canAttachPilot(): boolean {
+            if (!this.hasSomeTrait(Trait.Vehicle)) {
+                return false;
+            }
+
+            // TODO: we need logic for cards that override the pilot limit
+            if (this.upgrades.length > 0) {
+                const pilotUpgrades = this.upgrades.filter((upgrade) => upgrade.hasSomeTrait(Trait.Pilot));
+                return pilotUpgrades.length === 0;
```
AMMayberry1, line 687: nit: could consolidate into one line using `this.upgrades.some((upgrade) => upgrade.hasSomeTrait(Trait.Pilot))`

server/game/stateWatchers/CardsPlayedThisPhaseWatcher.ts
```diff
@@ -55,8 +55,8 @@ export class CardsPlayedThisPhaseWatcher extends StateWatcher<PlayedCardEntry[]>
                 currentState.concat({
                     card: event.card,
                     playEvent: event,
-                    parentCard: event.card.parentCard ?? null,
-                    parentCardInPlayId: event.card.parentCard?.canBeInPlay() ? event.card.parentCard.inPlayId : null,
+                    parentCard: event.card.isUpgrade() && event.card.isAttached() !== null ? event.card.parentCard : null,
```
AMMayberry1, line 58: Is the `!== null` a typo?

server/game/stateWatchers/CardsPlayedThisPhaseWatcher.ts
```diff
@@ -55,8 +55,8 @@ export class CardsPlayedThisPhaseWatcher extends StateWatcher<PlayedCardEntry[]>
...
+                    parentCardInPlayId: event.card.isUpgrade() && event.card.parentCard?.canBeInPlay() ? event.card.parentCard.inPlayId : null,
```
AMMayberry1, line 59: This _should_ work without the `event.card.parentCard?.canBeInPlay()` part, unless there's something weird going on with the typing for `parentCard`

test/server/core/abilities/keyword/Piloting.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+                expect(context.player1).toHaveExactPromptButtons(['Cancel', 'Play Iden Versio', 'Play Iden Versio with Piloting']);
+                context.player1.clickPrompt('Play Iden Versio with Piloting');
+                context.player1.clickCard(context.restoredArc170);
+                expect(context.player1.readyResourceCount).toBe(p1Resources - 3);
+                expect(context.restoredArc170.upgrades).toContain(context.idenInHand);
+                expect(context.player1).toHavePrompt('Choose which copy of Iden Versio, Adapt or Die to defeat');
+                context.player1.clickCard(context.idenInPlay);
+                expect(context.idenInPlay).toBeInZone('discard');
+                expect(context.idenInHand).toBeInZone('spaceArena');
+            });
```
AMMayberry1, line 163: These are solid, nice job. Looking over the Nexus doc, here's a few more I think we could add at this stage:

- Pilot leaders deployed as upgrades are vulnerable to upgrade removal
- Pilot leader upgrades affected by Bamboozle are defeated, and pilots that are in play as units count as having the keyword for Daimyō Boba
- Piloting and Smuggle are both alternate costs, and can't be played together. Choose one or the other!
- Pilot upgrades are defeated like any other upgrade if something happens to the attached unit
- Pilots that were defeated as upgrades can be returned to hand by Gar Saxon's deployed leader unit ability that grants a When Defeated to upgraded units, due to Last Known Information rules _(probably should go in the Gar Saxon tests)_
- SHD Survivors' Gauntlet may move pilot upgrades to a vehicle that already has a pilot upgrade attached

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
adbronico: I think it was due to believing it's likely that we'll have an event that can be attached as an upgrade someday

test/server/core/abilities/keyword/Piloting.spec.ts
AMMayberry1: Followup for the discussion in chat about testing effects that count or target units to confirm they don't see attached pilots as units

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
AMMayberry1: Oh hmm. It actually might not be worth accounting for that now due to stuff like this. If that happens we're going to have to heavily refactor EventCard anyway

server/game/cards/01_SOR/events/Bamboozle.ts
adbronico: I get a typing error if I do that, which I think is why the Smuggle check is there as well. Thoughts?

server/game/actions/PlayUpgradeAction.ts
adbronico: The other way runs into some issues with the canAttach check, but I can dig into it and see if I can straighten that out

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
adbronico: Do you think we should move it to InPlayCard?

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
adbronico: I added it where I think it makes sense but if you think there's a way to add it to the attachTo and canAttach methods let me know

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
adbronico: Do you think it makes sense to hold off on merging them until we reach the end of this little epic to make sure?

test/server/core/abilities/keyword/Piloting.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+        });
```
AMMayberry1, line 164: The Yoda PR made me realize we also should have some tests that the various state watchers see the pilots as upgrades, if we can find the right cards to test with

server/game/cards/01_SOR/events/Bamboozle.ts
AMMayberry1: Sorry I meant that what you have here is correct

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
AMMayberry1: Yeah I think so

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
AMMayberry1: I think we either don't include it at all or we merge them

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
AMMayberry1: The thing to do there might be to add a method called `checkIsAttachable()` which will throw by default but then we override it in UnitProperties and UpgradeCard to behave correctly

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
adbronico: Would this still go in PlayableOrDeployable or would it now go in InPlayCard? And how many other places would then use this? I don't know that fields I add to InPlayCard are accessible in PlayableOrDeployable - I had some issues when I tried to move parentCard, for example, but I might just need to do some casting?

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -649,10 +653,42 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
...
+         * @returns True if this is allowed to attach to the targetCard; false otherwise
+         */
+        public override canAttach(targetCard: Card, controller: Player = this.controller): boolean {
+            if (this.hasSomeKeyword(KeywordName.Piloting) && targetCard.isUnit()) {
+                return targetCard.canAttachPilot() && targetCard.controller === controller;
+            }
+            // TODO: Handle Phantom II and Sidon Ithano
+            return false;
+        }
+
```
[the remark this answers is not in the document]
adbronico: To confirm - the way I'm understanding this is that we need isUnit to return false when a pilot is attached as an upgrade, and the type to return as upgrade as well?

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
AMMayberry1: All the upgrade-related changes would move to InPlayCard so it would go along with them. And yeah we should never try to access properties in a derived class from the parent class, if we hit an issue where we need that to happen then we definitely need to back up and revisit that part of the design

AMMayberry1: Re: where the property would be used specifically, I think it would only be used in the new guard functions we're writing for e.g. `canAttach()`

server/game/core/card/propertyMixins/UnitProperties.ts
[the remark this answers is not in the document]
AMMayberry1: Yeah correct

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
adbronico: I'll leave this one open and let you resolve it, but I think with all of the code refactoring I've done since this review, it might be best to just start a new comment chain if we still want this?

test/server/core/abilities/keyword/Piloting.spec.ts
adbronico: Tests Added:
- Piloting and Smuggle are both alternate costs, and can't be played together. Choose one or the other!
- Pilot upgrades are defeated like any other upgrade if something happens to the attached unit
- Pilots that were defeated as upgrades can be returned to hand by Gar Saxon's deployed leader unit ability that grants a When Defeated to upgraded units, due to Last Known Information rules (probably should go in the Gar Saxon tests

To-Do:
- SHD Survivors' Gauntlet may move pilot upgrades to a vehicle that already has a pilot upgrade attached

Add later:
- Pilot leaders deployed as upgrades are vulnerable to upgrade removal
- Pilot leader upgrades affected by Bamboozle are defeated, and pilots that are in play as units count as having the keyword for Daimyō Boba

server/game/stateWatchers/UnitsDefeatedThisPhaseWatcher.ts
```diff
@@ -72,10 +73,16 @@ export class UnitsDefeatedThisPhaseWatcher extends StateWatcher<DefeatedUnitEntr
         // on card played, add the card to the player's list of cards played this phase
         this.addUpdater({
             when: {
-                onCardDefeated: (context) => context.card.isUnit(),
+                // This doesn't actually work for some reason, so I had to add the below workaround for now
+                onCardDefeated: (event) => EnumHelpers.isUnit(event.lastKnownInformation.type),
```
AMMayberry1, line 77: Ah I think I see the issue, you're treating the parameter like an `event` but it's a `context`. I think you just need `context.event`

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
adbronico: @AMMayberry1 What would the TODO say? I have a canBeUpgrade property already in Card.ts, so maybe I just move that into InPlayCard?

server/game/core/card/baseClasses/InPlayCard.ts
```diff
@@ -99,6 +117,11 @@ export class InPlayCard extends InPlayCardParent implements IInPlayCard {
         return this._pendingDefeat;
     }
 
+    public override getType(): CardType {
+        // TODO: should we add a CardType.PilotUpgrade?
```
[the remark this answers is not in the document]
adbronico: It made sense to me at the time, but now I'm wondering what we gain from it? It basically would only be used as a shorthand for a unit attached as an upgrade that is also a Pilot, so should we make it more generic like UnitUpgrade?

server/game/stateWatchers/UnitsDefeatedThisPhaseWatcher.ts
adbronico: Ah so I did try that - the same issue presents if you do context.event as well

server/game/core/card/baseClasses/InPlayCard.ts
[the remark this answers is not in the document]
AMMayberry1: Yeah I'm fine with that as well

AMMayberry1 (approved):

AMMayberry1 (approved):

PR #639, opened 2025-02-21
Che bato/deck validation endpoint

CheBato:


server/gamenode/GameServer.ts
```diff
@@ -177,12 +202,32 @@ export class GameServer {
         });
 
         app.post('/api/enter-queue', (req, res) => {
-            const { user, deck } = req.body;
-            const success = this.enterQueue(user, deck);
-            if (!success) {
-                return res.status(400).json({ success: false, message: 'Failed to enter queue' });
+            try {
+                const { user, deck } = req.body;
+                const validationResults = this.deckValidator.validateDeck(deck, SwuGameFormat.Premier);
```
AMMayberry1, line 207: Looks like most of this code block is duplicated between the two routes, can we consolidate it into a method?

server/utils/deck/DeckValidator.ts
```diff
@@ -70,6 +70,8 @@ export class DeckValidator {
         }
     }
 
+
+    /* TODO do we need this? */
```
AMMayberry1, line 74: This is so we can eventually have a page that shows all of the unimplemented cards 😁

server/utils/deck/DeckValidator.ts
```diff
@@ -85,17 +87,25 @@ export class DeckValidator {
     }
 
     // TODO: account for new bases that modify these values
-    public getMinimumSideboardedDeckSize(_deck: ISwuDbDecklist): number {
+    // TODO: return back to ISwuDbDecklist when we have checks for those
```
AMMayberry1, line 90: What does this mean?

server/utils/deck/DeckValidator.ts
```diff
@@ -85,17 +87,25 @@ export class DeckValidator {
...
+    public getMinimumSideboardedDeckSize(_deck: IDecklistInternal | ISwuDbDecklist): number {
         return 50;
     }
 
-    public validateDeck(deck: ISwuDbDecklist, format: SwuGameFormat): IDeckValidationFailures {
+    // Type guard to check if a deck is an ISwuDbDecklist
+    private isSwuDbDecklist(
+        deck: IDecklistInternal | ISwuDbDecklist
+    ): deck is ISwuDbDecklist {
```
AMMayberry1, line 98: I'm not sure if checks like this work for interfaces. It would be a little safer to have two different public methods like `validateSwudbDeck` and `validateInternalDeck` which each accept just one of the types, and then have a private method that handles the checks common to both types.

server/utils/deck/DeckValidator.ts
CheBato: gotcha!

AMMayberry1 (approved):

server/gamenode/GameServer.ts
```diff
@@ -176,20 +186,49 @@ export class GameServer {
...
+    // method for validating the deck via API
+    private async processDeckValidation(
+        deck: ISwuDbDecklist,
+        format: SwuGameFormat,
+        res: express.Response,
+        onValid: () => Promise<void> | void
+    ): Promise<void> {
+        const validationResults = this.deckValidator.validateSwuDbDeck(deck, format);
+        const hasBlockingFailures = Object.keys(validationResults).some(
+            (key) => key !== DeckValidationFailureReason.NotImplemented
```
danbastin, line 219: Do we need NotImplemented as a ValidationFailureReason? I think we can simply a lot of things if we keep validation errors to be blocking errors.

CheBato: i agree we could make a seperate function and return a list of unimplemented cards to the FE. What do you think @AMMayberry1 ?

AMMayberry1: Yeah that seems fine to me

AMMayberry1 (approved):

Issue #585, opened 2025-02-13
Yoda Leader ability isn't enabled by Leaders being defeated


adbronico:
I killed the enemy leader with my unit surviving, and on my next action tried to draw a card using Yoda's leader ability, but it didn't trigger. I'm assuming that the state watcher doesn't account for leaders leaving play

PR #643, opened 2025-02-22
Fix an issue with Yoda's leader action and defeated Leader units

adbronico:
Fixes:
https://github.com/SWU-Karabast/forceteki/issues/585


adbronico:
I tried the approach of adding a cardType to the card entry, but at the time the update runs, the leader has already been defeated and flipped back to leader side, so it wasn't quite registering correctly. I went with this approach but I'm open to changing it if you have any suggestions.

The other route would be to add the cardType back in and just set it to unit if it finds leader for the card

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -44,6 +44,14 @@ export class CardsLeftPlayThisPhaseWatcher extends StateWatcher<CardLeftPlayEntr
         return this.getCardsLeftPlay({ controller, filter }).length > 0;
     }
 
+    public someUnitLeftPlay({ controller, filter }: {
+        controller?: Player;
+        filter?: (event: CardLeftPlayEntry) => boolean;
+    }) {
+        // We check if a unit or leader left play because defeated leaders are already changed to leader side at time of registration
+        return this.getCardsLeftPlay({ controller, filter }).filter((card) => card.isUnit() || card.isLeader()).length > 0;
```
AMMayberry1, line 52: Could we instead add a new property to the entry type that records the type at time of defeat? That'll be useful for piloting

server/game/cards/03_TWI/leaders/YodaSensingDarkness.ts
```diff
@@ -23,13 +23,13 @@ export default class YodaSensingDarkness extends LeaderUnitCard {
             title: 'If a unit left play this phase, draw a card, then put a card from your hand on the top or bottom of your deck.',
             cost: AbilityHelper.costs.exhaustSelf(),
             immediateEffect: AbilityHelper.immediateEffects.conditional({
-                condition: () => this.cardsLeftPlayThisPhaseWatcher.someCardLeftPlay({ filter: (entry) => entry.card.isUnit() }),
+                condition: () => this.cardsLeftPlayThisPhaseWatcher.someUnitLeftPlay({}),
```
AMMayberry1, line 26: Does it work if we just provide no argument at all? Feels weird that we have to provide an empty object parameter

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
adbronico: @AMMayberry1 see https://github.com/SWU-Karabast/forceteki/pull/643#issuecomment-2675940960

It's already registered as a leader at that point, so we would need to manually set it there for leaders. Is that the approach we want to do? I'm also not sure how that will interact with pilots yet

AMMayberry1:
> I tried the approach of adding a cardType to the card entry, but at the time the update runs, the leader has already been defeated and flipped back to leader side, so it wasn't quite registering correctly. I went with this approach but I'm open to changing it if you have any suggestions.
> 
> 
> 
> The other route would be to add the cardType back in and just set it to unit if it finds leader for the card

Could you use event. lastKnownInformation? If that doesn't already have type we can add it

adbronico:
> > I tried the approach of adding a cardType to the card entry, but at the time the update runs, the leader has already been defeated and flipped back to leader side, so it wasn't quite registering correctly. I went with this approach but I'm open to changing it if you have any suggestions.
> > The other route would be to add the cardType back in and just set it to unit if it finds leader for the card
> 
> Could you use event.lastKnownInformation? If that doesn't already have type we can add it

I checked for LKI but it didn't have it when I debugged it. Do we need to set certain events to collect it?

server/game/cards/03_TWI/leaders/YodaSensingDarkness.ts
adbronico: So, it does seem to require that. It's a bit odd to me too

AMMayberry1 (approved):

PR #644, opened 2025-02-22
The invisible hand

ssoulier:
[The invisible Hand](https://swudb.com/card/TWI/234)

Screenshot 2025-02-22 at 09 01 42

### DIFF
--- a/server/game/cards/03_TWI/units/TheInvisibleHandImposingFlagship.ts
+++ b/server/game/cards/03_TWI/units/TheInvisibleHandImposingFlagship.ts
@@ -25,14 +25,16 @@ export default class TheInvisibleHandImposingFlagship extends NonLeaderUnitCard 
                 cardTypeFilter: WildcardCardType.Unit,
                 zoneFilter: WildcardZoneName.AnyArena,
                 cardCondition: (card) => card.hasSomeTrait(Trait.Separatist) && card.isPlayable() && !card.exhausted,
+                immediateEffect: AbilityHelper.immediateEffects.sequential([
+                    AbilityHelper.immediateEffects.exhaust((context) => ({
+                        target: context.target
+                    })),
+                    AbilityHelper.immediateEffects.damage((context) => ({
+                        target: context.player.opponent.base,
+                        amount: context.target.length,
+                    }))
+                ])
             },
-            then: (thenContext) => ({
-                title: 'Deal 1 damage to the defending player\'s base for each unit exhausted this way',
-                immediateEffect: AbilityHelper.immediateEffects.damage({
-                    target: thenContext.player.opponent.base,
-                    amount: thenContext.target.length,
-                })
-            })
         });
     }
 }
--- a/server/game/cards/03_TWI/units/TheInvisibleHandImposingFlagship.ts
+++ b/server/game/cards/03_TWI/units/TheInvisibleHandImposingFlagship.ts
@@ -1,0 +1,38 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { RelativePlayer, TargetMode, Trait, WildcardCardType, WildcardZoneName } from '../../../core/Constants';
+
+export default class TheInvisibleHandImposingFlagship extends NonLeaderUnitCard {
+    protected override getImplementationId () {
+        return {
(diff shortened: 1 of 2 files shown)


server/game/cards/03_TWI/units/TheInvisibleHandImposingFlagship.ts
```diff
@@ -0,0 +1,40 @@
...
+
+        this.addOnAttackAbility({
+            title: 'Exhaust any number of friendly Separatist units',
+            optional: true,
+            targetResolver: {
+                mode: TargetMode.Unlimited,
+                controller: RelativePlayer.Self,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: WildcardZoneName.AnyArena,
+                cardCondition: (card) => card.hasSomeTrait(Trait.Separatist) && card.isPlayable() && !card.exhausted,
```
ssoulier, line 27: My first test with the tokens does not work, because I do not think tokens are playable.
But without the `isPlayable` test and I can't call the `exhausted` property.
There must be something obvious I am missing, but I am stuck on this card because implementation because of this `exhausted` that I can test easily.

server/game/cards/03_TWI/units/TheInvisibleHandImposingFlagship.ts
```diff
@@ -0,0 +1,40 @@
...
+            immediateEffect: AbilityHelper.immediateEffects.createBattleDroid({ amount: 4 }),
+        });
+
+        this.addOnAttackAbility({
+            title: 'Exhaust any number of friendly Separatist units',
+            optional: true,
+            targetResolver: {
+                mode: TargetMode.Unlimited,
+                controller: RelativePlayer.Self,
+                cardTypeFilter: WildcardCardType.Unit,
```
ssoulier, line 25: Even if I specify that I want to filter on Unit, I still need to do `card.isUnit()` below to make sure I can call `card.exhausted`.
I would have expected the cardTypeFilter to be enough. Is it something we should update in the targetResolver logic?

AMMayberry1: It has to do with how the types are assigned in TypeScript. Since `card` is of type `Card` in `cardCondition`, there's no way for it to no that it is actually of type `IUnitCard` and has the `exhausted` property without doing the `isUnit()` check. So in this case it's easier to just remove `cardTypeFilter` since it's redundant and we have to have `card.isUnit()` anyway for things to work

server/game/cards/03_TWI/units/TheInvisibleHandImposingFlagship.ts
```diff
@@ -0,0 +1,40 @@
...
+                zoneFilter: WildcardZoneName.AnyArena,
```
AMMayberry1, line 26: nit: I think this is the default, we can remove it

test/server/cards/03_TWI/units/TheInvisibleHandImposingFlagship.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+                player2: {
+                    groundArena: ['droid-commando'],
+                    spaceArena: ['devastating-gunship']
+                }
+            });
+
+            const { context } = contextRef;
+            context.player1.clickCard(context.theInvisibleHand);
+            const battleDroidTokens = context.player1.findCardsByName('battle-droid');
+            expect(battleDroidTokens.length).toBe(4);
```
AMMayberry1, line 20: Need to check exhausted status

test/server/cards/03_TWI/units/TheInvisibleHandImposingFlagship.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+            context.moveToNextActionPhase();
+            expect(battleDroidTokens.every((token) => !token.exhausted)).toBeTrue();
+            context.player1.clickCard(context.theInvisibleHand);
+            context.player1.clickCard(context.player2.base);
+            expect(context.player1).toBeAbleToSelectExactly([
+                battleDroidTokens[0],
```
AMMayberry1, line 27: nit: could just do `...battleDroidTokens` to expand the array here

AMMayberry1 (approved):

AMMayberry1 (approved):

PR #647, opened 2025-02-23
add dorsal turret

ssoulier:
[Dorsal Turret](https://swudb.com/card/JTL/120/dorsal-turret)

Screenshot 2025-02-23 at 15 02 23


ssoulier:
Looks like CI does not have access to the card. Is there something I should do?

AMMayberry1:
> Looks like CI does not have access to the card. Is there something I should do?

Yeah can you update the card version number in card-data-version.txt, that will force it to pull the latest card data jsons

server/game/cards/04_JTL/upgrades/DorsalTurret.ts
```diff
@@ -0,0 +1,29 @@
...
+            id: '9338356823',
+            internalName: 'dorsal-turret'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card: Card) => card.hasSomeTrait(Trait.Vehicle));
+
+        this.addGainTriggeredAbilityTargetingAttached({
+            title: 'When this unit deals combat damage to a unit while attacking: Defeat that unit.',
```
AMMayberry1, line 18: nit: don't include the trigger condition text

server/game/cards/04_JTL/upgrades/DorsalTurret.ts
```diff
@@ -0,0 +1,29 @@
...
+            when: {
+                onDamageDealt: (event, context) =>
+                    event.type === DamageType.Combat &&
+                    event.damageSource.attack.attacker === context.source &&
+                    event.damageSource.damageDealtBy === context.source &&
+                    event.damageSource.attack.target?.isUnit()
+            },
+            immediateEffect: AbilityHelper.immediateEffects.defeat((context) => ({ target: context.event.damageSource.attack.target }))
```
AMMayberry1, line 26: nit: can do `context.event.card` here and on line 24, bit simpler

test/server/cards/04_JTL/upgrades/DorsalTurret.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+            expect(context.player1).toBeAbleToSelectExactly([
+                context.frontierAtrt,
+                context.greenSquadronAwing,
+                context.escortSkiff,
+                context.atst,
+            ]);
+            context.player1.clickCard(context.frontierAtrt);
+            context.player2.passAction();
+
+            // CASE 1: Frontier AT-RT attacks and survives
```
AMMayberry1, line 29: note: no need to update here since it's already done, but now that we aren't worried about test case count anymore these could all be separate test cases ([discord message link](https://discord.com/channels/1220057752961814568/1266424023290482812/1340492054509387870) if you missed it). Just a note for future PRs.

test/server/cards/04_JTL/upgrades/DorsalTurret.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+            context.player1.clickCard(context.battlefieldMarine);
+            expect(context.battlefieldMarine).toBeInZone('groundArena');
+            expect(context.frontierAtrt.damage).toBe(3);
+            context.moveToNextActionPhase();
+
+            // CASE 4: Frontier AT-RT defeat Leader unit
+            context.player1.clickCard(context.frontierAtrt);
+            context.player1.clickCard(context.lukeSkywalker);
+            expect(context.frontierAtrt).toBeInZone('discard');
+            expect(context.lukeSkywalker.deployed).toBeFalse();
```
AMMayberry1, line 70: Two more tests we should add to confirm the damage conditions:
- Use Strike True to deal damage, ability should not trigger
- Kind of a weird one - make a TIE Bomber with Perilous Position, Hardpoint Heavy Blaster, and Dorsal Turret attached. Have it attack into a space unit and deal the damage from Hardpoint to the space unit. The unit should _not_ be defeated by the Dorsal ability b/c no combat damage was dealt to it (TIE Bomber power is 0). Will need to leave a comment that the test will have to be updated when the TIE Bomber ability is implemented.

Issue #483, opened 2025-01-29
All abilities must be changed to use `context.player` instead of `context.source.controller` by default


AMMayberry1:
"When defeated" abilities must use `context.player` instead of `context.source.controller`. The reason is that, if the unit was stolen, then the player who triggered the "when defeated" will be the player who activated the ability (the stealing player) while `context.source.controller` is the _current_ controller of the card, which is the original owner.

The existing ones have been checked and updated, but we should retroactively try and replace all instances of this so that others will use it correctly in the future. Since `context.player` is almost never the wrong choice, we should just default to it by convention so we don't have to remember when we're in an edge case.

PR #654, opened 2025-02-24
Replace context.source.controller by context.player

VJubert:
#483


server/game/core/ability/TriggeredAbility.ts
```diff
@@ -96,6 +96,7 @@ export default class TriggeredAbility extends CardAbility {
     }
 
     protected override controllerMeetsRequirements(context): boolean {
+        // TODO CHECK WITH VELD IF NEEDED TO CHANGE
```
AMMayberry1, line 99: Yeah this one needs to stay as-is

AMMayberry1 (approved):

PR #657, opened 2025-02-25
Update Don't Get Cocky to use the new display prompt

moyerr:
Screenshot 2025-02-25 at 1 22 36 PM

### DIFF
--- a/server/game/cards/01_SOR/events/DontGetCocky.ts
+++ b/server/game/cards/01_SOR/events/DontGetCocky.ts
@@ -7,8 +7,6 @@ import type { IThenAbilityPropsWithSystems } from '../../../Interfaces';
 import * as Contract from '../../../core/utils/Contract';
 
 export default class DontGetCocky extends EventCard {
-    protected override readonly overrideNotImplemented: boolean = true;
-
     protected override getImplementationId() {
         return {
             id: '2202839291',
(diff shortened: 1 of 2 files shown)


server/game/cards/01_SOR/events/DontGetCocky.ts
```diff
@@ -46,7 +47,10 @@ export default class DontGetCocky extends EventCard {
                 activePromptTitle: `Current total cost: ${this.topXCardsTotalCost(cardsRevealedCount, contextWithUnitTarget)}\nSelect one:`,
                 mode: TargetMode.Select,
                 choices: {
-                    ['Reveal another card']: AbilityHelper.immediateEffects.reveal((context) => ({ target: context.player.getTopCardsOfDeck(7)[cardsRevealedCount] })),
+                    ['Reveal another card']: AbilityHelper.immediateEffects.reveal((context) => ({
+                        useDisplayPrompt: true,
+                        target: context.player.getTopCardsOfDeck(7)[cardsRevealedCount]
```
AMMayberry1, line 52: Now that we have the prompt, instead of just showing the latest card could we show all revealed up until now? I think that would make things a little easier to follow

moyerr: @AMMayberry1 Oh good point, I think that's possible, but I do think the chat reveal would be kinda weird. If my understanding is correct: for the first card, it will name the first card. For the second, it will name the first and second, and so on, until for the _nth_ card, the first one will have been named in chat _n_ times.

AMMayberry1: yeah I thought about that, I think it is slightly wonky but the impact isn't huge since we'd be emitting a reveal message for each card anyway. so most of them are just a bit longer now. personally I feel like it's an okay tradeoff

AMMayberry1 (approved):

PR #671, opened 2025-02-26
Find Mismatches and Display Conflicts

### LINKED ISSUE #636: Fix Card Data and Verify Variant Discrepancy Accuracies
Some data from the API (variant based) does not match from one printing of the card to another. The easiest place to see that is with the Chancellor Palpatine, Playing Both Sides. The Standard variant lists all 3 aspects (Cunning, Villany, and Heroism), whereas the other variants only list 2 (Cunning and Heroism). There are other cards, like Anakin Skywalker, What it Takes to Win, that have attributes like Overwhelm that don't match up.

We'll create a script that identifies where these discrepancies between printings exists that can then be manually reviewed and incorporated into the fetchcards script to make sure that we have as accurate data as we can.

Image

Image

Image

DonovanDevelops:
Run "npm run card-diff" to view the output. This should take in an array of cards and build the comparisons between all variants of the card.

Still needed: Check the card variants in "npm run get-cards" BEFORE filtering to one item per card (needs to be run as all variants returned from the API to compare all of the data between all variants)


DonovanDevelops:
@AMMayberry1 Added several updates and pushed so you could take a look.

This adds in a lot of extra changes, and addresses the following comments:
- First pass at separating the api calls into one function "retrieveCards" that takes in a "filtered" parameter currently used to indicate whether or not it duplicates would be removed. Required functions from fetchdata have been brought in as well.
- Created function "getInternalName" which takes in the title and subtitle and outputs the key to use across all functions. Can be used anywhere and adjusting the outputs will update everywhere.
- Did not update the property mess. Not 100% sure, but will investigate how it is done elsewhere to try and duplicate it.
- As advised, removed all of the matching functionality and other bloat from the compare function. Now just returns data that doesn't match.
- Added functionality to accept a string file location to write outputs to.
- Asked for clarification on the return type of the comparisons.
- Better error checking to avoid giant if blocks.

PR #676, opened 2025-02-27
Add Thrawn1 leader

AMMayberry1:
Added the Thrawn1 leader. Implemented support for having display text under card images in the display prompt, also did some bug fixing / enhancement on the select choice target resolver. Added some new test matchers to support checking the display text.


server/game/cards/01_SOR/leaders/GrandAdmiralThrawnPatientAndInsightful.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private addLookTrigger() {
+        this.addTriggeredAbility({
+            title: 'Look at the top card of each player\'s deck',
+            when: {
+                onPhaseStarted: (event) => event.phase === PhaseName.Action
+            },
+            immediateEffect: AbilityHelper.immediateEffects.lookAt((context) => {
+                const ownTopCard = context.source.controller.getTopCardOfDeck();
```
adbronico, line 76: Is this a situation where we would want to use context.player instead of context.source.controller? Ditto for the opponent top card

server/game/cards/01_SOR/leaders/GrandAdmiralThrawnPatientAndInsightful.ts
```diff
@@ -0,0 +1,96 @@
...
+        };
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addLookTrigger();
+
+        this.addActionAbility({
+            title: 'Reveal the top card of any player\'s deck',
+            cost: [AbilityHelper.costs.abilityResourceCost(1), AbilityHelper.costs.exhaustSelf()],
+            targetResolver: {
```
adbronico, line 19: Can we combine the shared bits of code in the targetResolver into a method since the reveal and exhaust logic is the same?

adbronico (changes requested):
A couple small/nit questions, but overall this looks good and I can approve once those are addressed

server/game/cards/01_SOR/leaders/GrandAdmiralThrawnPatientAndInsightful.ts
AMMayberry1: Oh yup good catch, I started this branch before the change

server/game/cards/01_SOR/leaders/GrandAdmiralThrawnPatientAndInsightful.ts
AMMayberry1: Gave it a shot but there's a weird type issue involving enums we've run into before that I still haven't found a clean way of getting around. Solving it would end up eating up as many lines of code as we currently have, I think. So just leaving this as-is.

adbronico (approved):
lgtm

PR #691, opened 2025-03-01
user leaving messages

danbastin:


server/gamenode/Lobby.ts
```diff
@@ -294,6 +294,14 @@ export class Lobby {
     }
 
     public removeUser(id: string): void {
+        const user = this.users.find((u) => u.id === id);
+        if (user) {
+            this.gameChat.addMessage(`${user.username} has left the lobby`);
```
AMMayberry1, line 299: Looks like we can do this and the leaves game message both, should we do only one or the other?

danbastin: Doesn't hurt to do both if a game is running. if we have a game we have a lobby so the message will be there if say there are spectators hat jump in and are able to fill the seat and the remaining players go back to the lobby. 

Is that what your asking?

AMMayberry1 (approved):

PR #695, opened 2025-03-01
draft stolen at hauler

ssoulier:
[card](https://swudb.com/card/JTL/221)

Screenshot 2025-03-01 at 09 02 50

### DIFF
--- a/server/game/cards/04_JTL/units/StolenATHauler.ts
+++ b/server/game/cards/04_JTL/units/StolenATHauler.ts
@@ -1,0 +1,39 @@
+import AbilityHelper from '../../../AbilityHelper';
+import type { Card } from '../../../core/card/Card';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { EffectName, RelativePlayer } from '../../../core/Constants';
+import type Game from '../../../core/Game';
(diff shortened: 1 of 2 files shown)


server/game/cards/04_JTL/units/StolenATHauler.ts
```diff
@@ -0,0 +1,39 @@
...
+        return {
+            id: '6272475624',
+            internalName: 'stolen-athauler'
+        };
+    }
+
+    private canPlayFromOpponentDiscard(): IOngoingEffectGenerator {
+        function createOngoingCardEffect(game: Game, source: Card, props: IOngoingCardEffectProps) {
+            props.targetController = RelativePlayer.Opponent;
+            return new OngoingCardEffect(game, source, props, new StaticOngoingEffectImpl(EffectName.CanPlayFromDiscard, null));
```
AMMayberry1, line 22: I think what we should do here is rename `CanPlayFromDiscard` to more explicitly say `OwnerCanPlayFromDiscard`. From there, we add a new effect named `OpponentCanPlayFromDiscard` and implement it in the engine.

This should be pretty straightforward if you just do a text search for the existing `CanPlayFromDiscard` / `OwnerCanPlayFromDiscard` and see all the places it is used, we'll just need to add similar code for `OpponentCanPlayFromDiscard` in each place

PR #718, opened 2025-03-02
MasAmeddaViceChair Card Implementation

griggsca91:
image


test/server/cards/01_SOR/units/MasAmeddaViceChair.spec.ts
```diff
@@ -0,0 +1,109 @@
+describe('Mas Amedda, Vice Chair', function() {
+    integration(function(contextRef) {
+        describe('Mas Amedda\'s Ability', function() {
+            beforeEach(async function () {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['mas-amedda#vice-chair'],
+                        hand: ['frontier-atrt'],
```
griggsca91, line 9: ahhh crap i put the test in the wrong folder need to fix

server/game/cards/03_TWI/units/MasAmeddaViceChair.ts
```diff
@@ -0,0 +1,34 @@
...
+            when: {
+                onCardPlayed: (event, context) =>
+                    event.card.isUnit() &&
+                    event.card.controller === context.player &&
+                    event.card !== context.source
+            },
+            optional: true,
+            immediateEffect: AbilityHelper.immediateEffects.exhaust(),
+            ifYouDo: {
+                title: 'If you do, search the top 4 cards of your deck for a unit, reveal it, and draw it',
```
AMMayberry1, line 24: nit: don't need to include the "if you do" in the title, once we start rendering these correctly we'll add that automatically

test/server/cards/03_TWI/units/MasAmeddaViceChair.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+
+                // Check cards in hand
+                expect(context.systemPatrolCraft).toBeInZone('hand');
+
+                // Check cards in deck
+                expect(context.player1.deck.length).toBe(5);
+                expect([context.clanWrenRescuer, context.concordDawnInterceptors, context.bountyPosting]).toAllBeInBottomOfDeck(context.player1, 3);
+            });
+
+            it('should be able to choose no cards', function() {
```
AMMayberry1, line 48: nit: no need to make changes since you've already done it, but for future search cards we typically just do the basic case (your first unit test). Just because the search system is encapsulated and all its edge cases are pretty well tested, so there's not a lot of need to spend the time writing out all the edge case tests for every card that does it.

griggsca91: Sounds good! Will keep that in mind

server/game/cards/03_TWI/units/MasAmeddaViceChair.ts
griggsca91: @AMMayberry1 removed the "if you do"

AMMayberry1 (approved):

Issue #712, opened 2025-03-02
Chewbacca pilot buffs the attached unit with its unit stats, not its upgrade stats


AMMayberry1:
![Image](https://github.com/user-attachments/assets/eb12aa4d-39dd-42a7-8662-3486e6889449)

PR #721, opened 2025-03-02
Fix issues related to pilots with differing unit and upgrade stats

adbronico:
Related to https://github.com/SWU-Karabast/forceteki/issues/712

In my pilot PR, the test units I used had the same HP/power. I had assumed upgradeHp and upgradePower were being used somewhere, but this fixes that bug from my initial PR and also wires up upgrades to use those fields as well.

I also randomly noticed that Perilous Position had the wrong test name, which I fixed.

### DIFF
--- a/server/game/core/card/CardInterfaces.ts
+++ b/server/game/core/card/CardInterfaces.ts
@@ -15,8 +15,8 @@ export interface ICardCanChangeControllers {
 
 /** IUpgradeCard definition (exists here to prevent import loops) */
 export interface IUpgradeCard extends IInPlayCard, ICardWithPrintedHpProperty, ICardWithPrintedPowerProperty, ICardWithCostProperty, ICardCanChangeControllers {
-    readonly upgradeHp?: number;
-    readonly upgradePower?: number;
+    readonly printedUpgradeHp: number;
+    readonly printedUpgradePower: number;
     getUpgradeHp(): number;
     getUpgradePower(): number;
 }
--- a/server/game/core/card/CardInterfaces.ts
+++ b/server/game/core/card/CardInterfaces.ts
@@ -15,4 +15,8 @@ export interface ICardCanChangeControllers {
 
 /** IUpgradeCard definition (exists here to prevent import loops) */
 export interface IUpgradeCard extends IInPlayCard, ICardWithPrintedHpProperty, ICardWithPrintedPowerProperty, ICardWithCostProperty, ICardCanChangeControllers {
+    readonly upgradeHp?: number;
+    readonly upgradePower?: number;
+    getUpgradeHp(): number;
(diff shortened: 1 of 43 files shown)


server/game/core/card/propertyMixins/PrintedHp.ts
```diff
@@ -3,7 +3,9 @@ import type { Card, CardConstructor } from '../Card';
 
 export interface ICardWithPrintedHpProperty extends Card {
     readonly printedHp: number;
+    readonly upgradeHp?: number;
```
AMMayberry1, line 6: Can we just put these on IUpgradeCard instead, since they're a requirement for that interface? Then the actual implementation logic and constructor checks would go in InPlayCard.

server/game/core/card/propertyMixins/PrintedPower.ts
```diff
@@ -18,10 +21,22 @@ export function WithPrintedPower<TBaseClass extends CardConstructor>(BaseClass:
 
             Contract.assertNotNullLike(cardData.power);
             this.printedPower = cardData.power;
+
+            if (this.isUpgrade()) {
+                Contract.assertNotNullLike(cardData.upgradePower);
```
AMMayberry1, line 26: We should have a similar check in UnitProperties.ts if the card has the piloting keyword (check needs to happen after the `super()` call which sets up the keywords)

server/game/core/card/CardInterfaces.ts
```diff
@@ -15,4 +15,8 @@ export interface ICardCanChangeControllers {
 
 /** IUpgradeCard definition (exists here to prevent import loops) */
 export interface IUpgradeCard extends IInPlayCard, ICardWithPrintedHpProperty, ICardWithPrintedPowerProperty, ICardWithCostProperty, ICardCanChangeControllers {
+    readonly upgradeHp?: number;
```
AMMayberry1, line 18: These shouldn't be `?`, same for the props on `InPlayCard`. Also I'm realizing these should probably be named `printedUpgradeHp` and `printedUpgradePower` to be clearer and consistent with IUnitCard

server/game/core/card/baseClasses/InPlayCard.ts
```diff
@@ -122,6 +125,20 @@ export class InPlayCard extends InPlayCardParent implements IInPlayCard {
 
         // this class is for all card types other than Base and Event (Base is checked in the superclass constructor)
         Contract.assertFalse(this.printedType === CardType.Event);
+
+        if (this.isUpgrade()) {
+            Contract.assertNotNullLike(cardData.upgradePower);
```
AMMayberry1, line 130: Need to check `upgradeHp` as well

AMMayberry1 (approved):

Issue #714, opened 2025-03-02
Scanning Officer error in logs


AMMayberry1:
![Image](https://github.com/user-attachments/assets/5611d79c-6976-48f6-89db-6bdb4969bd63)

adbronico:
Looks like it's related to an empty deck if Scanning Officer defeats a resource. I'll take a look at this.

PR #722, opened 2025-03-02
Fix an error when no units with Smuggle are revealed by Scanning Officer

adbronico:
Related to https://github.com/SWU-Karabast/forceteki/issues/714

I added a conditional to check that any cards had smuggle before trying to defeat and replace them.

I did try to use the "thenCondition" field but it threw an error in the IDE so I think there's some reason that isn't usable there.

### DIFF
--- a/server/game/cards/02_SHD/units/ScanningOfficer.ts
+++ b/server/game/cards/02_SHD/units/ScanningOfficer.ts
@@ -21,17 +21,21 @@ export default class ScanningOfficer extends NonLeaderUnitCard {
             then: (thenContext) => ({
                 title: 'Defeat each resource with Smuggle',
                 effect: 'defeat',
-                immediateEffect: AbilityHelper.immediateEffects.simultaneous(() => {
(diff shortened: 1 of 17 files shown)


server/game/cards/02_SHD/units/ScanningOfficer.ts
```diff
@@ -21,17 +21,21 @@ export default class ScanningOfficer extends NonLeaderUnitCard {
...
-                        AbilityHelper.immediateEffects.defeat({
-                            target: smuggleCards
-                        }),
-                        AbilityHelper.immediateEffects.resourceCard({
-                            targetPlayer: RelativePlayer.Opponent,
-                            target: thenContext.player.opponent.getTopCardsOfDeck(smuggleCards.length)
-                        })
-                    ];
+                immediateEffect: AbilityHelper.immediateEffects.conditional({
+                    condition: thenContext.events[0].cards.some((card) => card.hasSomeKeyword(KeywordName.Smuggle)),
```
AMMayberry1, line 25: vscode didn't have any issues for me with thencondition. sometimes the linting gets into a bad state, I've noticed.

image

AMMayberry1 (approved):

Issue #727, opened 2025-03-02
Set 1 Home One is giving itself too much restore


AMMayberry1:
Should not be targeting itself with the "gain restore" ability

ssoulier:
https://github.com/SWU-Karabast/forceteki/pull/728

PR #728, opened 2025-03-02
fix home one friendly restore

ssoulier:
Fixing Home One issue. It should not give +1 restore to itself.


ssoulier:
Fixing this issue: https://github.com/SWU-Karabast/forceteki/issues/727

AMMayberry1 (approved):

PR #736, opened 2025-03-02
SubStep Check Fixes

CDMayberry:
Fix for Qira Playing Her Part (and any Then usage). meetRequirements now only checks substeps for then/ifYouDo.

Additionally, I added some limited "debug" flags to the Game class and flag checks inside of the GamePipeline, intended for manual usage only. Essentially this allowed me to take a side-by-side comparison of the pipeline steps with and w/o a hand of cards (WinMerge for the win) to see where it diverged.

LMK if you'd prefer to remove the debug, I added some safety's but might still be good to keep it out unless we could enforce the flags at build time, IE ignore debug calls entirely unless we are running a single unit test.

image

The above example would enable the flag to write out each pipeline step, then disable it again, so it's scoped to specific actions. Also prevents people from leaving a debug flag active accidentally.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -33,10 +33,9 @@ export default tseslint.config(
                 ...globals.jasmine,
             },
 
-            ecmaVersion: 2020,
+            ecmaVersion: 2022,
             sourceType: "commonjs",
         },
-
         rules: {
             "jasmine/no-spec-dupes": 0,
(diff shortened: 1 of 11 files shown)


CDMayberry:
PS, updated the ESLint to ECMA 2022, to allow for # / private javascript fields.

AMMayberry1 (approved):

PR #737, opened 2025-03-02
Add basic piloting abilities

AMMayberry1:
No support yet for situations where a card can move from being a pilot to a unit while still on the field


test/server/cards/04_JTL/units/HeraSyndullaWeveLostEnough.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+                context.player1.clickPrompt('Play Hera Syndulla');
+
+                context.player2.passAction();
+
+                context.heraSyndulla.exhausted = false;
+                context.player1.clickCard(context.heraSyndulla);
+                context.player1.clickCard(context.p2Base);
+                expect(context.p1Base.damage).toBe(4);
+            });
+
```
adbronico, line 60: Can we add a test with Hera attached to a Restored ARC to make sure the numeric keywords' values stack?

adbronico (changes requested):
Add one test but looks really clean.

PR #742, opened 2025-03-03
Fix cost adjustment for upgrades

### LINKED ISSUE #495: Target-based cost adjustment for upgrades calculated wrong prior to play
Cost adjusters for upgrades that depend on the specific target - e.g., Darksaber or General Grievous - are currently not being calculated right when the card is still in hand. E.g., if you have exactly four resources and want to play Darksaber on a mando unit out-of-aspect, the game will not show Darksaber as playable. However, if you have six resources and play it, only four will be spent (the correct number).

~~Fixing this requires some updates to how we calculate cost adjustment, which is in turn currently blocked by ongoing work for the Dooku leader. Once that's merged in we can start in on this work at any time.~~

**EDIT:** this is no longer blocked

Frugghi:
Fixes #495

### DIFF
--- a/server/game/core/Player.js
+++ b/server/game/core/Player.js
@@ -17,7 +17,8 @@ const {
     KeywordName,
     WildcardCardType,
     Trait,
-    WildcardRelativePlayer
+    WildcardRelativePlayer,
+    Stage
 } = require('./Constants');
 
 const EnumHelpers = require('./utils/EnumHelpers');
@@ -80,6 +81,8 @@ class Player extends GameObject {
 
         /** @type {Deck} */
         this.decklistNames = null;
+
+        /** @type {CostAdjuster[]} */
         this.costAdjusters = [];
         this.abilityMaxByIdentifier = {}; // This records max limits for abilities
         this.promptedActionWindows = user.promptedActionWindows || {
@@ -821,11 +824,22 @@ class Player extends GameObject {
         return Math.max(cost, 0);
     }
 
-    getMatchingCostAdjusters(context, penaltyAspect = null, additionalCostAdjusters = null, ignoreExploit = false) {
+    /**
+     * @param {AbilityContext} context
+     * @param {Aspect=} penaltyAspect
+     * @param {CostAdjuster[]=} additionalCostAdjusters
+     * @param {boolean} ignoreExploit
+     * @returns {CostAdjuster[]}
+     */
+    getMatchingCostAdjusters(context, penaltyAspect = null, additionalCostAdjusters = [], ignoreExploit = false) {
         const allMatchingAdjusters = this.costAdjusters.concat(additionalCostAdjusters)
-            .filter((adjuster) =>
-                adjuster?.canAdjust(context.playingType, context.source, context, context.target, penaltyAspect)
-            );
+            .filter((adjuster) => {
+                if (context.stage === Stage.Cost && context.source.isUpgrade() && !context.target) {
+                    return context.game.getArenaUnits()
+                        .some((unit) => adjuster.canAdjust(context.playType, context.source, context, unit, penaltyAspect));
+                }
+                return adjuster.canAdjust(context.playType, context.source, context, context.target, penaltyAspect);
+            });
 
         if (ignoreExploit) {
             return allMatchingAdjusters.filter((adjuster) => !adjuster.isExploit());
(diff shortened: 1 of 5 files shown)


server/game/costs/PlayCardResourceCost.ts
```diff
@@ -95,7 +95,7 @@ export class PlayCardResourceCost<TContext extends AbilityContext = AbilityConte
             const amount = this.getAdjustedCost(context);
             context.costs.resources = amount;
 
-            event.context.player.markUsedAdjusters(context.playType, event.context.source, context);
+            event.context.player.markUsedAdjusters(context.playType, event.context.source, context, event.context.target);
```
Frugghi, line 98: This wasn't needed to fix the issue but I think it was wrong that it was missing!

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -114,7 +115,7 @@ export class PlayCardSystem<TContext extends AbilityContext = AbilityContext> ex
     }
 
     private buildPlayActionProperties(card: Card, properties: IPlayCardProperties, context: TContext, action: PlayCardAction = null) {
-        let costAdjusters = [this.makeCostAdjuster(properties.adjustCost, context)];
+        let costAdjusters = Helpers.asArray(this.makeCostAdjuster(properties.adjustCost, context));
```
Frugghi, line 118: This avoid having an array containing a `null` cost adjuster if `this.makeCostAdjuster` returns `null`.

server/game/core/Player.js
```diff
@@ -821,11 +824,22 @@ class Player extends GameObject {
...
+     */
+    getMatchingCostAdjusters(context, penaltyAspect = null, additionalCostAdjusters = [], ignoreExploit = false) {
         const allMatchingAdjusters = this.costAdjusters.concat(additionalCostAdjusters)
-            .filter((adjuster) =>
-                adjuster?.canAdjust(context.playingType, context.source, context, context.target, penaltyAspect)
-            );
+            .filter((adjuster) => {
+                if (context.stage === Stage.Cost && context.source.isUpgrade() && !context.target) {
+                    return context.game.getArenaUnits()
+                        .some((unit) => adjuster.canAdjust(context.playType, context.source, context, unit, penaltyAspect));
```
AMMayberry1, line 839: We might need a little more complexity here to account for attachment restrictions - e.g., is there anything here that prevents us from including a Mandalorian vehicle in the cost adjuster list for Darksaber even though it's not a legal target?

server/game/core/Player.js
```diff
@@ -821,11 +824,22 @@ class Player extends GameObject {
...
+     * @param {boolean} ignoreExploit
+     * @returns {CostAdjuster[]}
+     */
+    getMatchingCostAdjusters(context, penaltyAspect = null, additionalCostAdjusters = [], ignoreExploit = false) {
         const allMatchingAdjusters = this.costAdjusters.concat(additionalCostAdjusters)
-            .filter((adjuster) =>
-                adjuster?.canAdjust(context.playingType, context.source, context, context.target, penaltyAspect)
-            );
+            .filter((adjuster) => {
+                if (context.stage === Stage.Cost && context.source.isUpgrade() && !context.target) {
```
AMMayberry1, line 837: This currently won't work for pilots but that's fine for now since there aren't any that it would matter for. Let's leave a TODO though

test/server/cards/02_SHD/upgrades/TheDarksaber.spec.ts
```diff
@@ -56,6 +56,7 @@ describe('The Darksaber', function() {
 
             it('should ignore Aspect penalties when being attached to a friendly Mandalorian', function () {
                 const { context } = contextRef;
+                context.player1.setResourceCount(4);
                 context.player1.clickCard(context.theDarksaber);
```
AMMayberry1, line 60: We also need to test if the card is selectable by the player, since sometimes the highlighting we send to the FE doesn't work (the tests will still allow clicking the card in that case, the FE won't)

server/game/core/Player.js
Frugghi: Why this won't work for pilots? Isn't `context.source.isUpgrade()` true for pilots as well?

AMMayberry1: This will be called at the targeting phase when the card is still in the hand and still a unit, at which point this will be returning false still

test/helpers/PlayerInteractionWrapper.js
```diff
@@ -360,6 +360,7 @@ class PlayerInteractionWrapper {
             this.moveCard(card, 'resource');
             card.exhausted = false;
         });
+        this.game.continue();
```
AMMayberry1, line 363: Could we instead use `Util.refreshGameState()`, since we're actually triggering the game pipeline here there's a risk that this could have side effects

PR #744, opened 2025-03-03
#DeckAndSideboardMax

CheBato:
- returned the value of minimumDeckSize and maximumSideboard from the deckvalidator and supplied it to the FE.

### DIFF
--- a/server/utils/deck/DeckValidator.ts
+++ b/server/utils/deck/DeckValidator.ts
@@ -93,8 +93,14 @@ export class DeckValidator {
         return unimplementedCards;
     }
 
-    public getMinimumSideboardedDeckSize(baseData: ICardCheckData): number {
+    public getMinimumSideboardedDeckSize(baseId: string): number {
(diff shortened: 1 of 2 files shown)


server/utils/deck/DeckValidator.ts
```diff
@@ -93,10 +93,16 @@ export class DeckValidator {
...
     }
 
-    public getMinimumSideboardedDeckSize(baseData: ICardCheckData): number {
+    public getMinimumSideboardedDeckSize(baseId: string): number {
+        const baseData = this.getCardCheckData(baseId);
         return 50 + (baseData.minDeckSizeModifier ?? 0);
     }
 
+    // update this function if anything affects the sideboard count
+    public getMaxSideboardSize(): number {
```
AMMayberry1, line 102: What do we show for Open or Next Set formats where the sideboard is unlimited? I was assuming this would need to have similar logic to `checkMaxSideboardSize()`

AMMayberry1 (approved):

PR #753, opened 2025-03-03
allow 0 card on unlimited select

danbastin:
Card permitting selecting "any number" of cards should allow no cards to be selected.

### DIFF
--- a/server/game/core/cardSelector/UnlimitedCardSelector.js
+++ b/server/game/core/cardSelector/UnlimitedCardSelector.js
@@ -4,6 +4,10 @@ const BaseCardSelector = require('./BaseCardSelector.js');
  * A simple concrete implementation that doesn't impose any limits on the number of cards that can be selected
  */
 class UnlimitedCardSelector extends BaseCardSelector {
+    /** @override */
+    hasEnoughSelected(selectedCards, context) {
+        return this.optional || selectedCards.length >= 0;
+    }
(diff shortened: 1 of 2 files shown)


server/game/core/cardSelector/UnlimitedCardSelector.js
```diff
@@ -4,6 +4,10 @@ const BaseCardSelector = require('./BaseCardSelector.js');
  * A simple concrete implementation that doesn't impose any limits on the number of cards that can be selected
  */
 class UnlimitedCardSelector extends BaseCardSelector {
+    /** @override */
+    hasEnoughSelected(selectedCards, context) {
```
AMMayberry1, line 8: So this is definitely something that's currently awkward right now but we do have a `canChooseNoCards` property on the selector interface: https://github.com/SWU-Karabast/forceteki/blob/7df1bf69f6791615ab1dd7ba1e90daeade51375e/server/game/TargetInterfaces.ts#L131

We need to clean up our story around this but right now I _think_ it would be cleaner to just set that property to true on the EL implementation. If we got this route then we need to add in explicit support for shutting this _off_ if the property is set to false

AMMayberry1: Agreed generally that this is probably should default to allowing zero cards but I want to be careful about making that change

PR #764, opened 2025-03-04
add kimogila heavy fighter

ssoulier:
[Card](https://swudb.com/card/JTL/222/kimogila-heavy-fighter)

Screenshot 2025-03-04 at 20 40 11

### DIFF
--- a/server/game/cards/04_JTL/units/KimogilaHeavyFighter.ts
+++ b/server/game/cards/04_JTL/units/KimogilaHeavyFighter.ts
@@ -14,7 +14,6 @@ export default class KimogilaHeavyFighter extends NonLeaderUnitCard {
         return events
             .filter((event) => event.name === EventName.OnDamageDealt)
             .filter((event) => event.card.canBeExhausted())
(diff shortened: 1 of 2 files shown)


server/game/cards/04_JTL/units/KimogilaHeavyFighter.ts
```diff
@@ -0,0 +1,40 @@
...
+            id: '6757031085',
+            internalName: 'kimogila-heavy-fighter',
+        };
+    }
+
+    private filterUnitsDamagedThatCanBeExhausted(events) {
+        return events
+            .filter((event) => event.name === EventName.OnDamageDealt)
+            .filter((event) => event.card.canBeExhausted())
+            .filter((event) => !event.card.exhausted)
```
AMMayberry1, line 17: Let's remove this filter, not exactly sure how it could cause a problem but typically we should apply exhaust effect even if the target is already exhausted just for rules correctness reasons

AMMayberry1 (approved):

PR #765, opened 2025-03-04
Implementing bunch of JTL cards

aruizc:
image
image
image
image
image


server/game/cards/04_JTL/events/CoordinatedFront.ts
```diff
@@ -0,0 +1,43 @@
...
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Give a ground unit and a space unit +2/+2 for this phase',
+            optional: true,
+            targetResolvers: {
+                republicUnit: {
+                    activePromptTitle: 'Choose a ground unit',
+                    mode: TargetMode.Exactly,
+                    numCards: 1,
+                    canChooseNoCards: true,
```
AMMayberry1, line 22: It's a little wonky, but actually when there are multiple target resolvers there is an `optional` property available (this is not true when there's only target resolver). We should use that instead of using `TargetMode.Exactly`. So they will end up looking like this:

```ts
separatistUnit: {
    activePromptTitle: 'Give a space unit +2/+2 for this phase',
    optional: true,
    zoneFilter: ZoneName.SpaceArena,
    immediateEffect: AbilityHelper.immediateEffects.forThisPhaseCardEffect({
        effect: AbilityHelper.ongoingEffects.modifyStats({ power: 2, hp: 2 })
    })
},
```

Lmk if that doesn't work

server/game/cards/04_JTL/events/CoordinatedFront.ts
```diff
@@ -0,0 +1,43 @@
...
+        return {
+            id: '9595202461',
+            internalName: 'coordinated-front',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Give a ground unit and a space unit +2/+2 for this phase',
+            optional: true,
```
AMMayberry1, line 16: I _think_ we can remove this since we'll have optional in the target resolvers (see next comment)

test/server/cards/04_JTL/events/CoordinatedFront.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+            expect(context.tielnFighter.getHp()).toBe(3); // 1 Base HP + 2 from Coordinated Front ability
+
+            // Move to the next action phase
+            context.moveToNextActionPhase();
+
+            expect(context.superBattleDroid.getPower()).toBe(4); // 4 base power
+            expect(context.superBattleDroid.getHp()).toBe(3); // 3 base HP
+            expect(context.tielnFighter.getPower()).toBe(2); // 2 Base power
+            expect(context.tielnFighter.getHp()).toBe(1); // 1 Base HP
+        });
```
AMMayberry1, line 46: We've been having some problems with the choose no target button, do you mind adding two more test cases where we hit both of them in turn (can use a `beforeEach()` to reuse the test setup between them)

server/game/cards/04_JTL/events/ShootDown.ts
```diff
@@ -0,0 +1,51 @@
...
+        this.setEventAbility({
+            title: 'Deal 3 damage to space unit.',
+            targetResolver: {
+                zoneFilter: ZoneName.SpaceArena,
+                cardTypeFilter: WildcardCardType.Unit,
+                immediateEffect: AbilityHelper.immediateEffects.damage({
+                    amount: 3
+                })
+            },
+            then: (thenContext) => ({
```
AMMayberry1, line 26: We can change this to be an `ifYouDo` so it automatically does the damage resolution check for us

server/game/cards/04_JTL/events/ShootDown.ts
```diff
@@ -0,0 +1,51 @@
...
+                title: 'Deal 2 damage to a base',
+                optional: true,
+                thenCondition: () => thenContext.events.length > 0,
```
AMMayberry1, line 29: Will need to be updated to look like this:

```ts
ifYouDoCondition: (ifYouDoContext) => ifYouDoContext.events[0].willDefeat
```

The `willDefeat` property indicates that the unit was dealt enough damage to defeat it. Right now it's not doing the check if the unit was defeated, it's just checking if damage was dealt

test/server/cards/04_JTL/events/ShootDown.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+            it('should neither defeat a unit nor deal damage to a base', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.shootDown);
+
+                expect(context.player1).toBeAbleToSelectExactly([context.allianceXwing, context.tielnFighter, context.systemPatrolCraft]);
+                context.player1.clickCard(context.systemPatrolCraft);
+                expect(context.systemPatrolCraft.damage).toBe(3);
+                expect(context.player2).toBeActivePlayer();
+            });
```
AMMayberry1, line 58: Need to add a case where the unit is damaged but not defeated, base damage should not happen

server/game/cards/04_JTL/units/ChewbaccaFaithfulFirstMate.ts
```diff
@@ -0,0 +1,43 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'This unit can\'t be defeated or returned to hand by enemy card abilities',
+            ongoingEffect: [
+                AbilityHelper.ongoingEffects.cardCannot({
+                    cannot: AbilityRestriction.ReturnToHand,
+                    restrictedActionCondition: (context) => !(context.ability instanceof InitiateAttackAction) && context.player !== this.controller,
```
AMMayberry1, line 20: Don't need the `InitiateAttackAction` check, same for line 34

server/game/cards/04_JTL/units/CorporateDefenseShuttle.ts
```diff
@@ -0,0 +1,20 @@
...
+            id: '4332645242',
+            internalName: 'corporate-defense-shuttle',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'This unit can\'t attack',
+            ongoingEffect: [
+                AbilityHelper.ongoingEffects.cannotAttack()
```
AMMayberry1, line 16: nit: doesn't need to be an array

test/server/cards/04_JTL/events/ShootDown.spec.ts
aruizc: in that particular unit test I'm actually testing that scenario as System Patrol Craft is not defeated but damaged and dealing damage to a base is not triggered

AMMayberry1 (approved):

PR #768, opened 2025-03-05
InDebtToCrimsonDawn Implementation

griggsca91:
image


server/game/cards/04_JTL/upgrades/InDebtToCrimsonDawn.ts
```diff
@@ -0,0 +1,35 @@
...
+            id: '7962923506',
+            internalName: 'in-debt-to-crimson-dawn',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addTriggeredAbility({
+            title: 'Exhaust it unless its controller pay 2 resources',
+            when: {
+                onCardReadied: (event, context) => {
```
AMMayberry1, line 17: nit: don't really need the brackets since it's just returning a boolean, can just do `onCardReadied: (event, context) => context.source.parentCard === event.card`

server/game/cards/04_JTL/upgrades/InDebtToCrimsonDawn.ts
```diff
@@ -0,0 +1,35 @@
...
+                    return context.source.parentCard === event.card;
+                }
+            },
+            targetResolver: {
+                mode: TargetMode.Select,
+                choices: (context) => ({
+                    ['Pay 2 resource']: AbilityHelper.immediateEffects.payResourceCost({
```
AMMayberry1, line 24: nit, typo: pay 2 resources (plural)

server/game/cards/04_JTL/upgrades/InDebtToCrimsonDawn.ts
```diff
@@ -0,0 +1,35 @@
...
+                        target: context.source.parentCard.owner,
+                        amount: 2
+                    }),
+                    ['Exhaust it']: AbilityHelper.immediateEffects.exhaust({
```
AMMayberry1, line 28: Would be good if we could include the name of the unit being exhausted in the button string here, just for clarity

test/server/cards/04_JTL/upgrades/InDebtToCrimsonDawn.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+                player1: {
+                    spaceArena: ['green-squadron-awing'],
+                },
+                player2: {
+                    groundArena: [{ card: 'frontier-atrt', upgrades: [{ card: 'in-debt-to-crimson-dawn', ownerAndController: 'player1' }] }, 'consular-security-force'],
+                }
+            });
+        });
+
+        it('should prompt owner of exhausted unit on regroup to select an option', function () {
```
AMMayberry1, line 16: Not really a change but just a definition thing for clarity in case it comes up later - the owner of the card is the person who brought it in their deck, whereas the controller is the one who has control of it on the field rn. It's only relevant if a card has been stolen by the opponent but it's helpful to be aware of the distinction for when it comes up. So in this case technically it would be the controller (but it's fine if you leave the comment since it really doesn't matter here)

test/server/cards/04_JTL/upgrades/InDebtToCrimsonDawn.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+            const { context } = contextRef;
+
+            context.player1.passAction();
+
+            // CASE 1: Frontier AT-RT attacks and is exhausted
```
AMMayberry1, line 21: nit: we used to use comments like this with "CASE X" when we were compacting multiple tests into one test case (long story, we thought we needed to save github actions time), that convention isn't really needed anymore

test/server/cards/04_JTL/upgrades/InDebtToCrimsonDawn.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+            context.player1.clickPrompt('Done');
+            context.player2.clickPrompt('Done');
+
+
+            expect(context.player2).toHaveEnabledPromptButtons(['Pay 2 resource', 'Exhaust it']);
+            context.player2.clickPrompt('Pay 2 resource');
+
+            expect(context.frontierAtrt.exhausted).toBeFalse();
+            expect(context.player2.readyResourceCount).toBe(2);
+        });
```
AMMayberry1, line 63: We should add one more test with something like Bravado that readies the unit to make sure that also works

server/game/cards/04_JTL/upgrades/InDebtToCrimsonDawn.ts
griggsca91: any specific formatting?

like 
`${context.source.parentCard.title}: Exhaust it` 
`Exhaust ${context.source.parentCard.title}` 
`Exhaust "${context.source.parentCard.title}"`

test/server/cards/04_JTL/upgrades/InDebtToCrimsonDawn.spec.ts
griggsca91: ahhh will change.  technically correct is the best correct

server/game/cards/04_JTL/upgrades/InDebtToCrimsonDawn.ts
griggsca91: done!

server/game/cards/04_JTL/upgrades/InDebtToCrimsonDawn.ts
AMMayberry1: The middle one is our convention

griggsca91: gotcha, done!

AMMayberry1 (approved):

PR #773, opened 2025-03-05
Raddus - Holdos final command

priisig:
image


server/game/cards/04_JTL/units/RaddusHoldosFinalCommand.ts
```diff
@@ -0,0 +1,31 @@
...
+        return {
+            id: '8779760486',
+            internalName: 'raddus#holdos-final-command'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'While you control another resistance card, this unit gains Sentinel',
+            condition: (context) => context.player.anyCardsInPlay((card: Card) => card.hasSomeTrait(Trait.Resistance) && card !== context.source),
```
AMMayberry1, line 17: nit: I want to start moving people towards using the newer methods for searching for units / upgrades. In this case, it would look like:

```ts
context.player.hasSomeArenaCard({ trait: Trait.Resistance, otherThan: context.source });
```

server/game/cards/04_JTL/units/RaddusHoldosFinalCommand.ts
```diff
@@ -0,0 +1,31 @@
...
+            matchTarget: (card, context) => card === context.source,
+            ongoingEffect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Sentinel)
+        });
+        this.addWhenDefeatedAbility({
+            title: 'Deal damage equal to this unit\'s power to an enemy unit.',
+            targetResolver: {
+                cardCondition: (card, context) => card.isUnit() && card.controller !== context.source.controller,
```
AMMayberry1, line 24: nit: in this case we can remove the `isUnit()` check and use `cardTypeFilter: WildcardCardType.Unit`. We use `isUnit()` mostly when it's needed to do a cast and be able to read a property off of the card like `cost`

test/server/cards/04_JTL/units/RaddusHoldosFinalCommand.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+            it('should deal damage equal to its power when defeated', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.ruthlessRaider);
+                context.player1.clickCard(context.raddusHoldosFinalCommand);
+                context.player1.clickCard(context.raddusHoldosFinalCommand);
+                context.player2.clickCard(context.avengerHuntingStarDestroyer);
+                expect(context.raddusHoldosFinalCommand.zoneName).toBe('discard');
+                expect(context.avengerHuntingStarDestroyer.zoneName).toBe('discard');
```
AMMayberry1, line 59: Whenever we test a damage effect we try to use a card that won't be defeated by the damage so we can confirm the exact amount

test/server/cards/04_JTL/units/RaddusHoldosFinalCommand.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+                expect(context.ruthlessRaider.zoneName).toBe('discard');
+            });
```
AMMayberry1, line 61: Can we add another test where Raddus has some effect giving +power to confirm that gets used correctly for the ability

server/game/cards/04_JTL/units/RaddusHoldosFinalCommand.ts
AMMayberry1, line 24: Due to some rules correctness issues we hit, we've switch over to always using `context.player` instead of `context.source.controller`

AMMayberry1 (approved):

PR #774, opened 2025-03-05
Add JTL Millennium Falcon and R2-D2, and fix piloting

Frugghi:
| Millennium Falcon | R2-D2 |
| --- | --- |
| image | image |

This also includes a fix to ensure that pilot upgrades can be moved to vehicles that already have a pilot on them.


server/game/cards/04_JTL/units/R2D2Artooooooooo.ts
```diff
@@ -0,0 +1,25 @@
...
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addPilotingConstantAbilityTargetingAttached({
+            title: 'You may play or deploy 1 additional Pilot on this unit',
+            ongoingEffect: AbilityHelper.ongoingEffects.modifyPilotingLimit({ amount: 1 })
+        });
+
+        this.addConstantAbility({
```
Frugghi, line 19: Technically this should be `addPilotingConstantAbility` but that doesn't add the ability to the engine immediately so I used `addConstantAbility` because at the end of the day doesn't make a difference. Unless I'm missing some edge case.

test/server/core/abilities/keyword/Piloting.spec.ts
```diff
@@ -261,6 +261,33 @@ describe('Piloting keyword', function() {
                 expect(context.concordDawnInterceptors).toBeInZone('hand');
                 expect(context.idenVersio).toBeInZone('discard');
             });
+
+            it('can be moved to another vehicle with a Pilot ignoring the limit', async function () {
```
Frugghi, line 265: This has been confirmed [here](https://bsky.app/profile/davflamerock.bsky.social/post/3li3jgzevck27).

server/game/cards/04_JTL/units/R2D2Artooooooooo.ts
AMMayberry1: Ah yeah make sense, I think this is probably the best way to do it since it's sort of like a cost adjustment (happens at the play stage)

AMMayberry1 (approved):

PR #775, opened 2025-03-05
Torpedo Barrage implementation

griggsca91:
image


test/server/cards/04_JTL/events/TorpedoBarrage.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+                player2: {
+                    groundArena: [{ card: 'wampa', upgrades: ['shield'] }, { card: 'boba-fett#disintegrator', upgrades: ['boba-fetts-armor'] }],
+                    spaceArena: ['lurking-tie-phantom'],
+                    hand: ['vanquish'],
+                }
+            });
+
+            const { context } = contextRef;
+
+            const reset = () => {
```
AMMayberry1, line 18: Can remove this since it's not used in the test (and we don't use these reset handlers anymore in general, we instead break into smaller tests that don't need resetting)

test/server/cards/04_JTL/events/TorpedoBarrage.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+        it('Torpedo Barrage\'s ability should deal 5 indirect damage to a player', async function() {
+            await contextRef.setupTestAsync({
+                phase: 'action',
+                player1: {
+                    hand: ['torpedo-barrage'],
+                },
+                player2: {
+                    groundArena: [{ card: 'wampa', upgrades: ['shield'] }, { card: 'boba-fett#disintegrator', upgrades: ['boba-fetts-armor'] }],
+                    spaceArena: ['lurking-tie-phantom'],
+                    hand: ['vanquish'],
```
AMMayberry1, line 12: nit: our convention is to only include cards that are relevant for the test (this does include cards that exist just to validate that e.g. target filters are working right)

test/server/cards/04_JTL/events/TorpedoBarrage.spec.ts
griggsca91: ahhh that got left in there, will do!

test/server/cards/04_JTL/events/TorpedoBarrage.spec.ts
griggsca91: will do

griggsca91: done!

test/server/cards/04_JTL/events/TorpedoBarrage.spec.ts
griggsca91: done!

AMMayberry1 (approved):

PR #777, opened 2025-03-05
Bunch of cards (set4-bunch12)

VJubert:

### DIFF
--- a/server/game/cards/04_JTL/events/Commandeer.ts
+++ b/server/game/cards/04_JTL/events/Commandeer.ts
@@ -1,0 +1,38 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { PhaseName, Trait, WildcardCardType } from '../../../core/Constants';
+
+export default class Commandeer extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: '8105698374',
+            internalName: 'commandeer',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Take control of a non-leader Vehicle unit that costs 6 or less without a Pilot on it. At the start of the regroup phase, return that unit to its owner\'s hand.',
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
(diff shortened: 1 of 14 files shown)


AMMayberry1 (approved):

server/game/cards/04_JTL/events/Commandeer.ts
```diff
@@ -0,0 +1,38 @@
...
+                    AbilityHelper.immediateEffects.delayedCardEffect({
+                        title: 'Return that unit to its owner\'s hand.',
+                        when: {
+                            onPhaseStarted: (context) => context.phase === PhaseName.Regroup
+                        },
+                        immediateEffect: AbilityHelper.immediateEffects.returnToHand()
+                    })
+                ])
+            },
+            ifYouDo: (ifYouDoContext) => ({
```
AMMayberry1, line 32: I think we might need to rearrange things a bit here - so the `ifYouDo` will trigger if the most recent event in the ability succeeded, which in this case is the `delayedCardEffect`. I think we might need to do the takeControl as the second step in the `simultaneous()`, otherwise it may be possible for a player to use this to ready their own unit.

server/game/cards/04_JTL/units/AllegiantGeneralPrydeRuthlessAndLoyal.ts
```diff
@@ -0,0 +1,38 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '9611596703',
+            internalName: 'allegiant-general-pryde#ruthless-and-loyal',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Defeat a non-unique upgrade on the unit',
```
AMMayberry1, line 15: Wrong title

server/game/cards/04_JTL/units/RazorCrestRideForHire.ts
```diff
@@ -0,0 +1,31 @@
...
+            id: '1935873883',
+            internalName: 'razor-crest#ride-for-hire',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Return a non-leader unit that costs 2 or less or an exhausted non-leader unit that costs 4 or less to its owner\'s hand',
+            when: {
+                onCardPlayed: (event, context) =>
```
AMMayberry1, line 17: Caught this earlier and forgot to leave a comment - this should trigger any time a Pilot attaches, not just is played on it. That would include (a) Survivors' Gauntlet moving a card onto it, and (b) pilot leaders attaching to it

PR #787, opened 2025-03-05
implement barrel roll

ssoulier:
[Card](https://swudb.com/card/JTL/228/barrel-roll)

Screenshot 2025-03-05 at 20 57 49

### DIFF
--- a/server/game/cards/04_JTL/events/BarrelRoll.ts
+++ b/server/game/cards/04_JTL/events/BarrelRoll.ts
@@ -17,6 +17,7 @@ export default class BarrelRoll extends EventCard {
                 zoneFilter: ZoneName.SpaceArena,
                 immediateEffect: AbilityHelper.immediateEffects.attack(),
             },
+            // TODO: Use after instead of then when it's implemented
             then: (thenContext) => ({
                 title: 'Exhaust a space unit',
                 thenCondition: () => thenContext.events.length > 0,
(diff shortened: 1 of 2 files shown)


server/game/cards/04_JTL/events/BarrelRoll.ts
```diff
@@ -0,0 +1,31 @@
...
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setEventAbility({
+            title: 'Attack with a space unit',
+            targetResolver: {
+                zoneFilter: ZoneName.SpaceArena,
+                immediateEffect: AbilityHelper.immediateEffects.attack(),
+            },
```
AMMayberry1, line 19: Can you leave a TODO comment here: "Fix this card to use 'after' instead of 'then' once it's implemented". For this card I'm pretty sure it doesn't make a functional difference right now but something to remember for later definitely

test/server/cards/04_JTL/events/BarrelRoll.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+                    context.greenSquadronAwing,
+                    context.allianceXwing,
+                ]);
+
+                context.player1.clickCard(context.greenSquadronAwing);
+                expect(context.greenSquadronAwing.exhausted).toBe(true);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            xit('should not allow to attack with Blue Leader when it\'s in the ground areana', function () {
```
AMMayberry1, line 81: Let's not check this in, it will generate an annoying log message for everybody. Just a TODO is good enough

AMMayberry1 (approved):

PR #796, opened 2025-03-06
Update "lose keyword" effect implementation to match CR4 rules

AMMayberry1:
In CR3 and prior, if a keyword was lost due to an effect it could be regained by subsequent effects. This is no longer true. Updated the implementation and tests to reflect that.


moyerr (approved):
Thanks for getting these in ahead of the Kazuda changes! 👏

PR #806, opened 2025-03-06
add anakin pilot

ssoulier:
[Card](https://swudb.com/card/JTL/197)

Screenshot 2025-03-06 at 21 14 13


server/game/cards/04_JTL/units/AnakinsSkywalkerIllTrySpinning.ts
```diff
@@ -0,0 +1,32 @@
...
+            internalName: 'anakin-skywalker#ill-try-spinning',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addPilotingAbility({
+            type: AbilityType.Triggered,
+            title: 'Return this upgrade to its owner\'s hand',
+            when: {
+                onAttackCompleted: (event, context) => event.attack.attacker.upgrades.filter((upgrade) => upgrade === context.source).length > 0
```
ssoulier, line 18: There was maybe something more straightforward here, but I did not find it.

server/game/cards/04_JTL/units/AnakinsSkywalkerIllTrySpinning.ts
```diff
@@ -0,0 +1,32 @@
...
+            },
+            immediateEffect: AbilityHelper.immediateEffects.chooseModalEffects({
```
AMMayberry1, line 20: A little more straightforward to use a target resolver with `TargetMode.Select`, see K2SO as an example

server/game/cards/04_JTL/units/AnakinsSkywalkerIllTrySpinning.ts
adbronico: You could do:
event.attack.attacker.upgrades.some((upgrade) => upgrade === context.source)

server/game/cards/04_JTL/units/AnakinsSkywalkerIllTrySpinning.ts
```diff
@@ -0,0 +1,32 @@
...
+                amountOfChoices: 1,
+                choices: (context) => ({
+                    [`Return ${this.title} to your hand`]: AbilityHelper.immediateEffects.returnToHand({
```
adbronico, line 23: Change this.title to context.source.title

test/server/cards/04_JTL/units/AnakinSkywalkerIllTrySpinning.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+                        hand: ['anakin-skywalker#ill-try-spinning'],
+                        groundArena: ['escort-skiff'],
+                        spaceArena: ['green-squadron-awing']
+                    },
+                    player2: {
+                        groundArena: ['maul#shadow-collective-visionary'],
+                    }
+                });
+            });
+
```
adbronico, line 18: I would add a test that the ability doesn't trigger when he is a ground unit

As well as one that it can trigger twice per phase

adbronico (changes requested):

server/game/cards/04_JTL/units/AnakinsSkywalkerIllTrySpinning.ts
AMMayberry1: Easiest would be `event.attack.attacker === context.source.parentCard` I think

test/server/cards/04_JTL/units/AnakinSkywalkerIllTrySpinning.spec.ts
ssoulier: I don't think something prevents from doing it several times during the same phase.
I'll add a test to make sure we can do it.

server/game/cards/04_JTL/units/AnakinsSkywalkerIllTrySpinning.ts
ssoulier: This is interesting and worth debugging I think.
It's exactly the direction I took first and it never worked and I am not sure about the reason. I have only an intuition.
I think the targetResolver is too smart and understand than one of the choices do nothing (noAction). As a consequence, it picks the `returnToHand` always without prompting anything to the player.

So I tried the chooseModalEffects approach during my debugging. Because it works I kept this approach and stopped my debugging.

adbronico (dismissed):
lgtm (outside the modal updates)

server/game/cards/04_JTL/units/AnakinsSkywalkerIllTrySpinning.ts
AMMayberry1: Oh yeah there's a way to configure around that, just set `hasLegalTarget: true` in `noAction()`

ssoulier: Excellent. It's indeed what was missing. Now it works.

AMMayberry1 (approved):

PR #807, opened 2025-03-06
Implementing Trace Martez, Trusting Sister

supercodepoet:
image

### DIFF
--- a/server/game/cards/04_JTL/units/TraceMartezTrustingSister.ts
+++ b/server/game/cards/04_JTL/units/TraceMartezTrustingSister.ts
@@ -1,0 +1,29 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { AbilityType, WildcardRelativePlayer } from '../../../core/Constants';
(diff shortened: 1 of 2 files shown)


server/game/cards/04_JTL/units/TraceMartezTrustingSister.ts
```diff
@@ -0,0 +1,29 @@
...
+            type: AbilityType.Triggered,
+            title: 'Heal 2 total damage from any number of units',
+            when: {
+                onAttackDeclared: (event, context) => event.attack.attacker === context.source
+            },
+            immediateEffect: AbilityHelper.immediateEffects.distributeHealingAmong({
+                amountToDistribute: 2,
+                controller: WildcardRelativePlayer.Any,
+                canChooseNoTargets: true,
+                canDistributeLess: true,
```
AMMayberry1, line 24: It doesn't say "up to 2" so this is one of those weird ones where you have to either distribute 0 ("any number" includes 0) or distribute all of it, so this should be false

supercodepoet: gotcha, updated

PR #809, opened 2025-03-06
Bunch of JTL cards

Twet1:
image
image
image
image
image
image
image
image


server/game/cards/04_JTL/events/JamCommunications.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Look at an opponent\'s hand and discard an event from it',
+            immediateEffect: AbilityHelper.immediateEffects.lookAtAndSelectCard((context) => ({
+                target: context.player.opponent.hand,
+                canChooseNothing: false,
+                immediateEffect: AbilityHelper.immediateEffects.discardSpecificCard(),
+                cardCondition: (card) => !card.isUnit()
```
AMMayberry1, line 19: Specifically needs to be an event

server/game/cards/04_JTL/events/StayonTarget.ts
```diff
@@ -0,0 +1,37 @@
...
+                attackerLastingEffects: [
+                    { effect: AbilityHelper.ongoingEffects.modifyStats({ power: 2, hp: 0 }) },
+                    {
+                        effect: AbilityHelper.ongoingEffects.gainAbility({
+                            type: AbilityType.Triggered,
+                            title: 'When this unit deals damage to a base: Draw a card.',
+                            when: {
+                                onDamageDealt: (event, context) => (
+                                    event.damageSource?.attack?.attacker === context.source &&
+                                    ((event.type === DamageType.Combat && event.damageSource.attack.target?.isBase()) || event.type === DamageType.Overwhelm))
```
AMMayberry1, line 27: I think damage type doesn't matter, just needs to be to the base. We should test with something that does indirect damage to confirm that it works

server/game/cards/04_JTL/events/StayonTarget.ts
```diff
@@ -0,0 +1,37 @@
...
+        this.setEventAbility({
+            title: 'Attack with a Vehicle unit. For this attack, it gets +2/+0 and gains: "When this unit deals damage to a base: Draw a card"',
+            initiateAttack: {
+                attackerCondition: (card) => card.hasSomeTrait(Trait.Vehicle),
+                attackerLastingEffects: [
+                    { effect: AbilityHelper.ongoingEffects.modifyStats({ power: 2, hp: 0 }) },
+                    {
+                        effect: AbilityHelper.ongoingEffects.gainAbility({
+                            type: AbilityType.Triggered,
+                            title: 'When this unit deals damage to a base: Draw a card.',
```
AMMayberry1, line 23: nit: since this is still a triggered ability we don't need the "when" part

test/server/cards/04_JTL/events/StayonTarget.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+                const { context } = contextRef;
+
+                context.player1.passAction();
+                context.player2.clickCard(context.vanquish);
+                context.player2.clickCard(context.cloneCommanderCody);
+                context.player1.clickCard(context.stayOnTarget);
+                context.player1.clickCard(context.allianceXwing);
+                context.player1.clickCard(context.tielnFighter);
+                expect(context.player1.handSize).toBe(0);
+            });
```
AMMayberry1, line 53: As mentioned above, indirect damage test would be good. Also a test to make sure that if there are two separate instances of damage in the same attack, we get two separate draw effects

server/game/cards/04_JTL/units/EchoBaseEngineer.ts
```diff
@@ -0,0 +1,23 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '3660641793',
+            internalName: 'echo-base-engineer',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Give a Shield token',
```
AMMayberry1, line 15: nit: we could go ahead and put the "to a damaged Vehicle unit" in as well, for clarity

server/game/cards/04_JTL/events/StayonTarget.ts
Twet1: I removed the damage type and then all tests failed. With the checks in everything works but Indirect doesn't. I've basically mimicked the code from seventh sister integration

test/server/cards/04_JTL/events/SweepTheArea.spec.ts
moyerr, line 1: 🚀 🚀

server/game/cards/04_JTL/events/StayonTarget.ts
AMMayberry1: Seventh Sister is different because it says combat damage specifically. For this case you need something more like the Jango leader ability, I would start from that code and make the changes you need so that it's checking the card as the source instead of the player

AMMayberry1:
This PR has the cards I pulled out and merged (everything except Stay on Target and Echo Base Engineer): https://github.com/SWU-Karabast/forceteki/pull/837

PR #822, opened 2025-03-07
Implementing Covering the wing and Fight fire with fire

aruizc:
image
image

### DIFF
--- a/server/game/cards/04_JTL/events/CoveringTheWing.ts
+++ b/server/game/cards/04_JTL/events/CoveringTheWing.ts
@@ -13,16 +13,16 @@ export default class CoveringTheWing extends EventCard {
     public override setupCardAbilities() {
         this.setEventAbility({
             title: 'Create an X-Wing token',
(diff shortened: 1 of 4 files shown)


server/game/cards/04_JTL/events/CoveringTheWing.ts
```diff
@@ -0,0 +1,29 @@
...
+            internalName: 'covering-the-wing',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Create an X-Wing token',
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.createXWing(),
+                AbilityHelper.immediateEffects.selectCard({
```
AMMayberry1, line 18: Could we actually change this to be a `then` ability so that the optional part will behave a little more intuitively on the use side

AMMayberry1 (approved):

PR #826, opened 2025-03-08
Connection cleanup possibility

CheBato:
- Added some checks to check whether a lobby exists or not.
- Added Interval cleanup function for lobbies.

### DIFF
--- a/server/gamenode/GameServer.ts
+++ b/server/gamenode/GameServer.ts
@@ -396,6 +396,8 @@ export class GameServer {
             const lobby = this.lobbies.get(lobbyId);
             if (!lobby) {
                 logger.info('No lobby for', ioSocket.data.user.username, 'disconnecting');
+                // we forgot to remove the user there from the userLobbyMap
+                this.userLobbyMap.delete(user.id);
                 ioSocket.disconnect();
                 return;
             }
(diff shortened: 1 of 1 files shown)


server/gamenode/GameServer.ts
```diff
@@ -591,33 +595,78 @@ export class GameServer {
...
+        for (const [userId, lobbyId] of this.userLobbyMap.entries()) {
+            if (!this.lobbies.has(lobbyId)) {
+                this.userLobbyMap.delete(userId);
+            }
+        }
+
+        // Ensure all users in lobbies are mapped correctly
+        for (const [lobbyId, lobby] of this.lobbies.entries()) {
+            for (const user of lobby.users) {
+                if (this.userLobbyMap.get(user.id) !== lobbyId) {
```
danbastin, line 666: I think maybe we want to use userLobbyMap as the source of truth here. I think with some of these ghost lobbies we're ending up with a disassociated LobbyUser. If they don't have an entry in the userLobbyMap I think maybe it's better to clean them out of the lobby instead. I think with this is possible we hare changing the lobby association with a user who may have moved on to a new game.

server/gamenode/GameServer.ts
```diff
@@ -591,33 +595,78 @@ export class GameServer {
...
+                }
+            }, 20000);
+        } else {
+            // if lobby doesn't exist delete the mapping
+            this.userLobbyMap.delete(id);
+        }
+    }
+
+    /* possibly utility function for cleanup. */
+    private startPeriodicCleanup() {
```
danbastin, line 641: Seems like a good thing to start at server start.

PR #831, opened 2025-03-08
Han has his moments

ssoulier:
[Han Solo unit](https://swudb.com/card/JTL/203) 

Screenshot 2025-03-08 at 21 38 49

### DIFF
--- a/server/game/cards/04_JTL/units/HanSoloHasHisMoments.ts
+++ b/server/game/cards/04_JTL/units/HanSoloHasHisMoments.ts
@@ -1,7 +1,7 @@
+import AbilityHelper from '../../../AbilityHelper';
+import type { Attack } from '../../../core/attack/Attack';
 import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
-import { AbilityType } from '../../../enums/AbilityType';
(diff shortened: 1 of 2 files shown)


server/game/cards/04_JTL/units/HanSoloHasHisMoments.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public override setupCardAbilities () {
+        this.addPilotingAbility({
+            title: 'Attack with attached unit. If it\'s the Millennium Falcon, it deals its combat damage before the defender.',
+            type: AbilityType.Triggered,
+            when: {
+                onCardPlayed: (event, context) => event.card === context.source && event.playType === PlayType.Piloting
+            },
+            initiateAttack: {
```
AMMayberry1, line 21: Since we already know which card is going to do the attack (attached card), we should set it up so the player doesn't have to select it. This implementation is basically creating a target resolver that can only select one unit, instead we can jump straight to attacking by using immediateEffect instead of initiateAttack:

```ts
immediateEffect: AbilityHelper.immediateEfffects.attack((context) => ({
    target: context.source.parentCard
    attackerLastingEffects: ...
}));
```

AMMayberry1 (approved):

PR #834, opened 2025-03-09
stay on target split out

Twet1:
image

[AMMayberry1](https://github.com/AMMayberry1) [2 days ago](https://github.com/SWU-Karabast/forceteki/pull/809#discussion_r1984186842)
I think damage type doesn't matter, just needs to be to the base. We should test with something that does indirect damage to confirm that it works

Collaborator
Author
@[Twet1](https://github.com/Twet1) Twet1 [2 days ago](https://github.com/SWU-Karabast/forceteki/pull/809#discussion_r1984910683)
I removed the damage type and then all tests failed. With the checks in everything works but Indirect doesn't. I've basically mimicked the code from seventh sister integration

Collaborator
@[AMMayberry1](https://github.com/AMMayberry1) AMMayberry1 [yesterday](https://github.com/SWU-Karabast/forceteki/pull/809#discussion_r1985844778)
Seventh Sister is different because it says combat damage specifically. For this case you need something more like the Jango leader ability, I would start from that code and make the changes you need so that it's checking the card as the source instead of the player


Twet1:
@AMMayberry1 I have moved this, anyone able to take a look at it. Made some progress but stuck. I will sort the other code issues from bunch-of-jtl branch as messed up my commits.

Twet1:
As mentioned above, indirect damage test would be good. Also a test to make sure that if there are two separate instances of damage in the same attack, we get two separate draw effects

AMMayberry1 (approved):

PR #847, opened 2025-03-09
Connecting to the DYNAMODB Successful!

CheBato:
- Added service for the DynamoDB with basic add,edit,query,delete functions.
- Updated the environmental variables
- added a test POST call for the FE to contact the dynamoDB
- added 2 dependencies for the DB
- to be able to connect to the DB you need to add three fields in the local .env file.
AWS_REGION
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY

for the values contact an administrator.

Additionally here is the link to the document to set up a testing local database
https://docs.google.com/document/d/1juocZ5aC-3BGXeOw2W7EjHfEeKXvHCNGfwIoPyUwvFY/edit?usp=sharing

for testing the
NEXTAUTH_SECRET
env variable needs to be set to the same value as on the FE.


server/middleware/AuthMiddleWare.ts
```diff
@@ -0,0 +1,92 @@
...
+        } catch (error) {
+            logger.error('Auth middleware error:', error);
+            // Don't return an error, just continue without authenticated user
+            next();
+        }
+    };
+};
+
+// Middleware to require authentication
+export const requireAuth = (req: Request, res: Response, next: NextFunction) => {
```
AMMayberry1, line 83: Where does this get used?

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,119 @@
...
+            });
+
+            return await this.client.send(command);
+        } catch (error) {
+            logger.error(`DynamoDB putItem error: ${error}`);
+            throw error;
+        }
+    }
+
+    // Query items with the same partition key
```
AMMayberry1, line 59: What does partition key mean / how does it work? Not super familiar with dynamodb

server/middleware/AuthMiddleWare.ts
CheBato: This one is going to be used on API-calls that are gonna need authentication. We don't have them yet but once we set up the preference page and other profile page we will need to authenticate the user.

server/gamenode/GameServer.ts
```diff
@@ -131,6 +136,43 @@ export class GameServer {
...
+                // Get token from handshake auth
+                const token = socket.handshake.auth.token;
+
+                if (token) {
+                    try {
+                        // NEXTAUTH_SECRET should match your Next.js app
+                        const secret = process.env.NEXTAUTH_SECRET || 'your-nextauth-secret';
+
+                        // Decode without verification first to get the id
+                        const decoded = jwt.decode(token) as { id?: string; name?: string };
```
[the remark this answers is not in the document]
CheBato: Unfortunatelly since the authentication is adapted for different uses the middleware being for HTTP and this implementation for sockets we need to differentiate them. What we could do is find the similar lines of code and put them together into a function to reduce repetition of code.

server/services/DynamoDBService.ts
AMMayberry1: Oh interesting, I need to do some reading on nosql apparently. Thanks for the info!

server/gamenode/GameServer.ts
```diff
@@ -131,6 +136,80 @@ export class GameServer {
...
+                            }
+                        }
+                    } catch (tokenError) {
+                        logger.warn('Token handling error:', tokenError);
+                        // Fall back to query user
+                        if (queryUser) {
+                            socket.data.user = queryUser;
+                        }
+                    }
+                } else if (queryUser) {
```
AMMayberry1, line 200: What happens if queryUser is null here? I'm assuming we should fail somehow

server/gamenode/GameServer.ts
```diff
@@ -376,9 +491,15 @@ export class GameServer {
     // }
 
     public async onConnection(ioSocket) {
-        const user = JSON.parse(ioSocket.handshake.query.user);
+        // First check if the user is authenticated via JWT
+        const authenticatedUser = ioSocket.data.user;
+
+        const queryUser = ioSocket.handshake.query.user ? JSON.parse(ioSocket.handshake.query.user) : null;
+        const user = authenticatedUser || queryUser;
```
AMMayberry1, line 498: This code is pretty similar to the code in the socket middleware and has the same issue (user could end up being null). Probably we should make a private method for this.

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,191 @@
...
+            const result = await this.getItem(`USER#${userId}`, 'METADATA');
+            return result.Item as UserData | undefined;
+        } catch (error) {
+            logger.error(`Error getting user from DynamoDB: ${error}`);
+            throw error;
+        }
+    }
+
+    // Update user's last login time
+    public async updateUserLogin(userId: string) {
```
AMMayberry1, line 163: nit: this name sounds like it's updating the user's credentials. could we use something like `updateUserLoginTime` or `recordNewLogin` or something in that vein

server/middleware/AuthMiddleWare.ts
```diff
@@ -0,0 +1,79 @@
...
+            // 6. Attach the user data to the request for downstream routes
+            req.user = {
+                userId: user.id,
+                username: user.username,
+                email: user.email,
+                provider: user.provider,
+                image: user.avatarUrl
+            };
+        } catch (error) {
+            console.error('Auth middleware error:', error);
```
AMMayberry1, line 63: Is there a reason we don't throw or something here? iirc we're supposed to do `next(error)`

server/gamenode/GameServer.ts
```diff
@@ -131,6 +136,80 @@ export class GameServer {
...
+                                try {
+                                    const dbService = new DynamoDBService();
+                                    await dbService.updateUserLogin(socket.data.user.id);
+                                } catch (dbError) {
+                                    logger.warn('Failed to update user login time:', dbError);
+                                }
+                            }
+                        } catch (jwtError) {
+                            // If standard verification fails, the token might be from the session cookie
+                            logger.warn('Standard JWT verification failed:', jwtError.message);
```
AMMayberry1, line 175: Is there a way to check this without needing to resort trying and catching error? Typically having complex operations - especially outgoing requests - is not good to do inside of a `catch` block, because then you run the risk of throwing an exception from inside an exception handler and that gets really weird

server/gamenode/GameServer.ts
```diff
@@ -131,6 +136,80 @@ export class GameServer {
...
+                                if (user) {
+                                    socket.data.user = {
+                                        id: user.id,
+                                        username: user.username || 'Anonymous'
+                                    };
+                                }
+                            } else if (queryUser) {
+                                // Last resort: fall back to query user
+                                socket.data.user = queryUser;
+                            }
```
AMMayberry1, line 191: Left a couple notes already but a lot of these exception handlers will always fall through when it feels like they should have an option to either throw or call `next(error)` in certain cases

server/gamenode/GameServer.ts
```diff
@@ -131,6 +136,80 @@ export class GameServer {
...
+                // Get token from handshake auth
+                const token = socket.handshake.auth.token;
+                // Get user from query (for backward compatibility)
+                const queryUser = socket.handshake.query.user
+                    ? JSON.parse(socket.handshake.query.user as string) : null;
+
+                if (token) {
+                    try {
+                        // Verify JWT token
+                        const secret = process.env.NEXTAUTH_SECRET || '';
```
AMMayberry1, line 151: If `process.env.NEXTAUTH_SECRET` isn't provided, are we guaranteed to fail the `jwt.verify` call? If so we could shortcut past this by checking that earlier and maybe log a warning or something

server/gamenode/GameServer.ts
```diff
@@ -131,6 +136,80 @@ export class GameServer {
...
+                            } else if (queryUser) {
+                                // Last resort: fall back to query user
+                                socket.data.user = queryUser;
+                            }
+                        }
+                    } catch (tokenError) {
+                        logger.warn('Token handling error:', tokenError);
+                        // Fall back to query user
+                        if (queryUser) {
+                            socket.data.user = queryUser;
```
AMMayberry1, line 197: If it's possible, I think it might be time to go ahead and make a `User` wrapper class that we pass around internally during calls, with subclasses like `AuthenticatedUser` and `AnonymousUser`. And then they could have like `isAuthenticatedUser()` and `isAnonymousUser()` methods like we do with cards. With that we would want a factory that accepts the available auth data from the request and builds the right type of user wrapper.

Since we're going to start to have a lot of branching logic depending on the user type, I think making it very clear and consistent will prevent a lot of problems in advance.

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,191 @@
...
+                secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY || ''
+            }
+        });
+
+        this.client = DynamoDBDocumentClient.from(dbClient);
+    }
+
+    // Get an item by its primary key
+    public async getItem(pk: string, sk: string) {
+        try {
```
AMMayberry1, line 43: Instead of having to manually wrap each of these in a try-catch with the same catch logic, we could create a private method which accepts a handler and wraps it in the try-catch. I plan to do the same thing for GameServer and Lobby api routes once I have a little more time.

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,191 @@
...
+    private tableName = 'KarabastGlobalTable';
+
+    public constructor() {
+        // Configure the DynamoDB client
+        const dbClient = new DynamoDBClient({
+            region: process.env.AWS_REGION || 'us-east-1',
+            credentials: {
+                accessKeyId: process.env.AWS_ACCESS_KEY_ID || '',
+                secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY || ''
+            }
```
AMMayberry1, line 35: We probably need to create a second dynamodb that we use for local testing so that we can let people do dev work without needing to grant them rw access to actual user data

server/gamenode/GameServer.ts
CheBato: I've tried to resolve this by simplifying the try catch into only one where it is really needed. Let me know if this is better.

server/gamenode/GameServer.ts
CheBato: Absolutely as soon as i get the data we need to save into the dynamoDB I'll create wrappers for users.

server/gamenode/GameServer.ts
CheBato: I think it is redundant that we check this here since the socket authentication handles whether a user is null,JWT authenticated or anon.

server/services/DynamoDBService.ts
CheBato: Absolutely i have plans to create a testing database. I'll try to do that today when I get back.

server/middleware/AuthMiddleWare.ts
```diff
@@ -0,0 +1,59 @@
...
+                username: string;
+                email?: string;
+                image?: string;
+                provider?: string;
+            };
+        }
+    }
+}
+
+export const authMiddleware = () => {
```
AMMayberry1, line 20: To further disambiguate, could we call this `expressAuthMiddleware` and the other `SocketIoAuthenticationService` or something in that vein?

server/services/AuthenticationService.ts
```diff
@@ -0,0 +1,84 @@
...
+export class AuthService {
+    private dbService = new DynamoDBService();
+
+    /**
+     * Authenticate a user via JWT token
+     * @param token JWT token
+     * @returns The authenticated user or null if authentication failed
+     */
+    public async authenticateWithToken(token: string): Promise<{ id: string; username: string } | null> {
+        if (!token) {
```
AMMayberry1, line 13: I don't think there's a real risk here but just as a general rule if we're going to have a try block inside of a connection method like this, I'd prefer that we just put all code inside of it to remove the chance of like a weird unexpected null value error or something that could cause a server reset

server/services/AuthenticationService.ts
```diff
@@ -0,0 +1,84 @@
...
+
+            if (!user) {
+                // Create a new user if not found
+                const newUser = {
+                    id: userId,
+                    username: username,
+                    email: decoded.email || null,
+                    provider: decoded.provider || 'unknown',
+                    avatarUrl: decoded.picture || null,
+                    lastLogin: new Date().toISOString(),
```
AMMayberry1, line 40: I very much doubt it would ever be an issue but it might be good to have one variable like `const currentTime = new Date().toISOString()` and then set lastLogin and createdAt to that so that there's no risk of of e.g. lastLogin being an earlier time than createdAt.

server/middleware/AuthMiddleWare.ts
```diff
@@ -0,0 +1,59 @@
...
+                // No token found, so no user info. We let the request proceed without an attached user.
+                return next();
+            }
+            const basicUser = await authService.authenticateWithToken(token);
+            if (!basicUser) {
+                return res.status(401).json({ error: 'Authentication failed' });
+            }
+
+            // Get full user data for HTTP requests
+            const fullUser = await authService.getFullUserData(basicUser.id);
```
AMMayberry1, line 38: Is it needed to attach the additional user data such as email in all cases? It's not a major privacy risk if it gets leaked somehow but generally the less user data we can pass around the better

server/services/AuthenticationService.ts
```diff
@@ -0,0 +1,84 @@
...
+                return null;
+            }
+
+            const userId = decoded.id || decoded.sub;
+            const username = decoded.name || 'Anonymous';
+
+            // Try to get the user from the database
+            let user = await this.dbService.getUserById(userId);
+
+            if (!user) {
```
AMMayberry1, line 32: Does dynamodb have any concept of like a lock or a lease? There's a potential race condition here where another async call is already doing the create user call to dynamodb. I don't _think_ it's an issue since the operation is idempotent, but if we ever add a delete user operation or something it would be really good to have an already established pattern of using data locks (if possible).

If dynamodb doesn't support it directly we made want to add an option to do a simple lock by e.g. userid in our dynamo db service

server/middleware/AuthMiddleWare.ts
CheBato: it is not needed atm i've just been sending all the data. What we only need is the userId and username. I'll fix it up so we only supply these parameters.

server/services/AuthenticationService.ts
CheBato: sounds great

server/services/AuthenticationService.ts
CheBato: makes sense i'll run it all in the try/catch

server/middleware/AuthMiddleWare.ts
CheBato: ofcourse we can!

server/services/AuthenticationService.ts
CheBato: i'll have to take a look at the docs and figure out if they have a lock or lease. You're concern is completely justified and I've had scenarios where this was relevant in a Real world setting, when latency can be an issue.

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,351 @@
...
+     * Save a deck to the database, checking first if a deck with the same link already exists
+     * @param deckData The deck data to save
+     * @param user submitting user
+     * @returns Promise that resolves to the saved deck ID
+     */
+    public async saveDeck(deckData: IDeckData, user: User): Promise<string> {
+        try {
+            // Make sure the deck has the required fields
+            if (!deckData.userId || !deckData.deck) {
+                throw new Error('Deck data is missing required fields');
```
AMMayberry1, line 130: nit: error needs details on which field is missing

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,351 @@
...
+            }
+
+            // Check if a deck with this link already exists for this user
+            const deckLink = deckData.deck.deckLink;
+            if (deckLink) {
+                const existingDeck = await this.dbService.getDeckByLink(user.getId(), deckLink);
+
+                if (existingDeck) {
+                    // If the deck already exists, update it instead of creating a new one TODO delete this after dev
```
AMMayberry1, line 139: What's the TODO for? Don't quite follow

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,351 @@
...
+                    logger.info(`DeckService: Deck with link ${deckLink} already exists for user ${user.getUsername()}, updating existing deck`);
+
+                    // Use the existing deck's ID
+                    const updatedDeckData = {
+                        ...deckData,
+                        id: existingDeck.id
+                    };
+
+                    // Update existing deck
+                    await this.dbService.saveDeck(updatedDeckData);
```
AMMayberry1, line 149: If the deck already exists, are we actually changing anything here?

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,351 @@
...
+    /**
+     * Delete multiple decks for a user
+     * @param userId The user ID
+     * @param deckIds Array of deck IDs to delete
+     * @returns Promise that resolves to an array of successfully deleted deck IDs
+     */
+    public async deleteDecks(userId: string, deckIds: string[]): Promise<string[]> {
+        try {
+            if (!userId || !deckIds || deckIds.length === 0) {
+                logger.warn(`DeckService: Invalid parameters for delete operation. userId: ${userId}, deckIds: ${deckIds}`);
```
AMMayberry1, line 183: Definitely should be contract check

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,351 @@
...
+        }
+    }
+
+    /**
+     * Delete multiple decks for a user
+     * @param userId The user ID
+     * @param deckIds Array of deck IDs to delete
+     * @returns Promise that resolves to an array of successfully deleted deck IDs
+     */
+    public async deleteDecks(userId: string, deckIds: string[]): Promise<string[]> {
```
AMMayberry1, line 180: Are we planning to support batch deck deletion? If not then I think this would be better as a single-deck operation, just simpler that way

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,351 @@
...
+                    deletedDeckIds.push(deckId);
+                    logger.info(`DeckService: Successfully deleted deck ${deckId} for user ${userId}`);
+                } catch (error) {
+                    logger.error(`DeckService: Error deleting deck ${deckId} for user ${userId}:`, error);
+                    failedDeckIds.push(deckId);
+                }
+            }
+
+            if (failedDeckIds.length > 0) {
+                logger.warn(`DeckService: Failed to delete ${failedDeckIds.length} decks: ${failedDeckIds.join(', ')}`);
```
AMMayberry1, line 212: We already have an error log for each failure so I think this might be redundant. Definitely we would need to send an error message to the user, but that can wait until after the bigger discussion around error handling for all of this

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,351 @@
...
+                );
+
+                if (deck) {
+                    logger.info(`DeckService: Found deck with matching deckID/deckLID property: ${deck.id}`);
+                    return deck;
+                }
+            }
+
+            // Not found by any method
+            logger.info(`DeckService: Deck with ID ${deckId} not found for user ${userId} (checked both direct ID and deckID/deckLID properties)`);
```
AMMayberry1, line 344: nit: should be log.error

server/gamenode/GameServer.ts
```diff
@@ -280,6 +238,133 @@ export class GameServer {
             }
         });
 
+        // user DECKS
+        app.post('/api/get-decks', authMiddleware(), async (req, res, next) => {
+            const { decks } = req.body;
+            const user = req.user as User;
+            if (user.isAnonymousUser()) {
```
AMMayberry1, line 245: Are we still supporting decks from local storage?

server/gamenode/GameServer.ts
```diff
@@ -280,6 +238,133 @@ export class GameServer {
...
+                return res.status(403).json({
+                    success: false,
+                    message: 'User is not logged in'
+                });
+            }
+            // we try to sync the decks first
```
AMMayberry1, line 251: This will probably be kind of a slow operation, we shouldn't do it on every call to this API if we can help it. It should only need to happen on login, right?

server/gamenode/GameServer.ts
```diff
@@ -280,6 +238,133 @@ export class GameServer {
...
+                const usersDecks = await this.deckService.getUserDecksFavouritesFirst(user.getId());
+                return res.status(200).json(usersDecks);
+            } catch (err) {
+                logger.error('GameServer: Error in getting a users decks: ', err);
+                next(err);
+            }
+        });
+
+        // Add this to the setupAppRoutes method in GameServer.ts
+        app.get('/api/get-deck/:deckId', authMiddleware(), async (req, res, next) => {
```
AMMayberry1, line 268: nit: the convention I'm used to it would have this as `get-deck:deckId` instead of `get-deck/:deckId`, but if you've seen the latter more then feel free to keep it as-is

server/gamenode/GameServer.ts
```diff
@@ -280,6 +238,133 @@ export class GameServer {
...
+                next(err);
+            }
+        });
+
+        app.post('/api/save-deck', authMiddleware(), async (req, res, next) => {
+            try {
+                const { deck } = req.body;
+                const user = req.user as User;
+                if (user.isAnonymousUser()) {
+                    return res.status(403).json({
```
AMMayberry1, line 317: Should be 401 I think?

server/gamenode/GameServer.ts
```diff
@@ -280,6 +238,133 @@ export class GameServer {
...
+                        success: false,
+                        message: 'User is not logged in'
+                    });
+                }
+                // we save the deck
+                await this.deckService.saveDeck(deck, user);
+                return res.status(200).json({
```
AMMayberry1, line 324: nit: the typical pattern for a 200 code on a POST / PUT is to include the entity definition in the response

server/gamenode/GameServer.ts
```diff
@@ -280,6 +238,133 @@ export class GameServer {
...
+                    return res.status(401).json({
+                        success: false,
+                        message: 'Authentication required'
+                    });
+                }
+
+                if (!deckIds || !Array.isArray(deckIds) || deckIds.length === 0) {
+                    return res.status(400).json({
+                        success: false,
+                        message: 'deckIds must be a non-empty array of deck IDs'
```
AMMayberry1, line 350: Will this message be seen by the user? If so, we should have something more generic (since this is probably a failure on our end) and log this message

server/gamenode/GameServer.ts
```diff
@@ -289,19 +374,19 @@ export class GameServer {
...
+        app.post('/api/create-lobby', authMiddleware(), async (req, res, next) => {
             try {
-                const { user, deck, format, isPrivate, lobbyName } = req.body;
+                const { deck, format, isPrivate, lobbyName } = req.body;
+                const user = req.user;
                 // Check if the user is already in a lobby
-                const userId = typeof user === 'string' ? user : user.id;
-                if (!this.canUserJoinNewLobby(userId)) {
+                if (!this.canUserJoinNewLobby(user.getPlayerId())) {
                     return res.status(403).json({
```
AMMayberry1, line 383: I just caught this - definitely we shouldn't be returning 403 for this situation everywhere. Let's just leave a TODO comment though, this PR is plenty big already

server/gamenode/GameServer.ts
```diff
@@ -375,12 +459,12 @@ export class GameServer {
             }
         });
 
-        app.post('/api/enter-queue', async (req, res, next) => {
+        app.post('/api/enter-queue', authMiddleware(), async (req, res, next) => {
```
AMMayberry1, line 462: Not sure what the normal pattern is here - should it be creating a new middleware object every time?

server/gamenode/Lobby.ts
```diff
@@ -586,6 +588,124 @@ export class Lobby {
...
     }
 
+    /**
+     * Updates deck statistics when a game ends
+     * @param game The game that has ended
+     */
+    private async endGameUpdateStats(game: Game): Promise<void> {
+        try {
+            // Only update stats if the game has a winner
+            if (!game.winner || !game.finishedAt) {
```
AMMayberry1, line 598: Not sure if this is the case currently but we need to make sure that disconnects don't get counted for stats. Also we should come up with a policy for how to handle concedes - potentially could depend on how long the game has gone?

server/gamenode/Lobby.ts
```diff
@@ -586,6 +588,124 @@ export class Lobby {
...
+            if (players.length !== 2) {
+                logger.warn(`Lobby ${this.id}: Cannot update stats for game with ${players.length} players`);
+                return;
+            }
+
+            const [player1, player2] = players;
+
+            // Determine the winner and loser
+            let winner, loser;
+            if (game.winner.includes(player1.name)) {
```
AMMayberry1, line 615: Hmm, probably `game.winner` and `game.loser` shouldn't be arrays

server/gamenode/Lobby.ts
```diff
@@ -586,6 +588,124 @@ export class Lobby {
...
+                winner = player1;
+                loser = player2;
+            } else if (game.winner.includes(player2.name)) {
+                winner = player2;
+                loser = player1;
+            }
+
+            // If we have a draw (or couldn't determine winner/loser), set as draw
+            const isDraw = !winner || !loser || game.winner.length > 1;
```
AMMayberry1, line 624: Definitely we need to update the game-end logic to make it clearer that a draw happened

server/gamenode/Lobby.ts
```diff
@@ -586,6 +588,124 @@ export class Lobby {
...
+                logger.warn(`Lobby ${this.id}: Could not get DeckService`);
+                return;
+            }
+            if (player1User.socket.user.isAuthenticatedUser()) {
+                // Update stats for player 1's deck
+                if (isDraw) {
+                    await deckService.updateDeckStats(
+                        player1User.socket.user.getId(),
+                        player1User.deck.id,
+                        'draw',
```
AMMayberry1, line 653: Could we set this to a variable instead of repeating the `updateDeckStats()` call?

server/gamenode/Lobby.ts
```diff
@@ -586,6 +588,124 @@ export class Lobby {
...
+            const player2LeaderId = player2User.deck.leader.id;
+            const player2BaseId = player2User.deck.base.id;
+
+            // Get the deck service
+            const deckService = this.server.getDeckService();
+            if (!deckService) {
+                logger.warn(`Lobby ${this.id}: Could not get DeckService`);
+                return;
+            }
+            if (player1User.socket.user.isAuthenticatedUser()) {
```
AMMayberry1, line 647: Would be good to just make a private method for this so we don't repeat the code for player1 and player2

server/services/DynamoDBInterfaces.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface IOpponentStat {
+    leaderId: string;
+    baseId: string;
+    wins: number;
+    losses: number;
+    draws: number;
+}
+
+// Updated stats interface
+export interface IDeckStats {
```
AMMayberry1, line 18: We may want to have a standard naming convention to indicate when an interface represents a dynamoDB data type. Maybe `IDeckStatsEntity` or some pattern like that

server/services/DynamoDBInterfaces.ts
```diff
@@ -0,0 +1,74 @@
+export interface IUserData {
+    id: string;
+    username: string;
+    preferences?: Record<string, any>;
+    playerId?: string;
+}
+
+// Interface for opponent stats
+export interface IOpponentStat {
+    leaderId: string;
```
AMMayberry1, line 10: Is this an internal guid or the FFG ID we are currently using for cards? If it's the latter, we need to switch this to internalName since the FFG IDs are inconsistent and we are moving away from them soon

server/services/DynamoDBInterfaces.ts
```diff
@@ -0,0 +1,74 @@
...
+    player1: string;
+    player2: string;
+    firstTurn: string;
+    winner: string;
+    winnerBaseHealthRemaining: number;
+    player1LeaderId: string;
+    player1BaseId: string;
+    player2LeaderId: string;
+    player2BaseId: string;
+    timestamp?: string;
```
AMMayberry1, line 73: Why does this need to be optional?

server/socket.js
```diff
@@ -67,7 +67,7 @@ class Socket extends EventEmitter {
     }
 
     onAuthenticate(token) {
-        jwt.verify(token, env.secret, (err, user) => {
+        jwt.verify(token, env.NEXTAUTH_SECRET, (err, user) => {
```
AMMayberry1, line 70: Does this need to be `| ''` like in the other instance where it's used?

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * Factory class responsible for creating the appropriate User instance
+ * based on authentication status and data
+ */
+export class UserFactory {
+    private dynamoDbService: DynamoDBService;
+    private static instance: UserFactory;
+
+    // Singleton pattern
+    public static getInstance(): UserFactory {
```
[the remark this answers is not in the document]
CheBato: No i don't think it needs to be. I was trying to debug and fix an issue when react FE in developer mode concurrently calls the same API call within milliseconds as to help you find async issues. I did fix it with the concurrent error handling.

server/services/DynamoDBService.ts
```diff
@@ -1,39 +1,40 @@
...
 export class DynamoDBService {
     private client: DynamoDBDocumentClient;
     private tableName: string;
     private isLocalMode: boolean;
 
+    // singleton instance
+    private static instance: DynamoDBService;
+
+    // Static method to get the singleton instance
+    public static getInstance(localMode?: boolean): DynamoDBService {
```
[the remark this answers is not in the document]
CheBato: i'll check it out and rewrite it.

server/services/DynamoDBService.ts
```diff
@@ -1,39 +1,40 @@
...
+        if (!DynamoDBService.instance) {
+            DynamoDBService.instance = new DynamoDBService(localMode);
```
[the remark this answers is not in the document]
CheBato: yes i absolutely agree. I will use the env variable for development mode.

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,225 @@
...
+     */
+    private async authenticateWithToken(token: string): Promise<{ id: string; username: string; playerId: string } | null> {
+        try {
+            if (!token) {
+                return null;
+            }
+
+            const secret = process.env.NEXTAUTH_SECRET || '';
+
+
```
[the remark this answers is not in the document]
CheBato: will remove whitespace

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,225 @@
...
+        return new AnonymousUser(id || uuid(), name || 'Anonymous');
+    }
+
+    /**
+     * Change username with rate limiting (30 days)
+     * @param userId The user ID
+     * @param newUsername The new username
+     * @returns Object with success status and message
+     */
+    public async changeUsername(userId: string, newUsername: string): Promise<{ success: boolean; message: string }> {
```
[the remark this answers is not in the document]
CheBato: atm its not but it will be later on since we have plans for users to be able to change their usernames.

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,225 @@
...
+     * @param token JWT token
+     * @returns The authenticated user or null if authentication failed
+     */
+    private async authenticateWithToken(token: string): Promise<{ id: string; username: string; playerId: string } | null> {
+        try {
+            if (!token) {
+                return null;
+            }
+
+            const secret = process.env.NEXTAUTH_SECRET || '';
```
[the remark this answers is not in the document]
CheBato: in this case currently no it isn't possible to succeed but I will remove the or

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,225 @@
...
+
+
+            const decoded = jwt.verify(token, secret) as any;
+
+            if (!decoded || (!decoded.id && !decoded.sub)) {
+                return null;
+            }
+
+            const userId = decoded.id || decoded.sub;
```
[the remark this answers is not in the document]
CheBato: its a secondary parameter that i forgot to remove.

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,351 @@
...
+            if (result === 'win') {
+                opponentStat.wins += 1;
+            } else if (result === 'loss') {
+                opponentStat.losses += 1;
+            } else {
+                opponentStat.draws += 1;
+            }
+
+            // Save updated stats
+            await this.dbService.updateItem(
```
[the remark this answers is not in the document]
CheBato: alright i'll move it to the dbservice.

server/gamenode/GameServer.ts
```diff
@@ -152,39 +147,29 @@ export class GameServer {
...
-                if (!authenticated) {
-                    const queryUser = socket.handshake.query.user
-                        ? JSON.parse(socket.handshake.query.user as string)
-                        : null;
-
-                    if (queryUser) {
-                        socket.data.user = queryUser;
-                        authenticated = true;
-                    }
+                const queryUser = socket.handshake.query.user
```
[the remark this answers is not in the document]
CheBato: it probably can yes i'll move it to the userFactory

server/gamenode/GameServer.ts
CheBato: at the moment yes we are.

server/gamenode/GameServer.ts
CheBato: True it only needs to happen once.

server/gamenode/GameServer.ts
CheBato: that makes sense

server/gamenode/GameServer.ts
CheBato: will fix them up. Thank you for the headsup.

server/gamenode/GameServer.ts
CheBato: It might be yes depending on how we will tell the user when something goes wrong.

server/services/DynamoDBInterfaces.ts
CheBato: its the FFG ID. That makes sense i'm assuming we will have to change how we retrieve image files from the s3bucket then since its currently with their IDs.

server/services/DynamoDBInterfaces.ts
CheBato: your guess is as good as mine I must have been seeing things at this point.

server/socket.js
CheBato: no it does not.

server/services/DynamoDBInterfaces.ts
CheBato: that makes sense

server/gamenode/GameServer.ts
CheBato: its not a problem i'll change it :)

server/gamenode/Lobby.ts
CheBato: makes sense. I'm just worried how we will differentiate a connection disconnect from a user just leaving by shutting down the browser. If you know what I mean since a lot of people rather then playing till the end just closes the browser.

server/gamenode/Lobby.ts
CheBato: i'm not quite following in what way do you mean a variable here?

server/utils/user/UserFactory.ts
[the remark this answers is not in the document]
AMMayberry1: Will it be called directly as an API? I'm trying to understand why the return value is formatted like an API response, that would normally be the responsibility of whatever code is calling this method

server/utils/user/UserFactory.ts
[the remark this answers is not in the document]
AMMayberry1: Yeah, probably should be a contract check then

test/server/database/testLocalDBConnection.ts
```diff
@@ -1,5 +1,5 @@
-import { DynamoDBClient, ListTablesCommand, CreateTableCommand } from '@aws-sdk/client-dynamodb';
-import { DynamoDBDocumentClient, PutCommand, GetCommand } from '@aws-sdk/lib-dynamodb';
+import { DynamoDBClient, ListTablesCommand, CreateTableCommand, DeleteTableCommand } from '@aws-sdk/client-dynamodb';
```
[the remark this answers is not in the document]
AMMayberry1: You can just do a `git mv` command to rename it but just be prepared for some weirdness on pushing / pulling

server/utils/deck/DeckService.ts
AMMayberry1: Oh I think it's good to have actually, this call won't happen that much compared to normal game stuff

server/utils/deck/DeckService.ts
AMMayberry1: Oh I see. There's probably a cleaner flow then where we could just have one call to `dbService.saveDeck()` and construct the properties object by either taking the existing one and updating it (existing deck) or else filling in missing values if it's a new deck

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,351 @@
...
+        try {
+            // First, try direct lookup by ID
+            let deck = await this.dbService.getDeck(userId, deckId);
+
+            // If found directly, return it
+            if (deck) {
+                return deck;
+            }
+
+            // If not found directly, try to find by deckID or deckLID property
```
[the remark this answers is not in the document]
AMMayberry1: Oh gotcha, I would use that name instead then. Just a bit clearer

server/gamenode/GameServer.ts
AMMayberry1: Makes sense. The typical pattern we use for APIs at my work is to always show the user a generic "server failure" message unless there's some action on their end that they should do differently. Then we log the details in an internal error log

server/gamenode/Lobby.ts
AMMayberry1: Yeah that's fair, probably something we should raise with the group. Especially the people looking at stats stuff would want to have input I would guess.

One thing I can imagine is creating a button in the game end window that the player can use to manually choose whether to count the results or not if it was a concede / disconnect, but that feels like probably too much work for the end user.

server/gamenode/Lobby.ts
AMMayberry1: Just need to fiddle with the game-end logic a little bit. I might see if I can do that as a separate PR in advance since you have a lot to do with this PR already

server/gamenode/Lobby.ts
AMMayberry1: Looking at it a little closer I might modify the request - I mean generally something like maybe have variables `player1WinLossDraw` and `player2WinLossDraw` and then set up a method `updateDeckStatsForUser(updatingPlayer, opponent, winLossDraw)` that we call with the right values. This if / else stuff duplicates a lot of code and is kind of hard to follow

server/services/DynamoDBInterfaces.ts
AMMayberry1: Yep, that's correct. Does that impact anything here? Not sure if this data goes directly to the FE anywhere

server/socket.js
AMMayberry1: Makes sense, probably want a contract check on in then

server/services/DynamoDBInterfaces.ts
CheBato: What this does impact is the stats screen where we display the image of the opponents leader and base.

server/gamenode/GameServer.ts
CheBato: they wouldn't this would fail for them. For the time being I'm leaving this open to anons until we are ready to push the update and close off the deck section on the FE.

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,225 @@
...
+                return this.createAnonymousUser();
+            }
+
+            const userData = await this.dynamoDbService.getUserProfile(basicUser.id);
+            if (!userData) {
+                logger.warn(`User profile not found for authenticated user ${basicUser.id}`);
+                return this.createAnonymousUser();
+            }
+            return new AuthenticatedUser({ ...userData, playerId: basicUser.playerId });
+        } catch (error) {
```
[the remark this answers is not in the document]
CheBato: What i've done now is that we will throw an error if there is not userData and only authenticate the user as anon if the first check fails.

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Creates a user instance from a JWT token
+     * @param token JWT token
+     * @returns A Promise that resolves to a User instance (either Authenticated or Anonymous)
+     */
+    public async createUserFromToken(token: string): Promise<User> {
+        try {
+            const basicUser = await this.authenticateWithToken(token);
+            if (!basicUser) {
```
AMMayberry1, line 26: Could probably just do this all in one line as a contract check, same at 32

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    /**
+     * Update user preferences
+     * @param userId The user ID
+     * @param preferences The updated preferences object
+     * @returns True if update was successful
+     */
+    public async updateUserPreferences(userId: string, preferences: Record<string, any>): Promise<boolean> {
+        try {
```
AMMayberry1, line 115: Also here feels like these should just allow exceptions to float up and let the request fail

server/gamenode/GameServer.ts
```diff
@@ -194,6 +233,177 @@ export class GameServer {
...
+                const usersDecks = await this.deckService.getUserDecks(user.getId());
+                return res.status(200).json(usersDecks);
+            } catch (err) {
+                logger.error('GameServer: Error in getting a users decks: ', err);
+                next(err);
+            }
+        });
+
+        // Add this to the setupAppRoutes method in GameServer.ts
+        app.get('/api/get-deck:deckId', authMiddleware(), async (req, res, next) => {
```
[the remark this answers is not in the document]
CheBato: yes correct!

server/services/DynamoDBInterfaces.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// Define user interface
+export interface IUserProfileDataEntity extends IUserDataEntity {
+    id: string;
+    username: string;
+    lastLogin: string;
+    createdAt: string;
+    usernameSetAt?: string; // When username was set/changed
+    preferences?: Record<string, any>;
```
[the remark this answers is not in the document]
CheBato: what would be your preference here. Caz mentioned in his document that this will be an object.

server/services/DynamoDBInterfaces.ts
```diff
@@ -0,0 +1,73 @@
...
+    };
+    stats?: IDeckStatsEntity;
+}
+
+// Interface for game record
+export interface IGameRecordEntity {
+    id: string;
+    player1: string;
+    player2: string;
+    firstTurn: string;
```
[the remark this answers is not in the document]
CheBato: This is a record of who started the first turn (had the initiative at the beginning). Here is the reference to the DB schema that Caz wrote https://docs.google.com/document/d/1EzSnr40xbcasSv1pA4J7zCojujSa5ar2_tqTcGaPzxs/edit?tab=t.0

server/gamenode/GameServer.ts
```diff
@@ -194,6 +233,177 @@ export class GameServer {
...
+                    message: 'Server error'
+                });
+            }
+            // we retrieve the decks for the FE
+            try {
+                const usersDecks = await this.deckService.getUserDecks(user.getId());
+                return res.status(200).json(usersDecks);
+            } catch (err) {
+                logger.error('GameServer: Error in getting a users decks: ', err);
+                next(err);
```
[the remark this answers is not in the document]
CheBato: this will fail the request. What it does is forwards the error to a default error handler. Which returns a 500 server error.

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,355 @@
...
+     * @returns The processed deck data
+     */
+    public async convertOpponentStatsForFe(deck: IDeckDataEntity, cardDataGetter: CardDataGetter): Promise<IDeckDataEntity> {
+        // If there are no stats or no opponent stats, return the deck as is
+        if (!deck.stats || !deck.stats.statsByMatchup || !deck.stats.statsByMatchup.length) {
+            return deck;
+        }
+
+        // Process each opponent stat
+        deck.stats.statsByMatchup = await Promise.all(
```
AMMayberry1, line 341: Small optimization - we could maybe remove the `async` from this method if we set this promise to return a `deck` using a chained `.then()`. Not 100% sure that works, but I think it does. And if so, we can just do `return Promise.all(...).then(...);`

Generally when it's possible to return the Promise directly instead of awaiting that's preferable as it removes one layer of async state machine

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,207 @@
...
+            await this.dynamoDbService.updateUserProfile(userId, {
+                username: newUsername,
+                usernameLastUpdatedAt: new Date().toISOString()
+            });
+
+            logger.info(`Username for ${userId} changed to ${newUsername}`);
+            return true;
+        } catch (error) {
+            logger.error('Error changing username:', error);
+            return false;
```
AMMayberry1, line 99: Throw

server/utils/user/UserFactory.ts
```diff
@@ -0,0 +1,207 @@
...
+            // This should never happen but if it does its so we don't create two accounts
+            if (error.name === 'ConditionalCheckFailedException') {
+                // This could happen if there's a race condition while creating user records
+                // or if the unique constraints we're enforcing are violated for other reasons
+                logger.error(`DynamoDB conditional check failed during user authentication: ${error}`);
+                return null;
+            }
+            // This catches both JWT verification errors and database errors
+            logger.error('Authentication error:', error);
+            return null;
```
AMMayberry1, line 204: I could go either way on this one but still feels like it should throw to indicate that an unexpected failure happened, and then let the caller deal with it by catching if they want to

AMMayberry1:
Let's migrate that cosmosdb testing document into the existing setup wiki since there's already a page for setting up local testing. That way it's all in one place

server/gamenode/GameServer.ts
```diff
@@ -241,6 +243,52 @@ export class GameServer {
...
+                if (user.isAnonymousUser()) {
+                    logger.error(`GameServer (change-username): Anonymous user ${user.getId()} is attempting to change username`);
+                    return res.status(401).json({
+                        success: false,
+                        message: 'Authentication required to change username'
+                    });
+                }
+
+                // Validate the new username
+                if (!newUsername || typeof newUsername !== 'string') {
```
AMMayberry1, line 264: _sigh_ we should probably go ahead and implement a profanity filter now. Maybe we could snag the one that Petranaki has set up

server/gamenode/GameServer.ts
```diff
@@ -241,6 +243,52 @@ export class GameServer {
...
+                // Call the changeUsername method
+                const result = await this.userFactory.changeUsername(user.getId(), newUsername);
+                if (result) {
+                    return res.status(200).json({
+                        succeess: true,
+                        message: 'Username successfully changed',
+                        username: result
+                    });
+                }
+                logger.error(`GameServer (change-username): User ${user.getId()} did not wait till 1h has passed from last username change`);
```
AMMayberry1, line 281: We need to set a constant value in UserFactory indicating both the amount of time (as a Date or a Number or something) and the string representation. Right now if we change the value in UserFactory it will be easy to forget to come update these strings

AMMayberry1:
> Let's migrate that cosmosdb local testing document into the existing setup wiki since there's already a page for setting up local testing. That way it's all in one place

Actually, also we should check in at least one shell script that runs the commands from the from the doc (docker compose up and maybe test as well)

server/env.ts
```diff
@@ -9,7 +9,10 @@ const parsedEnv = z
         GAME_NODE_HOST: z.string(),
         GAME_NODE_NAME: z.string(),
         GAME_NODE_SOCKET_IO_PORT: z.coerce.number().int(),
-        SECRET: z.string()
+        AWS_REGION: z.string(),
+        AWS_ACCESS_KEY_ID: z.string(),
+        AWS_SECRET_ACCESS_KEY: z.string(),
+        NEXTAUTH_SECRET: z.string(),
```
AMMayberry1, line 15: Is there a way to make these optional? Local dev run breaks if they're not provided currently

server/gamenode/GameServer.ts
CheBato: Do you know where i might find that?

server/gamenode/GameServer.ts
CheBato: i think the way we need to set this up is going to be different aswell since Caz has the idea that we let them change their username for 1h and then lock it for 4 months. Would it make sense to display this logger.error in the userFactory rather then here where it is easy to get the constant values.

CheBato:
> > Let's migrate that cosmosdb local testing document into the existing setup wiki since there's already a page for setting up local testing. That way it's all in one place
> 
> Actually, also we should check in at least one shell script that runs the commands from the from the doc (docker compose up and maybe test as well)

hehe I like that you mention cosmosdb I think we are using DynamoDB or is this the same thing?

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,428 @@
...
+                logger.error(`Failed to ensure DynamoDB local table exists: ${err}`);
+            });
+        }
+    }
+
+    /**
+     * Ensures the table exists in DynamoDB Local with the appropriate GSI
+     * This is only used in local development mode
+     */
+    private async ensureTableExists(): Promise<void> {
```
AMMayberry1, line 64: Just caught this - all of the async methods need to have that in their name, like `ensureTableExistsAsync`. That will help with validating they're used correctly.

I also have an optional config in the linter that will check for missing `await` on promises, it's not on by default but we definitely should do a manual run at some point before we merge this in since there's a lot of new async code

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,428 @@
...
+                accessKeyId: process.env.AWS_ACCESS_KEY_ID,
+                secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY
+            };
+        }
+
+        const dbClient = new DynamoDBClient(dbClientConfig);
+        this.client = DynamoDBDocumentClient.from(dbClient);
+
+        if (this.isLocalMode) {
+            this.ensureTableExists().catch((err) => {
```
AMMayberry1, line 54: I think we should put this somewhere where it can be properly awaited, it might cause headaches here. We can either have some helper code somewhere that initializes it separately, or we can create a `static async createAsync()` method like with GameServer

server/services/DynamoDBService.ts
AMMayberry1: Also this is probably better named `ensureLocalTableExistsAsync`

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,428 @@
...
+                            }
+                        }
+                    ],
+                    BillingMode: 'PAY_PER_REQUEST'
+                })
+            );
+
+            logger.info(`Created DynamoDB local table '${this.tableName}' with GSI`);
+        } catch (error) {
+            logger.error(`Error creating local DynamoDB table: ${error}`);
```
AMMayberry1, line 117: For dummies like me who forget to start the docker container, let's add a check for `error.code === 'ECONNREFUSED'` and emit a special log message reminding the user to start the container

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    /**
+     * A utility method to wrap DB operations in a try-catch block
+     */
+    private async executeDbOperation<T>(operation: () => Promise<T>, errorMessage: string): Promise<T> {
+        try {
+            return await operation();
+        } catch (error) {
+            logger.error(`${errorMessage}: ${error}`);
```
AMMayberry1, line 129: Note for all of the error logs: I found today that just stringify-ing them like this doesn't give much output. Brian actually added some helper logic that will pretty-print error messages in local and also format them correctly in AWS. Can something like

```ts
logger.error('An error occurred while communicating with DynamoDB', { error: { message: error.message, stack: error.stack } });`

File: server/gamenode/GameServer.ts (hunk 64 above)
Comment (AMMayberry1) [reply server/gamenode/GameServer.ts:264]: I do not, sorry

File: server/gamenode/GameServer.ts (hunk 65 above)
Comment (AMMayberry1) [reply server/gamenode/GameServer.ts:281]: Yeah not a bad idea

File: server/services/DynamoDBService.ts (hunk 69 above)
Comment (AMMayberry1) [reply server/services/DynamoDBService.ts:117]: Actually I guess for a lot of testing we don't need the dynamodb so people might leave it off on purpose. Maybe then this is just a warning log with a gentle reminder that the container might not be turned on

File: server/gamenode/GameServer.ts (hunk 64 above)
Comment (AMMayberry1) [reply server/gamenode/GameServer.ts:264]: Ah wait found it: https://github.com/search?q=repo%3ASWU-Petranaki%2FSWUOnline%20profanity&type=code

File: server/services/DynamoDBInterfaces.ts (hunk 58 above)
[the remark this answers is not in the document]
Comment (AMMayberry1) [reply server/services/DynamoDBInterfaces.ts:31]: Finally took a look at the doc - he wrote it as "object" but I don't know if that means it necessarily needs to be a generic `Record<string, any>` object. Like we could still define an interface type for it like with all this other stuff and it would still be an object.

Also looks like he wrote the same thing for the "deck" property so I'd assume we're fine to do it the same way for both

File: server/services/DynamoDBInterfaces.ts (hunk 59 above)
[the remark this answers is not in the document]
Comment (AMMayberry1) [reply server/services/DynamoDBInterfaces.ts:66]: I'd recommend that we change it to "firstInitiativePlayer" or something similar

**Comment by AMMayberry1** [comment]
> > > Let's migrate that cosmosdb local testing document into the existing setup wiki since there's already a page for setting up local testing. That way it's all in one place
> > 
> > 
> > Actually, also we should check in at least one shell script that runs the commands from the from the doc (docker compose up and maybe test as well)
> 
> hehe I like that you mention cosmosdb I think we are using DynamoDB or is this the same thing?

They're different, I use cosmosdb at work so that's just stuck in my head lol

File: server/gamenode/Lobby.ts [hunk 71]
```diff
@@ -737,6 +754,87 @@ export class Lobby {
...
+
+    /**
+     * Private method to update a players stats
+     */
+    private async updatePlayerStatsAsync(playerUser: PlayerDetails, opponentPlayerUser: PlayerDetails, score: ScoreType) {
+        // Get the deck service
+        const opponentPlayerLeaderId = await this.cardDataGetter.getCardBySetCodeAsync(opponentPlayerUser.leaderID);
+        const opponentPlayerBaseId = await this.cardDataGetter.getCardBySetCodeAsync(opponentPlayerUser.baseID);
+
+        Contract.assertTrue(playerUser.user.isAuthenticatedUser());
```
AMMayberry1, line 766: I was hitting this contract check in local testing when playing with an unauthenticated user - that's an unexpected situation so it shouldn't be throwing an exception. Either we shouldn't call this method if the player isn't authenticated (and keep the contract check) or else check at the top of the method and just return

server/gamenode/GameServer.ts
AMMayberry1: Also a note on the text - I didn't realize this gets shown directly to the user, should be something like "Username can only be changed once every <amount of time>"

server/utils/deck/DeckService.ts
```diff
@@ -0,0 +1,362 @@
...
+            }
+
+            return decks
+                .map((deckData) => ({
+                    // Extract just the inner deck object which matches ILocalStorageDeckData
+                    ...deckData.deck,
+                    deckID: deckData.id
+                }));
+        } catch (error) {
+            logger.error(`Error retrieving decks for user ${userId}:`, { error: { message: error.message, stack: error.stack } });
```
AMMayberry1, line 37: I had an issue where a deck got saved in an incomplete state (there was a contract failure on the BE side, I've left a comment on the needed fix) and now decks can't load for the player at all. Specifically the ID field is missing.

 I'm realizing we probably need to think at least a bit about what happens when data in the DB gets mangled - for the case of decks, when we find that a deck is missing a required field we should probably just do a delete to clean it uo and then move on with sending back the other decks to the FE.

server/env.ts
```diff
@@ -13,6 +13,8 @@ const parsedEnv = z
         AWS_ACCESS_KEY_ID: z.string().optional(),
         AWS_SECRET_ACCESS_KEY: z.string().optional(),
         NEXTAUTH_SECRET: z.string(),
+        SECRET: z.string(),
```
AMMayberry1, line 16: optional?

server/services/DynamoDBService.ts
```diff
@@ -340,7 +359,16 @@ class DynamoDBService {
...
         return await this.executeDbOperationAsync(async () => {
             const result = await this.getItemAsync(`USER#${userId}`, `DECK#${deckId}`);
-            return result.Item as IDeckDataEntity | undefined;
+            try {
+                iDeckDataEntitySchema.parse(result.Item);
+                return result.Item as IDeckDataEntity;
+            } catch (error) {
+                if (error instanceof z.ZodError) {
+                    logger.error(`Retrieved deck ${deckId} has validation issues:`, error.format());
+                    await this.deleteItemAsync(`USER#${userId}`, `DECK#${deckId}`);
```
AMMayberry1, line 368: In cases like this where there is some complex action we want to take on failure, it's better to use an option like `zod.safeParse()` which will _not_ throw on failure but instead give us back a report. That we can check for it, do the delete async operation (not inside an exception handler), and then throw if we want.

Definitely we should factor this logic into its own private method since it's shared across multiple functions

server/env.ts
AMMayberry1: Looks like this one is still not optional, it causes things to break on start. All the env vars should either be optional or have a default value set in `run-server.js`.

A good test is to remove or rename your local .env file, then run the local server and confirm that the server can start in a "blank" state (even though logins obviously won't work)

server/env.ts
```diff
@@ -9,7 +9,11 @@ const parsedEnv = z
         GAME_NODE_HOST: z.string(),
         GAME_NODE_NAME: z.string(),
         GAME_NODE_SOCKET_IO_PORT: z.coerce.number().int(),
-        SECRET: z.string(),
+        AWS_REGION: z.string().optional(),
+        AWS_ACCESS_KEY_ID: z.string().optional(),
+        AWS_SECRET_ACCESS_KEY: z.string().optional(),
+        NEXTAUTH_SECRET: z.string(),
```
AMMayberry1, line 15: Looks like this one is still not optional, it causes things to break on start. All the env vars should either be optional or have a default value set in run-server.js.

A good test is to remove or rename your local .env file, then run the local server and confirm that the server can start in a "blank" state (even though logins obviously won't work)

server/env.ts
```diff
@@ -9,7 +9,11 @@ const parsedEnv = z
...
+        SECRET: z.string().optional(),
```
AMMayberry1, line 16: Is this used for anything at this point?

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,535 @@
...
+                        }
+                    ],
+                    BillingMode: 'PAY_PER_REQUEST'
+                })
+            );
+
+            logger.info(`Created DynamoDB local table '${this.tableName}' with GSI`);
+        } catch (error) {
+            if (error.code === 'ECONNREFUSED') {
+                logger.warn('A gentle reminder that the docker container for the DynamoDB might not be turned on');
```
AMMayberry1, line 167: nit: might be good to clearly lead with "unable to form a connection to the local dynamodb container."

server/services/DynamoDBService.ts
```diff
@@ -0,0 +1,535 @@
...
+            }
+            logger.error('Error creating local DynamoDB table:', { error: { message: error.message, stack: error.stack } });
```
AMMayberry1, line 169: Let's emit this only if it's not the ECONNREFUSED case, to avoid confusing new people who won't be using the local dynamodb server

package.json
```diff
@@ -16,6 +16,8 @@
     "get-cards": "node ./scripts/fetchdata.js"
   },
   "dependencies": {
+    "@aws-sdk/client-dynamodb": "^3.758.0",
+    "@aws-sdk/lib-dynamodb": "^3.758.0",
```
AMMayberry1, line 20: Also need to commit the updated package-lock.json

server/env.ts
CheBato: nope :D

server/gamenode/GameServer.ts
CheBato: @AMMayberry1  this is the last piece that I'm trying to puzzle out.

AMMayberry1:
I think the package-lock.json is stale, I pulled and did `npm install` and it had new changes

server/services/DynamoDBService.ts
```diff
@@ -265,7 +265,6 @@ class DynamoDBService {
             pk: `USER#${userData.id}`,
             sk: 'PROFILE',
             ...userData,
-            usernameSetAt: userData.usernameLastUpdatedAt || new Date().toISOString(),
```
AMMayberry1: Just checking, why is this being removed?

server/utils/user/UserFactory.ts
```diff
@@ -193,8 +277,9 @@ export class UserFactory {
                 username: username,
                 lastLogin: new Date().toISOString(),
                 createdAt: new Date().toISOString(),
-                usernameSetAt: new Date().toISOString(),
-                preferences: { cardback: null },
+                usernameLastUpdatedAt: new Date().toISOString(),
+                welcomeMessage: true,
+                preferences: { cardback: null, isAdmin: false },
```
AMMayberry1, line 282: Since `isAdmin` is a really sensitive value, I'm actually inclined to maybe remove it from here so we don't ever accidentally set it to true with like a write mistake or something. Probably more secure to have a separate dynamodb entry that lists the specific admins by userId, since that's much harder to make mistakes with. We can do that part later though.

server/services/DynamoDBService.ts
CheBato: because we have userData.usernameLastUpdatedAt, they were basically the same property

PR #853, opened 2025-03-10
Better guard against user joining new lobby while in existing one

AMMayberry1:


server/gamenode/GameServer.ts
```diff
@@ -609,13 +645,6 @@ export class GameServer {
             // Check if the user is still disconnected after the timer
             if (lobby.getUserState(id) === 'disconnected') {
                 this.userLobbyMap.delete(id);
-                lobby.removeUser(id);
```
brianries, line 612: I think you're missing a 
this.removeUserMaybeCleanupLobby(lobby, id);  here?

danbastin:
Good start to refactoring some of this. Seems to work very cleanly as I just jumpped in and out of a lobby through the join button. I tried doing similar shenanigans with the invite link and the users lobby listing overrides the invite and just sends them to their current lobby. Both tabs don't update at the same time, but disconnect events behave as expected.

CheBato:
I couldn't add the comment in the code since it isn't something we changed but we need to start checking if the lobby exists. this is in the onConnection. In the case where lobby doesn't exist we don't remove the user from the this.userLobbyMap.
```ts
// 1. If user is already in a lobby
        if (this.userLobbyMap.has(user.id)) {
            const lobbyId = this.userLobbyMap.get(user.id);
            const lobby = this.lobbies.get(lobbyId);
            if (!lobby) {
                logger.info('No lobby for', ioSocket.data.user.username, 'disconnecting');
                ioSocket.disconnect();
                return;
            }
            // we get the user from the lobby since this way we can be sure it's the correct one.
            const socket = new Socket(ioSocket);
            lobby.addLobbyUser(user, socket);
            socket.send('connectedUser', user.id);
```

PR #867, opened 2025-03-10
Implementing All Wings Report In

aruizc:
image

### DIFF
--- a/server/game/cards/04_JTL/events/AllWingsReportIn.ts
+++ b/server/game/cards/04_JTL/events/AllWingsReportIn.ts
@@ -1,6 +1,6 @@
 import AbilityHelper from '../../../AbilityHelper';
 import { EventCard } from '../../../core/card/EventCard';
-import { RelativePlayer, TargetMode, ZoneName } from '../../../core/Constants';
+import { RelativePlayer, TargetMode, WildcardCardType, ZoneName } from '../../../core/Constants';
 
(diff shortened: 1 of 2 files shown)


server/game/cards/04_JTL/events/AllWingsReportIn.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Exhaust up to 2 friendly space units. For each unit exhausted this way, create an X-Wing token',
+            targetResolver: {
+                mode: TargetMode.UpTo,
+                numCards: 2,
+                controller: RelativePlayer.Self,
+                zoneFilter: ZoneName.SpaceArena,
+                cardCondition: (card) => card.isUnit() && !card.exhausted,
```
AMMayberry1, line 21: It will be a little bit trickier than this, we can't just filter out exhausted cards because you are technically allowed to target them for the exhaust effect. But then we need to check and count only the number of units where the exhaust actually did happen.

This probably looks like an `ifYouDo` step that checks `context.events` and only counts ones that are marked as Resolved, but you may need to play around with it a bit. And then obviously we'll want a test case where you choose one ready unit and one exhausted unit to exhaust, and only one x-wing is created

AMMayberry1 (approved):

Issue #860, opened 2025-03-10
Investigate War Juggernaut prompting


AMMayberry1:
We had one user complaint that War Juggernaut's prompt was confusing. Would be good for someone to try it out real quick and see if there are any easy ways we could make it more intuitive.

moyerr:
Here's what it looks like. Maybe the prompt text could be updated to something like "Choose units to deal 1 damage to"
<img width="1499" alt="Image" src="https://github.com/user-attachments/assets/5481898c-5608-4ada-9b03-338c9c11df6e" />

PR #874, opened 2025-03-11
Update War Juggernaut Prompt

adbronico:
Proposed partial fix for:
https://github.com/SWU-Karabast/forceteki/issues/860

I suspect the FE needs some updates to make it more visible too

### DIFF
--- a/server/game/cards/04_JTL/units/WarJuggernaut.ts
+++ b/server/game/cards/04_JTL/units/WarJuggernaut.ts
@@ -22,7 +22,7 @@ export default class WarJuggernaut extends NonLeaderUnitCard {
         });
 
         this.addWhenPlayedAbility({
-            title: 'Deal 1 damage to each of any number of units.',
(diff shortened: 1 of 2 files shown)


server/game/cards/04_JTL/units/WarJuggernaut.ts
```diff
@@ -22,7 +22,7 @@ export default class WarJuggernaut extends NonLeaderUnitCard {
         });
 
         this.addWhenPlayedAbility({
-            title: 'Deal 1 damage to each of any number of units.',
+            title: 'Choose units to deal 1 damage to.',
```
AMMayberry1, line 25: Would be better to leave this as-is and add this string as `targetResolver.activePromptTitle`. The ability title is what will appear if there are multiple triggers in the same window, and the original title is more descriptive in that situation imo.

AMMayberry1 (approved):

PR #879, opened 2025-03-11
implement ImproprietyAmongThieves

priisig:
also added isExhausted() function to UnitProperties

### DIFF
--- a/server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
+++ b/server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
@@ -45,7 +45,7 @@ export default class ImproprietyAmongThieves extends EventCard {
                             },
                             immediateEffect: AbilityHelper.immediateEffects.takeControlOfUnit({
                                 target: context.targets.friendlyUnit,
-                                newController: context.targets.friendlyUnit.controller.opponent,
+                                newController: context.targets.friendlyUnit.owner,
                             }),
                         })),
                         AbilityHelper.immediateEffects.delayedCardEffect((context) => ({
@@ -55,7 +55,7 @@ export default class ImproprietyAmongThieves extends EventCard {
                             },
                             immediateEffect: AbilityHelper.immediateEffects.takeControlOfUnit({
                                 target: context.targets.enemyUnit,
-                                newController: context.targets.enemyUnit.controller.opponent,
+                                newController: context.targets.enemyUnit.owner,
                             }),
                         }))
                     ]),
--- a/server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
+++ b/server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
@@ -1,0 +1,69 @@
+import { EventCard } from '../../../core/card/EventCard';
+import AbilityHelper from '../../../AbilityHelper';
+import { PhaseName, RelativePlayer, WildcardCardType, WildcardZoneName } from '../../../core/Constants';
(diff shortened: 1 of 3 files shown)


adbronico (dismissed):
Looks pretty good, there are some small changes and a few more tests I'd like to see

server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
```diff
@@ -0,0 +1,66 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '1302133998',
+            internalName: 'impropriety-among-thieves'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Choose a ready non-leader unit controlled by each player. ',
```
AMMayberry1, line 16: nit: empty space at the end of the string

server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
```diff
@@ -0,0 +1,66 @@
...
+            targetResolvers: {
+                friendlyUnit: {
+                    activePromptTitle: 'Choose a ready friendly non-leader unit',
+                    controller: RelativePlayer.Self,
+                    cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                    zoneFilter: WildcardZoneName.AnyArena,
+                    cardCondition: (card: IUnitCard) => !card.isExhausted(),
```
AMMayberry1, line 23: We try avoid doing casts like this and instead prefer something like `(card) => card.canBeExhausted() && !card.exhausted`, it's just safer. Same note for the other target resolver.

server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
```diff
@@ -0,0 +1,66 @@
...
+                    dependsOn: 'friendlyUnit',
+                    activePromptTitle: 'Choose a ready enemy non-leader unit',
+                    controller: RelativePlayer.Opponent,
+                    cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                    zoneFilter: WildcardZoneName.AnyArena,
+                    cardCondition: (card: IUnitCard) => !card.isExhausted(),
+                    immediateEffect: AbilityHelper.immediateEffects.simultaneous([
+                        AbilityHelper.immediateEffects.takeControlOfUnit((context) => ({
+                            target: context.targets.friendlyUnit,
+                            newController: context.targets.friendlyUnit.controller.opponent,
```
AMMayberry1, line 35: nit: it shouldn't matter too much here but we generally prefer to use `context.player` over `context.[card].controller`. So in this case it would be `newController: context.player.opponent` and for the other it would be `newController: context.player`

test/server/cards/03_TWI/events/ImproprietyAmongThieves.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+                        player2: {
+                            groundArena: ['seasoned-shoretrooper'],
+                            leader: { card: 'sabine-wren#galvanized-revolutionary', deployed: true },
+                        }
+                    });
+                });
+
+                it('does nothing', () => {
+                    const { context } = contextRef;
+                    context.superlaserTechnician.exhaust();
```
AMMayberry1, line 22: nit: instead of doing it this way can just set exhausted directly in the test setup with `{ card: 'superlaser-technican', exhausted: true }`. Same note for other tests.

AMMayberry1 (approved):

PR #887, opened 2025-03-11
Corvus draft

ssoulier:
[Card](https://www.swudb.com/card/JTL/038/corvus-inferno-squadron-raider)

Screenshot 2025-03-11 at 18 30 07


test/server/cards/04_JTL/units/CorvusInfernoSquadronRaider.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.corvus);
+                expect(context.player1).toBeAbleToSelectExactly([context.astromechPilot, context.determinedRecruit]);
+                expect(context.player1).toHavePassAbilityButton();
+                context.player1.clickPrompt('Pass');
+                expect(context.corvus).toHaveExactUpgradeNames([]);
+                expect(context.greenSquadronAwing).toHaveExactUpgradeNames(['determined-recruit']);
+                expect(context.astromechPilot).toHaveExactUpgradeNames(['devotion', 'protector']);
+            });
```
AMMayberry1, line 62: Couple more quick test cases it would be great to have for safety, since this is kind of a new feature:

- Confirm that non-Pilot upgrades aren't targetable
- Confirm that pilots on opponent units aren't targetable
- Can take a pilot from a friendly ground vehicle
- Check that if we have a Pilot on a unit that is stolen with Change of Heart, we can still grab that pilot with Corvus (since the upgrade controller hasn't changed). Would be best if this was a ground vehicle
- Test with a leader pilot. Currently we can attach leader pilots using the test setup, need to do it manually

AMMayberry1 (approved):

PR #890, opened 2025-03-12
The Ghost, Heart of the Family

moyerr:

### DIFF
--- a/server/game/cards/04_JTL/units/TheGhostHeartOfTheFamily.ts
+++ b/server/game/cards/04_JTL/units/TheGhostHeartOfTheFamily.ts
@@ -15,7 +15,9 @@ export default class TheGhostHeartOfTheFamily extends NonLeaderUnitCard {
             title: 'Each other friendly Spectre unit gains this unit\'s keywords',
             matchTarget: (card, context) => card !== context.source && card.isUnit() && card.hasSomeTrait(Trait.Spectre),
             ongoingEffect: AbilityHelper.ongoingEffects.gainKeywords((target, context) => {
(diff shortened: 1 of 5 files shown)


server/game/cards/04_JTL/units/TheGhostHeartOfTheFamily.ts
```diff
@@ -0,0 +1,30 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addConstantAbility({
+            title: 'Each other friendly Spectre unit gains this unit\'s keywords',
+            matchTarget: (card, context) => card !== context.source && card.isUnit() && card.hasSomeTrait(Trait.Spectre),
+            ongoingEffect: AbilityHelper.ongoingEffects.gainKeywords((target, context) => {
+                const blankedKeywords = context.source.getOngoingEffectValues(EffectName.LoseKeyword);
+                return context.source.getOngoingEffectValues(EffectName.GainKeyword)
```
AMMayberry1, line 19: Is there a reason we couldn't just do `return context.source.keywords;`?

moyerr: @AMMayberry1 Okay so I didn't know that existed, but after taking a look, yeah. We cannot use that because it gives us a `KeywordInstance`, which does not provide an interface for grabbing the original properties of the keyword. And we need the properties to be able to use the `GainKeyword` effect.

I briefly went down the road of adding this interface to `KeywordInstance`, but with all the different keyword property types, I quickly found myself in typescript hell. I think it would be a nice improvement if we could add this though!

AMMayberry1 (approved):

PR #902, opened 2025-03-12
Add Admiral Yularen, Targeting Computer, and Death Star Plans

adbronico:
Also:
- Fixed a small bug with Krennic's state watcher
- Added a new test helper method for checking what an upgrade is attached to

### DIFF
--- a/server/game/cards/04_JTL/upgrades/DeathStarPlans.ts
+++ b/server/game/cards/04_JTL/upgrades/DeathStarPlans.ts
@@ -40,8 +40,7 @@ export default class DeathStarPlans extends UpgradeCard {
             ongoingEffect: AbilityHelper.ongoingEffects.decreaseCost({
                 amount: 2,
                 cardTypeFilter: WildcardCardType.NonLeaderUnit,
-                match: (card) => this.isFirstUnitPlayedThisPhase(card),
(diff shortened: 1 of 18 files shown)


server/game/cards/04_JTL/upgrades/DeathStarPlans.ts
```diff
@@ -0,0 +1,56 @@
...
+            }
+        });
+
+        this.addGainConstantAbilityTargetingAttached({
+            title: 'The first unit you play each round costs 2 less',
+            ongoingEffect: AbilityHelper.ongoingEffects.decreaseCost({
+                amount: 2,
+                cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                match: (card) => this.isFirstUnitPlayedThisPhase(card),
+                limit: AbilityHelper.limit.perRound(1)
```
AMMayberry1, line 44: I thought we concluded it shouldn't have the limit so that it can trigger once for each player?

AMMayberry1: Oh huh, looks like it still works. Maybe there's a limit reset happening under the hood that shouldn't be.

AMMayberry1 (approved):

PR #907, opened 2025-03-13
FaceOff Implementation

priisig:


server/game/cards/04_JTL/events/FaceOff.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'If no player has taken the initiative this phase, you may ready an enemy unit. If you do, ready a friendly unit in the same arena',
+            optional: true,
+            targetResolver: {
+                condition: (context) => !context.game.isInitiativeClaimed,
+                activePromptTitle: 'Choose an enemy unit to ready',
+                controller: RelativePlayer.Opponent,
+                cardCondition: (card) => card.canBeExhausted() && card.exhausted,
```
AMMayberry1, line 21: The way effects work in SWU is a little weird, but it is legal to try and ready a unit that is already readied. If we do, then the `ifYouDo` won't fire, but it's still legal to choose them technically. So basically we shouldn't have this condition here or on line 28

server/game/cards/04_JTL/events/FaceOff.ts
```diff
@@ -0,0 +1,35 @@
...
+                immediateEffect: AbilityHelper.immediateEffects.ready((context) => context.target),
+            },
+            ifYouDo: (ifYouDoContext) => ({
+                title: 'Ready a friendly unit in the same arena',
+                targetResolver: {
+                    controller: RelativePlayer.Self,
+                    cardCondition: (card) => card.canBeExhausted() && card.exhausted && card.zone === ifYouDoContext.target.zone,
```
AMMayberry1, line 28: We need to use `zoneName` instead of `zone` for comparing, the latter could cause issues in some cases since we'd be comparing `Zone` objects directly

AMMayberry1:
Reminder to please include the card image in the PR description to make reviewing a little faster. Thanks!

AMMayberry1 (approved):

PR #915, opened 2025-03-14
Add lobby owner check

AMMayberry1:


server/gamenode/Lobby.ts
```diff
@@ -45,6 +46,7 @@ export class Lobby {
     private readonly cardDataGetter: CardDataGetter;
     private readonly deckValidator: DeckValidator;
     private readonly testGameBuilder?: any;
+    private readonly server: GameServer;
```
CheBato, line 49: This is something i was trying to avoid to do. Basically have GameServer in the lobby since my fear is that we'll accidentally loop call at some point. What do you think about it?

CheBato:
I'm just checking whether it would make sense to change this check in the lobby to set lobbyOwner to null? Would that break anything?
```ts
public removeUser(id: string): void {
        const user = this.users.find((u) => u.id === id);
        if (user) {
            this.gameChat.addMessage(`${user.username} has left the lobby`);
        } else {
            return;
        }
        if (this.game) {
            this.game.addMessage(`${user.username} has left the game`);
            const winner = this.users.find((u) => u.id !== id);
            if (winner) {
                this.game.endGame(this.game.getPlayerById(winner.id), `${user.username} has conceded`);
            }
            this.sendGameState(this.game);
        }

        if (this.lobbyOwnerId === id) {
            const newOwner = this.users.find((u) => u.id !== id);
            if (newOwner) {
                this.lobbyOwnerId = newOwner.id;
            }
        }
        this.users = this.users.filter((u) => u.id !== id);
        logger.info(`Lobby ${this.id}: removing user: ${user.username}, id: ${user.id}. User list size = ${this.users.length}`);

        this.sendLobbyState();
    }
```

PR #919, opened 2025-03-14
Add JTL Chimaera and uncomment Gar Saxon tests

adbronico:


server/game/gameSystems/UseWhenDefeatedSystem.ts
```diff
@@ -0,0 +1,74 @@
...
+
+            Object.assign(promptProperties, { choices, handlers });
+
+            event.context.game.promptWithHandlerMenu(player, promptProperties);
+        }
+    }
+
+    // since the actual effect of the bounty is resolved in a sub-window, we don't check its effects here
+    public override canAffect(card: Card, context: TContext, additionalProperties: any = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        return card.canRegisterTriggeredAbilities() && card.getTriggeredAbilities().some((ability) => ability.isWhenDefeated);
```
AMMayberry1, line 63: We also need to filter on the ability being able to be resolved - this is done using `ability.meetsRequirements(ability.createContext(context.player) === ''`

server/game/gameSystems/UseWhenDefeatedSystem.ts
```diff
@@ -0,0 +1,74 @@
...
+    protected override readonly targetTypeFilter = [WildcardCardType.Unit]; // TODO - add Upgrades for Thrawn
+
+    protected override defaultProperties: IUseWhenDefeatedProperties = {
+        triggerAll: false
+    };
+
+    public eventHandler(event): void {
+        const whenDefeatedSource = event.whenDefeatedSource;
+        const triggerAll = event.triggerAll;
+        const whenDefeatedAbilities = whenDefeatedSource.getTriggeredAbilities().filter((ability) => ability.isWhenDefeated);
```
AMMayberry1, line 24: Typically this is the kind of thing we'd attach directly to the event in `addPropertiesToEvent` instead of needing to compute it again here

server/game/gameSystems/UseWhenDefeatedSystem.ts
```diff
@@ -0,0 +1,74 @@
...
+            Object.assign(promptProperties, { choices, handlers });
+
+            event.context.game.promptWithHandlerMenu(player, promptProperties);
+        }
+    }
+
+    // since the actual effect of the bounty is resolved in a sub-window, we don't check its effects here
+    public override canAffect(card: Card, context: TContext, additionalProperties: any = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        return card.canRegisterTriggeredAbilities() && card.getTriggeredAbilities().some((ability) => ability.isWhenDefeated);
+        // && super.canAffect(card, context, additionalProperties, mustChangeGameState); TODO: add this back in if it doesn't break anything
```
AMMayberry1, line 64: Why was this removed in the first place? Did it break something specific?

test/server/cards/04_JTL/units/ChimaeraReinforcingTheCenter.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+                context.player1.passAction();
+                context.player2.clickCard(context.directHit);
+                context.player2.clickCard(context.chimaera);
+
+                const tieFighters = context.player1.findCardsByName('tie-fighter');
+                expect(tieFighters.length).toBe(2);
+                expect(tieFighters).toAllBeInZone('spaceArena');
+                expect(tieFighters.every((tieFighter) => tieFighter.exhausted)).toBeTrue();
+                expect(context.player2.getArenaCards().length).toBe(0);
+            });
```
AMMayberry1, line 330: Let's also add Raddus since that reads off stats from the unit, which is something we don't have yet

server/game/gameSystems/UseWhenDefeatedSystem.ts
AMMayberry1: Actually to amend that - I guess we only do that check if `mustChangeGameState` is anything other than `None`.

server/game/gameSystems/UseWhenDefeatedSystem.ts
adbronico: No, I just wasn't sure if it was needed since the bounty one didn't have it

server/game/gameSystems/UseWhenDefeatedSystem.ts
adbronico: I added a test for Hevy and he is selectable so I'm not sure if we need this?

AMMayberry1: Oh I'm saying he should _not_ be selectable because his ability will do nothing if there are no enemy ground units

server/game/gameSystems/UseWhenDefeatedSystem.ts
AMMayberry1: It's definitely not a huge difference, more of a consistency thing. The general usage pattern is that the event gets configured when it's generated and then the resolution is just acting on the properties attached to the event. But also typically we try to compute all of the properties and attach them to the event at the beginning in case we need to check or modify them during resolution.

server/game/gameSystems/UseWhenDefeatedSystem.ts
adbronico: Hmm should we filter those out? It's still available to use technically even if there's no targets

I guess it only matters if they print a "if you used a when defeated this phase" card

AMMayberry1 (approved):

PR #923, opened 2025-03-14
Fix #918 - Luke JTL Pilot upgrade should not be able to be defeated by card abilities

### LINKED ISSUE #918: JTL Luke leader can be defeated by upgrade removal
A user reported that Aggression was able to defeat JTL Luke leader when it shouldn't be able to

aruizc:
Adding Unit Test for Luke Pilot upgrade not able to be defeated by card enemy abilities


AMMayberry1:
Ended up fixing as part of the Chewie fix PR here: https://github.com/SWU-Karabast/forceteki/pull/990

PR #932, opened 2025-03-15
Initial Game Object state and registration

CDMayberry:
Just wanted to get this in as a foundation for game state and starting on Undo, as I keep having branches get too far out of sync.


server/game/core/GameObject.ts
```diff
@@ -1,28 +1,43 @@
...
+    public set id(value) {
+        this.state.id = value;
     }
 
     public constructor(
-        public game: Game,
-        name: string
+        game: Game,
+        name: string,
+        id?: string
```
CDMayberry, line 35: @AMMayberry1 This ID field might be the problem. I didn't love that name is also ID, _sometimes_, but it might be best to remove this part and leave it. I thought I had it defaulted to name and I don't think anything started to use the new id argument, but maybe best to just remove it and see.

CDMayberry: There were only 4 tests that broke, and all broke due to "unable to find card by name" ("Finn, This is a Rescue" and "Capture system"), hence thinking this might be related.

server/game/core/GameObjectBase.ts
```diff
@@ -0,0 +1,47 @@
...
+    protected state: T;
+
+    /** ID given by the game engine. */
+    public get uuid() {
+        return this.state.uuid;
+    }
+
+    public set uuid(value) {
+        Contract.assertIsNullLike(this.state.uuid, `Tried to set the engine ID of a object that already contains an ID: ${this.state.uuid}`);
+        this.state.uuid = value;
```
AMMayberry1, line 24: Could we change the name from uuid to something else if we're not going with standard uuid format?

server/game/core/GameObject.ts
AMMayberry1: **EDIT**: _I found the error and it's not specifically related to id, see second comment below. This one is still relevant though._

Yeah so there are a few different uses of "ID" it might be good to disambiguate at this point:

- `uuid`: unique identifier per object within the context of `Game`
- `Card.id`: an internal id from FFG used to identify a printing of a card. so this may be shared by multiple cards, and is also going to be removed in a future pass soon - we realized that it is different for different printings of the same card and therefore not useful as an identifier.
- `Player.id`: an external ID generated at the GameServer / Lobby level to identify a site user
- There's also a similar concept of a card ID that you'll see in `Deck` and `DeckValidator` but we can safely ignore that here, that's outside the context of `Game` for the most part

So realistically, I don't think we actually _need_ this concept of an ID on GameObject since other stuff that inherits from it doesn't need it or uses it in ways that are not consistent with each other. Potentially we could just promote the `uuid` concept to be that and then `Card` and `Player` can have an implementation of it at their own level.

If that's something it would be easier to do in a separate PR that's potentially fine (I can help), but we need to check this one against the FE to make sure everything works since player and card ID need to have their behavior unchanged to communicate data to the FE.

server/game/core/Game.js
```diff
@@ -1338,6 +1353,23 @@ class Game extends EventEmitter {
...
+        );
+
+        gameObject.uuid = 'GameObject_' + this.state.nextId;
+        this.state.nextId += 1;
+        this.allGameObjects.push(gameObject);
+        this.gameObjectMapping.set(gameObject.uuid, gameObject);
+    }
+
+    /** @param {GameObjectBase[]} gameObjects */
+    registerGameObjects(gameObjects) {
```
AMMayberry1, line 1369: nit: we could just have one `registerGameObjects` that accepts `GameObjectBase | GameObjectBase[]` and then just use `Helpers.asArray()` to coerce the input into being an array that we iterate over for creation

server/game/core/GameObject.ts
```diff
@@ -1,28 +1,43 @@
...
     public get name() {
-        return this.nameField;
+        return this.state.nameField;
+    }
+
+    public get id() {
+        return this.state.id;
+    }
+
+    public set id(value) {
```
AMMayberry1, line 28: Do we need to have a setter for this? I was assuming it would be readonly

server/game/core/GameObject.ts
AMMayberry1: Ah okay so the issue is way deep in the test setup infrastructure. Basically it looks like player.nameField is coming up undefined here, it was accessing it not realizing it was private: https://github.com/SWU-Karabast/forceteki/blob/be68c4d526a7ff40e7a2fa264a2eaf52e079c4c0/test/helpers/GameStateBuilder.js#L276

It says `player.player.nameField` b/c the outermost player is actually a `PlayerInteractionWrapper` object. I confirmed that all tests pass if you change all usages of that to `player.player.name`, there are a handful of spots.

I have the fix done locally if you want me to just push it up

server/game/core/GameObject.ts
CDMayberry: Only for the Player object, which as mentioned could be moved out.

server/game/core/GameObjectBase.ts
CDMayberry: I was definitely thinking that but a part of me wants to separate "rename all" from the PR that will touch every file 😅. I was trying to convert the Player object to TS and that caused 100+ files changes so I decided I'd come back for that lol.

server/game/core/Game.js
CDMayberry: Moved this into GameObject manager and merged into one "register" function.

server/game/core/GameObjectBase.ts
```diff
@@ -0,0 +1,47 @@
...
+export abstract class GameObjectBase<T extends IGameObjectBaseState = IGameObjectBaseState> {
+    protected state: T;
+
+    /** ID given by the game engine. */
+    public get uuid() {
+        return this.state.uuid;
+    }
+
+    public set uuid(value) {
+        Contract.assertIsNullLike(this.state.uuid, `Tried to set the engine ID of a object that already contains an ID: ${this.state.uuid}`);
```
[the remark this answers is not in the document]
CDMayberry: That is possible but the intention behind this is if we ever want to "recreate" a game object, we need to be able to set the UUID externally. Hence why the game object registers itself with a higher class and lets it handle the UUID setup; Whether it's creating new UUID or spawning the object and giving it an existing state and UUID.

server/game/core/Game.js
```diff
@@ -363,6 +368,16 @@ class Game extends EventEmitter {
         return this.allArenas.getCards({ condition: predicate });
     }
 
+    /**
+     *
+     * @template {GameObject} T
+     * @param {import('./GameObjectBase.js').GameObjectRef<T>} gameObjectRef
+     * @returns {T}
```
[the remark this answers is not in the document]
CDMayberry: Done and moved into manager class

server/game/core/Game.js
```diff
@@ -67,6 +70,8 @@ class Game extends EventEmitter {
         this.owner = details.owner;
         this.started = false;
         this.playStarted = false;
+        this.allGameObjects = [];
+        this.gameObjectMapping = new Map();
```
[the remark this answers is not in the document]
CDMayberry: Done

server/game/core/GameObject.ts
```diff
@@ -1,28 +1,43 @@
...
 
-export abstract class GameObject {
-    public uuid = uuidv4();
-    protected id: string;
+export interface IGameObjectState extends IGameObjectBaseState {
+    id: string;
+    nameField: string;
+}
+
+// TODO: Rename to TargetableGameObject? Or something to imply this is a object with effects (as opposed to an Ability).
```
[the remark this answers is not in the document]
CDMayberry: Yeah that's a good idea. I might do this under a "Great Rename" PR because unfortunately JS files aren't caught and it throws errors everwhere.

AMMayberry1 (approved):

PR #933, opened 2025-03-15
Added spectator support.

CheBato:
Added the ability for users to spectate currently ongoing games. You require two different browsers to test this.
- First browser for first player
- incognito tab for the second player
- second browser for the spectator

once you start the game the spectator can join and watch the game. Once the game ends he can wait to see if the players will reset the game.

### DIFF
--- a/server/gamenode/GameServer.ts
+++ b/server/gamenode/GameServer.ts
@@ -118,7 +118,6 @@ export class GameServer {
         app.use(cors(corsOptions));
 
         this.setupAppRoutes(app);
-        this.startPeriodicCleanup();
         app.use((err, req, res, next) => {
             logger.error('GameServer: Error in API route:', err);
             res.status(err.status || 500).json({
@@ -783,47 +782,6 @@ export class GameServer {
             }, 20000);
         } catch (err) {
             logger.error('Error in onSocketDisconnected:', err);
-        }
-    }
-
-
-    /* possibly utility function for cleanup. */
-    private startPeriodicCleanup() {
-        try {
-            setInterval(() => {
-                // Check for empty lobbies or inconsistencies
-                this.ensureMappingConsistency();
-
-                // cleanup ghost lobbies.
-                for (const [lobbyId, lobby] of this.lobbies.entries()) {
-                    if (lobby.isEmpty()) {
-                        logger.info(`Found empty lobby with id ${lobbyId} cleaning up.`);
-                        this.lobbies.delete(lobbyId);
-                    }
-                }
-            }, 60000); // Check every minute
-        } catch (err) {
-            logger.error('Error in startPeriodicCleanup:', err);
-        }
-    }
-
-    private ensureMappingConsistency() {
-        // Remove any userLobbyMap entries that point to non-existent lobbies
-        for (const [userId, lobbymap] of this.userLobbyMap.entries()) {
-            if (!this.lobbies.has(lobbymap.lobbyId)) {
-                logger.info(`Found lobby that doesn't exist removing lobby ${lobbymap.lobbyId} from userLobbyMap.`);
-                this.userLobbyMap.delete(userId);
-            }
-        }
-
-        // Ensure all users in lobbies are in the userLobbyMap correctly
-        for (const [, lobby] of this.lobbies.entries()) {
-            for (const user of lobby.users) {
-                if (this.userLobbyMap.get(user.id)?.lobbyId === lobby.id) {
-                    logger.info(`Found misplaced user with id ${user.id} removing from lobby ${lobby.id}.`);
-                    this.removeUserMaybeCleanupLobby(lobby, user.id);
-                }
-            }
         }
     }
 }
--- a/server/gamenode/GameServer.ts
+++ b/server/gamenode/GameServer.ts
@@ -518,10 +518,10 @@ export class GameServer {
                 return;
             }
             const socket = new Socket(ioSocket);
-            lobby.addSpectator(user, socket);
             socket.registerEvent('disconnect', () => {
                 this.onSpectatorDisconnect(user.id);
             });
+            lobby.addSpectator(user, socket);
             return;
         }
 
--- a/server/gamenode/GameServer.ts
+++ b/server/gamenode/GameServer.ts
@@ -86,6 +86,7 @@ export class GameServer {
 
     private readonly lobbies = new Map<string, Lobby>();
     private readonly userLobbyMap = new Map<string, string>();
+    private readonly spectatorLobbyMap = new Map<string, string>();
     private readonly io: IOServer;
     private readonly cardDataGetter: CardDataGetter;
     private readonly deckValidator: DeckValidator;
@@ -155,6 +156,34 @@ export class GameServer {
                 return res.json(this.deckValidator.getUnimplementedCards());
             } catch (err) {
                 logger.error('GameServer: Error in setupAppRoutes:', err);
+                next(err);
+            }
+        });
+
+        app.post('/api/spectate-game', (req, res, next) => {
+            try {
+                const { gameId, user } = req.body;
+                const lobby = this.lobbies.get(gameId);
+
+                // if they are in a game already we give them 403 forbidden
+                if (this.userLobbyMap.get(user.id)) {
+                    return res.status(403).json({
+                        success: false,
+                        message: 'User is already in a game'
+                    });
+                }
+                if (!lobby || !lobby.hasOngoingGame() || lobby.isPrivate) {
+                    return res.status(404).json({
+                        success: false,
+                        message: 'Game not found or does not allow spectators'
+                    });
+                }
+
+                // Register this user as a spectator for the game
+                this.spectatorLobbyMap.set(user.id, lobby.id);
+                return res.status(200).json({ success: true });
+            } catch (err) {
+                logger.error('GameServer: Error in spectate-game:', err);
                 next(err);
             }
         });
@@ -449,6 +478,7 @@ export class GameServer {
     public async onConnection(ioSocket) {
         const user = JSON.parse(ioSocket.handshake.query.user);
         const requestedLobby = JSON.parse(ioSocket.handshake.query.lobby);
+        const isSpectator = ioSocket.handshake.query.spectator === 'true';
 
         if (user) {
             ioSocket.data.user = user;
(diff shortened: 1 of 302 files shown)


server/gamenode/GameServer.ts
```diff
@@ -86,6 +86,7 @@ export class GameServer {
 
     private readonly lobbies = new Map<string, Lobby>();
     private readonly userLobbyMap = new Map<string, string>();
+    private readonly spectatorLobbyMap = new Map<string, string>();
```
AMMayberry1, line 89: Adding another map is risky because it creates another way we could have desynced state - i.e., a user could have an entry both in user and spectator maps and it's something we have to enforce and do correctness checks on.

Could we instead maybe use one map for both and change the value type from `string` to some object like `{ lobbyId, playerOrSpectator }` where playerOrSpectator is an enum type?

server/gamenode/GameServer.ts
```diff
@@ -449,6 +478,7 @@ export class GameServer {
     public async onConnection(ioSocket) {
         const user = JSON.parse(ioSocket.handshake.query.user);
         const requestedLobby = JSON.parse(ioSocket.handshake.query.lobby);
+        const isSpectator = ioSocket.handshake.query.spectator === 'true';
```
AMMayberry1, line 481: Just for my awareness - how is this determined on the FE side?

server/gamenode/Lobby.ts
```diff
@@ -26,6 +26,13 @@ interface LobbyUser {
     deckValidationErrors?: IDeckValidationFailures;
     importDeckValidationErrors?: IDeckValidationFailures;
 }
+interface LobbySpectator {
```
AMMayberry1, line 29: Probably good to have this be a base interface that LobbyUser inherits from since they have the same properties

server/gamenode/Lobby.ts
```diff
@@ -159,6 +171,33 @@ export class Lobby {
         this.updateUserLastActivity(user.id);
     }
 
+    public addSpectator(user: User, socket: Socket): void {
+        const existingSpectator = this.spectators.find((s) => s.id === user.id);
```
AMMayberry1, line 175: We also should do a check to see if the user id exists in the `users` map. Not sure what the correct thing to do is in that situation though.

Also the same check for addLobbyUser - need to make sure they aren't already in the spectators map.

server/gamenode/Lobby.ts
```diff
@@ -520,12 +561,27 @@ export class Lobby {
...
+
     public sendGameState(game: Game): void {
         for (const user of this.users) {
             if (user.state === 'connected' && user.socket) {
                 user.socket.send('gamestate', game.getState(user.id));
             }
         }
+        for (const user of this.spectators) {
+            if (user.socket) {
+                user.socket.send('gamestate', game.getState(user.id));
```
AMMayberry1, line 582: Have we done some checking to validate what the spectator can and can't see?

server/gamenode/GameServer.ts
```diff
@@ -45,6 +45,16 @@ interface QueuedPlayer {
     user: User;
 }
 
+enum UserRole {
+    Player = 'player',
+    Spectator = 'spectator'
+}
+
+interface LobbyMapping {
```
AMMayberry1, line 53: nit: interface names should start with I (realized we've been slipping on this lately)

server/gamenode/GameServer.ts
```diff
@@ -494,31 +503,31 @@ export class GameServer {
         // 0. If user is spectator
         if (isSpectator) {
             // Check if user is registered as a spectator
-            if (!this.spectatorLobbyMap.has(user.id)) {
+            if (!this.userLobbyMap.has(user.id) || this.userLobbyMap.get(user.id).role !== UserRole.Spectator) {
```
AMMayberry1, line 506: nit: can simplify this a bit and not do the same map lookup

```ts
const lobbyUserEntry = this.userLobbyMap.get(user.id);
if (!lobbyUserEntry || lobbyUserEntry.role !== UserRole.Spectator) {
```

AMMayberry1: Actually since this `this.userLobbyMap.get(user.id)` happens several times in the other branches, could just do it at the top of the function

server/gamenode/GameServer.ts
```diff
@@ -494,31 +503,31 @@ export class GameServer {
...
                 this.onSpectatorDisconnect(user.id);
             });
+            lobby.addSpectator(user, socket);
             return;
         }
 
         // 1. If user is already in a lobby
         if (this.userLobbyMap.has(user.id)) {
-            const lobbyId = this.userLobbyMap.get(user.id);
+            const lobbyId = this.userLobbyMap.get(user.id).lobbyId;
```
AMMayberry1, line 530: We should double-check here that the user entry is in Player mode, not Spectator. If they are we probably just remove them from `userLobbyMap` completely and fail the connection, like in the other case

server/gamenode/GameServer.ts
```diff
@@ -566,13 +575,13 @@ export class GameServer {
             if (!user.username) {
                 const newUser = { username: 'Player2', id: user.id };
                 lobby.addLobbyUser(newUser, socket);
-                this.userLobbyMap.set(newUser.id, lobby.id);
+                this.userLobbyMap.set(newUser.id, { lobbyId: lobby.id, role: UserRole.Player });
```
AMMayberry1, line 578: Just realized there's another mapping issue here - what do we do if the user is in already in a lobby and joins another via link? I guess again we just boot them out of both and remove them from the map. This applies regardless of whether they're anon

server/gamenode/GameServer.ts
```diff
@@ -723,15 +732,15 @@ export class GameServer {
 
     public onSpectatorDisconnect(id: string) {
         try {
-            if (!this.spectatorLobbyMap.has(id)) {
+            if (!this.userLobbyMap.has(id)) {
```
AMMayberry1, line 735: Same optimization note here and in `onSocketDisconnected` as earlier - can do one map get and then a null check on the value

server/gamenode/GameServer.ts
```diff
@@ -770,4 +779,41 @@ export class GameServer {
             logger.error('Error in onSocketDisconnected:', err);
         }
     }
+
+
+    /* possibly utility function for cleanup. */
+    private startPeriodicCleanup() {
+        setInterval(() => {
```
AMMayberry1, line 786: This needs a try / catch

server/gamenode/GameServer.ts
CheBato: yup that sounds like a good idea.

server/gamenode/GameServer.ts
```diff
@@ -289,7 +327,7 @@ export class GameServer {
     }
 
     private canUserJoinNewLobby(userId: string) {
-        const previousLobbyForUser = this.userLobbyMap.get(userId);
+        const previousLobbyForUser = this.userLobbyMap.get(userId).lobbyId;
```
AMMayberry1, line 330: This caused an error on my local when I tried to create a game, because `this.userLobbyMap.get(userId)` is `undefined`. Needs to be `this.userLobbyMap.get(userId)?.lobbyId`, would be good to do a pass through both files and see if any other similar instances need updating

server/gamenode/Lobby.ts
```diff
@@ -396,7 +402,11 @@ export class Lobby {
 
     public getUserState(id: string): string {
         const user = this.users.find((u) => u.id === id);
-        return user ? user.state : null;
+        if (user) {
+            return user.state;
+        }
+        const spectator = this.spectators.find((u) => u.id === id);
+        return spectator ? spectator.state : null;
```
AMMayberry1, line 409: nit: simpler as `return spectator?.state;`

server/gamenode/Lobby.ts
```diff
@@ -195,11 +200,12 @@ export class Lobby {
 
     public removeSpectator(id: string): void {
         const spectator = this.spectators.find((s) => s.id === id);
-        if (spectator) {
-            this.spectators = this.spectators.filter((s) => s.id !== id);
-            logger.info(`Lobby ${this.id}: removing spectator: ${spectator.username}, id: ${spectator.id}. Spectator count = ${this.spectators.length}`);
-            this.sendLobbyState();
+        if (!spectator) {
+            return;
```
AMMayberry1, line 204: Can we add a log message here that just says "attempted to remove spectator with \<spectator id\> from lobby \<lobby id\> but they were not found"

server/gamenode/GameServer.ts
```diff
@@ -119,7 +118,7 @@ export class GameServer {
         app.use(cors(corsOptions));
 
         this.setupAppRoutes(app);
-
+        this.startPeriodicCleanup();
```
AMMayberry1, line 121: This is a nontrivial change so we'll want to ask everyone to do another round of local tests to validate that it doesn't break anything

server/gamenode/GameServer.ts
```diff
@@ -699,4 +785,45 @@ export class GameServer {
...
+                this.userLobbyMap.delete(userId);
+            }
+        }
+
+        // Ensure all users in lobbies are in the userLobbyMap correctly
+        for (const [, lobby] of this.lobbies.entries()) {
+            for (const user of lobby.users) {
+                if (!this.userLobbyMap.get(user.id)) {
+                    logger.info(`Found misplaced user with id ${user.id} removing from lobby ${lobby.id}.`);
+                    lobby.removeUser(user.id);
```
AMMayberry1, line 824: Probably should be `removeUserMaybeCleanupLobby()` instead

server/gamenode/GameServer.ts
```diff
@@ -699,4 +785,45 @@ export class GameServer {
...
+            if (!this.lobbies.has(lobbymap.lobbyId)) {
+                logger.info(`Found lobby that doesn't exist removing lobby ${lobbymap.lobbyId} from userLobbyMap.`);
+                this.userLobbyMap.delete(userId);
+            }
+        }
+
+        // Ensure all users in lobbies are in the userLobbyMap correctly
+        for (const [, lobby] of this.lobbies.entries()) {
+            for (const user of lobby.users) {
+                if (!this.userLobbyMap.get(user.id)) {
```
AMMayberry1, line 822: Just caught this - one other thing we should check is not just that the user exists but that the lobby map shows them as being in this lobby. I.e., `this.userLobbyMap.get(user.id)?.lobbyId === lobby`. If it is _not_ the case, then they should be removed from the lobby (we assume that whatever is in the lobby map is the most correct value)

server/gamenode/GameServer.ts
CheBato: I agree absolutely

AMMayberry1 (approved):

Issue #878, opened 2025-03-11
Improve Chat log for Kazuda leader


moyerr:
Currently, there is no indication which unit the player chose to blank with Kazuda's effect. Because of this, the Kazuda player could easily forget which unit they blanked, and the opposing player does not know the full board state. It would be nice to have a visual indicator on board, but as a lower-effort first step, we should log the blanked cards to chat. 

Here is an example of current board and chat after using Kazuda's ability:

<img width=800 src="https://github.com/user-attachments/assets/3c510071-0161-4017-865d-92942f48e2c2" />

PR #934, opened 2025-03-15
Update Kazuda game message and basic damage messages

AMMayberry1:
Did some light cleanup on the game message system and made it so that the Kazuda effect now says the name of the target, as well as "basic" damage cases in which only a single target is being damaged with a fixed amount (eg Daring Raid, basic attacks).


AMMayberry1:
Fixes #878
Fixes #910

PR #951, opened 2025-03-17
Implement Brain Invaders

moyerr:

### DIFF
--- a/server/game/core/ability/ActionAbility.ts
+++ b/server/game/core/ability/ActionAbility.ts
@@ -4,7 +4,6 @@ import { PhaseName } from '../Constants.js';
 import type { IActionAbilityProps } from '../../Interfaces.js';
 import type { Card } from '../card/Card.js';
 import type Game from '../Game.js';
-import { EpicActionLimit } from './AbilityLimit.js';
 
 /**
  * Represents an action ability provided by card text.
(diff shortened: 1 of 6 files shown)


server/game/core/ability/ActionAbility.ts
```diff
@@ -34,6 +35,10 @@ export class ActionAbility extends CardAbility {
 
     public readonly condition?: (context?: AbilityContext) => boolean;
 
+    public get isEpicAction(): boolean {
+        return this.limit instanceof EpicActionLimit;
+    }
```
moyerr, line 40: @AMMayberry1 This was the easiest way I could find to determine whether a given action ability is an Epic Action. Please lemme know if there's a better way!

AMMayberry1: Yeah this is probably the easiest way. One thing though is that I prefer to avoid using `instanceof` whenever possible b/c it requires that we do a full import instead of `import type`, and the more we do that the more likely we are to start hitting import loops.

I always prefer to add a checker method, like `isEpicActionLimit(): this is EpicActionLimit` on `IAbilityLimit`

moyerr:
Closing while I poke around with some of the subtleties of gaining abilities while a card is blanked.

PR #953, opened 2025-03-17
Implemented Padme Amidala Pursuing Peace

notyalcs:
image


server/game/cards/03_TWI/units/PadmeAmidalaPursuingPeace.ts
```diff
@@ -0,0 +1,32 @@
+import AbilityHelper from "../../../AbilityHelper";
+import { NonLeaderUnitCard } from "../../../core/card/NonLeaderUnitCard";
+import { AbilityType, RelativePlayer, WildcardCardType } from "../../../core/Constants";
+
+export default class PadmeAmidalaPursuingPeace extends NonLeaderUnitCard {
+    static implemented: boolean;
```
notyalcs, line 6: I can't find this in other PRs but the docs say to include `<classname>.implemented = true;` which seems to require this be here explicitly. I assume it gets removed automatically on publish to prod or something but I don't know whether this is supposed to be public or private or something else entirely

AMMayberry1 (approved):

PR #954, opened 2025-03-17
Card and Zone Initial State Changes 

CDMayberry:
#### Initial Card State Setup
I've setup basic state definitions for all of the card classes. A number of the higher objects don't have state just overrides, so I was able to skip those. 

Mixins did make it harder, but if you pass in a state interface to the initial class being mixed in it still can work.
```ts
const NonLeaderUnitCardParent = WithUnitProperties(WithStandardAbilitySetup(InPlayCard<INonLeaderUnitCardState>));
```

I kinda got the mixins themselves working to extend the state object definition, but if you take a look at it, it's _ugly_. But it's the best I could do without doing a deep dive into TS typings 😅.

I don't think I captured _all_ state for Cards, but this at least puts the backbone in, so I don't have to worry about branch desyncs.

#### Game Object Manager
GameObjectManager's get method now throws an error if it is given a valid obj/UUID but it can't find it. Because GameObjectRefs should only ever be created from existing objects (ie they should only ever be created from a obj.getRef() call), if the engine is given an UUID and can't find it, something is fundamentally broken (something should be in the state object but isn't) and we need to know ASAP. 

Related, but the GameObjectRef system works flawlessly so far, so that's good news.

#### Zone Changes
Zones are now derived from GameObjectBase.  This means we can use GameObjectRef<Zone> (see Card.ts for an example). Converted the two base classes (SimpleZone and ConcreteArenaZone) that track cards to use the state object.

ConcreteArenaZone had a noticeable rework, it now keeps all of the cards in a list but now maintains the Map object as "cached" value. I added a "onAfterStateSetup" method to GameObjectBase to call on objects after an Undo is triggered, so that they can refresh any values that are computations based on state (such as as the Map in this case). This does mean ConcreteArenaZone is now maintaining two sets of objects, which is not my favorite. I could probably add simplify this down further but it's functional for now.

### DIFF
--- a/server/game/core/GameObjectBase.ts
+++ b/server/game/core/GameObjectBase.ts
@@ -7,6 +7,7 @@ export interface IGameObjectBaseState {
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars, unused-imports/no-unused-vars
 export interface GameObjectRef<T extends GameObjectBase = GameObjectBase> {
+    __isRef: true;
     uuid: string;
 }
 
@@ -29,6 +30,7 @@ export abstract class GameObjectBase<T extends IGameObjectBaseState = IGameObjec
     ) {
         // @ts-expect-error state is a generic object that is defined by the deriving classes, it's essentially w/e the children want it to be.
         this.state = {};
+        // All state defaults *must* happen before registration, so we can't rely on the derived constructor to set the defaults as register will already be called.
         this.onSetupDefaultState();
         this.game.gameObjectManager.register(this);
     }
@@ -37,11 +39,16 @@ export abstract class GameObjectBase<T extends IGameObjectBaseState = IGameObjec
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     protected onSetupDefaultState() { }
 
+    /** Sets the state.  */
     public setState(state: T) {
         this.state = state;
     }
 
+    /** A function for game to call on all objects after all state has been set. for example, to cache calculated values. */
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public onAfterSetState() { }
+
     public getRef<T extends GameObjectBase = this>(): GameObjectRef<T> {
-        return { uuid: this.state.uuid };
+        return { __isRef: true, uuid: this.state.uuid };
     }
 }
--- a/server/game/core/GameObjectManager.ts
+++ b/server/game/core/GameObjectManager.ts
@@ -15,7 +15,11 @@ export class GameObjectManager {
         this._gameObjectMapping = new Map<string, GameObjectBase>();
     }
 
-    public get<T extends GameObjectBase>(gameObjectRef: GameObjectRef<T>): T {
+    public get<T extends GameObjectBase>(gameObjectRef: GameObjectRef<T>): T | null {
+        if (!gameObjectRef?.uuid) {
+            return null;
+        }
+
         const ref = this._gameObjectMapping.get(gameObjectRef.uuid);
         Contract.assertNotNullLike(ref, `Tried to get a Game Object but the UUID is not registered ${gameObjectRef.uuid}. This *VERY* bad and should not be possible w/o breaking the engine, stop everything and fix this now.`);
         return ref as T;
--- a/server/game/core/GameObjectManager.ts
+++ b/server/game/core/GameObjectManager.ts
@@ -15,8 +15,10 @@ export class GameObjectManager {
         this._gameObjectMapping = new Map<string, GameObjectBase>();
     }
 
-    public get<T extends GameObjectBase>(gameObjectRef: GameObjectRef<T>): T | undefined {
-        return this._gameObjectMapping.get(gameObjectRef.uuid) as (T | undefined);
+    public get<T extends GameObjectBase>(gameObjectRef: GameObjectRef<T>): T {
+        const ref = this._gameObjectMapping.get(gameObjectRef.uuid);
+        Contract.assertNotNullLike(ref, `Tried to get a Game Object but the UUID is not registered ${gameObjectRef.uuid}. This *VERY* bad and should not be possible w/o breaking the engine, stop everything and fix this now.`);
+        return ref as T;
     }
 
     public register(gameObject: GameObjectBase | GameObjectBase[]) {
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -114,11 +114,26 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
     protected actionAbilities: ActionAbility[] = [];
     protected constantAbilities: IConstantAbility[] = [];
     protected disableWhenDefeatedCheck = false;
-    protected movedFromZone?: ZoneName = null;
     protected triggeredAbilities: TriggeredAbility[] = [];
 
     private nextAbilityIdx = 0;
     private _zone: Zone;
+
+    protected get hiddenForController() {
+        return this.state.hiddenForController;
+    }
+
+    protected get hiddenForOpponent() {
+        return this.state.hiddenForOpponent;
+    }
+
+    protected get movedFromZone() {
+        return this.state.movedFromZone;
+    }
+
+    protected set movedFromZone(value: ZoneName | null) {
+        this.state.movedFromZone = value;
+    }
 
 
     // ******************************************** PROPERTY GETTERS ********************************************
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -1,6 +1,7 @@
 import type { IActionAbilityProps, IConstantAbilityProps, ISetId, Zone, ITriggeredAbilityProps } from '../../Interfaces';
 import { ActionAbility } from '../ability/ActionAbility';
 import type PlayerOrCardAbility from '../ability/PlayerOrCardAbility';
+import type { IOngoingEffectSourceState } from '../ongoingEffect/OngoingEffectSource';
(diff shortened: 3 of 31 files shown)


server/game/core/card/Card.ts
```diff
@@ -125,11 +149,15 @@ export class Card extends OngoingEffectSource {
     }
 
     public get controller(): Player {
-        return this._controller;
+        return this.game.gameObjectManager.get(this.state.controllerRef);
+    }
+
+    public set controller(value: Player) {
```
AMMayberry1, line 155: If there's no technical issue with this being protected, I would strongly prefer that for rules enforcement reasons

server/game/core/card/propertyMixins/PrintedPower.ts
```diff
@@ -1,14 +1,18 @@
 import * as Contract from '../../utils/Contract';
-import type { Card, CardConstructor } from '../Card';
+import type { Card, CardConstructor, ICardState } from '../Card';
 
 export interface ICardWithPrintedPowerProperty extends Card {
     readonly printedPower: number;
     getPower(): number;
 }
 
+// export interface IWithPrintedPowerState extends ICardState {
```
AMMayberry1, line 9: No longer needed?

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -36,8 +36,14 @@ import type { ILeaderCard } from './LeaderProperties';
 import type { ILeaderUnitCard } from '../LeaderUnitCard';
 import type { PilotLimitModifier } from '../../ongoingEffect/effectImpl/PilotLimitModifier';
 import type { AbilityContext } from '../../ability/AbilityContext';
+import type { GameObjectRef } from '../../GameObjectBase';
 
 export const UnitPropertiesCard = WithUnitProperties(InPlayCard);
+export interface IUnitPropertiesCardState extends IInPlayCardState {
+    // placeholder code, not used.
```
AMMayberry1, line 43: To clarify, what's the use for this?

server/game/core/zone/ConcreteArenaZone.ts
```diff
@@ -2,29 +2,35 @@ import type { IInPlayCard } from '../card/baseClasses/InPlayCard';
...
+export abstract class ConcreteArenaZone extends ConcreteOrMetaArenaZone<IConcreteArenaZoneState> implements IAddRemoveZone {
     public override readonly hiddenForPlayers: null;
     public abstract override readonly name: ZoneName;
     public override readonly owner: Game;
 
-    protected _cards = new Map<Player, IInPlayCard[]>();
+    // STATE: This is a "cached" state field, it's a field derived from one or more values on the state object, and needs to be updated after a state change.
+    protected _playerCards = new Map<Player, IInPlayCard[]>();
 
     public override get cards(): IInPlayCard[] {
```
AMMayberry1, line 27: I know you explained but I'm still kind of confused what's going on with this change - do we now need to maintain the state of both `this.state.cards` and `this._playerCards` separately?

server/game/core/zone/SimpleZone.ts
```diff
@@ -1,47 +1,57 @@
...
 /**
  * Base class for zones that are basically just a list of cards with no special behavior
  */
-export abstract class SimpleZone<TCard extends Card> extends ZoneAbstract<TCard> implements IAddRemoveZone {
+export abstract class SimpleZone<TCard extends Card> extends ZoneAbstract<TCard, ISimpleZoneState<TCard>> implements IAddRemoveZone {
     public abstract override readonly name: ZoneName;
     public override readonly owner: Player;
 
-    protected _cards: TCard[] = [];
+    // protected _cards: TCard[] = [];
```
AMMayberry1, line 21: nit: can remove I assume?

server/game/core/GameObjectManager.ts
```diff
@@ -15,8 +15,14 @@ export class GameObjectManager {
         this._gameObjectMapping = new Map<string, GameObjectBase>();
     }
 
-    public get<T extends GameObjectBase>(gameObjectRef: GameObjectRef<T>): T | undefined {
-        return this._gameObjectMapping.get(gameObjectRef.uuid) as (T | undefined);
+    public get<T extends GameObjectBase>(gameObjectRef: GameObjectRef<T>): T | null {
+        if (!gameObjectRef?.uuid) {
+            return null;
```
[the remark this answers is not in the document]
CDMayberry: I was sure and didn't have it initially. The reason I added it was so that the logic naturally fits into the original usage w/o needing to make any major changes.

so here's an example I ran into:
```ts
private _zone: Zone | null;

get zone() {
	return this._zone;
}
```
I'd have to now add a "if null" check here if were to call the get method but it disallowed. Where as it is now, I can simply drop in the get call and it works w/o needing changes to the overall flow.

```ts
private _zoneRef: GameObjectRef<Zone> | null;

get zone() {
	return this.game.gameObjectManager.get(this._zoneRef);
}
```

*BUT*, honestly being explicit with nulls is _always_ safer, even if it's a little more work on the devs part. I am up for either, I just opted for "ease of drop-in" first.

server/game/core/card/Card.ts
```diff
@@ -35,9 +36,24 @@ import type { ICardWithTriggeredAbilities } from './propertyMixins/TriggeredAbil
...
+import type { GameObjectRef } from '../GameObjectBase';
 
 // required for mixins to be based on this class
-export type CardConstructor = new (...args: any[]) => Card;
+export type CardConstructor<T extends ICardState = ICardState> = new (...args: any[]) => Card<T>;
+
+export interface ICardState extends IOngoingEffectSourceState {
+
+    facedown: boolean;
+    hasImplementationFile: boolean;
```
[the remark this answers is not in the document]
CDMayberry: Ah right, yeah that was probably just me finding an example to use, can remove it.

server/game/core/GameObjectBase.ts
```diff
@@ -37,11 +48,16 @@ export abstract class GameObjectBase<T extends IGameObjectBaseState = IGameObjec
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     protected onSetupDefaultState() { }
```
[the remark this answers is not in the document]
CDMayberry: Nah, it's just a C# thing, both event listeners and optional, overridable "during/before/after" methods often prefix with "on". I can change to just setupDefaultState.

server/game/core/card/Card.ts
CDMayberry: Ah my bad, just duplicated the getter accessibility.

server/game/core/card/Card.ts
```diff
@@ -211,7 +240,7 @@ export class Card extends OngoingEffectSource {
         this._backSideTitle = cardData.backSideTitle;
         this._unique = cardData.unique;
 
-        this._controller = owner;
+        this.state.controllerRef = owner.getRef();
```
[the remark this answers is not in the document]
CDMayberry: Ah yeah, I did do this elsewhere, I think I initially changed the state, fixed this, then added the setter. Will change over.

server/game/core/GameObjectBase.ts
```diff
@@ -5,8 +5,18 @@ export interface IGameObjectBaseState {
...
+ * Never create an object with this interface manually, instead always use {@link GameObjectBase.getRef} to create an instance.
+ * @template T The template itself is unused, but it can provide some type safety, or at least awareness,
+ * of what type the GameObjectRef was created from. See the Card.controller set property for an example.
+ * @example this.state.controllerRef = player.getRef();
+ * // ... elsewhere
+ * const player = this.game.gameObjectManager.get(this.state.controllerRef);
+ */
 // eslint-disable-next-line @typescript-eslint/no-unused-vars, unused-imports/no-unused-vars
 export interface GameObjectRef<T extends GameObjectBase = GameObjectBase> {
+    __isRef: true;
```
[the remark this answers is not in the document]
CDMayberry: Ah, I just needed to add a field so that it doesn't "overlap" with the GameObject interface, and this was just an older JS way of doing that. 

Double underscore was often added by an external package to a object so that it can store data on the object w/o colliding with normal naming conventions. But that's not really needed these days.

server/game/core/card/propertyMixins/PrintedPower.ts
CDMayberry: Correct, another one of those "added to see if I could make it work, then realized printed is a fixed value", so there was no actual need for this. A part of me was just keeping it around for reference but don't need it anymore.

server/game/core/zone/SimpleZone.ts
CDMayberry: Yes, will get it removed.

server/game/core/card/propertyMixins/UnitProperties.ts
CDMayberry: I think this was originally a placeholder interface because I knew the captureZone would need to be converted to state eventually, but then I just did it. Most likely I just forgot to remove the comment 😅

server/game/core/zone/ConcreteArenaZone.ts
AMMayberry1: Yeah that sounds better to me, just for clarity and reducing the chance of accidental desync if someone else is working on it who doesn't understand the interaction

server/game/core/card/propertyMixins/PrintedPower.ts
AMMayberry1: Yeah sorry, forgot this is still draft or I wouldn't have commented on these

AMMayberry1 (approved):

PR #957, opened 2025-03-18
Fix exception loop

AMMayberry1:


server/game/core/utils/Contract.ts
```diff
@@ -20,7 +19,7 @@ class AssertContractCheckImpl implements IContractCheckImpl {
         }
 
         if (process.env.ENVIRONMENT === 'development') {
-            assert.fail(`Contract assertion failure: ${message}`);
+            throw new Error(`Contract assertion failure: ${message}`);
```
danbastin, line 22: Can we just simplify this to always throw regardless of environment or is this meant to be temporary?

PR #961, opened 2025-03-18
set4TokenInTest

Twet1:
Tie fighter and Xwing tokens available in test suite
Updated - PlayerInteractionWrapper.js and Util.js

Added a test case to check. Let me know if this needs removing or moving elsewhere


test/server/cards/04_JTL/tokens/TokenTest.spec.ts
```diff
@@ -0,0 +1,24 @@
+describe('X Wing and TIE Fighter Tokens', function() {
```
AMMayberry1, line 1: Actually typically we don't write explicit test cases for this, just having other cases that use it is enough. Would you mind searching through the other JTL card tests and finding some that are creating token units using cards when they should be using the test setup? If we can find a few of those and update them that will be good enough

Twet1:
Embedded the tests inside Shaaki Ti and Nameless Valor of set 3!

PR #972, opened 2025-03-19
Allow Pilots to be played via Ezra and Fine Addition

adbronico:


server/game/cards/03_TWI/events/AFineAddition.ts
```diff
@@ -34,18 +34,26 @@ export default class AFineAddition extends EventCard {
         this.setEventAbility({
             title: 'If an enemy unit was defeated this phase, play an upgrade from your hand or from any player\'s discard pile, ignoring its aspect penalty.',
             targetResolver: {
-                cardTypeFilter: CardType.BasicUpgrade,
                 zoneFilter: [ZoneName.Discard, ZoneName.Hand],
-                cardCondition: (card, context) => this.wasEnemyUnitDefeatedThisPhaseForPlayer(context.player.opponent) && this.checkZoneAndOwnershipOfCard(card, context.player),
+                cardCondition: (card, context) => this.wasEnemyUnitDefeatedThisPhaseForPlayer(context.player.opponent) && this.checkZoneAndOwnershipOfCard(card, context.player) &&
+                  (card.isUpgrade() || (card.isUnit() && card.hasSomeKeyword(KeywordName.Piloting))),
```
AMMayberry1, line 39: Since this same condition might come up on other cards (e.g. Fenn Rau), could we add something to `playCard()` itself that will handle this under the hood so we don't have to do this same complex checking logic in multiple places?

I'm not exactly sure what the best interface is but we could add e.g. an optional `playAsType` on `PlayCardSystem` which can be `WildcardCardType.Unit`, `WildcardCardType.Upgrade`, or `CardType.Event`.

The comments below outline how that would work together with potential changes to the get play action methods.

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -150,6 +154,11 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
             return [];
         }
 
+        if (playType === PlayType.PlayFromOutOfPlay && this.hasSomeKeyword(KeywordName.Piloting)) {
```
AMMayberry1, line 157: See comment above - this method is pretty specific to the play type, so I think that this logic should live in the caller and this method only constructs a piloting ability if it's explicitly asked to (the existing behavior)

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -105,8 +105,12 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
             playCardActions = this.buildPlayCardActions(PlayType.Smuggle, propertyOverrides);
         }
 
-        if (this.zoneName === ZoneName.Discard && this.hasOngoingEffect(EffectName.CanPlayFromDiscard)) {
-            playCardActions = this.buildPlayCardActions(PlayType.PlayFromOutOfPlay, propertyOverrides);
+        if (this.zoneName === ZoneName.Discard) {
+            if (this.hasOngoingEffect(EffectName.CanPlayFromDiscard)) {
+                playCardActions = this.buildPlayCardActions(PlayType.PlayFromOutOfPlay, propertyOverrides);
+            } else if (propertyOverrides != null && propertyOverrides.canPlayFromAnyZone && this.hasSomeKeyword(KeywordName.Piloting)) {
```
AMMayberry1, line 111: So we shouldn't be checking the property overrides here I think. There are two slightly different pathways for playing from discard:

1. A card has e.g. Cobb Vanth effect that lets it be played as an action for this phase. In that situation, the ActionWindow calls `getPlayCardActions()` on every card in play or in discard and sees if it has any results. In that case we shouldn't be checking the overrides here, we should just check for piloting and return it with the regular discard play action. That way if the card has `CanPlayFromDiscard` applied and it has piloting, we will always get both actions in the action window.

2. A card effect specifically says "play a card from discard" e.g. A Fine Addition or Palp's Return. In this case, the PlayCardSystem will explicitly call `getPlayCardFromOutOfPlayActions()` and in there we should handle the piloting check.

It would probably end up looking like:

```ts
public getPlayCardFromOutOfPlayActions(propertyOverrides: IPlayCardActionOverrides = null) {
    Contract.assertFalse(
        [ZoneName.Hand, ZoneName.SpaceArena, ZoneName.GroundArena].includes(this.zoneName),
        `Attempting to get "play from out of play" actions for card ${this.internalName} in invalid zone: ${this.zoneName}`
    );

   const playCardActions = [this.buildPlayCardActions(PlayType.PlayFromOutOfPlay, propertyOverrides)];

   if (this.hasSomeKeyword(KeywordName.Piloting)) {
      playCardActions.push(this.buildPlayCardActions(PlayType.Piloting, propertyOverrides));
   }

    return playCardActions ;
}
```

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -110,10 +110,19 @@ export class PlayCardSystem<TContext extends AbilityContext = AbilityContext> ex
         // filter out actions that don't match the expected playType or aren't legal in the current play context (e.g. can't be paid for)
         return availableCardPlayActions.filter((action) => {
             const newContext = action.createContext(context.player);
-            return action.playType === properties.playType && action.meetsRequirements(newContext, properties.ignoredRequirements) === '';
+            return this.checkActionPlayType(properties.playType, action.playType) && action.meetsRequirements(newContext, properties.ignoredRequirements) === '';
```
AMMayberry1, line 113: The changes mentioned above would make it so we don't have to do this awkward check here. The pattern is that if a card is being played from out of play, we call `card.getPlayCardFromOutOfPlayActions(overrideProperties)` which would then (with the changes above) return both the piloting and regular play options.

If we add some way to specify "only play upgrades" or "only play units" on the PlayCardSystem properties, we could then filter here to remove one of those options before returning.

test/server/cards/03_TWI/events/AFineAddition.spec.ts
```diff
@@ -112,5 +112,100 @@ describe('A Fine Addition', function () {
...
+            context.player1.clickCard(context.daggerSquadronPilot);
+            expect(context.player1).toBeAbleToSelectExactly([context.cartelTurncoat]);
+            context.player1.clickCard(context.cartelTurncoat);
+            expect(context.daggerSquadronPilot).toBeAttachedTo(context.cartelTurncoat);
+
+            expect(context.player1.exhaustedResourceCount).toBe(6);
+            expect(context.player2).toBeActivePlayer();
+        });
+
+        it('A Fine Addition\'s ability should play a pilot as an upgrade from your opponent\'s discard, ignoring aspect penalty, if an enemy was defeated this phase', async function () {
```
AMMayberry1, line 179: I think this test is a copy of the above?

adbronico: No - one is from the opponent's discard

server/game/cards/03_TWI/events/AFineAddition.ts
adbronico: So, I added the cardCondition here to help filter out selectable targets. How would that happen with the reworked system?

AMMayberry1:
I realized we need to hit a few more cards:

- YMOH
- Cobb Vanth
- Dryden Vos maybe? Need to ask judge chat on this one I think

server/game/gameSystems/PlayCardSystem.ts
adbronico: We actually need the first check too to ensure Piloting doesn't allow units to be played as units

server/game/cards/03_TWI/events/AFineAddition.ts
AMMayberry1: If `PlayCardSystem.canAffect()` returns true / false, it filters out the card the same as if we had a filter in target resolver. So if we give it the logic to know how to do the above check then we don't need to write it in the target resolver for each relevant card implementation

server/game/core/ability/PlayCardAction.ts
```diff
@@ -18,6 +18,7 @@ import type Player from '../Player';
 
 export interface IPlayCardActionPropertiesBase {
     playType: PlayType;
+    playAsType?: WildcardCardType.Upgrade | WildcardCardType.Unit | CardType.Event;
```
AMMayberry1, line 21: Why is this here? Looks like it's not used

server/game/gameSystems/PlayMultipleCardsFromDeckSystem.ts
```diff
@@ -26,7 +28,8 @@ export class PlayMultipleCardsFromDeckSystem<TContext extends AbilityContext = A
                     nested: true,
                     adjustCost: {
                         costAdjustType: CostAdjustType.Free
-                    }
+                    },
+                    playAsType: propertiesOrPropertyFactory instanceof Function ? propertiesOrPropertyFactory().playAsType : propertiesOrPropertyFactory.playAsType,
```
AMMayberry1, line 32: I don't think this should be needed, it should be happening automatically. Was this a fix for an error or something?

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -822,7 +823,9 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
             Contract.assertTrue(this.canBeUpgrade);
             if (targetCard.isUnit()) {
                 if (context.playType === PlayType.Piloting && this.hasSomeKeyword(KeywordName.Piloting)) {
-                    return targetCard.canAttachPilot(this, context.playType) && targetCard.controller === controller;
+                    // This is needed for abilities that let you play Pilots from the opponent's discard
+                    const canPlayFromAnyZone = (context.ability as PlayUpgradeAction).canPlayFromAnyZone;
```
AMMayberry1, line 827: This isn't a safe assumption to make - there are abilities like Survivors' Gauntlet that can attach upgrades but aren't play card actions. If the below check works when `canPlayFromAnyZone` is null then we should be good, but probably should remove / change the cast to be clearer

server/game/gameSystems/PlayCardSystem.ts
AMMayberry1: Ahh okay, makes sense now

server/game/gameSystems/PlayMultipleCardsFromDeckSystem.ts
adbronico: Pilots could be played as upgrades. There was no logic checking that they could only be played as units

server/game/core/card/propertyMixins/UnitProperties.ts
adbronico: Isn't the play type check keeping that from being a concern?

AMMayberry1: Oh yeah fair. I think this was written before the update that changed how Survivors' Gauntlet interacts with pilots, I was thinking we had adjusted that already. Per the new rules we'll need to change this flow so that the canAttachPilot() check happens even if it's not a play action but I'll just make a GH issue for that

server/game/gameSystems/PlayMultipleCardsFromDeckSystem.ts
AMMayberry1: Ah yeah I misread what this was doing, it makes a little more sense now. Unfortunately we can't invoke `propertiesOrPropertyFactory()` safely here because it expects a context object. Looks like there's no current card that uses that or else this would've broken tests, but there's nothing stopping a future card from using it.

Probably the best thing to do is move this whole thing into an override of `generatePropertiesFromContext()` which just sets the value of `properties.selectedCardsImmediateEffect` like above, that's probably where it should've been in the first place

server/game/core/card/propertyMixins/UnitProperties.ts
adbronico: Do we need to make any changes here then? Or just write up that GH issue?

AMMayberry1: We're good for now, I'll write something up

server/game/gameSystems/PlayMultipleCardsFromDeckSystem.ts
```diff
@@ -19,22 +19,23 @@ export interface IPlayMultipleCardsFromDeckProperties<TContext extends AbilityCo
 }
 
 export class PlayMultipleCardsFromDeckSystem<TContext extends AbilityContext = AbilityContext> extends SearchDeckSystem<TContext, IPlayMultipleCardsFromDeckProperties<TContext>> {
-    // constructor needs to do some extra work to ensure that the passed props object ends up as valid for the parent class
-    public constructor(propertiesOrPropertyFactory: IPlayMultipleCardsFromDeckProperties<TContext> | ((context?: AbilityContext) => IPlayMultipleCardsFromDeckProperties<TContext>)) {
-        const propsWithViewType = GameSystem.appendToPropertiesOrPropertyFactory<ISearchDeckProperties<TContext>, 'selectedCardsImmediateEffect'>(propertiesOrPropertyFactory,
+    public override generatePropertiesFromContext(context: TContext, additionalProperties = {}): IPlayMultipleCardsFromDeckProperties<TContext> {
+        const properties = super.generatePropertiesFromContext(context, additionalProperties) as IPlayMultipleCardsFromDeckProperties<TContext>;
+
+        const propsWithViewType = GameSystem.appendToPropertiesOrPropertyFactory<ISearchDeckProperties<TContext>, 'selectedCardsImmediateEffect'>(properties,
```
AMMayberry1, line 25: `appendToPropertiesOrPropertyFactory` isn't needed here and might have some weird behaviors. Since at this point we have a concrete `properties` object (it can no longer be a function), we can just do:

```ts
const propsWithViewType = { ...properties, selectedCardsImmediateEffect: <the PlayCardSystem stuff> };
```

AMMayberry1 (approved):

PR #973, opened 2025-03-19
Avenger - Wording change

Twet1:
Noticed wording issue during game
image

### DIFF
--- a/server/game/cards/01_SOR/units/AvengerHuntingStarDestroyer.ts
+++ b/server/game/cards/01_SOR/units/AvengerHuntingStarDestroyer.ts
@@ -18,6 +18,7 @@ export default class AvengerHuntingStarDestroyer extends NonLeaderUnitCard {
                 onCardPlayed: (event, context) => event.card === context.source
             },
             targetResolver: {
+                activePromptTitle: 'Choose a non-leader unit to defeat',
                 choosingPlayer: RelativePlayer.Opponent,
                 controller: RelativePlayer.Opponent,
                 cardTypeFilter: WildcardCardType.NonLeaderUnit,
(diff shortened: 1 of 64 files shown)


server/game/cards/01_SOR/units/AvengerHuntingStarDestroyer.ts
```diff
@@ -12,7 +12,7 @@ export default class AvengerHuntingStarDestroyer extends NonLeaderUnitCard {
 
     public override setupCardAbilities() {
         this.addTriggeredAbility({
-            title: 'Choose a friendly non-leader unit to defeat',
+            title: 'opponent chooses a non-leader unit they control to defeat',
```
adbronico, line 15: Capitalize the o

adbronico (changes requested):

adbronico (dismissed):
looks good to me, wait on Veld to merge

server/game/cards/01_SOR/units/AvengerHuntingStarDestroyer.ts
```diff
@@ -12,7 +12,7 @@ export default class AvengerHuntingStarDestroyer extends NonLeaderUnitCard {
 
     public override setupCardAbilities() {
         this.addTriggeredAbility({
-            title: 'Choose a friendly non-leader unit to defeat',
+            title: 'Opponent chooses a non-leader unit they control to defeat',
```
AMMayberry1, line 15: If we're making this change here we'll also want to set `activePromptTitle` on the targetResolver so that the opponent doesn't get a confusing message

Twet1: @AMMayberry1 The prompt that made me first notice it, is because of Piett giving Ambush. Once clicked it then prompts NAP to "Choose a friendly non leader unit to defeat." Which is correct for NAP. 

Now what happens is AP gets a prompt - 
image

and NAP player gets prompt - 
image

So I think it comes down to more how we want to handle the context used at which point.

AMMayberry1: @Twet1 yeah I believe that if you change the `targetResolver.activePromptTitle` is what the opponent will see when they are choosing. So if you change it to be "Choose a unit to defeat" or "Choose a unit to defeat for Power of the Dark Side" then that will give a more intuitive message to the player who is making the choice

AMMayberry1 (approved):

PR #978, opened 2025-03-20
Logger switch for CloudWatch newlines

brianries:
Added a switch to the logger to do literal newlines for CloudWatch when running under the AWS environment.  This will collapse a stack trace into a singular entries with explicit new line characters (hopefully rendering correctly when selected via LogInsight)


brianries:
Put this into draft, debating switching to JSON structured logging for CloudWatch and regular printf in console

AMMayberry1 (approved):

PR #979, opened 2025-03-21
Add Shadow Caster

adbronico:


test/server/cards/04_JTL/units/ShadowCasterJustBusiness.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+                context.player1.clickPrompt('Trigger');
+
+                expect(context.player1.findCardsByName('clone-trooper').length).toBe(2);
+
+                expect(context.player1).toHavePassAbilityPrompt('Use the When Defeated ability again');
+                context.player1.clickPrompt('Trigger');
+                expect(context.player1.findCardsByName('clone-trooper').length).toBe(3);
+
+                expect(context.player1).toBeActivePlayer();
+            });
```
AMMayberry1, line 172: We definitely need some tests with Thrawn since that interaction could get kind of thorny. Specifically we need to make sure Shadow Caster doesn't trigger off of the Thrawn ability activating

test/server/cards/04_JTL/units/ShadowCasterJustBusiness.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+
+                const { context } = contextRef;
+
+                context.player1.passAction();
+                context.player2.clickCard(context.daringRaid);
+                context.player2.clickCard(context.rhokaiGunship);
+
+                context.player1.clickCard(context.p2Base);
+
+                expect(context.player1).toHavePassAbilityPrompt('Use the When Defeated ability again');
```
AMMayberry1, line 23: Do you think it would be useful to add the ability name to this message? I could go either way

adbronico: I wanted to but I wasn't sure if it was possible given this is the title of the tiggered ability

test/server/cards/04_JTL/units/ShadowCasterJustBusiness.spec.ts
adbronico: There is a test for Thrawn at the bottom of the file, I just forgot to rename it. I'll update that

test/server/cards/04_JTL/units/ShadowCasterJustBusiness.spec.ts
AMMayberry1: Oh yeah fair, we'd need to add some piping for that. Can do it sometime later

test/server/cards/04_JTL/units/ShadowCasterJustBusiness.spec.ts
AMMayberry1: Ohh missed that, nice. I think we should add one more where we trigger in the other order (Shadow Caster first then Thrawn) just to make sure that flow also works

AMMayberry1 (approved):

PR #987, opened 2025-03-22
More JSON Logger updates

brianries:
Updated the log format to support json objects, and updated the Lobby to use the new format.

here is an example of the logFormatter 'printf' style
image

and here is the pure json that cloudwatch would receive
image

### DIFF
--- a/server/gamenode/Lobby.ts
+++ b/server/gamenode/Lobby.ts
@@ -163,7 +163,7 @@ export class Lobby {
             deckValidationErrors: deck ? this.deckValidator.validateInternalDeck(deck.getDecklist(), this.gameFormat) : {},
             deck
         }));
-        logger.info(`Lobby ${this.id}: Creating username: ${user.username}, id: ${user.id} and adding to users list (${this.users.length} user(s))`);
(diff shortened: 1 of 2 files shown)


server/gamenode/Lobby.ts
```diff
@@ -163,7 +163,7 @@ export class Lobby {
             deckValidationErrors: deck ? this.deckValidator.validateInternalDeck(deck.getDecklist(), this.gameFormat) : {},
             deck
         }));
-        logger.info(`Lobby ${this.id}: Creating username: ${user.username}, id: ${user.id} and adding to users list (${this.users.length} user(s))`);
+        logger.info(`Creating username: ${user.username}, id: ${user.id} and adding to users list (${this.users.length} user(s))`, { lobbyId: this.id });
```
AMMayberry1, line 166: Username / user ID is something else it would be good to log automatically. Eventually we can try to make a logger wrapper / helper that automates some of that.

AMMayberry1 (approved):

server/gamenode/Lobby.ts
brianries: Added userName and userId to all the json payloads

brianries:
Updated json:

image

AMMayberry1 (approved):

PR #1006, opened 2025-03-27
Fix a few chat log issues

Frugghi:


test/server/cards/03_TWI/events/DropIn.spec.ts
```diff
@@ -17,6 +17,9 @@ describe('Drop In', function () {
             expect(cloneTroopers).toAllBeInZone('groundArena');
             expect(cloneTroopers.every((cloneTrooper) => cloneTrooper.exhausted)).toBeTrue();
             expect(context.player2.getArenaCards().length).toBe(0);
+            expect(context.getChatLogs(1)).toEqual([
+                'player1 plays Drop In to create 2 cloneTroopers',
```
adbronico, line 21: I think we should try to get this to say "Clone Troopers" instead if possible

Frugghi: That would be ideal but I'm not sure how to get access to the token's title from there. Any ideas?

AMMayberry1: We can just have a switch statement that maps token type to a proper name. Could just do it directly in the create token system or maybe add a helper method somewhere that makes sense

test/server/cards/01_SOR/units/EzraBridger.spec.ts
```diff
@@ -34,7 +34,8 @@ describe('Ezra Bridger', function() {
 
                 expect(context.player1).toHaveExactSelectableDisplayPromptCards([context.momentOfPeace]);
                 expect(context.player1).toHaveExactDisplayPromptPerCardButtons(['Play it', 'Discard it', 'Leave it on top of your deck']);
-                expect(context.getChatLogs(1)).not.toContain('Ezra Bridger sees Moment of Peace');
+                expect(context.getChatLogs(1)).toContain('player1 looks at a card');
+                expect(context.getChatLogs(1)).not.toContain('player1 sees Moment of Peace');
```
AMMayberry1, line 38: nit: we can just do `.not.toContain('Moment of Peace');` since that will catch more potential problematic cases

test/server/cards/03_TWI/events/DropIn.spec.ts
Frugghi: @AMMayberry1 Do you prefer that over the current solution to use `cardDataGetter`? It seems to me that using `cardDataGetter` would be more flexible.

AMMayberry1: @Frugghi I don't quite follow sorry, could you elaborate? Not sure what solution with `cardDataGetter` you're referring to

AMMayberry1: Ahh sorry, missed that. Yeah this is good

AMMayberry1 (approved):

PR #1014, opened 2025-03-28
Added testcases requested by todos

priisig:


test/server/cards/01_SOR/units/MaceWinduPartyCrasher.spec.ts
```diff
@@ -73,31 +73,29 @@ describe('Mace Windu, Party Crasher', function() {
...
+                    player1: {
+                        groundArena: [{ card: 'mace-windu#party-crasher', upgrades: ['fallen-lightsaber'] }]
+                    },
+                    player2: {
+                        groundArena: ['jawa-scavenger']
+                    }
+                });
+            });
+
+            it('will not ready him if the unit is defeated by an on-attack ability', function () {
```
AMMayberry1, line 89: This test should remain commented out since it's actually wrong - the Mace ability is supposed to trigger in this situation but currently doesn't. Maybe we can just remove the commented-out test completely and leave a clearer TODO, also I'll create a GH issue

test/server/cards/02_SHD/events/CalculatedLethality.spec.ts
```diff
@@ -72,7 +72,28 @@ describe('Calculated Lethality', function () {
...
+
+        describe('Calculated Lethality\'s ability', function () {
+            it('should not be able to defeat lurking tie phantom', async function () {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        hand: ['calculated-lethality'],
+                    },
+                    player2: {
+                        spaceArena: ['lurking-tie-phantom'],
```
AMMayberry1, line 85: The specific test we wanted to do here is to confirm that the engine handles this ruling correctly. So we need to put some upgrades on LTP

image

test/server/cards/02_SHD/leaders/BobaFettDaimyo.spec.ts
```diff
@@ -130,6 +130,28 @@ describe('Boba Fett, Daimyo', function () {
         });
 
         // TODO should add tests with timely intervention or ecl
```
AMMayberry1, line 132: nit: can remove the comment now

Issue #1005, opened 2025-03-26
Energy Conversion Lab can give Ambush to a unit that cannot (yet) be played


moyerr:
### Description

As seen in [Veiled Shot's recent Han2 video](https://youtu.be/0ChbWmcDaIs?si=as4ZIzzdGRvl0hn0&t=1526):

ECL is able to give a unit ambush, even if the unit is not able to be played (yet). If the card is played later in the same action phase, it will have ambush.

### Steps to Reproduce

1. Start with 4 resources
2. Click on Energy Conversion Lab
3. Click on Poe Dameron unit in hand (he cannot be played because we don't have enough resources)
4. (Opponent takes an action)
5. Click on Han Solo, Worth the Risk (leader)
6. Play Poe Dameron for a 1 resource discount

**Result**: Poe Dameron enters play with Ambush

### Game Setup

```json
{
    "phase": "action",
    "player1": {
        "hand": [
            "poe-dameron#quick-to-improvise"
        ],
        "base": "energy-conversion-lab",
        "leader": "han-solo#worth-the-risk",
        "resources": 4
    },
    "player2": {
        "groundArena": [
            "rugged-survivors"
        ]
    }
}
```

AMMayberry1:
`simultaneous()` and `sequential()` by default will fire if at least one of the effects inside has a legal target. We need to add a property on both that allows configuring them to require that all effects have a target to prevent situations like this.

PR #1016, opened 2025-03-29
All targets legal

ssoulier:
Fixing this issue: https://github.com/SWU-Karabast/forceteki/issues/1005

### DIFF
--- a/server/game/cards/01_SOR/bases/EnergyConversionLab.ts
+++ b/server/game/cards/01_SOR/bases/EnergyConversionLab.ts
@@ -18,10 +18,14 @@ export default class EnergyConversionLab extends BaseCard {
                 cardTypeFilter: CardType.BasicUnit,
                 controller: RelativePlayer.Self,
                 zoneFilter: ZoneName.Hand,
-                immediateEffect: AbilityHelper.immediateEffects.simultaneous([
-                    AbilityHelper.immediateEffects.playCardFromHand(),
-                    AbilityHelper.immediateEffects.forThisPhaseCardEffect({ effect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Ambush) }),
-                ])
+                immediateEffect: AbilityHelper.immediateEffects.simultaneous(
+                    [
+                        AbilityHelper.immediateEffects.playCardFromHand(),
+                        AbilityHelper.immediateEffects.forThisPhaseCardEffect({ effect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Ambush) }),
+                    ],
+                    false, // We do not ignore targeting requirements
+                    true // All effect must have a legal target
+                )
             }
         });
     }
--- a/server/game/gameSystems/SequentialSystem.ts
+++ b/server/game/gameSystems/SequentialSystem.ts
@@ -12,7 +12,7 @@ export interface ISequentialSystemProperties<TContext extends AbilityContext = A
     gameSystems: GameSystem<TContext>[];
 
     // If true, all game systems must be legal for the entire sequence to be legal
-    allGameSystemsMustBeLegal?: boolean;
+    everyGameSystemMustBeLegal?: boolean;
 }
 
 // TODO: add a variant of this (or a configuration option) for repeating the same action a variable number of times
@@ -25,14 +25,14 @@ export interface ISequentialSystemProperties<TContext extends AbilityContext = A
  */
 export class SequentialSystem<TContext extends AbilityContext = AbilityContext> extends AggregateSystem<TContext, ISequentialSystemProperties<TContext>> {
     protected override readonly eventName: MetaEventName.Sequential;
-    protected readonly allGameSystemsMustBeLegal: boolean;
-    public constructor(gameSystems: ISystemArrayOrFactory<TContext>, allGameSystemsMustBeLegal: boolean = false) {
+    protected readonly everyGameSystemMustBeLegal: boolean;
+    public constructor(gameSystems: ISystemArrayOrFactory<TContext>, everyGameSystemMustBeLegal: boolean = false) {
         if (typeof gameSystems === 'function') {
             super((context: TContext) => ({ gameSystems: gameSystems(context) }));
         } else {
             super({ gameSystems });
         }
-        this.allGameSystemsMustBeLegal = allGameSystemsMustBeLegal;
+        this.everyGameSystemMustBeLegal = everyGameSystemMustBeLegal;
     }
 
     // eslint-disable-next-line @typescript-eslint/no-empty-function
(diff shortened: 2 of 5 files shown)


server/game/gameSystems/SequentialSystem.ts
```diff
@@ -10,6 +10,9 @@ import type { Player } from '../core/Player';
 
 export interface ISequentialSystemProperties<TContext extends AbilityContext = AbilityContext> extends IGameSystemProperties {
     gameSystems: GameSystem<TContext>[];
+
+    // If true, all game systems must be legal for the entire sequence to be legal
+    everyGameSystemMustBeLegal?: boolean;
```
ssoulier, line 15: @AMMayberry1 Do you have in mind a card where we can test this new sequential parameter?

server/game/cards/01_SOR/bases/EnergyConversionLab.ts
```diff
@@ -18,10 +18,14 @@ export default class EnergyConversionLab extends BaseCard {
...
-                immediateEffect: AbilityHelper.immediateEffects.simultaneous([
-                    AbilityHelper.immediateEffects.playCardFromHand(),
-                    AbilityHelper.immediateEffects.forThisPhaseCardEffect({ effect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Ambush) }),
-                ])
+                immediateEffect: AbilityHelper.immediateEffects.simultaneous(
+                    [
+                        AbilityHelper.immediateEffects.playCardFromHand(),
+                        AbilityHelper.immediateEffects.forThisPhaseCardEffect({ effect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Ambush) }),
+                    ],
+                    false, // We do not ignore targeting requirements
```
ssoulier, line 26: I would have preferred to use an objects to name each property explicitly but it looked like a significant refactoring that is probably not desired.

server/game/gameSystems/SequentialSystem.ts
```diff
@@ -22,19 +25,30 @@ export interface ISequentialSystemProperties<TContext extends AbilityContext = A
...
     public override eventHandler() {}
 
+    private allGameSystemsAreLegal(context: TContext, additionalProperties = {}): boolean {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        return properties.gameSystems.every((gameSystem) => gameSystem.hasLegalTarget(context, additionalProperties));
+    }
+
     public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext, additionalProperties = {}): void {
         const properties = this.generatePropertiesFromContext(context, additionalProperties);
+        if (this.everyGameSystemMustBeLegal && !this.allGameSystemsAreLegal(context, additionalProperties)) {
```
ssoulier, line 48: This implementation is maybe naive. Seems working but as I do not master well all the subtleties of the engine I may have missed something

server/game/gameSystems/SequentialSystem.ts
AMMayberry1: Hmm nothing currently

server/game/cards/01_SOR/bases/EnergyConversionLab.ts
AMMayberry1: Actually if you're interested that would be a nice small project, realistically this should be using a property object like everything else but we haven't super needed it before now. It would mostly involve a small refactor here and then making some adjustments to the `simultaneous()` and `sequential()` helper methods in GameSystemLibrary so they can do some formatting to allow either a list of systems or a property object.

Lmk if that's something you'd be interested in, I think it would actually be fairly straightforward and make these a little cleaner. Either way though, let's get this PR closed out with the smallest amount of work we can since it's fixing a bug, and if you're up for that refactor it can be a followup.

server/game/gameSystems/SequentialSystem.ts
AMMayberry1: Looks right to me 😁

server/game/gameSystems/SimultaneousSystem.ts
```diff
@@ -14,17 +14,21 @@ export interface ISimultaneousSystemProperties<TContext extends AbilityContext =
...
+    // If true, all game systems must be legal for the entire "simultaneous" to be legal
+    everyGameSystemMustBeLegal?: boolean;
 }
 
 export class SimultaneousGameSystem<TContext extends AbilityContext = AbilityContext> extends AggregateSystem<TContext, ISimultaneousSystemProperties<TContext>> {
     protected override readonly eventName: MetaEventName.Simultaneous;
-
-    public constructor(gameSystems: ISystemArrayOrFactory<TContext>, ignoreTargetingRequirements = false) {
+    protected readonly everyGameSystemMustBeLegal: boolean;
+    public constructor(gameSystems: ISystemArrayOrFactory<TContext>, ignoreTargetingRequirements = false, everyGameSystemMustBeLegal = false) {
```
AMMayberry1, line 25: These two booleans are mutually exclusive - we shouldn't be able to set both to true. Can we set up an exported enum, maybe in AggregateSystem since it's the parent class? It could be something like `TargetingRequirements` or `EffectRequirements`. Should use the same one for SequentialSystem

server/game/cards/01_SOR/bases/EnergyConversionLab.ts
ssoulier: Very happy to try. I won't be fast (lack of time at the moment + lack of typescript skills) but I look into it and may ask a few questions along the way.

Feel free to merge as-is to solve the bug and I will work on the refactoring on a new PR.

server/game/gameSystems/SimultaneousSystem.ts
ssoulier: Should I do it in the refactoring PR?

AMMayberry1: Hey sorry missed this - I could go either way, but at the barest minimum we need a contract check here which prevents both booleans from being true at the same time.

server/game/cards/01_SOR/bases/EnergyConversionLab.ts
AMMayberry1: Sounds great!

server/game/gameSystems/SimultaneousSystem.ts
ssoulier: I updated the PR (and sorry for the delay, I was at professional event for 3 days).
Was it what you had in mind to merge it?
Then I'll start working on the small refactoring.

AMMayberry1: Looks good to me! And no worries, hope the event was good 😄

PR #1025, opened 2025-03-31
Update Ambush to match CR4

Frugghi:
Update Ambush to not ready the unit before the attack.


server/game/actions/InitiateAttackAction.ts
```diff
@@ -23,7 +23,7 @@ interface IInitiateAttackProperties extends IAttackProperties {
  */
 export class InitiateAttackAction extends PlayerAction {
     public constructor(game: Game, card: Card, private attackProperties?: IInitiateAttackProperties) {
-        const exhaustCost = attackProperties?.allowExhaustedAttacker
+        const exhaustCost = (attackProperties?.allowExhaustedAttacker || attackProperties?.isAmbush)
```
AMMayberry1, line 26: nit: could we simplify this a bit by just setting `allowExhaustedAttacker: true` in the ambush setup?

AMMayberry1 (approved):

PR #1035, opened 2025-04-03
Game and other state, and unit test improvements

CDMayberry:
### Highlights
* A number of Game.js fields moved into a state object
* Player Decklist converted to use GameObjectRef and moved into state
* Card.parentCard converted to GameObjectRef and moved into state
* Fixed, then disabled the "filter card from list" in Game.js, it wasn't doing anything before so it's disabled for now.
* Converted undo test to a custom global uit function
* Added a JS private #experimental object in the Game for flagging the system to use undo logic. That way I can enable it in tests now and in the future for local builds only.

### Code I'm Not So Proud Of
Soooo I made some methods public in ActionPhase to make clearing the ActionPhase possible. It's a little horrifying but it did work for tests. And the way I clear the pipeline also was quick and dirty.

### Testing Notes
the new uit function only works with tests that do the setup in a beforeEach. In part because the rollback only is working for the ActionPhase, but also because if the setup is within an _it_ call, it'll bump into the rollback which happens before the test runs and can cause complications. So, the uit method will simply treat the test as a normal test if the setup call hasn't happened yet.

To manually run all tests in undo, simply put `global.it = global.uit;` after the uit definition. uit will have false positives, but the failures we do run into will help us hunt down more issues with undo.

### DIFF
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -125,17 +125,18 @@ class Game extends EventEmitter {
         this.playStarted = false;
         this.gameObjectManager = new GameStateManager(this);
         this.createdAt = new Date();
-        this.currentActionWindow = null;
         // Debug flags, intended only for manual testing, and should always be false. Use the debug methods to temporarily flag these on.
         this.#debug = { pipeline: false };
 
-        /** @type { EventWindow } */
-        this.currentEventWindow = null;
-
-        this.currentAttack = null;
         this.manualMode = false;
         this.gameMode = details.gameMode;
+
+        /** @type { EventWindow } */
+        this.currentEventWindow = null;
+        this.currentAttack = null;
         this.currentPhase = null;
+        this.currentActionWindow = null;
+        this.currentOpenPrompt = null;
 
         /** @type { import('./GameStateManager.js').IGameState } */
         this.state = {
@@ -151,7 +152,6 @@ class Game extends EventEmitter {
         this.movedCards = [];
         // STATE TODO: Move the generator logic into the state object.
         this.randomGenerator = seedrandom();
-        this.currentOpenPrompt = null;
         this.cardDataGetter = details.cardDataGetter;
         this.playableCardTitles = this.cardDataGetter.playableCardTitles;
 
@@ -1526,9 +1526,19 @@ class Game extends EventEmitter {
     //     return snapshot.id;
     // }
 
-    // rollbackToSnapshot(snapshotId) {
-    //     this.gameObjectManager.rollbackToSnapshot(snapshotId);
-    // }
+    /**
+     * @param {number} snapshotId
+     */
+    rollbackToSnapshot(snapshotId) {
+        if (this.pipeline.currentStep instanceof ActionPhase) {
+            const actionPhase = this.pipeline.currentStep;
+            this.gameObjectManager.rollbackToSnapshot(snapshotId);
+            actionPhase.resetPhase();
+            actionPhase.queueNextAction();
+        } else {
+            throw new Error('Cannout Undo outside of Action Phase');
+        }
+    }
 
     // TODO: Make a debug object type.
     /**
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -41,6 +41,7 @@ const { WildcardCardType } = require('./Constants');
 const { validateGameConfiguration, validateGameOptions } = require('./GameInterfaces.js');
 const { GameStateManager } = require('./GameStateManager.js');
 const { ActionWindow } = require('./gameSteps/ActionWindow.js');
+const { GameObjectBase } = require('./GameObjectBase.js');
 
 class Game extends EventEmitter {
     #debug;
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -100,7 +100,7 @@ class Game extends EventEmitter {
     }
 
     set roundNumber(value) {
-        Contract.assertNonNegative(value, 'Round Number must be non-zero.');
+        Contract.assertNonNegative(value, 'Round Number must be non-zero: ' + value);
         this.state.roundNumber = value;
     }
 
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -1395,6 +1395,7 @@ class Game extends EventEmitter {
         );
 
         const player = token.owner;
+        // Functionality did nothing previously, now that it's fixed, disabling until we're ready to activate again.
         // this.filterCardFromList(token, this.state.allCards);
         // this.filterCardFromList(token, player.decklist.tokens);
         // this.filterCardFromList(token, player.decklist.allCards);
@@ -1416,10 +1417,18 @@ class Game extends EventEmitter {
      * @param {import('./GameObjectBase.js').GameObjectRef[]} list
      */
     filterCardFromList(removeCard, list) {
-        const index = list.findIndex((x) => x.uuid === removeCard.uuid);
-        Contract.assertNonNegative(index, `Tried to remove card from list but ${removeCard?.title} was not found.`);
+        const indexes = [];
 
-        list.splice(index, 1);
+        for (let i = list.length - 1; i >= 0; i--) {
+            const ref = list[i];
+            if (ref.uuid === removeCard.uuid) {
+                indexes.push(i);
+            }
+        }
+
+        for (let index of indexes) {
+            list.splice(index, 1);
+        }
     }
 
     // formatDeckForSaving(deck) {
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -1535,6 +1535,9 @@ class Game extends EventEmitter {
             this.gameObjectManager.rollbackToSnapshot(snapshotId);
             actionPhase.resetPhase();
             actionPhase.queueNextAction();
+            // continue to the takeSnapshot step and then the ActionWindow step.
+            actionPhase.continue();
+            actionPhase.continue();
         } else {
             throw new Error('Cannout Undo outside of Action Phase');
         }
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -1619,7 +1619,7 @@ class Game extends EventEmitter {
      * @param {() => void} fcn
      */
     debugPipeline(fcn) {
-        this.#debug.pipeline = Helpers.isDevelopment;
+        this.#debug.pipeline = Helpers.isDevelopment();
         try {
             fcn();
         } finally {
@@ -1632,7 +1632,7 @@ class Game extends EventEmitter {
      * @param {() => any} fcn
      */
     enableUndo(fcn) {
-        this.#experimental.undo = Helpers.isDevelopment;
+        this.#experimental.undo = Helpers.isDevelopment();
         try {
             return fcn();
         } finally {
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -45,6 +45,7 @@ const { GameObjectBase } = require('./GameObjectBase.js');
 
 class Game extends EventEmitter {
     #debug;
+    #experimental;
 
     get isInitiativeClaimed() {
         return this.state.isInitiativeClaimed;
@@ -135,6 +136,8 @@ class Game extends EventEmitter {
 
         // Debug flags, intended only for manual testing, and should always be false. Use the debug methods to temporarily flag these on.
         this.#debug = { pipeline: false };
+        // Experimental flags, intended only for manual testing. Use the enable methods to temporarily flag these on during tests.
+        this.#experimental = { undo: false };
 
         this.manualMode = false;
         this.gameMode = details.gameMode;
@@ -1550,7 +1553,7 @@ class Game extends EventEmitter {
     }
 
     takeSnapshot() {
-        if (this.pipeline.currentStep instanceof ActionPhase) {
+        if (this.#experimental.undo && this.pipeline.currentStep instanceof ActionPhase) {
             return this.gameObjectManager.takeSnapshot();
         }
 
@@ -1561,13 +1564,12 @@ class Game extends EventEmitter {
      * @param {number} snapshotId
      */
     rollbackToSnapshot(snapshotId) {
-        if (this.pipeline.currentStep instanceof ActionPhase) {
+        if (this.#experimental.undo && this.pipeline.currentStep instanceof ActionPhase) {
             const actionPhase = this.pipeline.currentStep;
             this.gameObjectManager.rollbackToSnapshot(snapshotId);
             actionPhase.resetPhase();
             actionPhase.queueNextAction();
-            // continue to the takeSnapshot step and then the ActionWindow step.
-            actionPhase.continue();
+            // continue the action phase again.
             actionPhase.continue();
             return true;
         }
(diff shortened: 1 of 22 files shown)


server/game/core/GameStateManager.ts
```diff
@@ -62,18 +79,28 @@ export class GameStateManager {
...
-    // TODO: Where are all of the places GameObjects are stored? Obviously here, but what about Token GOs? Attack GOs? We need to ensure those are all GameObjectRefs, or we'll start building up garbage.
-    public rollbackToSnapshot(snapshotId: number) {
+    /**
+     *
+     * @param snapshotId The specific snapshotId to return to. If not provided, will return to the last snapshot.
+     */
+    public rollbackToSnapshot(snapshotId?: number) {
+        if (snapshotId == null) {
+            Contract.assertTrue(this.snapshots.length > 1, 'No snapshots to rollback to.');
+            // We take a snapshot at the start of someone's turn, so hitting undo would mean we need to back to the second most recent snapshot.
```
AMMayberry1, line 89: We might need a little bit of tracking to remember each player's most recent snapshot separately. In this case, if p1 wants to undo their action but then p2 completely finishes an action before p1 hits the button, then the number of states we need to roll back by will have increased by 1

server/game/core/card/Card.ts
```diff
@@ -975,7 +975,7 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
         return (
             this.unique &&
             this.game.allCards.some(
-                (card) =>
+                (card: any) =>
```
AMMayberry1, line 978: Could probably set the type here as `Card`

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -927,14 +941,19 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor<TSta
...
         }
 
-        public get lastPlayerToModifyHp() {
-            Contract.assertTrue(this.isInPlay());
-            return this._lastPlayerToModifyHp;
+        protected override afterSetState(oldState) {
+            super.afterSetState(oldState);
+            // STATE: I don't wholly trust this covers all cases, but it's a good start at least.
+            if (oldState.zone?.uuid !== this.state.zone.uuid) {
+                const oldZone = this.game.gameObjectManager.get(oldState.zone) as Zone;
```
AMMayberry1, line 953: nit: could we make this method templated so that it's `gameObjectManager.get<Zone>(oldState.zone)`? I think that's a pretty typical pattern for something like this

server/game/core/gameSteps/phases/Phase.ts
```diff
@@ -7,7 +7,8 @@ import type { IStep } from '../IStep';
 import { TriggerHandlingMode } from '../../event/EventWindow';
 
 export abstract class Phase extends BaseStepWithPipeline {
-    public steps: IStep[] = [];
+    private steps: IStep[] = [];
+    private endStep: IStep;
```
AMMayberry1, line 11: Just curious, any specific reason this was needed?

server/game/core/gameSteps/phases/Phase.ts
```diff
@@ -48,4 +50,9 @@ export abstract class Phase extends BaseStepWithPipeline {
             this.game.emit(EventName.OnPhaseEndedCleanup, { phase: this.name });
         }
     }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public resetPhase(): void {
+
```
AMMayberry1, line 56: nit: empty space

server/utils/deck/Deck.ts
```diff
@@ -9,6 +9,17 @@ import type { ITokenCard } from '../../game/core/card/propertyMixins/Token';
 import type { IBaseCard } from '../../game/core/card/BaseCard';
 import type { ILeaderCard } from '../../game/core/card/propertyMixins/LeaderProperties';
 import { cards } from '../../game/cards/Index';
+import type { GameObjectRef } from '../../game/core/GameObjectBase';
+
+export interface IDeskList {
```
AMMayberry1, line 14: nit, typo: "Desk"

server/utils/deck/Deck.ts
```diff
@@ -9,6 +9,17 @@ import type { ITokenCard } from '../../game/core/card/propertyMixins/Token';
...
+    deckCards: GameObjectRef<IPlayableCard>[];
+    outOfPlayCards: any[];
+    outsideTheGameCards: GameObjectRef<Card>[];
+    tokens: GameObjectRef<ITokenCard>[];
+    base: GameObjectRef<IBaseCard> | undefined;
+    leader: GameObjectRef<ILeaderCard> | undefined;
+    allCards: GameObjectRef<Card>[];
+}
 
 export class Deck {
```
AMMayberry1, line 24: Since this gets used in the Lobby flow and we don't have tests for that yet, it would be good to do a quick smoke check with the local dev setup that both a lobby game and a test setup game work. Lmk if you need a pointer to that, can't remember if you've done it before

test/helpers/IntegrationHelper.d.ts
```diff
@@ -142,4 +133,13 @@ declare namespace jasmine {
...
+}
+
+/**
+ * Define a single spec. A spec should contain one or more expectations that test the state of the code.
+ * A spec whose expectations all succeed will be passing and a spec with any failures will fail.
+ * @param expectation Textual description of what this spec is checking
+ * @param assertion Function that contains the code of your test. If not provided the test will be pending.
+ * @param timeout Custom timeout for an async spec.
+ */
+declare function uit(expectation: string, assertion?: jasmine.ImplementationCallback, timeout?: number): void;
```
AMMayberry1, line 145: Could we go back to `undoIt`, or really anything with `undo` in the name? I think that would be a bit clearer

test/helpers/IntegrationHelper.js
```diff
@@ -90,9 +69,6 @@ global.integration = function (definitions) {
             const setupGameStateWrapperAsync = async (options) => {
                 await gameStateBuilder.setupGameStateAsync(newContext, options);
                 gameStateBuilder.attachAbbreviatedContextInfo(newContext, contextRef);
-                if (options.testUndo) {
```
AMMayberry1, line 93: I know there's an issue with tests that have the `setupStateAsync` baked directly into the case, could we solve that by just keeping this hook (or something similar) in here?

test/scenarios/undo/Undo.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+                    player1: {
+                        groundArena: [
+                            { card: '21b-surgical-droid' },
+                            { card: 'r2d2#ignoring-protocol', damage: 3 },
+                            { card: 'c3po#protocol-droid', damage: 1 }],
+                    },
+                    player2: {
+                        groundArena: [{ card: 'wampa', damage: 2 }]
+                    },
+                    testUndo: true
```
AMMayberry1, line 67: Looks like this parameter isn't used anymore - is there still value in having these tests copied here?

server/game/core/Game.js
```diff
@@ -1495,10 +1605,27 @@ class Game extends EventEmitter {
...
+    /**
+     * Should only be used for manual testing inside of unit tests, *never* committing any usage into main.
+     * @param {() => void} fcn
+     */
+    enableUndo(fcn) {
+        this.#experimental.undo = true;
+        try {
+            return fcn();
+        } finally {
+            this.#experimental.undo = false;
```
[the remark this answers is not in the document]
CDMayberry: It does happen here? The finally is just to ensure the undo is unflagged before the exception leaves the scope.

server/game/core/GameObjectBase.ts
```diff
@@ -18,10 +22,13 @@ export interface IGameObjectBaseState {
 export interface GameObjectRef<T extends GameObjectBase = GameObjectBase> {
     isRef: true;
     uuid: string;
+
+    /** This property is for debugging purposes only and will be wiped during rollback. Do not use it anywhere in the code. */
+    // gameObject: never;
```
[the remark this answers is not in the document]
CDMayberry: Yes, sorry meant to remove that. initially I added that to inform people that this is bad and it shouldn't be referenced by marking it as never, but I then thought it might be better to just not have them be aware of it to begin with.

server/game/core/GameStateManager.ts
CDMayberry: I just was thinking about that. If it's null we probably also need a way to tell _who_ is calling rollback.

server/game/core/card/Card.ts
CDMayberry: No but this is probably worth noting: This actually "any" before I made my changes. My changes actually made it infer as Card, but the methods it's calling aren't all available on the base Card class.

server/game/core/gameSteps/phases/Phase.ts
CDMayberry: Ah no, so I was going down a rabbit whole trying to figure out why something wasn't working, and I was starting to separate out the steps more and more. I realized it was a issue somewhere and left this. Will remove endStep.

test/helpers/IntegrationHelper.d.ts
CDMayberry: kk will move to undoIt, jasmine convention just keeps it really simple (xit, fit, etc.)

server/utils/deck/Deck.ts
CDMayberry: Yes I've done it before, I'll start it up and resolve convo assuming it works as expected.

test/helpers/IntegrationHelper.js
CDMayberry: We might be able to, but I wanted to avoid doing snapshots on any game setup call. That said, maybe I can have the undoIt call put in a flag that says "this is undoIt, setupGameAsync can take a snapshot" rather than having a manual flag. I'll give that a go.

CDMayberry: Wait hm. The problem there is we we do hit the gameSetup call on the second run, we basically override the undo happening because we just forcibly reset the state at that point. I might need to have it do game setup on the first call, but on the second assertion call basically have it flag as "don't do the normal game setup, just do the rollback to the snapshot after the first setup"

server/game/core/card/Card.ts
AMMayberry1: Oh yeah now that I look at it, this would hit an error if it were ever actually called - looks like it's not though, so we can just remove it outright

server/game/core/Game.js
```diff
@@ -1456,15 +1552,29 @@ class Game extends EventEmitter {
         return {};
     }
 
-    // takeSnapshot() {
-    //     const snapshot = this.gameObjectManager.takeSnapshot();
-    //     this.gameObjectManager.pushSnapshot(snapshot);
-    //     return snapshot.id;
-    // }
+    takeSnapshot() {
+        if (this.#experimental.undo && this.pipeline.currentStep instanceof ActionPhase) {
```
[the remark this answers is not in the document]
CDMayberry: Ok, added it as a optional function to the interface and now Game just checks if the function exists on the step.

server/game/core/card/Card.ts
CDMayberry: Hm so this function is in use by DrawSpecificCardSystem.

server/game/core/GameStateManager.ts
CDMayberry: Alright, removed the nullability, instead there is now a rollbackToPlayerSnapshot which finds the previous snapshot depending on if the calling player is the active player.

server/utils/deck/Deck.ts
CDMayberry: Alright no issues in the lobby. Also, I think Lobby uses it's own player object, so I think this is good regardless.

server/game/core/GameStateManager.ts
```diff
@@ -67,29 +68,43 @@ export class GameStateManager {
...
 
         return snapshot.id;
     }
 
+    public rollbackToPlayerSnapshot(player: Player) {
+        Contract.assertPositiveNonZero(this.snapshots.length, `Tried to to rollback to ${player.name}'s snapshot but there are none.`);
+
+        // This assumes snapshots are taken at the start of a player's turn.
+        let snapshot: IGameSnapshot;
+        if (snapshot && this.game.state.actionPhaseActivePlayer.uuid !== player.uuid) {
```
AMMayberry1, line 90: Ah yeah, I forgot to mention this but this still doesn't work in all cases. If player 1 takes initiative (passing for the rest of the phase), then player 2 is free to take sequential actions until they're done.

Between that and the fact that we will eventually need snapshots for e.g. start of previous turn, start of next turn, might be good to have some kind of mapping layer on top of the snapshots we can index using enum values or something.

This could all potentially go in the next PR depending on how much you want in this one.

server/game/core/utils/Helpers.ts
```diff
@@ -254,4 +254,12 @@ function mergeProperty<TPropertySet extends { [key in TPropName]?: TMergePropert
     return { ...propertySet, [newPropName]: mergeFn(oldPropValue, newPropValue) };
 }
 
-export type DistributiveOmit<T, K extends keyof T> = T extends any ? Omit<T, K> : never;
+export function objectForEach<T extends Record<any, any>, TK extends Extract<keyof T, string> = Extract<keyof T, string>>(obj: T, fcn: (prop: TK, value?: T[TK]) => void) {
+    for (const prop in obj) {
+        if (Object.prototype.hasOwnProperty.call(obj, prop)) {
```
AMMayberry1, line 259: Just for my education, what's this doing? Haven't seen it used before, seems like a cool trick

CDMayberry: It's a safeguard, it prevents inherited properties from being looped through, such as toString. In essence it ensures that you are only looping through the properties you directly added to the object.

server/game/core/GameStateManager.ts
```diff
@@ -65,43 +88,122 @@ export class GameStateManager {
...
+            Contract.fail('Invalid rollback settings provided.');
         }
 
-        Contract.assertNotNullLike(snapshot, `Tried to get ${player.name}'s previous snapshot, but no snapshot was found.`);
-        Contract.assertNotNullLike(snapshot.gameState.actionPhaseActivePlayer.uuid === player.uuid, () => `Expected player ${player.name} to have been the active player for the rollback snapshot, but a snapshot for ${this.get(snapshot.gameState.actionPhaseActivePlayer).name} was found at the expected index.`);
-
         return this.rollbackToSnapshot(snapshot.id);
     }
 
+    private getSnapshotForPlayer(playerRef: GameObjectRef<Player>): IGameSnapshot {
```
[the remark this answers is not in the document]
CDMayberry: Oh fair.

server/game/core/utils/Helpers.ts
AMMayberry1: Makes sense. What is `Object.prototype.hasOwnProperty.call`? I'm familiar with `Object.prototype.hasOwnProperty` but not this form

CDMayberry: It's a way of calling a function that defines what `this` is. Essentially we use the function directly on the prototype and call it _on_ the object, call is rather than call it from the object (IE `object.hasOwnProperty`). This was the old, and still often recommended way, for getting around an object potentially being null.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call

In Node 2022+ and modern web, there is now a static `Object`.hasOwn` function that does this very thing without the ugliness of accessing the prototype and using call; we just still have a older minimum version for the project so we can't use it yet.

AMMayberry1 (approved):

server/game/core/card/Card.ts
AMMayberry1: Missed this was still open - it's fine as-is, can sort it out later

PR #1037, opened 2025-04-03
Fix uniqueness check when multiple unique units enter play

### LINKED ISSUE #1029: Two copies of a unique unit rescued simultaneously breaks game
Player encountered a situation where their opponent had capture two copies of the same unique unit. When the captor was defeated and the captured units rescued, the game gave the prompt for choosing a unique unit to defeat (as expected) but this caused a game lock for some reason.

Player report:

I killed a Cad Bane unit who held two Kuiils. I had to kill one Kuill, but now it wants be to kill one again. Clicking on the second Kuiil doesn't do anything

Image

moyerr:
Fixes #1029 

#### Prompt to defeat multiple copies
Screenshot 2025-04-02 at 9 32 17 PM

#### Prompt state when 2 copies are selected
Screenshot 2025-04-02 at 9 32 30 PM

### DIFF
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -1235,24 +1235,20 @@ class Game extends EventEmitter {
 
     /** Goes through the list of cards moved during event resolution and does a uniqueness rule check for each */
     checkUniqueRule() {
-        const dedupedMovedCards = this.movedCards.reduce((acc, card) => {
-            // Only prompt once per copy of a unique card controlled by the same player
-            const existingCard = acc.find((otherCard) =>
-                otherCard.title === card.title &&
-                otherCard.subtitle === card.subtitle &&
-                otherCard.controller === card.controller
-            );
+        const checkedCards = new Array();
 
-            if (!existingCard) {
-                acc.push(card);
-            }
-
-            return acc;
-        }, []);
-
-        for (const movedCard of dedupedMovedCards) {
+        for (const movedCard of this.movedCards) {
             if (EnumHelpers.isArena(movedCard.zoneName) && movedCard.unique) {
-                movedCard.checkUnique();
+                const existingCard = checkedCards.find((otherCard) =>
+                    otherCard.title === movedCard.title &&
+                    otherCard.subtitle === movedCard.subtitle &&
+                    otherCard.controller === movedCard.controller
+                );
+
+                if (!existingCard) {
+                    checkedCards.push(movedCard);
+                    movedCard.checkUnique();
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


server/game/core/Game.js
```diff
@@ -1235,7 +1235,22 @@ class Game extends EventEmitter {
...
-        for (const movedCard of this.movedCards) {
+        const dedupedMovedCards = this.movedCards.reduce((acc, card) => {
+            // Only prompt once per copy of a unique card controlled by the same player
+            const existingCard = acc.find((otherCard) =>
+                otherCard.title === card.title &&
+                otherCard.subtitle === card.subtitle &&
+                otherCard.controller === card.controller
+            );
+
+            if (!existingCard) {
```
AMMayberry1, line 1246: We have to be a little bit careful here because of Clone, since it can copy a unique card but will not itself have the `unique` property so it would not trigger the unique rule defeat.

Concretely the scenario I could imagine is:

- P2 has two Kuiil and one Clone captured
- All three are rescued
- The Clone is declared to be a Kuiil on re-entering play
- We select Clone as the `existingCard` here, since it will present as having the same title and subtitle
- `checkUnique()` on Clone is never called b/c Clone itself is not considered unique, so the two "real" Kuiils both stay on the field

test/server/core/card/UniqueRule.spec.ts
```diff
@@ -381,5 +381,180 @@ describe('Uniqueness rule', function() {
                 expect(context.cellBlockGuard).toBeInZone('groundArena');
             });
         });
+
+        describe('When multiple copies of a unique unit are rescued simultaneously,', function() {
+            it('the player is prompted to choose which copy to defeat', async function () {
```
AMMayberry1, line 386: For this and the following test, can we add a second test where we choose to defeat the other copy just to make sure there's not some issue that's sensitive to ordering or something like that

server/game/core/Game.js
```diff
@@ -1235,7 +1235,22 @@ class Game extends EventEmitter {
...
+                acc.push(card);
+            }
+
+            return acc;
+        }, []);
```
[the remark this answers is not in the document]
moyerr: Was able to reduce some looping by performing the logic in the original for-loop, but still needed to use `find` since we aren't comparing the whole object for this check.

The number of cards that can move at once will always be relatively small in real-game scenarios, so this should never blow up too much.

test/server/core/card/UniqueRule.spec.ts
moyerr: Done! ✅

AMMayberry1 (approved):

Issue #1032, opened 2025-04-02
Assajj JTL with Corvus


Twet1:
When Assajj deployed as a ground unit, and you play Corvus. She should be able to attach to the Corvus as a pilot. report is that this interaction killed her. 

Using the unit tests, Vonreg works from ground to Corvus as expected.

PR #1046, opened 2025-04-05
test asajj can be attached to corvus

ssoulier:
Test https://github.com/SWU-Karabast/forceteki/issues/1032

Seems working just fine.

(used cursor.com to write the test)


AMMayberry1:
Hey I realized that the original GH issue was stale, we got more details from the player later on what the actual issue was. It's in this issue: https://github.com/SWU-Karabast/forceteki/issues/1047

Do you mind if we close down this PR and I can maybe link this branch in the GH issue as a starting point?

AMMayberry1:
Resolved in another PR

PR #1048, opened 2025-04-07
Add Darth Maul unit and multi-attack support

adbronico:


server/game/actions/InitiateAttackAction.ts
```diff
@@ -28,6 +28,12 @@ export class InitiateAttackAction extends PlayerAction {
             : exhaustSelf();
 
         super(game, card, 'Attack', [exhaustCost], {
+            mode: TargetMode.BetweenVariable,
+            minNumCardsFunc: () => 1,
+            maxNumCardsFunc: (context) => context.source.getMaxUnitAttackLimit(),
+            canUseSingleSelectMode: true,
+            useSingleSelectModeFunc: (attacker, possibleTargets) => attacker.getMaxUnitAttackLimit() === 1 || possibleTargets.length === 1 || possibleTargets.some((card) => card.isBase()) && possibleTargets.filter((card) => card.isUnit()).length === 1,
```
AMMayberry1, line 35: This method shouldn't be required to validate that there is only one legal target (i.e. possibleTargets === 1). That opens us up to bugs if someone provides a wrong implementation for this method that doesn't correctly check number of targets and we end up trying to set SingleCardSelector with multiple targets.

I think to guard against that, this method should be set up so it can assume it's only being called in the case that there is only one legal target

server/game/actions/InitiateAttackAction.ts
```diff
@@ -28,6 +28,12 @@ export class InitiateAttackAction extends PlayerAction {
             : exhaustSelf();
 
         super(game, card, 'Attack', [exhaustCost], {
+            mode: TargetMode.BetweenVariable,
+            minNumCardsFunc: () => 1,
+            maxNumCardsFunc: (context) => context.source.getMaxUnitAttackLimit(),
+            canUseSingleSelectMode: true,
```
AMMayberry1, line 34: I think we can probably do without this boolean and just check whether useSingleSelectModeFunc exists? Unless there's a case where we need that flexibility

server/game/core/attack/Attack.ts
```diff
@@ -40,10 +40,53 @@ export class Attack {
...
+    public getSingleTarget(): IAttackableCard {
+        Contract.assertTrue(this.targets.length === 1, 'Expected one target but there are multiple');
+        return this.targets[0];
+    }
+
+    public getAllTargets(): IAttackableCard[] {
+        return this.targets;
+    }
+
+    public isTargetUnit(allowMultipleMatches: boolean = false): boolean {
```
AMMayberry1, line 52: Realizing that naming-wise these two might be clearer as `targetIsUnit()` and `targetIsUnitWithCondition()`.

Honestly, I wonder if the usage would be clear enough that we could just fold them into one named `targetIsUnit()` which has an optional condition filter? Not sure if it would be obvious how to use though. If we do keep `targetIsUnit()` separate, then definitely it should just call `targetIsUnitWithCondition()` under the hood so we don't have duplicate implementations

server/game/cards/01_SOR/upgrades/JediLightsaber.ts
```diff
@@ -21,10 +21,10 @@ export default class JediLightsaber extends UpgradeCard {
 
             // need to check if the target is a base - if so, don't apply the stat modifier effect
             immediateEffect: AbilityHelper.immediateEffects.conditional({
-                condition: (context) => context.event.attack.target.isBase(),
-                onFalse: AbilityHelper.immediateEffects.forThisPhaseCardEffect((context) => ({
+                condition: (context) => context.event.attack.isTargetUnit(),
+                onTrue: AbilityHelper.immediateEffects.forThisPhaseCardEffect((context) => ({
                     effect: AbilityHelper.ongoingEffects.modifyStats({ power: -2, hp: -2 }),
-                    target: (context.event.attack as Attack).target
+                    target: (context.event.attack as Attack).getAllTargets()
```
AMMayberry1, line 27: Do you think this would look cleaner as a property getter so it's `attack.targets` instead of `attack.getAllTargets()`? I could go either way

server/game/cards/03_TWI/events/CornerThePrey.ts
```diff
@@ -15,8 +15,8 @@ export default class CornerThePrey extends EventCard {
             initiateAttack: {
                 attackerLastingEffects: (_context, attack) =>
                     ({
-                        condition: () => !attack.target.isBase(),
-                        effect: AbilityHelper.ongoingEffects.modifyStats({ power: attack.target.damage, hp: 0 })
+                        condition: () => !attack.getAllTargets().some((target) => target.isBase()),
```
AMMayberry1, line 18: Did you mean `attack.getAllTargets().some((target) => !target.isBase())`?

server/game/cards/03_TWI/leaders/AnakinSkywalkerWhatItTakesToWin.ts
```diff
@@ -19,7 +19,7 @@ export default class AnakinSkywalkerWhatItTakesToWin extends LeaderUnitCard {
                 cardTypeFilter: WildcardCardType.Unit,
                 immediateEffect: AbilityHelper.immediateEffects.attack({
                     attackerLastingEffects: {
-                        condition: (attack) => attack.target.isUnit(),
+                        condition: (attack) => attack.isTargetUnitWithCondition(() => true, true),
```
AMMayberry1, line 22: Should this just be `isTargetUnit()`?

server/game/core/attack/Attack.ts
```diff
@@ -40,10 +40,53 @@ export class Attack {
...
+            }
+            if (!target.isInPlay()) {
+                continue;
+            }
+            if (condition(target)) {
+                matchingUnits.push(target);
+            }
+        }
+
+        if (!allowMultipleMatches) {
```
AMMayberry1, line 82: Should this maybe be filtering on number of targets instead of number of matched targets? I assume the intent is to enforce cases where there is only one attack target, so if the filter happens to filter out one of them then this might still erroneously pass

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -909,6 +907,18 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor<TSta
             }
         }
 
+        public getMaxUnitAttackLimit(): number {
+            let attackLimit = 1;
+            if (this.hasOngoingEffect(EffectName.CanAttackMultipleUnitsSimultaneously)) {
+                this.getOngoingEffectValues(EffectName.CanAttackMultipleUnitsSimultaneously).forEach((effect) => {
```
AMMayberry1, line 913: nit: we try to avoid `forEach` for debugging reasons, prefer regular for loop

server/game/gameSystems/AttackStepsSystem.ts
```diff
@@ -220,7 +238,12 @@ export class AttackStepsSystem<TContext extends AbilityContext = AbilityContext>
     }
 
     public override checkEventCondition(event, additionalProperties): boolean {
-        return this.canAffect(event.target, event.context, additionalProperties);
+        for (const target of event.target) {
```
AMMayberry1, line 241: nit: in case it's ever somehow not an array, let's use `Helpers.asArray(event.target)`

server/game/gameSystems/DamageSystem.ts
```diff
@@ -191,24 +191,26 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext, TPro
         Contract.assertTrue(context.source.isUnit());
         Contract.assertNotNullLike(card);
 
-        let damageDealtBy: IUnitCard;
+        let damageDealtBy: IUnitCard[];
 
         if (properties.source) {
-            Contract.assertTrue(properties.source.isUnit());
-            damageDealtBy = properties.source;
+            if (Array.isArray(properties.source)) {
```
AMMayberry1, line 197: nit: could simplify this into one check by first creating `const araSource = Helpers.asArray(properties.source)`

test/gameSetups/testGameTemplate.json
```diff
@@ -1,21 +1,15 @@
 {
```
AMMayberry1, line 1: Can you undo the changes to this file, they should've been blocked by .gitignore

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.darthMaul);
+            expect(context.player1).toHaveEnabledPromptButton('Cancel');
+            expect(context.player1).not.toHaveEnabledPromptButton('Done');
+            expect(context.player1).toBeAbleToSelectExactly([context.moistureFarmer, context.wampa, context.p2Base]);
+            context.player1.clickCard(context.p2Base);
+
+            expect(context.player1).toBeAbleToSelectNoneOf([context.moistureFarmer, context.wampa]);
```
AMMayberry1, line 156: Would prefer to use `.toBeAbleToSelectExactly([context.p2Base]))` just to make sure something weird doesn't happen with the targeting list

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+
+            expect(context.player1).toBeAbleToSelectExactly([context.moistureFarmer, context.wampa]);
+            context.player1.clickCard(context.wampa);
+            context.player1.clickPrompt('Done');
+
+            expect(context.darthMaul.damage).toBe(4);
+            expect(context.moistureFarmer).toBeInZone('discard');
+            expect(context.wampa).toBeInZone('discard');
+            expect(context.p2Base.damage).toBe(0);
+        });
```
AMMayberry1, line 85: Few more basic targeting test cases it would be good to have:

- More than two sentinels
- More than two non-sentinel targets
- Select only one target when multiple are available
- Can't attack ground and space (just add a space unit to an existing test)

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+            context.player1.clickCard(context.cantinaBraggart);
+            context.player1.clickPrompt('Done');
+
+            expect(context.darthMaul.damage).toBe(0);
+            expect(context.moistureFarmer).toBeInZone('discard');
+            expect(context.cantinaBraggart).toBeInZone('discard');
+
+            // Maul should have dealt 1 Overwhelm from Moisture Farmer and 2 Overwhelm from Cantina Braggart
+            expect(context.p2Base.damage).toBe(3);
+        });
```
AMMayberry1, line 296: Let's add a timing test where both targets have a when-defeated, just to confirm they both go off at the same time

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.darthMaul);
+            context.player1.clickCard(context.wampa);
+            context.player1.clickCard(context.moistureFarmer);
+            context.player1.clickPrompt('Done');
+
+            expect(context.darthMaul.damage).toBe(0);
+            expect(context.shield).not.toBeAttachedTo(context.darthMaul);
+        });
```
AMMayberry1, line 206: Let's add tests for defenders with shield - a test where only one has a shield and a test where they both have shields.

Also we should add tests for other prevention effects like the Finn unit, to ensure that it only affects the unit selected for the prevention (could also use Boba Fett's Armor)

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+            context.player1.setDistributeDamagePromptState(new Map([
+                [context.moistureFarmer, 2],
+                [context.cantinaBraggart, 1],
+            ]));
+
+            // Moisture Farmer should have taken 10 new damage - 2 from the Flamethrower, 8 from the attack
+            expect(context.moistureFarmer.damage).toBe(11);
+        });
+
+        it('should deal Overwhelm damage from both targets if attacking with Overwhelm', async function () {
```
AMMayberry1, line 268: We need to test the cases where Maul defeats one or both defenders (test case for each) and the overwhelm damage goes straight through. Fallen Lightsaber is good for this

server/game/actions/InitiateAttackAction.ts
adbronico: So this existed because there are situations where you want to be able to use single select, but not be forced to use it, but also times when you don't want to enable single select.

For the latter, Exploit is the key case. I think allowing it to auto resolve when Exploit is chosen and there is only one valid target is a bit weird so I introduced this to allow that case to still exist.

server/game/cards/01_SOR/upgrades/JediLightsaber.ts
adbronico: Maybe we poll the dev team? I'm also pretty indifferent

server/game/cards/03_TWI/events/CornerThePrey.ts
adbronico: I think they give the same result but yours would in theory be more efficient

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+            context.player1.clickPrompt('Done');
+
+            expect(context.darthMaul.damage).toBe(0);
+            expect(context.moistureFarmer).toBeInZone('discard');
+            expect(context.cantinaBraggart).toBeInZone('discard');
+
+            // Maul should have dealt 1 Overwhelm from Moisture Farmer and 2 Overwhelm from Cantina Braggart
+            expect(context.p2Base.damage).toBe(3);
+        });
+    });
```
AMMayberry1, line 297: Thought of one more - let's test with something that can't be targeted like one of the Sabine units

server/game/actions/InitiateAttackAction.ts
AMMayberry1: Could that still be accomplished by checking if `useSingleSelectModeFunc == null`?

server/game/cards/03_TWI/events/CornerThePrey.ts
AMMayberry1: They're a bit different - the current one will return false if any target is a base, my suggested change will return true if any target is not a base

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
AMMayberry1: *defeats one or both at the on-attack stage

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
AMMayberry1, line 297: Just realized, need to test with Ambush

server/game/cards/01_SOR/upgrades/JediLightsaber.ts
adbronico: I thought about this and I think there's two reasons I don't like using getters:
1. We need to do the Contract check inside getSingleTarget. Can we do that inside a getter?
2. target vs. targets could be easy to mistype compared to the methods

server/game/cards/03_TWI/leaders/AnakinSkywalkerWhatItTakesToWin.ts
adbronico: No, we need it like this in case an Anakin buffs a Maul to attack two units

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
adbronico: Added all but the last one, which is already covered

server/game/cards/01_SOR/upgrades/JediLightsaber.ts
AMMayberry1: Sounds good to me. For future reference on #1, yeah you totally can. Getters are basically just normal methods with a special signature.

server/game/core/attack/Attack.ts
```diff
@@ -80,7 +65,7 @@ export class Attack {
         }
 
         if (!allowMultipleMatches) {
```
AMMayberry1, line 67: If we're changing the behavior I feel like this name doesn't fit anymore, might want something like `allowMultipleTargets`

test/server/cards/03_TWI/units/DarthMaulRevengeAtLast.spec.ts
```diff
@@ -294,5 +553,33 @@ describe('Darth Maul, Revenge At Last', function() {
             // Maul should have dealt 1 Overwhelm from Moisture Farmer and 2 Overwhelm from Cantina Braggart
             expect(context.p2Base.damage).toBe(3);
         });
+
+        it('should deal Overwhelm damage from both targets if both died in the attack', async function () {
```
AMMayberry1, line 557: Would be good to have another test where only one unit dies just to confirm nothing weird happens with like attack resolution

AMMayberry1 (commented):
nit: maybe we can call this something like `directOverwhelmDamage` or something to make it clearer that it doesn't include "normal" overwhelm damage?

server/game/core/attack/AttackFlow.ts
```diff
@@ -55,48 +55,65 @@ export class AttackFlow extends BaseStepWithPipeline {
             return;
         }
 
-        let overwhelmDamageOnly = false;
-        if (!this.attack.isAttackTargetLegal()) {
-            if (!this.attack.hasOverwhelm()) {
-                this.context.game.addMessage('The attack does not resolve because the defender is no longer in play');
-                return;
+        const inPlayTargets = [];
+        let overwhelmDamage = 0;
```
AMMayberry1, line 59: nit: maybe we can call this something like directOverwhelmDamage or something to make it clearer that it doesn't include "normal" overwhelm damage?

AMMayberry1, line 59: nit: maybe we can call this something like directOverwhelmDamage or something to make it clearer that it doesn't include "normal" overwhelm damage?

AMMayberry1 (approved):

PR #1055, opened 2025-04-08
Add Hotshot Maneuver

Frugghi:
image

### DIFF
--- a/server/game/cards/04_JTL/events/HotshotManeuver.ts
+++ b/server/game/cards/04_JTL/events/HotshotManeuver.ts
@@ -1,0 +1,47 @@
+import AbilityHelper from '../../../AbilityHelper';
+import type { AbilityContext } from '../../../core/ability/AbilityContext';
+import type { Card } from '../../../core/card/Card';
+import { EventCard } from '../../../core/card/EventCard';
+import { RelativePlayer, TargetMode, WildcardCardType } from '../../../core/Constants';
+
+export default class HotshotManeuver extends EventCard {
+    protected override getImplementationId () {
+        return {
+            id: '6544277158',
+            internalName: 'hotshot-maneuver',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setEventAbility({
+            title: 'For each “On Attack” abilities on a friendly unit, deal 2 damage to a different enemy unit',
+            targetResolver: {
+                activePromptTitle: 'Choose a friendly unit',
+                cardTypeFilter: WildcardCardType.Unit,
+                controller: RelativePlayer.Self,
+                immediateEffect: AbilityHelper.immediateEffects.sequential([
+                    AbilityHelper.immediateEffects.conditional({
+                        condition: (context) => this.numberOfTargets(context) > 0,
+                        onFalse: AbilityHelper.immediateEffects.noAction({ hasLegalTarget: true }),
+                        onTrue: AbilityHelper.immediateEffects.selectCard({
+                            activePromptTitle: (context) => (this.numberOfTargets(context) === 1 ? 'Choose an enemy unit to deal 2 damage to' : `Choose ${this.numberOfTargets(context)} enemy units to deal 2 damage to`),
+                            mode: TargetMode.ExactlyVariable,
+                            numCardsFunc: (context) => this.numberOfTargets(context),
+                            cardTypeFilter: WildcardCardType.Unit,
+                            controller: RelativePlayer.Opponent,
+                            innerSystem: AbilityHelper.immediateEffects.damage({ amount: 2 }),
+                        })
+                    }),
+                    AbilityHelper.immediateEffects.attack(),
+                ]),
+            },
+        });
+    }
+
+    private numberOfTargets<T extends Card>(context: AbilityContext<T>): number {
+        return Math.min(
+            context.player.opponent.getArenaUnits().length,
+            context.target.getTriggeredAbilities().reduce((total: number, ability) => total + (ability.isOnAttackAbility ? 1 : 0), 0)
+        );
+    }
+}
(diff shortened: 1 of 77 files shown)


server/game/cards/04_JTL/events/HotshotManeuver.ts
```diff
@@ -0,0 +1,47 @@
...
+                        onTrue: AbilityHelper.immediateEffects.selectCard({
+                            activePromptTitle: (context) => (this.numberOfTargets(context) === 1 ? 'Choose an enemy unit to deal 2 damage to' : `Choose ${this.numberOfTargets(context)} enemy units to deal 2 damage to`),
+                            mode: TargetMode.ExactlyVariable,
+                            numCardsFunc: (context) => this.numberOfTargets(context),
+                            cardTypeFilter: WildcardCardType.Unit,
+                            controller: RelativePlayer.Opponent,
+                            innerSystem: AbilityHelper.immediateEffects.damage({ amount: 2 }),
+                        })
+                    }),
+                    AbilityHelper.immediateEffects.attack(),
```
Frugghi, line 35: I couldn't figure out how to make this work with a `then` 😕

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -326,8 +326,8 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor<TSta
         }
 
         protected addOnAttackAbility(properties: Omit<ITriggeredAbilityProps<this>, 'when' | 'aggregateWhen'>): void {
-            const triggeredProperties = { ...properties, when: { onAttackDeclared: (event, context) => event.attack.attacker === context.source } };
-            this.addTriggeredAbility(triggeredProperties);
+            const when: WhenTypeOrStandard = { [StandardTriggeredAbilityType.OnAttack]: true };
+            this.addTriggeredAbility({ ...properties, when });
```
AMMayberry1, line 330: One thing that will be super helpful going forward is to add an automated check that we are correctly registering the on-attack ability for all cards. If you look at the original when-defeated PR for Krennic (https://github.com/SWU-Karabast/forceteki/pull/893/files#diff-4cbd89bdc5f79a3adeb0c3eb44e82c004c5eb919360fd32aa34803adaef0c7fe) you can see where I added a check on InPlayCard which does a text search to confirm that if the card has "when defeated" in the text that it also has an "when defeated" ability registration.

This obviously has some limits, so I also added an override option to disable the error on specific cards if it's a false positive (you can see it on e.g. Inspiring Mentor). But I think especially for veteran devs who are used to writing the on-attack or when-defeated stuff a certain way, it's good to have the check to ensure that they are correctly using the new properties. Otherwise it'll be pretty hard to catch when we miss using the correct property.

server/game/cards/04_JTL/events/HotshotManeuver.ts
```diff
@@ -0,0 +1,47 @@
...
+            id: '6544277158',
+            internalName: 'hotshot-maneuver',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.setEventAbility({
+            title: 'For each “On Attack” abilities on a friendly unit, deal 2 damage to a different enemy unit',
+            targetResolver: {
+                activePromptTitle: 'Choose a friendly unit',
```
AMMayberry1, line 19: I think it might be better without this, by default the prompt will show the ability title now

server/game/cards/04_JTL/events/HotshotManeuver.ts
AMMayberry1: Interesting - how exactly did you have it set up and what was the issue you were hitting? I was assuming this would be a pretty straightforward one to split, might be a bug somewhere

Frugghi: I tried different ways but the issue was that it was ignoring the target I was passing (I tried using `target` and `attacker`) so the user was prompted to pick an attacker.

server/game/core/card/baseClasses/InPlayCard.ts
```diff
@@ -330,6 +330,12 @@ export class InPlayCard<T extends IInPlayCardState = IInPlayCardState> extends I
             !this.triggeredAbilities.some((ability) => ability.isWhenDefeated),
             `Card ${this.internalName} has one or more 'When Defeated' keywords in its text but no corresponding ability definition or set property 'disableWhenDefeatedCheck' to true on card implementation`
         );
+        Contract.assertFalse(
+            !this.disableOnAttackCheck &&
+            cardText && Helpers.hasSomeMatch(cardText, /\bOn Attack\b/g) &&
```
AMMayberry1, line 335: If you use the regex above for the "When Defeated", it will automatically omit gained abilities since those don't start the ability trigger on a newline. We won't need as many manual disables in that case

server/game/cards/04_JTL/events/HotshotManeuver.ts
AMMayberry1: Oh huh, I'll take a look tomorrow and see if I spot any issues

AMMayberry1: I tried this and it passed the current test suite:

```ts
this.setEventAbility({
    title: 'For each “On Attack” abilities on a friendly unit, deal 2 damage to a different enemy unit',
    targetResolver: {
            condition: (context) => this.numberOfTargets(context) > 0,
            onFalse: AbilityHelper.immediateEffects.noAction({ hasLegalTarget: true }),
            onTrue: AbilityHelper.immediateEffects.selectCard({
                activePromptTitle: (context) => (this.numberOfTargets(context) === 1 ? 'Choose an enemy unit to deal 2 damage to' : `Choose ${this.numberOfTargets(context)} enemy units to deal 2 damage to`),
                mode: TargetMode.ExactlyVariable,
                numCardsFunc: (context) => this.numberOfTargets(context),
                controller: RelativePlayer.Opponent,
                innerSystem: AbilityHelper.immediateEffects.damage({ amount: 2 }),
            })
        }),
    },
    then: (thenContext) => ({
        title: 'Attack with the selected unit',
        immediateEffect: AbilityHelper.immediateEffects.attack({
            target: thenContext.target
        })
    })
});
```

AMMayberry1 (approved):

PR #1057, opened 2025-04-08
Che bato/bug reporting

CheBato:
FE PR:  https://github.com/SWU-Karabast/forceteki-client/pull/283

Steps to setup the webhook: 
step 1 - On discord you need to be server admin.
step 2 - Click on the **server name** on the top left corner and select **Server Settings** (image)

step2

step 3 - In  the server settings menu select **integrations** (image)

step3

step 4 - click on webhooks

step 5 - click on New Webhook

step 6 - Click on the newly created hook

step 7 - change the name and the channel this webhook should post to we can add a bug picture aswell.

step 8 - Copy Webhook URL

step 9 - Paste the webhook URL to the env with key DISCORD_BUG_REPORT_WEBHOOK_URL="<your webhook URL>"

and now you're done it should now send


server/gamenode/Lobby.ts
```diff
@@ -761,4 +765,54 @@ export class Lobby {
...
+            if (!description || description.trim().length === 0) {
+                socket.send('bugReportResult', {
+                    success: false,
+                    message: 'Bug description cannot be empty'
+                });
+                return;
+            }
+
+            // Limit description length
+            if (description.length > 500) {
```
AMMayberry1, line 782: 500 characters is a bit low, we should allow people to write verbose descriptions if they want. It's just going into a log line so we don't really have any technical limitations on the size here from that angle, we can probably just go with whatever max is allowed by Discord

server/gamenode/Lobby.ts
```diff
@@ -761,4 +765,54 @@ export class Lobby {
...
+            // we find the user
+            const existingUser = this.users.find((u) => u.id === socket.user.id);
+            existingUser.reportedBugs += success ? 1 : 0;
+            this.sendLobbyState();
+        } catch (error) {
+            logger.error('Error processing bug report', {
+                error: { message: error.message, stack: error.stack },
+                lobbyId: this.id,
+                userId: socket.user.id
+            });
```
AMMayberry1, line 815: The log line is good to have but I think this should then throw

server/utils/bugreport/BugReportHandler.ts
```diff
@@ -0,0 +1,307 @@
...
+            });
+
+            return true;
+        } catch (error) {
+            logger.error('Failed to send bug report to Discord', {
+                error: { message: error.message, stack: error.stack },
+                lobbyId: bugReport.lobbyId
+            });
+
+            return false;
```
AMMayberry1, line 72: Same note here, this can just throw

server/utils/bugreport/BugReportHandler.ts
```diff
@@ -0,0 +1,307 @@
...
+            player2: this.capturePlayerState(player2)
+        };
+    }
+
+    /**
+     * Captures a player's state for a bug report
+     * @param player The player object
+     * @returns A simplified player state representation
+     */
+    private capturePlayerState(player: any): IPlayerBugReportState {
```
AMMayberry1, line 193: I was thinking we will need something like this eventually, so this is great to see! Long-term I want to kind of refactor the game state stuff out of the test suite and merge it with this so it's all using the same interfaces but that's a project for another time. For now, I think it would make the most sense for these methods to live on their respective object classes (Player and Card) kind of the way that `getSummary()` does. That way if we have other scenarios that need to "serialize" the state it's in a place that's easy to find

server/utils/bugreport/BugReportInterfaces.ts
```diff
@@ -0,0 +1,38 @@
+
+export interface IBugReportCardState {
+    card: string;
+    damage?: number;
+    upgrades?: string[];
+    deployed?: boolean;
+    exhausted?: boolean;
+}
+
+export interface IPlayerBugReportState {
```
AMMayberry1, line 10: Actually, looking at this now it might make sense to go ahead and start merging the interfaces since the test interfaces are pretty similar to this already. Let's do the following:

1. Move the `BugReportState` interfaces to somewhere more generic like `Interfaces.ts`
2. Rename them something more generic like `SerializedState` or something
3. Refactor the interfaces like `SwuSetupTestOptions` in IntegrationHelper.d.ts to extend these and then add on the test-specific stuff like `autoSingleTarget`. Note that the import style in that file is a little weird, make sure you use the same format for importing or else things will break

test/gameSetups/testGameTemplate.json
```diff
@@ -1,21 +1,99 @@
 {
     "phase": "action",
     "player1": {
-        "hand": ["the-emperors-legion", "wilderness-fighter"],
```
AMMayberry1, line 4: Do you mind reverting this back to the original? There's some stuff missing now like upgrades and leader deployed status (could also add those if you want)

server/game/core/Player.ts
```diff
@@ -1303,6 +1304,78 @@ export class Player extends GameObject<IPlayerState> {
...
+    public capturePlayerState(player: string): IPlayerSerializedState {
+        const state: IPlayerSerializedState = {};
+        try {
+            // Hand cards
+            if (this.handZone && this.handZone.cards && this.handZone.cards.length > 0) {
+                state.hand = this.handZone.cards.map((card) => card.internalName);
+            }
+
+            // Ground arena units
+            if (this.game && this.game.groundArena) {
```
AMMayberry1, line 1321: Instead of checking `this.game` multiple times, can we just do a contract check at the top? I'm assuming there has to be a game present for this method to work

server/game/core/Player.ts
```diff
@@ -1303,6 +1304,78 @@ export class Player extends GameObject<IPlayerState> {
...
+                const groundArenaCards = this.game.groundArena.getCards({ controller: this });
+                if (groundArenaCards && groundArenaCards.length > 0) {
```
AMMayberry1, line 1323: nit: I think the getCards() method will always return at least an empty array so we probably don't need this if block, same for space arena

server/game/core/Player.ts
```diff
@@ -1303,6 +1304,78 @@ export class Player extends GameObject<IPlayerState> {
...
+
+            // Initiative
+            if (this.game && this.game.initiativePlayer === this) {
+                state.hasInitiative = true;
+            }
+        } catch (error) {
+            logger.error('Error capturing player state', {
+                error: { message: error.message, stack: error.stack },
+                playerId: this.id
+            });
```
AMMayberry1, line 1373: This can throw, same for Card

server/game/core/card/Card.ts
```diff
@@ -1125,6 +1133,78 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
...
     }
 
+    /**
+     * Captures a card's state
+     * @param card The card object
+     * @param player Either player1 or player2
+     * @param currentPlayerID The current players id
+     * @returns A simplified card state representation
+     */
+    public captureCardState(card: any, player: string, currentPlayerID: string): string | ISerializedCardState {
```
AMMayberry1, line 1143: Seems a bit weird that we're passing a card object here, shouldn't it be operating on `this`? Same note for Player

server/game/Interfaces.ts
```diff
@@ -275,6 +275,49 @@ export interface ISetId {
...
+    player1?: IPlayerSerializedState;
+    player2?: IPlayerSerializedState;
+}
+
+export interface ISerializedReportState {
+    description: string;
+    gameState: ISerializedGameState;
+    reporter: {
+        id: string;
+        username: string;
```
AMMayberry1, line 314: It would be useful if we could factor into the game state json somehow so we can easily tell which player is the reporter and which is the opponent

server/game/core/Player.ts
```diff
@@ -1303,6 +1304,78 @@ export class Player extends GameObject<IPlayerState> {
...
+                const spaceArenaCards = this.game.spaceArena.getCards({ controller: this });
+                if (spaceArenaCards && spaceArenaCards.length > 0) {
+                    state.spaceArena = spaceArenaCards
+                        .filter((card) => !card.isLeaderUnit() && card.captureCardState(card, player, this.id) !== null)
+                        .map((card) => card.captureCardState(card, player, this.id));
+                }
+            }
+
+            // Discard pile
+            if (this.discardZone && this.discardZone.cards && this.discardZone.cards.length > 0) {
```
AMMayberry1, line 1341: Similar note on all of these... I don't _think_ we need all the if checks? If a zone is in an invalid state then we probably just want to throw anyway

server/game/core/card/Card.ts
```diff
@@ -1125,6 +1133,78 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
...
+        if (!card || (typeof card.isAttached === 'function' && card.isAttached())) {
```
AMMayberry1, line 1144: What's this `isAttached` stuff doing?

server/utils/bugreport/BugReportHandler.ts
```diff
@@ -0,0 +1,189 @@
...
+     * @param bugReport The bug report data
+     * @returns Promise that resolves to true if successful, false otherwise
+     */
+    public async sendBugReportToDiscord(bugReport: ISerializedReportState): Promise<boolean> {
+        try {
+            // Always log the bug report
+            logger.info(`Bug report received from user ${bugReport.reporter.username}`, {
+                lobbyId: bugReport.lobbyId,
+                reporterId: bugReport.reporter.id,
+                description: bugReport.description
```
AMMayberry1, line 32: Would be good to also dump a compressed game state json in the logs (i.e., stringify with no newlines)

server/gamenode/Lobby.ts
```diff
@@ -767,4 +771,46 @@ export class Lobby {
...
+            const bugReport = this.server.bugReportHandler.createBugReport(
+                description,
+                gameState,
+                socket.user,
+                this.id,
+                this.game?.id
+            );
+
+            // Send to Discord
+            const success = await this.server.bugReportHandler.sendBugReportToDiscord(bugReport);
```
AMMayberry1, line 802: We should throw if success is false

server/gamenode/Lobby.ts
```diff
@@ -767,4 +771,46 @@ export class Lobby {
...
+            // we find the user
+            const existingUser = this.users.find((u) => u.id === socket.user.id);
+            existingUser.reportedBugs += success ? 1 : 0;
+            this.sendLobbyState();
```
AMMayberry1, line 806: Do we need a state update here?

server/gamenode/Lobby.ts
```diff
@@ -767,4 +771,46 @@ export class Lobby {
             }
         }
     }
+
+    // Add this method to the Lobby class
+    private async reportBug(socket: Socket, description: string): Promise<void> {
+        try {
+            // Validate description
+            if (!description || description.trim().length === 0) {
+                socket.send('bugReportResult', {
```
AMMayberry1, line 780: We should make it consistent how we report an error to the FE on this path - in some cases we throw and here we do this `socket.send()`

server/utils/bugreport/BugReportHandler.ts
```diff
@@ -0,0 +1,189 @@
...
+            timestamp: new Date().toISOString()
+        };
+    }
+
+    /**
+     * Captures the current game state for a bug report
+     * @param game The current game instance
+     * @returns A simplified game state representation
+     */
+    public captureGameState(game: any): ISerializedGameState {
```
AMMayberry1, line 162: This should probably live on the Game class

AMMayberry1, line 162: This is admittedly a bit of a bigger ask but we should at least investigate it now. It would be good to add a check on all of our tests to ensure that there is no game state that causes an error when serializing (e.g., I know what we have doesn't work right for piloting leaders but idk if it would cause an exception). That can be done by calling this `captureGameState` method in the `afterEach()` method in IntegrationHelper.

Something like this:

```js
try {
  context.game.captureGameState();
} catch (error) {
  throw new TestSetupError('Failed to correctly serialize post-test game state');
}
```

If you find there are a lot of cases that break then maybe I can invest some time in helping, lmk how it goes either way

server/env.ts
```diff
@@ -9,7 +9,8 @@ const parsedEnv = z
         GAME_NODE_HOST: z.string(),
         GAME_NODE_NAME: z.string(),
         GAME_NODE_SOCKET_IO_PORT: z.coerce.number().int(),
-        SECRET: z.string()
+        SECRET: z.string(),
+        DISCORD_BUG_REPORT_WEBHOOK_URL: z.string(),
```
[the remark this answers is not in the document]
CheBato: yessir!

test/gameSetups/testGameTemplate.json
CheBato: oh ofcourse my mistake i didn't think it got pushed

server/game/core/Player.ts
CheBato: sounds good

server/game/core/card/Card.ts
CheBato: `this `  works with Player and i'm using it there. Here its a bit more tricky because some parameters aren't available from the base Card.

server/game/Interfaces.ts
CheBato: i think we can, do we want the reporter to be consistently player1?

server/game/core/Player.ts
CheBato: yeah i definatelly went overboard here.

server/game/core/card/Card.ts
CheBato: So basically, if we retrieve all the cards in a zone, we also get "experience cards, shield cards, pilot upgrades" and other upgrades. What happened was that they would be on a unit and also in the ground/space arena as a separate card. This way I clear them out.

server/utils/bugreport/BugReportHandler.ts
CheBato: what would be considered a compressed game state json?

server/gamenode/Lobby.ts
CheBato: yes we do, otherwise the FE doesn't retrieve the amount of bugs a user has reported.

server/gamenode/Lobby.ts
CheBato: i'm more inclined on throwing rather then sending a different message. I'll unify.

server/utils/bugreport/BugReportHandler.ts
CheBato: makes sense i'll do that.

server/utils/bugreport/BugReportHandler.ts
brianries, line 32: We're going to want to cleanup the description message so it doesn't break the JSON payload.  We should remove double quotes, not sure if we have to remove anything else or if JSON is happy with anything inside double quotes? (except quotes)?

server/utils/bugreport/BugReportHandler.ts
```diff
@@ -0,0 +1,189 @@
...
+    ISerializedReportState
+} from '../../game/Interfaces';
+
+export class BugReportHandler {
+    private discordWebhookUrl: string;
+
+    public constructor() {
+        // Use provided webhook URL or load from environment variables
+        this.discordWebhookUrl = process.env.DISCORD_BUG_REPORT_WEBHOOK_URL;
+        if (!this.discordWebhookUrl) {
```
brianries, line 16: Could we move this URL check much later in the function, and log the game state inside it when it fails? That way we get a full CONSOLE dump of what we would have sent in developer mode?

server/game/core/card/Card.ts
AMMayberry1: Ah gotcha, in that case we'll need to it kind of like `getSummary()` where it's got overrides across multiple classes

server/game/Interfaces.ts
AMMayberry1: That works fine, if we do that then we probably should also give it a name like "reportingPlayer" and "opponent"

server/game/core/card/Card.ts
AMMayberry1: Oh gotcha, I see now

server/utils/bugreport/BugReportHandler.ts
AMMayberry1: Oh I just mean no whitespace - i.e. set this `space` parameter to null: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#space

server/gamenode/Lobby.ts
AMMayberry1: Oh no that makes sense, all good

server/game/Interfaces.ts
CheBato: i'm not sure how the the Test framework will react to that since I believe its set up for player1 and player2. But I can probably change it to work with reporting player and opponent aswell.

server/utils/bugreport/BugReportHandler.ts
CheBato: We actually do the check in the method sendBugReportToDiscord where we first log the full console dump before checking the url.

CheBato: well i added the gameStateJson now to the logger.info there.

server/utils/bugreport/BugReportHandler.ts
CheBato: setup a cleaner function

scripts/run-server.js
```diff
@@ -12,6 +12,7 @@ function createEnvFile() {
         SECRET: 'verysecret',
         // set this to "true" to force the local client to pull card data jsons from AWS instead of local folder
         FORCE_REMOTE_CARD_DATA: 'false',
+        DISCORD_BUG_REPORT_WEBHOOK_URL: null,
```
AMMayberry1, line 15: A default value for an env var should be `''` because null will get converted to the string `'null'`. environment variable values can only be strings

server/game/core/Game.js
```diff
@@ -1636,6 +1636,44 @@ class Game extends EventEmitter {
...
+        if (players.length !== 2) {
+            return {
+                phase: this.currentPhase,
+                player1: {},
+                player2: {}
+            };
+        }
+        let player1;
+        let player2;
+        if (players[0].id === reportingPlayer) {
```
AMMayberry1, line 1663: nit: let's do it as a switch where we explicitly check the ID matches one of the players in the game, and do contract.fail in the default case in case e.g. a spectator is somehow able to open the prompt

server/game/core/card/Card.ts
```diff
@@ -1135,73 +1135,56 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
...
-                cardState.exhausted = card.exhausted;
-            }
-
-            // Capture upgrades
-            if (card.upgrades && card.upgrades.length > 0) {
-                cardState.upgrades = card.upgrades.map((upgrade) => this.captureCardState(upgrade, player, currentPlayerID));
+            if (this.isUpgrade()) {
+                return null;
             }
+            if (this.isLeaderUnit() || this.isUnit()) {
```
AMMayberry1, line 1147: This pattern works but it's not recommended, it's bad practice for a superclass to have to identify which of its child class types it is and then act differently based on that information. We still do it in some places in the code but I'm trying to move us away from it.

The much better pattern here is to move the code specific to a certain subclass to live inside of that subclass, like how we did with the overrides of `Card.getSummary()`. That way also when we're coding and making change inside that subclass, we can see if we need to make changes to this code instead of needing to remember to come up here and check on updating it.

server/gamenode/Lobby.ts
```diff
@@ -797,20 +793,33 @@ export class Lobby {
                 this.id,
                 this.game?.id
             );
-
             // Send to Discord
             const success = await this.server.bugReportHandler.sendBugReportToDiscord(bugReport);
+            if (!success) {
+                throw new Error('Bug report failed to send to discord.');
```
AMMayberry1, line 799: nit: should include some details from the response

server/gamenode/Lobby.ts
```diff
@@ -772,21 +772,17 @@ export class Lobby {
...
     private async reportBug(socket: Socket, description: string): Promise<void> {
         try {
             // Validate description
             if (!description || description.trim().length === 0) {
-                socket.send('bugReportResult', {
-                    success: false,
-                    message: 'Bug description cannot be empty'
-                });
-                return;
+                throw new Error('description is invalid');
```
AMMayberry1, line 780: Will this be shown to the user? If so, I'd recommend keeping the same string you had before

AMMayberry1: Actually disregard this comment, I see now that it will only happen if the FE side does something wrong

server/game/Interfaces.ts
```diff
@@ -284,6 +284,7 @@ export interface ISerializedCardState {
     exhausted?: boolean;
     capturedUnits?: ({ card: string; owner: string } | string)[];
     flipped?: boolean;
+    keywords?: string[];
```
AMMayberry1, line 287: I'm not sure how I feel about this one since it's not something that would make sense as part of a test setup

AMMayberry1 (approved):

PR #1060, opened 2025-04-09
First pass at implementing the Force Token

moyerr:
### Developer Notes

- The Force Token is implemented as a new `TokenCard` type. Previously, we had only seen Token Units and Token Upgrades, so this more general token card type required a few updates for the engine to support it.
- When the Force Token is in play, it belongs to the `BaseZone`, so there are some minor engine updates to support this.
- Relevant cards are implemented with temporary IDs until the official SWU database adds these cards

### Relevant Cards

| Force Token | Shadowed Undercity | Darth Maul |
| :---: | :---: | :---: |
|  |  |  |


server/game/cards/05_LOF/leaders/DarthMaulSithRevealed.ts
```diff
@@ -0,0 +1,30 @@
...
+            ],
+            targetResolver: {
+                mode: TargetMode.Exactly,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: WildcardZoneName.AnyArena,
+                controller: WildcardRelativePlayer.Any,
+                numCards: 2,
+                immediateEffect: AbilityHelper.immediateEffects.damage({ amount: 1 })
+            }
+        });
```
moyerr, line 28: So the target resolver doesn't actually work as intended here (will not let the user dismiss if there is only 1 unit on board). But the main thing I wanted to test with this was the ability cost, and that is working as expected.

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -187,6 +189,9 @@ export function createXWing<TContext extends AbilityContext = AbilityContext>(pr
 export function createTieFighter<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<ICreateTieFighterProperties, TContext> = {}) {
     return new CreateTieFighterSystem<TContext>(propertyFactory);
 }
+export function createForceToken<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<ICreateForceTokenProperties, TContext> = {}) {
```
AMMayberry1, line 192: It's probably good that the system name is more literal but I think this definitely needs to be called `theForceIsWithYou` 😁

server/game/gameSystems/CreateForceTokenSystem.ts
```diff
@@ -0,0 +1,62 @@
...
+                continue;
+            }
+            event.generatedTokens.push(context.game.generateToken(player, TokenCardName.Force));
+        }
+
+        event.setContingentEventsGenerator((event) => {
+            const events = [];
+
+            for (const token of event.generatedTokens) {
+                const moveCardEvent = new MoveCardSystem({
```
AMMayberry1, line 45: The `MoveCardSystem` is kind of the catchall for generic move cases that don't have their own special rules, which this does (see also DefeatSystem as another example). So in this case let's just do the move in the eventHandler.

server/game/gameSystems/UseTheForceSystem.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { Player } from '../core/Player';
+
+export class UseTheForceSystem<TContext extends AbilityContext = AbilityContext, TProperties extends IPlayerTargetSystemProperties = IPlayerTargetSystemProperties> extends PlayerTargetSystem<TContext, TProperties> {
+    public override name = 'useTheForce';
+    protected override eventName = EventName.OnCardLeavesPlay;
+
+    public override eventHandler(event): void {
+        const forceToken = event.context.player.baseZone.forceToken;
+
+        if (forceToken) {
```
AMMayberry1, line 14: This should be a contract check, we shouldn't get here if the player doesn't have a force token (the canAffect check should block it)

server/game/gameSystems/UseTheForceSystem.ts
```diff
@@ -0,0 +1,26 @@
...
+            forceToken.moveTo(ZoneName.OutsideTheGame);
+        }
+    }
+
+    public override canAffectInternal(player: Player, context: TContext, additionalProperties: any = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        if (!context.player.hasTheForce()) {
```
AMMayberry1, line 20: Little bit of nuance here - we should do this check only if `mustChangeGameState !== GameStateChangeRequired.None`. At the targeting stage of resolution there (usually) isn't a requirement that the ability have any effect. This is more of a thing for abilities that target cards since players are pretty much always legally targetable for effects, but still good to be consistent just in case.

server/game/cards/05_LOF/bases/ShadowedUndercity.ts
```diff
@@ -0,0 +1,23 @@
...
+            internalName: 'shadowed-undercity',
+        };
+    }
+
+    protected override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'The Force is with you',
+            when: {
+                onAttackDeclared: (event, context) => event.attack.attacker.hasSomeTrait(Trait.Force) &&
+                  event.attack.attacker.controller === context.source.owner
```
AMMayberry1, line 18: nit: shouldn't ever matter here, but for consistency it should be `context.player` instead of `context.source.owner`

server/game/cards/05_LOF/leaders/DarthMaulSithRevealed.ts
AMMayberry1: Yeah, looks like in the past we've done a hack involving ExactlyVariable (see General Grievous, Trophy Collector). We can just do that here for now, eventually we probably should add a property to Exactly which will handle this more gracefully.

server/game/cards/05_LOF/tokens/Force.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default class Force extends TokenCard {
+    protected override getImplementationId() {
+        return {
+            id: 'temp-force-id',
+            internalName: 'force',
+        };
+    }
+
+    public override isForceToken(): this is ITokenCard {
```
AMMayberry1, line 12: Should return `this is Force`

server/game/cards/05_LOF/tokens/Force.ts
```diff
@@ -0,0 +1,15 @@
+import type { ITokenCard } from '../../../core/card/propertyMixins/Token';
+import { TokenCard } from '../../../core/card/TokenCards';
+
+export default class Force extends TokenCard {
```
AMMayberry1, line 4: I think `ForceToken` is probably better just so it's clear wherever it's used

server/game/core/card/Card.ts
```diff
@@ -325,22 +325,19 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
...
 
+        if (printedTypes[0] === 'token') {
             switch (printedTypes[1]) {
                 case 'unit':
                     return CardType.TokenUnit;
                 case 'upgrade':
                     return CardType.TokenUpgrade;
                 default:
-                    throw new Error(`Unexpected token types: ${printedTypes}`);
+                    return CardType.TokenCard;
```
AMMayberry1, line 337: I always prefer for the cases to explicitly handle all expected options and the default to be a failure. In this case there could be an issue if there's a typo like `"token upargde"`.

If the Force token's type ends up being literally just `"token"` then we probably want to do a length check first to see if there's a second word in the text, and only do the switch at that point

server/game/core/zone/BaseZone.ts
```diff
@@ -37,6 +43,18 @@ export class BaseZone extends ZoneAbstract<ILeaderCard | IBaseCard, IBaseZoneSta
         this.state.leader = value?.getRef();
     }
 
+    public get forceToken(): ITokenCard | null {
+        return this.game.gameObjectManager.get(this.state.forceToken);
+    }
+
+    public set forceToken(value: ITokenCard | null) {
```
AMMayberry1, line 50: This should be handled the same way as the Leader card where it's a private set that's accessed via a method with some sanity checks

server/game/core/zone/BaseZone.ts
```diff
@@ -10,19 +12,23 @@ import { ZoneAbstract } from './ZoneAbstract';
 
 export interface IBaseZoneState extends IGameObjectBaseState {
     leader: GameObjectRef<ILeaderCard> | null;
+    forceToken: GameObjectRef<ITokenCard> | null;    // Use abstraction or concrete type?
 }
 
+type BaseZoneCard = ILeaderCard | IBaseCard | TokenCard;
```
AMMayberry1, line 18: nit: we always start interface type names with an I ("IBaseZoneCard")

server/game/gameSystems/CreateForceTokenSystem.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+
+        event.generatedTokens = [];
+
+        for (const player of Helpers.asArray(properties.target)) {
+            if (player.hasTheForce()) {
+                continue;
+            }
+            event.generatedTokens.push(context.game.generateToken(player, TokenCardName.Force));
```
AMMayberry1, line 38: I kind of feel like maybe the Force token should be handled differently since it's guaranteed to be a one-of. Like maybe we generate one for each player at the beginning of the game and it lives in the OutsideTheGameZone. We just need to adjust the token cleanup logic to not delete it automatically like we do with unit / upgrade tokens. Honestly, since it literally never goes anywhere other than base zone we could just "stash" it here by having a boolean flag controlling whether it's visible or not.

The advantage of enforcing it as a singleton like this is we never run the risk of accidentally having multiples floating around for a given player somehow

server/game/core/Player.ts
```diff
@@ -223,6 +223,10 @@ export class Player extends GameObject<IPlayerState> {
         return this.game.allArenas.hasSomeCard({ ...filter, type: WildcardCardType.Upgrade, controller: this });
     }
 
+    public hasTheForce(): boolean {
```
AMMayberry1, line 226: This could also be a property getter (`public get hasTheForce()`), depending on your preference. Some other similar stuff is already set that way

server/game/core/card/TokenCards.ts
```diff
@@ -26,3 +28,7 @@ export class TokenUpgradeCard extends TokenUpgradeParent implements ITokenUpgrad
         return true;
     }
 }
+
+export class TokenCard extends TokenCardParent implements ITokenCard {
```
AMMayberry1, line 32: We can just replace TokenCardParent with TokenCard completely, it only served to be a concrete parent for the other two to derive from

server/game/core/utils/EnumHelpers.ts
```diff
@@ -164,6 +164,7 @@ export const isToken = (cardType: CardTypeFilter) => {
         case WildcardCardType.Token:
         case CardType.TokenUpgrade:
         case CardType.TokenUnit:
+        case CardType.TokenCard:
```
AMMayberry1, line 167: Haven't checked myself yet, but please do a careful pass through the CardType-related methods in here and make sure they all correctly handle this new type

server/game/gameSystems/CreateForceTokenSystem.ts
```diff
@@ -0,0 +1,62 @@
...
+    protected override eventName = EventName.OnTokensCreated;
+
+    // event handler doesn't do anything since the token is generated in updateEvent
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public override eventHandler(event, additionalProperties = {}): void { }
+
+    public override getEffectMessage(context: TContext): [string, any[]] {
+        return ['the force is with {0}', [context.player]];
+    }
+
```
AMMayberry1, line 22: Need to add a `canAffectInternal` here to prevent gaining the Force again when you already have it. It would use the same rule regarding `mustChangeGameState` as the note I left on the UseTheForceSystem

server/game/gameSystems/CreateForceTokenSystem.ts
AMMayberry1: Actually on second thought, just leaving it in the base zone and hiding it probably wouldn't work since the card itself still needs to report as being in a specific zone

server/game/gameSystems/UseTheForceSystem.ts
moyerr: I see many other systems do that check with `(properties.isCost || mustChangeGameState !== GameStateChangeRequired.None)`, should that be the case here too, since this is almost always a cost?

server/game/gameSystems/CreateForceTokenSystem.ts
moyerr: So just to clarify, no changes needed here?

server/game/core/card/LOFCommonBase.ts
```diff
@@ -0,0 +1,16 @@
+import AbilityHelper from '../../AbilityHelper';
+import { BaseCard } from './BaseCard';
+import { Trait } from '../Constants';
+
+export class LOFCommonBase extends BaseCard {
```
moyerr, line 5: Open to feedback on this class's name. Also had to put it outside of the `server/game/cards/` directory because Index.ts was picking it up and throwing an error for lack of `id`

moyerr:
This is blocked from merging until the the new cards are added to the SWU database.

server/game/gameSystems/UseTheForceSystem.ts
AMMayberry1: Oh yes good call, that's correct

server/game/gameSystems/CreateForceTokenSystem.ts
AMMayberry1: Oh sorry - no so I do think it would be good to make this a singleton and not generate a new one each time, I'm just not 100% sure what the best pattern for moving it around is. Feels like just letting live in OutsideTheGameZone is fine, though like I mentioned we'll need to exempt it from the end-of-round token cleanup logic

server/game/core/card/TokenCards.ts
AMMayberry1: Ah sorry, I didn't look close enough and forgot the pattern that we use for tokens. I think what you had originally is actually pretty much correct, my apologies. The one change would be that I think it should be `AsToken(Card)`, since it doesn't need any of the methods on `InPlayCard`. Lmk if that breaks something in the inheritance tree though, if it gets painful then it's not worth too much effort.

server/game/gameSystems/CreateForceTokenSystem.ts
moyerr: Do you think implementing triggered abilities for the Force (e.g. "When you gain the Force", "When you use the force") would be the same complexity with the singleton pattern, versus standard token creation?

AMMayberry1: I think they'd be even simpler than they are now since they're just moving a card around instead of dealing with token generation

AMMayberry1:
Just remembered that we need to add a boolean to `Player.getStateSummary()` to indicate whether the player has their force token or not. This is what gets sent to the FE so it can render the token. Make sure to merge main first, I just added another property there as well - the player aspects, which are needed for deciding whether to render the light side or dark side token

server/game/core/card/TokenCards.ts
moyerr: Yep, looks like it may be a bit more of a lift to use `Card` instead of `InPlayCard`, so gonna keep typing the same for now

server/game/gameSystems/CreateForceTokenSystem.ts
moyerr: Done! Lemme know if you want me to adjust anything about that specific implementation strategy

server/game/core/card/LOFCommonBase.ts
AMMayberry1: Name should probably be like "ForceBase" or something, just to be specific. 

I would also like to put it somewhere different, `core/` is really only for the very fundamental stuff.

Maybe we could set up a convention where we have a folder named `cards/05_LOF/common/` or something like that and then we just updated `cards/Index.ts` to filter out anything in a folder named `common/`. I imagine we'll need stuff like this again on occasion, so it'll be good to have a pattern for it

AMMayberry1, line 5: Let's make the class abstract

server/game/gameSystems/CreateForceTokenSystem.ts
```diff
@@ -12,9 +11,19 @@ export class CreateForceTokenSystem<TContext extends AbilityContext = AbilityCon
...
-    public override eventHandler(event, additionalProperties = {}): void { }
+    public override eventHandler(event, additionalProperties = {}): void {
+        const context = event.context;
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+
+        for (const player of Helpers.asArray(properties.target)) {
+            const forceTokens = player.outsideTheGameZone
+                .getCards({ condition: (card) => card.isForceToken() });
+
+            if (forceTokens.length > 0) {
```
AMMayberry1, line 22: This should probably be a contract check, if we get here and don't find a force token there's a problem

server/game/gameSystems/CreateForceTokenSystem.ts
```diff
@@ -24,39 +33,17 @@ export class CreateForceTokenSystem<TContext extends AbilityContext = AbilityCon
...
 
-            return events;
-        });
+        event.tokenType = TokenCardName.Force;
     }
 
-    protected override addPropertiesToEvent(event: any, player: Player, context: TContext, additionalProperties?: {}): void {
-        super.addPropertiesToEvent(event, player, context, additionalProperties);
+    public override canAffectInternal(player: Player, context: TContext, additionalProperties: any = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        if (mustChangeGameState !== GameStateChangeRequired.None && context.player.hasTheForce) {
```
AMMayberry1, line 43: We'll need to do the `isCost` check here as well

server/game/gameSystems/UseTheForceSystem.ts
```diff
@@ -11,13 +12,15 @@ export class UseTheForceSystem<TContext extends AbilityContext = AbilityContext,
     public override eventHandler(event): void {
         const forceToken = event.context.player.baseZone.forceToken;
 
-        if (forceToken) {
-            forceToken.moveTo(ZoneName.OutsideTheGame);
-        }
+        Contract.assertNotNullLike(forceToken, 'Force token should not be null for this player.');
```
AMMayberry1, line 15: nit: we should include the player id and / or name in the error output

server/utils/deck/DeckValidator.ts
```diff
@@ -12,7 +12,8 @@ enum SwuSet {
     SOR = 'sor',
     SHD = 'shd',
     TWI = 'twi',
-    JTL = 'jtl'
+    JTL = 'jtl',
+    LOF = 'lof',
```
AMMayberry1, line 16: Need to make a corresponding change in `fetchdata.js` for pulling down the card data

server/game/cards/05_LOF/bases/JediTemple.ts
```diff
@@ -0,0 +1,10 @@
+import { LOFCommonBase } from '../../../core/card/LOFCommonBase';
+
+export default class JediTemple extends LOFCommonBase {
```
AMMayberry1, line 3: We should probably leave a comment note on all of these that the tests are in the LOFCommonBase tests

test/server/cards/05_LOF/bases/LOFCommonBase.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+                context.player2.passAction();
+
+                context.player1.clickCard(context.countDooku);
+                context.player1.clickCard(context.p2Base);
+
+                // Only Player 1 should have The Force token
+                expect(context.player1.hasTheForce).toBe(true);
+                expect(context.player2.hasTheForce).toBe(false);
+            });
```
AMMayberry1, line 102: Kind of an edge case, but let's add one more to vaidate that it doesn't generate the token when a unit with an attached Force pilot (unit or leader) attacks

test/server/gameSystems/TheForce.spec.ts
```diff
@@ -1,44 +0,0 @@
-describe('The Force', function() {
```
AMMayberry1: It would actually be good to keep this in just to check some basic stuff - right now mostly to confirm that the token doesn't get cleaned up between phases if it's in the OutsideTheGameZone

test/server/cards/05_LOF/leaders/DarthMaulSithRevealed.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+                        ]
+                    },
+                    player2: {
+                        groundArena: ['battlefield-marine', 'consular-security-force'],
+                    }
+                });
+
+                const { context } = contextRef;
+
+                // Attack with the Guardian of the Whills to gain the Force
```
AMMayberry1, line 22: Since it _should_ be pretty easy, would you mind extending the test infrastructure so that it's possible to just declare the Force token for a player in the test setup?

Make sure to merge main before you do, we just did a big change to some of the interfaces. Also if it ends up being a headache then don't worry about it for now, we can push it off to a future PR.

AMMayberry1: Err actually we haven't merged the relevant change yet I just realized, so hold off on this. It should be in by tomorrow.

AMMayberry1 (dismissed):

AMMayberry1 (approved):

PR #1068, opened 2025-04-11
Fix Relentless blanking smuggle keyword in resources

### LINKED ISSUE #1039: Relentless prevents events with Smuggle from being playable
Report from Trys2Hard:

Opponent played relentless, I wanted to play timely intervention from resources for 2 to do nothing and then play vanquish but it wouldn't let me smuggle the timely.

Image

Image

Frugghi:
Fixes #1039

### DIFF
--- a/server/game/core/ability/KeywordInstance.ts
+++ b/server/game/core/ability/KeywordInstance.ts
@@ -1,7 +1,8 @@
 import type { IAbilityPropsWithType, ITriggeredAbilityBaseProps } from '../../Interfaces';
 import type { Card } from '../card/Card';
-import { EffectName, type Aspect, type KeywordName } from '../Constants';
+import { EffectName, RelativePlayer, type Aspect, type KeywordName } from '../Constants';
 import * as Contract from '../utils/Contract';
+import * as EnumHelpers from '../utils/EnumHelpers';
 
 export class KeywordInstance {
     public readonly name: KeywordName;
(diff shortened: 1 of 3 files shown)


server/game/core/ability/KeywordInstance.ts
```diff
@@ -1,15 +1,20 @@
...
+import * as EnumHelpers from '../utils/EnumHelpers';
 
 export class KeywordInstance {
     public readonly name: KeywordName;
 
     private readonly card: Card;
 
     public get isBlank() {
-        if (this.card.hasOngoingEffect(EffectName.Blank)) {
+        if (EnumHelpers.isHiddenFromOpponent(this.card.zoneName, RelativePlayer.Self) && this.hasCostValue()) {
```
AMMayberry1, line 13: What exactly did the issue end up being? This fix feels like a bit of a band-aid, I'm wondering if we could do a more holistic fix (potentially in a later PR if needed)

Frugghi: Yes, I agree that this feels a band-aid fix but I wanted to keep it simple because it's a fix for just one card.

The issue is that Relentless blanks events in any zone, so events in your deck, hand, resources, etc. will be blank before you play them, therefore they lose the smuggle keyword. The correct way to do this would be to blank the card only after it is played/costs are paid but I didn't find an obvious/simple way to do that because of the logic to resolve early targets.

In general I think it's safe to say that a keyword can't be blanked if the card is in a hidden zone but, for example, this won't work if in the future they introduce a "play from discard" keyword.

PR #1069, opened 2025-04-12
Fix playing events owned by other players

### LINKED ISSUE #1051: Cards stolen with DJ can't be played using Tech
See below, from the report I expect there's some contract check failing (or some other exception) which is blocking the play from happening after the pay step has already resolved.

Player report (unverified):

playing on PC. Having issues when playing events from my resource row with Tech that I have stolen from my opponent with DJ. Resources get spent, card gets moved to opponents discard, but effect doesn't happen.

Frugghi:
Fixes #1051

### DIFF
--- a/server/game/actions/PlayEventAction.ts
+++ b/server/game/actions/PlayEventAction.ts
@@ -18,7 +18,7 @@ export class PlayEventAction extends PlayCardAction {
             this.copyContextTargets(context, abilityContext);
         }
 
-        context.game.queueStep(new AbilityResolver(context.game, abilityContext, false, null, this.earlyTargetResults));
+        context.game.queueStep(new AbilityResolver(context.game, abilityContext, false, null, this.earlyTargetResults, ['player']));
     }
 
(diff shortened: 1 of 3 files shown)


server/game/actions/PlayEventAction.ts
```diff
@@ -18,7 +18,7 @@ export class PlayEventAction extends PlayCardAction {
             this.copyContextTargets(context, abilityContext);
         }
 
-        context.game.queueStep(new AbilityResolver(context.game, abilityContext, false, null, this.earlyTargetResults));
+        context.game.queueStep(new AbilityResolver(context.game, abilityContext, false, null, this.earlyTargetResults, ['player']));
```
AMMayberry1, line 21: Do you have a sense of why this wasn't needed for the other card types?

Frugghi: Because events are moved to the discard pile before resolving the ability so when the ability resolver runs the controller of the event is the owner and not the current player.

PR #1095, opened 2025-04-15
Bugfix: Hello There targets cards that entered play

### LINKED ISSUE #1088: Hello There doesn't work on leaders deployed this phase
It's currently using the card played watcher but we should be using the entered play watcher

moyerr:
Fixes #1088

### Testing Notes

Added the following test cases:
 - ✅  It can target a leader deployed this phase
 - ✅ It can target a token unit that entered play this phase
 - ✅ It can target a unit that was rescued from capture this phase
 - ✅ It cannot target leaders, tokens, or units that entered play the previous phase

### DIFF
--- a/server/game/stateWatchers/CardsEnteredPlayThisPhaseWatcher.ts
+++ b/server/game/stateWatchers/CardsEnteredPlayThisPhaseWatcher.ts
@@ -36,6 +36,11 @@ export class CardsEnteredPlayThisPhaseWatcher extends StateWatcher<EnteredCardEn
             .map((entry) => entry.card);
     }
 
+    /** Check the list of cards that entered play in the state if we found cards that match filters */
+    public someCardPlayed(filter: (entry: EnteredCardEntry) => boolean): boolean {
+        return this.getCardsPlayed(filter).length > 0;
+    }
+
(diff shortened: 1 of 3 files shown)


server/game/stateWatchers/CardsEnteredPlayThisPhaseWatcher.ts
```diff
@@ -36,6 +36,11 @@ export class CardsEnteredPlayThisPhaseWatcher extends StateWatcher<EnteredCardEn
             .map((entry) => entry.card);
     }
 
+    /** Check the list of cards that entered play in the state if we found cards that match filters */
+    public someCardPlayed(filter: (entry: EnteredCardEntry) => boolean): boolean {
```
AMMayberry1, line 40: Just caught this - I think naming the above method `getCardsPlayed` was a copy-paste error, could we rename these methods to be `getCardsEnteredPlay` and `someCardEnteredPlay`?

moyerr: Oh yeah, 💯 good catch

AMMayberry1 (approved):

PR #1098, opened 2025-04-15
#resolved issue where undeployed leaders weren't getting saved

CheBato:
#resolved issue where undeployed leaders weren't getting saved

### DIFF
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -1153,7 +1153,7 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
             }
             const currentCardState = this.getCardState();
             if (this.isLeader() && !currentCardState.deployed) {
-                return this.internalName;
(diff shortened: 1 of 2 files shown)


server/game/core/card/Card.ts
```diff
@@ -1152,7 +1152,9 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
                 return null;
             }
             const currentCardState = this.getCardState();
-
+            if (this.isLeader() && !currentCardState.deployed) {
```
AMMayberry1, line 1155: Needs to go in the leader card implementations, `LeaderUnitCard` and `DoubleSidedLeaderCard`

CheBato: oh so we do a property like deployed/undeployed?

CheBato:
Check why initiative is setting true for both players.

AMMayberry1:
Would also be good to return exhausted resource state if possible

PR #1102, opened 2025-04-17
Prevent willDefeat triggers if event has been replaced

### LINKED ISSUE #1096: Shoot Down second part ability trigger if shield is destroyed
**Describe the bug**
Shoot Down deal 2 damage if shield was defeat but it should not

**To Reproduce**
Add this tests : 
```
  it('should not deal damage to base if shield was defeat', function () {

                context.player1.passAction();
                context.player2.clickCard(context.momentOfPeace);
                context.player2.clickCard(context.tielnFighter);

                context.player1.clickCard(context.shootDown);
                context.player1.clickCard(context.tielnFighter);

            });
```

Frugghi:
Fixes #1096

### DIFF
--- a/server/game/core/event/GameEvent.ts
+++ b/server/game/core/event/GameEvent.ts
@@ -194,6 +194,15 @@ export class GameEvent {
         return undefined;
     }
 
+    /**
+     * Returns the event or its replacement effect events that are resolved.
+     *
+     * If the event is not resolved, it returns an empty array.
+     * If the event is resolved, it returns only itself.
+     * If the event has been replaced, it returns all replacement effect events that are resolved.
+     *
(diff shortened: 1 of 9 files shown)


server/game/core/event/GameEvent.ts
```diff
@@ -193,4 +193,16 @@ export class GameEvent {
         }
         return undefined;
     }
+
+    public get resolvedEvents(): GameEvent[] {
```
AMMayberry1, line 197: Could we add a clarifying jsdoc comment to this and `AbilityContext.resolvedEvents` explaining that it is returning the events that actually resolved, which might mean it's returning the replacement effect. And / or rename this to something longer like `getEventsOrReplacedEvents()`

AMMayberry1 (approved):

PR #1116, opened 2025-04-23
Legends of the Force | Drain Essence & Darth Sidious

moyerr:
darth-maul-deck-cards

### DIFF
--- a/server/game/cards/05_LOF/events/DrainEssence.ts
+++ b/server/game/cards/05_LOF/events/DrainEssence.ts
@@ -1,0 +1,28 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { WildcardCardType, WildcardRelativePlayer, WildcardZoneName } from '../../../core/Constants';
+
+export default class DrainEssence extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: 'temp-drain-essence-id',
+            internalName: 'drain-essence'
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Deal 2 damage to a unit. The Force is with you.',
+            immediateEffect: AbilityHelper.immediateEffects.simultaneous([
+                AbilityHelper.immediateEffects.selectCard({
+                    activePromptTitle: 'Select a unit to deal 2 damage to',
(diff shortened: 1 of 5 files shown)


server/game/cards/05_LOF/events/DrainEssence.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Deal 2 damage to a unit. The Force is with you.',
+            immediateEffect: AbilityHelper.immediateEffects.simultaneous([
+                AbilityHelper.immediateEffects.selectCard({
+                    activePromptTitle: 'Select a unit to deal 2 damage to',
+                    controller: WildcardRelativePlayer.Any,
+                    zoneFilter: WildcardZoneName.AnyArena,
```
AMMayberry1, line 20: nit: I think these are the defaults for controller and zoneFilter

server/game/cards/05_LOF/units/DarthSidiousThePhantomMenace.ts
```diff
@@ -0,0 +1,30 @@
...
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Use The Force',
+            optional: true,
+            immediateEffect: AbilityHelper.immediateEffects.useTheForce(),
+            ifYouDo: {
+                title: 'Defeat each non-Sith unit with 3 or less remaining HP',
+                immediateEffect: AbilityHelper.immediateEffects.defeat((context) => ({
+                    target: context.game.getArenaUnits({
+                        condition: (card) => card.isUnit() &&
```
AMMayberry1, line 22: nit: is the `isUnit()` needed here? I thought we had the types set up so that `condition` would accept an `IUnit`

moyerr: I do get a type error here when i try to remove it, so it does seem this check helps solve the type puzzle

> Property 'remainingHp' does not exist on type 'Card<ICardState>'

AMMayberry1: Ah okay, yeah looks like I didn't quite finish up with that. I'll tackle it later, so this is good for now. Thanks!

AMMayberry1 (approved):

PR #1126, opened 2025-04-27
Enforce upgrades attachment condition

### LINKED ISSUE #994: Enforce attachment restrictions for effects like Survivors' Gauntlet
FFG issued a late rules update saying that attachment restrictions must now be respected when moving attachments. This includes restrictions on e.g. Vehicle or not, as well as pilot count.

Frugghi:
Fixes #994

### DIFF
--- a/test/server/core/abilities/keyword/Piloting.spec.ts
+++ b/test/server/core/abilities/keyword/Piloting.spec.ts
@@ -261,13 +261,14 @@ describe('Piloting keyword', function() {
                 expect(context.idenVersio).toBeInZone('discard');
             });
 
-            it('can be moved to another vehicle with a Pilot ignoring the limit', async function () {
+            it('cannot be moved to another vehicle with a Pilot ignoring the limit', async function () {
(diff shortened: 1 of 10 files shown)


test/server/core/abilities/keyword/Piloting.spec.ts
```diff
@@ -261,13 +261,14 @@ describe('Piloting keyword', function() {
                 expect(context.idenVersio).toBeInZone('discard');
             });
 
-            it('can be moved to another vehicle with a Pilot ignoring the limit', async function () {
+            it('cannot be moved to another vehicle with a Pilot ignoring the limit', async function () {
```
AMMayberry1, line 264: Let's also maybe add some tests in the Upgrade.spec.ts suite with Survivors' Gauntlet and / or Evidence of the Crime to confirm that we can't move upgrades with specific restrictions around vehicles, leaders, etc

AMMayberry1 (approved):

PR #1129, opened 2025-04-30
Fix Satine Hp calculation and several cards that could be incorrectly triggered when playing a pilot

adbronico:


server/game/cards/02_SHD/leaders/BobaFettDaimyo.ts
```diff
@@ -16,7 +16,7 @@ export default class BobaFettDaimyo extends LeaderUnitCard {
             optional: true,
             when: {
                 onCardPlayed: (event, context) =>
-                    event.card.isUnit() && event.player === context.player && event.card.keywords.length > 0
+                    event.card.isUnit() && event.playType !== PlayType.Piloting && event.player === context.player && event.card.keywords.length > 0
```
AMMayberry1, line 19: There's now an `event.cardTypeWhenInPlay` property, let's start switching to using that instead of `event.card.isUnit() && event.playType !== PlayType.Piloting` since it captures everything

server/game/cards/03_TWI/units/SatineKryzeCommittedToPeace.ts
```diff
@@ -20,7 +20,7 @@ export default class SatineKryzeCommittedToPeace extends NonLeaderUnitCard {
                 title: 'Discard cards from an opponent\'s deck equal to half this unit\'s remaining HP, rounded up',
                 cost: AbilityHelper.costs.exhaustSelf,
                 immediateEffect: AbilityHelper.immediateEffects.discardFromDeck((context) => {
-                    const remainingHp = context.source.isUnit() ? context.source.getHp() : 0;
+                    const remainingHp = context.source.isUnit() ? (context.source.getHp() - context.source.damage) : 0;
```
AMMayberry1, line 23: nit: simpler to use `context.source.remainingHp`

AMMayberry1 (approved):

PR #1135, opened 2025-05-01
Add Third Sister leader

Frugghi:
| --- | ---- |
|  |  |


server/game/cards/05_LOF/leaders/ThirdSisterSeethingWithAmbition.ts
```diff
@@ -0,0 +1,59 @@
...
+        this.addOnAttackAbility({
+            title: 'The next unit you play this phase gains Hidden',
+            immediateEffect: AbilityHelper.immediateEffects.delayedPlayerEffect({
+                title: 'The next unit you play this phase gains Hidden',
+                when: {
+                    onCardPlayed: this.isUnitPlayedEvent,
+                },
+                duration: Duration.UntilEndOfPhase,
+                limit: AbilityLimit.perGame(1),
+                immediateEffect: AbilityHelper.immediateEffects.forThisPhaseCardEffect((context) => ({
```
Frugghi, line 46: I realized this is wrong, the on attack ability gives Hidden beyond the current phase which can be important for cards that check the existence of a keyword. Will fix it later!

server/game/core/ongoingEffect/effectImpl/DynamicOngoingEffectImpl.ts
```diff
@@ -19,9 +19,9 @@ export default class DynamicOngoingEffectImpl<TValue> extends StaticOngoingEffec
     }
 
     public override apply(target) {
-        // TODO: these two calls were in the reverse order in l5r, not sure if that was required for some reason
-        this.recalculate(target);
         super.apply(target);
+
+        this.recalculate(target);
```
AMMayberry1, line 24: I guess it was required 😅 what was the issue here?

test/server/core/abilities/keyword/Hidden.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+
+            context.player1.clickCard(context.anakinSkywalker);
+
+            context.player2.clickCard(context.battlefieldMarine);
+            expect(context.player2).toBeAbleToSelectExactly([context.wampa, context.p1Base]);
+            context.player2.clickPrompt('Cancel');
+
+            context.player2.clickCard(context.sanctionersShuttle);
+            context.player2.clickCard(context.anakinSkywalker);
+            expect(context.anakinSkywalker).toBeCapturedBy(context.sanctionersShuttle);
```
AMMayberry1, line 231: nit: we do now have capture support in test setups finally. e.g.: `{ card: 'cartel-spacer', capturedUnits: ['millennium-falcon#piece-of-junk'] }`

test/server/cards/05_LOF/leaders/ThirdSisterSeethingWithAmbition.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.idenVersio]);
+                context.player1.clickCard(context.idenVersio);
+
+                expect(context.idenVersio).toBeInZone('groundArena');
+                expect(context.idenVersio.hasSomeKeyword('hidden')).toBeTrue();
+                expect(context.player1.exhaustedResourceCount).toBe(4);
+
+                context.player2.clickCard(context.wampa);
+                expect(context.player2).toBeAbleToSelectExactly([context.p1Base]);
+                context.player2.clickCard(context.p1Base);
```
AMMayberry1, line 31: Can we play a second unit and make sure it does not gain hidden (unless that's covered and I missed it)

server/game/core/ongoingEffect/effectImpl/DynamicOngoingEffectImpl.ts
Frugghi: The issue was that if give `Hidden` to a unit already in play using a dynamic effect (e.g. The Ghost), if you call `recalculate` before `apply`, the "while in play" abilities will be refreshed before target gains the keyword so the ability for `Hidden` won't be added.

server/game/core/ongoingEffect/OngoingEffectEngine.ts
```diff
@@ -60,7 +60,7 @@ export class OngoingEffectEngine {
         }
         const effectTriggers = effectsToTrigger.map((effect) => {
             const properties = effect.impl.getValue();
-            const context = effect.context;
+            const context = effect.context.createCopy({ events });
```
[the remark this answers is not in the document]
Frugghi: This is to allow to access the events from Third Sister because she needs to know which unit was played! `effect.context` is a generic context that doesn't contain the events that triggered the effect.

server/game/gameSystems/DeployLeaderSystem.ts
```diff
@@ -36,6 +36,7 @@ export class DeployLeaderSystem<TContext extends AbilityContext = AbilityContext
 
     protected override updateEvent(event, card: Card, context: TContext, additionalProperties: any = {}) {
         super.updateEvent(event, card, context, additionalProperties);
+        event.type = DeployType.LeaderUnit;
```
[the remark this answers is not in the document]
Frugghi: This isn't actually needed, I've added it just because `DeployAndAttachPilotLeaderSystem` sets it and so that the state watcher could be used to differentiate between a deploy as a unit or as an upgrade. It doesn't matter at all for now so I can also remove it.

test/server/core/abilities/keyword/Hidden.spec.ts
Frugghi: Oh yeah but here I want to play Anakin first to check that it can't be attack, then capture it and rescue it in the same phase to ensure that it can be attacked because it's a new copy.

PR #1139, opened 2025-05-02
savage opress

VJubert:

### DIFF
--- a/test/server/cards/05_LOF/units/SavageOpressImbuedWithHate.spec.ts
+++ b/test/server/cards/05_LOF/units/SavageOpressImbuedWithHate.spec.ts
@@ -1,0 +1,117 @@
+describe('Savage Opress, Imbued with Hate', () => {
+    integration(function (contextRef) {
+        describe('Savage Opress\' when played ability', () => {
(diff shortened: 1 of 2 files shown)


test/server/cards/05_LOF/units/SavageOpressImbuedWithHate.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+                        groundArena: ['savage-opress#imbued-with-hate'],
+                        hasForceToken: true,
+                        base: { card: 'echo-base', damage: 0 }
+                    },
+                    player2: {
+                        hand: ['rivals-fall'],
+                        hasInitiative: true,
+                    }
+                });
+            });
```
moyerr, line 70: We may want a case that uses No Glory to ensure the ability resolves for the opponent

AMMayberry1: +1. We're doing this as standard practice for all "when defeated" abilities now

AMMayberry1 (approved):

PR #1141, opened 2025-05-02
LOF | Darth Maul's Lightsaber

moyerr:
image

### DIFF
--- a/server/game/cards/05_LOF/upgrades/DarthMaulsLightsaber.ts
+++ b/server/game/cards/05_LOF/upgrades/DarthMaulsLightsaber.ts
@@ -1,0 +1,31 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { UpgradeCard } from '../../../core/card/UpgradeCard';
+import { KeywordName, Trait } from '../../../core/Constants';
+
+export default class DarthMaulsLightsaber extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '2167393423',
+            internalName: 'darth-mauls-lightsaber',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card) => !card.hasSomeTrait(Trait.Vehicle));
+
+        this.addWhenPlayedAbility({
+            title: 'Attack with Darth Maul. For this attack, he gains overwhelm and can\'t attack bases.',
+            optional: true,
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => context.source.parentCard?.title === 'Darth Maul',
+                onTrue: AbilityHelper.immediateEffects.attack((attackCtx) => ({
+                    target: attackCtx.source.parentCard,
+                    targetCondition: (target) => target.isUnit(),
+                    attackerLastingEffects: {
+                        effect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Overwhelm)
+                    }
+                }))
+            })
+        });
+    }
+}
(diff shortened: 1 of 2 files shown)


server/game/cards/05_LOF/upgrades/DarthMaulsLightsaber.ts
```diff
@@ -0,0 +1,32 @@
...
+
+        this.addWhenPlayedAbility({
+            title: 'Attack with Darth Maul. For this attack, he gains overwhelm and can\'t attack bases.',
+            optional: true,
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => context.source.parentCard?.title === 'Darth Maul',
+                onTrue: AbilityHelper.immediateEffects.attack((attackCtx) => ({
+                    target: attackCtx.source.parentCard,
+                    targetCondition: (target) => target.isUnit(),
+                    optional: false,
```
AMMayberry1, line 24: nit: this is the default I believe

server/game/cards/05_LOF/upgrades/DarthMaulsLightsaber.ts
```diff
@@ -0,0 +1,32 @@
...
+export default class DarthMaulsLightsaber extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '2167393423',
+            internalName: 'darth-mauls-lightsaber',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setAttachCondition((card) => !card.hasSomeTrait(Trait.Vehicle));
```
AMMayberry1, line 14: There's an additional restriction that it has to be a friendly unit

test/server/cards/05_LOF/upgrades/DarthMaulsLightsaber.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+                    maulUnit,
+                    context.cellBlockGuard,
+                    context.phaseiiiDarkTrooper
+                ]);
+
+                context.player1.clickCard(maulLeader);
+                expect(maulLeader).toHaveExactUpgradeNames(['darth-mauls-lightsaber']);
+
+                // No ability is triggered
+                expect(context.player1).not.toHavePassAbilityPrompt(prompt);
```
AMMayberry1, line 226: I'm a bit lost - why is the ability not triggered in these tests?

test/server/cards/05_LOF/upgrades/DarthMaulsLightsaber.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+                    context.phaseiiiDarkTrooper
+                ]);
+
+                context.player1.clickCard(maulUnit);
+                expect(maulUnit).toHaveExactUpgradeNames(['darth-mauls-lightsaber']);
+
+                // No ability is triggered
+                expect(context.player1).not.toHavePassAbilityPrompt(prompt);
+                expect(context.player2).toBeActivePlayer();
+            });
```
AMMayberry1, line 251: We should test that attaching to a non-Maul unit doesn't trigger the ability (unless it's already there and I missed it)

server/game/cards/05_LOF/upgrades/DarthMaulsLightsaber.ts
moyerr: It actually inherits the optionality from the enclosing context, which in this case results in a double-optional ability. It will ask to trigger the when-played, then ask again if you want to initiate the attack. Felt redundant and weird, so I explicitly made the attack non-optional if you've already opted into the triggered ability.

server/game/cards/05_LOF/upgrades/DarthMaulsLightsaber.ts
moyerr: Oh good call!

test/server/cards/05_LOF/upgrades/DarthMaulsLightsaber.spec.ts
moyerr: Because these cards are named "Maul", not ""Darth Maul"

test/server/cards/05_LOF/upgrades/DarthMaulsLightsaber.spec.ts
moyerr: Yeah, i'll add one more. Thanks!

PR #1152, opened 2025-05-05
Implement Obi-Wan Kenobi, Courage Makes Heroes

moyerr:
| Leader Side | Unit Side |
| :---: | :---: |
| image | image |


server/game/cards/05_LOF/leaders/ObiWanKenobiCourageMakesHeroes.ts
```diff
@@ -0,0 +1,49 @@
...
+            cost: [
+                AbilityHelper.costs.exhaustSelf(),
+                AbilityHelper.costs.useTheForce()
+            ],
+            targetResolver: {
+                activePromptTitle: 'Choose unit to give Experience token to',
+                mode: TargetMode.Single,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: WildcardZoneName.AnyArena,
+                controller: WildcardRelativePlayer.Any,
```
AMMayberry1, line 25: nit: I think you actually don't need any of these, they're either defaults or they'll be handled automatically by the validation checks for `giveExperience()`. Same note for the deployed side

test/server/cards/05_LOF/leaders/ObiWanKenobiCourageMakesHeroes.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+                        groundArena: [
+                            'battlefield-marine',
+                        ]
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.obiwanKenobi);
+                expect(context.player1).toHaveExactPromptButtons([
```
AMMayberry1, line 22: Just a note for the future, it can be convenient to set the resource count below the deploy threshold so you don't have to deal with this

server/game/cards/05_LOF/leaders/ObiWanKenobiCourageMakesHeroes.ts
```diff
@@ -0,0 +1,49 @@
...
+                cardCondition: (card) => card.isUnit() &&
+                  !card.upgrades.some((upgrade) => upgrade.title === 'Experience'),
```
AMMayberry1, line 27: We have an `isShield()` property on Shield.ts, let's add a similar one for Experience (and on Card.ts by extension). I'm honestly surprised it hasn't come up before

test/server/cards/05_LOF/leaders/ObiWanKenobiCourageMakesHeroes.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine]);
+                context.player1.clickCard(context.battlefieldMarine);
+
+                expect(context.obiwanKenobi.exhausted).toBeTrue();
+                expect(context.player1.hasTheForce).toBeFalse();
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['experience']);
+            });
+
+            it('does not allow giving an Experience token to a unit that already has one', async function () {
+                await contextRef.setupTestAsync({
```
AMMayberry1, line 38: nit: this is personal preference and no need to change the current file, but also feel free to have one big shared test setup in a `beforeEach()` that covers several cases in one setup (the `it()` blocks would still be separated). Especially for little stuff like this it can save time and make the file easier to read

server/game/cards/05_LOF/leaders/ObiWanKenobiCourageMakesHeroes.ts
```diff
@@ -0,0 +1,49 @@
...
+                AbilityHelper.costs.exhaustSelf(),
+                AbilityHelper.costs.useTheForce()
+            ],
+            targetResolver: {
+                activePromptTitle: 'Choose unit to give Experience token to',
+                mode: TargetMode.Single,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: WildcardZoneName.AnyArena,
+                controller: WildcardRelativePlayer.Any,
+                cardCondition: (card) => card.isUnit() &&
```
AMMayberry1, line 26: nit: isUnit() check shouldn't be needed, since only units are legal targets for giveExperience(). Same on the deployed side

test/server/cards/05_LOF/leaders/ObiWanKenobiCourageMakesHeroes.spec.ts
moyerr: Oh that's a great idea 💡

server/game/cards/05_LOF/leaders/ObiWanKenobiCourageMakesHeroes.ts
moyerr: This is one of those situations where `isUnit()` is helping to solve the type puzzle. Without it, `upgrades` (or the newly added `hasExperience()`) is not available because it's not guaranteed to be a unit card.

AMMayberry1 (approved):

PR #1156, opened 2025-05-05
Darth Revan leader

AMMayberry1:
| Leader Side | Unit Side |
| :---: | :---: |
| image | image |

### DIFF
--- a/server/game/cards/02_SHD/events/TimelyIntervention.ts
+++ b/server/game/cards/02_SHD/events/TimelyIntervention.ts
@@ -13,6 +13,7 @@ export default class TimelyIntervention extends EventCard {
     public override setupCardAbilities() {
         this.setEventAbility({
             title: 'Play a unit from your hand. Give it ambush for this phase',
+            cannotTargetFirst: true,
             targetResolver: {
                 cardTypeFilter: CardType.BasicUnit,
                 controller: RelativePlayer.Self,
@@ -22,7 +23,7 @@ export default class TimelyIntervention extends EventCard {
                     AbilityHelper.immediateEffects.forThisPhaseCardEffect({
                         effect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Ambush)
                     }),
-                ])
+                ], false, true)
             }
         });
     }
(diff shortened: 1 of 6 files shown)


test/server/cards/05_LOF/leaders/DarthRevanScourgeOfTheOldRepublic.spec.ts
```diff
@@ -0,0 +1,384 @@
+describe('Darth Revan, Scourge of the Old Republic', function () {
```
Frugghi, line 1: Should we have a test with Maul defeating two units at the same time?

server/game/cards/05_LOF/leaders/DarthRevanScourgeOfTheOldRepublic.ts
```diff
@@ -0,0 +1,47 @@
...
+            internalName: 'darth-revan#scourge-of-the-old-republic',
+        };
+    }
+
+    protected override setupLeaderSideAbilities() {
+        this.addTriggeredAbility({
+            title: 'Exhaust this leader',
+            when: {
+                onCardDefeated: (event, context) =>
+                    // TODO: update trigger condition so that defender being defeated by attacker at the 'on attack' stage will also work
```
moyerr, line 17: Is this something that is not currently possible? If it requires more work, I wonder if it would be a good case for a  `pending` spec? Something like:

```ts
it('should trigger if an enemy unit is defeated by an on attack ability before combat damage is dealt', function() {
    pending('Update onCardDefeated trigger logic to support defeats during the on attack stage');
});

AMMayberry1: I'm still not a huge fan of the pending idea, we have a lot of TODOs like this so they'd pile up pretty fast. For lists that get that long I'd prefer to track using something that's designed for it and allows sorting and prioritizing like GH issues

moyerr: Sounds good, makes sense!

moyerr (approved):
Nice work on this, crazy the amount of tests these triggered leader abilities need 💪

PR #1166, opened 2025-05-06
Add Cure Wounds, Death Field, Jedi Consular, Darth Revan's Lightsabers

Frugghi:
| Cure Wounds | Death Field |
| --- | --- |
| image | image |

| Jedi Consular | Darth Revan's Lightsabers |
| --- | --- |
| image | image |

### DIFF
--- a/server/game/cards/05_LOF/upgrades/DarthRevansLightsabers.ts
+++ b/server/game/cards/05_LOF/upgrades/DarthRevansLightsabers.ts
@@ -1,0 +1,25 @@
+import type { AbilityContext } from '../../../core/ability/AbilityContext';
+import type { Card } from '../../../core/card/Card';
+import { UpgradeCard } from '../../../core/card/UpgradeCard';
+import { KeywordName, Trait } from '../../../core/Constants';
+import type { Player } from '../../../core/Player';
+
+export default class DarthRevanLightsabers extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '9566815036',
+            internalName: 'darth-revans-lightsabers',
+        };
+    }
+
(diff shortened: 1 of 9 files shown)


server/game/cards/05_LOF/upgrades/DarthRevansLightsabers.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export default class DarthRevanLightsabers extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '9566815036',
+            internalName: 'darth-revans-lightsabers',
+        };
+    }
+
+    public override canAttach(targetCard: Card, _context: AbilityContext, _controller: Player): boolean {
```
AMMayberry1, line 15: nit: instead of overriding this can just do

```ts
    this.setAttachCondition((card: Card) => !card.hasSomeTrait(Trait.Vehicle));
    // ...
```

test/server/cards/05_LOF/units/JediConsular.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+                });
+            });
+
+            it('should allow the controller to play a unit with a discount of 1', function () {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.jediConsular);
+                expect(context.player1).toHaveEnabledPromptButtons(['Attack', 'Play a unit from your hand. It costs 2 less']);
+
+                context.player1.clickPrompt('Play a unit from your hand. It costs 2 less');
```
AMMayberry1, line 23: We've been having some issues recently with stuff that plays a card from hand not filtering targets correctly. Can we adjust this test to set a resource amount and add some cards that shouldn't be playable due to cost, just to make sure they aren't selectable

PR #1167, opened 2025-05-07
Fix cards similar to Timely Intervention

### LINKED ISSUE #1086: Finish ECL hotfix work
ECL had a major bug where it would allow the player to click any card in the hand regardless of whether it could be played. We did a couple of PRs to try and address this, it is _mostly_ fixed at this point:

https://github.com/SWU-Karabast/forceteki/pull/1016
https://github.com/SWU-Karabast/forceteki/pull/1085

There are two outstanding issues that need to be addressed:

1. ECL is supposed to be able to be used as a "soft pass" by activating it and then doing nothing, even if there are no legal targets in the hand (this is legal because the hand is hidden information). Currently it cannot be, it will not be clickable if there are no legal targets.

2. The changes applied to the SimultaneousSystem were not applied to the SequentialSystem, we need to get both into the same place. Probably they should be using a single enum property instead of multiple boolean flags.

Frugghi:
Apply a fix similar to #1160 to other cards like Timely Intervention. It also fixes a contract assertion that was happening if you selected `Choose nothing` when using one of these abilities.

Fixes #1086

### DIFF
--- a/server/game/abilities/EpicActionAbility.ts
+++ b/server/game/abilities/EpicActionAbility.ts
@@ -1,0 +1,15 @@
+import { EpicActionLimit } from '../core/ability/AbilityLimit';
+import { ActionAbility } from '../core/ability/ActionAbility';
+import type { Card } from '../core/card/Card';
+import type Game from '../core/Game';
+import type { IEpicActionProps } from '../Interfaces';
+
+export class EpicActionAbility extends ActionAbility {
+    public constructor(game: Game, card: Card, properties: IEpicActionProps) {
+        super(game, card, { ...properties, limit: new EpicActionLimit() });
+    }
+
+    public override isEpicAction(): this is EpicActionAbility {
+        return true;
+    }
+}
--- a/server/game/abilities/EpicActionAbility.ts
+++ b/server/game/abilities/EpicActionAbility.ts
@@ -1,3 +1,4 @@
+import type { AbilityContext } from '../core/ability/AbilityContext';
 import { EpicActionLimit } from '../core/ability/AbilityLimit';
 import { ActionAbility } from '../core/ability/ActionAbility';
 import type { Card } from '../core/card/Card';
@@ -7,9 +8,15 @@ import type { IEpicActionProps } from '../Interfaces';
 export class EpicActionAbility extends ActionAbility {
     public constructor(game: Game, card: Card, properties: IEpicActionProps) {
         super(game, card, { ...properties, limit: new EpicActionLimit() });
+
+        this.canResolveWithoutLegalTargets = true;
     }
 
     public override isEpicAction(): this is EpicActionAbility {
         return true;
+    }
+
+    public override meetsRequirements(context: AbilityContext, ignoredRequirements: string[] = [], thisStepOnly: boolean = false): string {
+        return super.meetsRequirements(context, [...ignoredRequirements, 'gameStateChange'], thisStepOnly);
     }
 }
--- a/server/game/core/ability/abilityTargets/CardTargetResolver.ts
+++ b/server/game/core/ability/abilityTargets/CardTargetResolver.ts
@@ -84,8 +84,7 @@ export class CardTargetResolver extends TargetResolver<ICardTargetsResolver<Abil
 
     public override hasLegalTarget(context: AbilityContext) {
         const player = this.getChoosingPlayer(context);
-        return this.selector.hasEnoughTargets(context, player) ||
-          (context.ability?.isEpicAction() && this.isChoosingFromHidden(this.selector.getAllLegalTargets(context, player), context));
+        return this.selector.hasEnoughTargets(context, player);
     }
 
     public getAllLegalTargets(context: AbilityContext): Card[] {
--- a/server/game/core/ability/abilityTargets/CardTargetResolver.ts
+++ b/server/game/core/ability/abilityTargets/CardTargetResolver.ts
@@ -83,7 +83,9 @@ export class CardTargetResolver extends TargetResolver<ICardTargetsResolver<Abil
     }
 
     public override hasLegalTarget(context: AbilityContext) {
-        return this.selector.hasEnoughTargets(context, this.getChoosingPlayer(context));
+        const player = this.getChoosingPlayer(context);
+        return this.selector.hasEnoughTargets(context, player) ||
+          (context.ability?.isEpicAction() && this.isChoosingFromHidden(this.selector.getAllLegalTargets(context, player), context));
     }
 
     public getAllLegalTargets(context: AbilityContext): Card[] {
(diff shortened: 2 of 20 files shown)


server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -83,7 +83,9 @@ export class CardTargetResolver extends TargetResolver<ICardTargetsResolver<Abil
     }
 
     public override hasLegalTarget(context: AbilityContext) {
-        return this.selector.hasEnoughTargets(context, this.getChoosingPlayer(context));
+        const player = this.getChoosingPlayer(context);
+        return this.selector.hasEnoughTargets(context, player) ||
+          (context.ability?.isEpicAction() && this.isChoosingFromHidden(this.selector.getAllLegalTargets(context, player), context));
```
AMMayberry1, line 88: I was thinking about it, and I guess the reason that you are allowed to trigger ECL and then do nothing is that consuming the epic action limit technically counts as changing game state.

It might future-proof us a bit if we just wire that up explicitly. It shouldn't be too hard, the current flow is that `PlayerOrCardAbility.meetsRequirements()` will call into the overload `CardAbilityStep.hasAnyLegalEffects()` and from there we can check. Tbh, we could just have an overload of it in `BaseCard.ts` that just returns true so long as the epic action is still available.

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -656,11 +657,21 @@ export function selectPlayer<TContext extends AbilityContext = AbilityContext>(p
...
 //     return new SelectTokenAction(propertyFactory);
 // }
-export function sequential<TContext extends AbilityContext = AbilityContext>(gameSystems: ISystemArrayOrFactory<TContext>, everyGameSystemMustBeLegal: boolean = false) {
-    return new SequentialSystem<TContext>(gameSystems, everyGameSystemMustBeLegal);
-} // takes an array of gameActions, not a propertyFactory
-export function simultaneous<TContext extends AbilityContext = AbilityContext>(gameSystems: ISystemArrayOrFactory<TContext>, ignoreTargetingRequirements = null, everyGameSystemMustBeLegal: boolean = false) {
-    return new SimultaneousGameSystem<TContext>(gameSystems, ignoreTargetingRequirements, everyGameSystemMustBeLegal);
+export function sequential<TContext extends AbilityContext = AbilityContext>(propertyFactory: PropsFactory<ISequentialSystemProperties<TContext> | GameSystem<TContext>[], TContext>) {
+    return new SequentialSystem<TContext>((context: TContext) => {
+        const props = typeof propertyFactory === 'function' ? propertyFactory(context) : propertyFactory;
```
AMMayberry1, line 662: If the property is a factory, we don't necessarily want to unroll it here since it's intended to dynamically adjust as `context` changes. The `GameSystem.appendToPropertiesOrPropertyFactory()` helper covers that case (can see examples of usage in this file).

server/game/gameSystems/GameSystemLibrary.ts
```diff
@@ -656,11 +657,21 @@ export function selectPlayer<TContext extends AbilityContext = AbilityContext>(p
...
+        return !Array.isArray(props) ? props : {
```
AMMayberry1, line 663: nit: I think if we do the array check first we don't need to unroll or modify the property object in the case where the input isn't an array

server/game/gameSystems/SimultaneousOrSequentialSystem.ts
```diff
@@ -0,0 +1,75 @@
...
+    public override eventHandler() {}
+
+    public override getInnerSystems(properties: TProps) {
+        return properties.gameSystems;
+    }
+
+    public override hasLegalTarget(context: TContext, additionalProperties: Partial<TProps> = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+
+        if (properties.targetingEnforcement === TargetingEnforcement.EnforceAll) {
```
AMMayberry1, line 39: Should be there IgnoreAll checks in this class?

server/game/gameSystems/GameSystemLibrary.ts
Frugghi: I've rewritten this to avoid using the function every time. `GameSystem.appendToPropertiesOrPropertyFactory()` doesn't work here because we are not appending stuff to the properties, this is just to support initialising the system by passing an array instead of the full properties object.

server/game/gameSystems/SimultaneousOrSequentialSystem.ts
```diff
@@ -0,0 +1,75 @@
...
+import { GameStateChangeRequired } from '../core/Constants';
+import type { GameObject } from '../core/GameObject';
+import type { GameSystem, IGameSystemProperties } from '../core/gameSystem/GameSystem';
+import { AggregateSystem } from '../core/gameSystem/AggregateSystem';
+import type { Player } from '../core/Player';
+
+export interface ISimultaneousOrSequentialSystemProperties<TContext extends AbilityContext = AbilityContext> extends IGameSystemProperties {
+    gameSystems: GameSystem<TContext>[];
+
+    targetingEnforcement?: TargetingEnforcement;
```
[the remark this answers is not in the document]
Frugghi: Yeah, I think I agree that it would make sense. It would add a little bit of noise in the code but it would make things clearer for sure, it's not really obvious what's the default behaviour and it's easy to forget.

server/game/gameSystems/SimultaneousOrSequentialSystem.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export enum TargetingEnforcement {
```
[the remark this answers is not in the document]
Frugghi: I've renamed it but I'm still not 100% happy with it. I just don't have good ideas, naming is hard!

server/game/abilities/EpicActionAbility.ts
```diff
@@ -7,9 +8,15 @@ import type { IEpicActionProps } from '../Interfaces';
 export class EpicActionAbility extends ActionAbility {
     public constructor(game: Game, card: Card, properties: IEpicActionProps) {
         super(game, card, { ...properties, limit: new EpicActionLimit() });
+
+        this.canResolveWithoutLegalTargets = true;
```
AMMayberry1, line 12: What does this do? Not seeing it referenced anywhere

AMMayberry1 (approved):

server/game/abilities/EpicActionAbility.ts
Frugghi: It's used by `AbilityResolver`, it was needed to allow using an epic ability to soft pass,

PR #1172, opened 2025-05-08
#Added checks for user logins.

CheBato:
- Added a new check that lets us need to retrieve the user only when he first logs in. Afterwards we always check the query whether it already contains the data for the user.

### DIFF
--- a/server/gamenode/GameServer.ts
+++ b/server/gamenode/GameServer.ts
@@ -11,6 +11,7 @@ import { logger } from '../logger';
 import { Lobby, MatchType } from './Lobby';
 import Socket from '../socket';
 import type { User } from '../utils/user/User';
+import { AuthenticatedUser } from '../utils/user/User';
 import * as env from '../env';
 import type { Deck } from '../utils/deck/Deck';
 import type { CardDataGetter } from '../utils/cardData/CardDataGetter';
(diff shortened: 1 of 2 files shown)


server/middleware/AuthMiddleWare.ts
```diff
@@ -44,8 +45,18 @@ export const authMiddleware = (routeName = null) => {
...
                 logger.info(`Auth ${routeName}: token found in cookies. Authenticating user.`);
             }
+            if (req.body.user?.authenticated) {
+                const userEntity = {
+                    username: req.body.user.username,
+                    id: req.body.user.id,
+                    preferences: req.body.user.preferences,
+                    welcomeMessageSeen: req.body.user.welcomeMessageSeen
+                };
+                req.user = new AuthenticatedUser(userEntity);
```
AMMayberry1, line 55: Is there a reason this can't also just be done as `new AuthenticatedUser(req.body.user?.authenticated)`?

server/gamenode/GameServer.ts
```diff
@@ -367,7 +390,7 @@ export class GameServer {
         });
 
         // Add this to the setupAppRoutes method in GameServer.ts
-        app.get('/api/get-deck/:deckId', authMiddleware(), async (req, res, next) => {
+        app.post('/api/get-deck/:deckId', authMiddleware(), async (req, res, next) => {
```
AMMayberry1, line 393: I missed that this was a get api originally - was there a specific reason for that? post does make the most sense I think

AMMayberry1 (approved):

PR #1173, opened 2025-05-08
Implementing Grogu Mysterious Child

aruizc:
image


server/game/cards/05_LOF/units/GroguMysteriousChild.ts
```diff
@@ -0,0 +1,34 @@
...
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addActionAbility({
+            title: 'Heal up to 2 damage from a unit',
+            cost: AbilityHelper.costs.exhaustSelf(),
+            targetResolver: {
+                cardCondition: (card) => card.isUnit(),
+                immediateEffect: AbilityHelper.immediateEffects.heal({
```
AMMayberry1, line 19: This isn't quite correct, the player is technically allowed to choose how much healing they do. You can refer to Kashyyyk Defender as an example, it's extremely similar

PR #1174, opened 2025-05-08
kanan ezra luke

VJubert:


server/game/cards/05_LOF/units/LukeSkywalkerAHerosBeginning.ts
```diff
@@ -0,0 +1,34 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '7338701361',
+            internalName: 'luke-skywalker#a-heros-beginning',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Use the Force. If you do, give an Experience token and a Shield token to this unit',
```
AMMayberry1, line 15: nit: eventually I'll update the ability title generation to automatically append the "if you do" part, so we shouldn't include it in the base title

AMMayberry1 (dismissed):

AMMayberry1 (approved):

PR #1175, opened 2025-05-09
Implementing Sifo Dyas

aruizc:
image


server/game/cards/05_LOF/units/SifoDyasCommissioningAnArmy.ts
```diff
@@ -0,0 +1,43 @@
...
+        this.cardsPlayedThisPhaseWatcher = AbilityHelper.stateWatchers.cardsPlayedThisPhase(registrar, this);
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenDefeatedAbility({
+            title: 'Search the top 8 cards of your deck for any number of Clone units with combined cost 4 or less and discard them. For this phase, you may play those cards from your discard pile for free.',
+            immediateEffect: AbilityHelper.immediateEffects.deckSearch({
+                searchCount: 8,
+                targetMode: TargetMode.Unlimited,
+                cardCondition: (card) => card.isUnit() && card.hasSomeTrait(Trait.Clone) && card.cost <= 4,
```
AMMayberry1, line 27: This condition isn't correct, it will allow any card with cost <= 4. The card should enforce that the total cost of all selected units is less than 4 - see U-Wing for how to do this

test/server/cards/05_LOF/units/SifoDyasCommissioningAnArmy.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+                expect(context.player1).toHavePrompt('Select all cards to reveal');
+
+                expect(context.player1).toHaveExactDisplayPromptCards({
+                    selectable: [context.crosshair, context.batchBrothers, context.coruscantGuard, context.cloneHeavyGunner],
+                    invalid: [context.pykeSentinel, context.atst, context.wrecker, context.fives]
+                });
+                context.player1.clickPrompt('Take nothing');
+
+                expect(context.player1).toBeActivePlayer();
+            });
```
AMMayberry1, line 170: Two more tests that we need:
1. Play a unit for free from discard, then defeat it and confirm that it can't be played from discard again
2. Check that the opponent can't play the discarded units

AMMayberry1 (commented):
This condition isn't correct, it searches for a single card with cost <= 4. The card should allow selecting any number so long as the total cost is less than 4 - see U-Wing for how to do this

server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -47,6 +48,8 @@ export interface ISearchDeckProperties<TContext extends AbilityContext = Ability
     /** Used for filtering selection based on things like trait, type, etc. */
     cardCondition?: (card: Card, context: TContext) => boolean;
     chooseNothingImmediateEffect?: GameSystem<TContext>;
+    multiSelectCondition?: (card: Card, currentlySelectedCards: Card[], context: TContext) => boolean;
+    playAsType?: WildcardCardType.Upgrade | WildcardCardType.Unit | CardType.Event;
```
AMMayberry1, line 52: As discussed in chat, we should remove this

AMMayberry1 (approved):

PR #1208, opened 2025-05-14
Mind Trick, Pong Krell, Kylo Ren

VJubert:


server/game/cards/05_LOF/events/MindTrick.ts
```diff
@@ -0,0 +1,37 @@
...
+            id: 'mind-trick-id',
+            internalName: 'mind-trick',
+        };
+    }
+
+    protected override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Exhaust any number of units with a combined power of 4 or less. If you control a Force unit, those units lose all abilities and can\'t gain abilities for this phase',
+            targetResolver: {
+                activePromptTitle: 'Exhaust any number of units with a combined power of 4 or less',
```
AMMayberry1, line 20: nit: we don't need these as much anymore, the ability title will show up at the targeting stage

server/game/cards/05_LOF/units/KyloRenIKnowYourStory.ts
```diff
@@ -0,0 +1,30 @@
...
+            internalName: 'kylo-ren#i-know-your-story',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Use the Force',
+            when: {
+                onCardPlayed: (event, context) => {
+                    return event.player === context.player &&
```
AMMayberry1, line 17: nit: don't need the `return` format here, could just do `onCardPlayed: (event, context) => ({ ... })`

server/game/cards/05_LOF/units/KyloRenIKnowYourStory.ts
```diff
@@ -0,0 +1,30 @@
...
+                      event.card.isUpgrade() &&
```
AMMayberry1, line 18: I don't _think_ it would ever come up here but just to be safe we should probably use `event.playAsType === WildcardCardType.Upgrade`. In case there's ever like a reverse pilot that could be played as an upgrade on a unit

server/game/cards/05_LOF/units/KyloRenIKnowYourStory.ts
```diff
@@ -0,0 +1,30 @@
...
+                      event.attachTarget === context.source;
+                },
+            },
+            optional: true,
+            immediateEffect: AbilityHelper.immediateEffects.useTheForce(),
+            ifYouDo: {
+                title: 'Draw a card',
+                immediateEffect: AbilityHelper.immediateEffects.draw({ amount: 1 })
```
AMMayberry1, line 26: nit: I think `amount: 1` is the default

AMMayberry1 (approved):

PR #1210, opened 2025-05-15
Add LoF Rare bases

adbronico:


server/game/cards/05_LOF/bases/MysticMonastery.ts
```diff
@@ -0,0 +1,19 @@
...
+            id: 'mystic-monastery-id',
+            internalName: 'mystic-monastery',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addActionAbility({
+            title: 'The Force is with you',
+            limit: AbilityHelper.limit.perGame(3),
+            immediateEffect: AbilityHelper.immediateEffects.theForceIsWithYou()
```
AMMayberry1, line 16: We should add a visual representation of how many uses are left, but in the meantime we should log something in chat. That should be doable with effect / effectArgs

server/game/cards/05_LOF/bases/VergenceTemple.ts
```diff
@@ -0,0 +1,25 @@
...
+    protected override getImplementationId () {
+        return {
+            id: 'vergence-temple-id',
+            internalName: 'vergence-temple',
+        };
+    }
+
+    public override setupCardAbilities () {
+        this.addTriggeredAbility({
+            title: 'If you control a unit with 4 or more remaining HP, the Force is with you',
```
AMMayberry1, line 15: nit: don't need the "if" part, since this won't trigger if the condition isn't met

test/server/cards/05_LOF/bases/TempleOfDestruction.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+            context.player1.passAction();
+
+            context.player2.clickCard(context.wampa);
+            context.player2.clickCard(context.p1Base);
+
+            expect(context.player1.hasTheForce).toBe(false);
+            expect(context.player2.hasTheForce).toBe(false);
+        });
+
+        it('should give the Force when a friendly unit deals 3 or more combat damage to an enemy base', async function () {
```
AMMayberry1, line 46: Few other cases to check:
- Make sure that an event like FACIBI doesn't trigger it
- Overwhelm damage should work
- A unit ability like Synara San doesn't trigger

test/server/cards/05_LOF/bases/TombOfEilram.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+            context.player1.clickCard(context.tombOfEilram);
+            expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine]);
+            context.player1.clickCard(context.battlefieldMarine);
+
+            expect(context.player1.hasTheForce).toBe(true);
+            expect(context.player2.hasTheForce).toBe(false);
+            expect(context.player2).toBeActivePlayer();
+        });
+
+        it('should gain the Force when used', async function () {
```
AMMayberry1, line 54: nit: repeated test name

PR #1216, opened 2025-05-15
Qui Gon Jinn Lightsaber

aruizc:
image


server/game/cards/05_LOF/upgrades/QuiGonJinnsLightsaber.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    public override setupCardAbilities() {
+        this.addWhenPlayedAbility({
+            title: 'Exhaust any number of units with combined cost 6 or less.',
+            optional: true,
+            targetResolver: {
+                activePromptTitle: 'Exhaust any number of units with combined cost 6 or less',
+                mode: TargetMode.Unlimited,
+                canChooseNoCards: true,
+                multiSelectCardCondition: (card, selectedCards) => card.isUnit() && this.costSum(selectedCards.concat(card)) <= 6 && card.exhausted === false,
```
AMMayberry1, line 29: We shouldn't use the `card.exhausted === false` filter here, per the rules you're allowed to exhaust cards that are already exhausted. Let's add a test case confirming that

test/server/cards/05_LOF/upgrades/QuiGonJinnsLightsaber.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+                        groundArena: [
+                            'cell-block-guard',
+                            'phaseiii-dark-trooper',
+                            'darth-vader#commanding-the-first-legion',
+                        ]
+                    }
+                });
+            });
+
+            it('leader, it allows him to exhaust any number of units with combined cost 6 or less', function () {
```
AMMayberry1, line 32: Need a test that nothing happens when played on a unit that isn't Qui-Gon

server/game/cards/05_LOF/upgrades/QuiGonJinnsLightsaber.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export default class QuiGonJinnsLightsaber extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '3445044882',
+            internalName: 'quigon-jinns-lightsaber',
+        };
+    }
+
+    public override canAttach(targetCard: Card, _context: AbilityContext, controller: Player): boolean {
```
AMMayberry1, line 17: Instead of canAttach we can use `this.setAttachCondition()`. See any of the other Lightsaber cards for an example. Once changed we won't need the `isUnit()` restriction either

server/game/cards/05_LOF/upgrades/QuiGonJinnsLightsaber.ts
```diff
@@ -0,0 +1,44 @@
...
+export default class QuiGonJinnsLightsaber extends UpgradeCard {
+    protected override getImplementationId() {
+        return {
+            id: '3445044882',
+            internalName: 'quigon-jinns-lightsaber',
+        };
+    }
+
+    public override setupCardAbilities() {
+        const selectedCards: Card[] = [];
```
AMMayberry1, line 16: Sorry, just caught this - this doesn't work right. `selectedCards` will always be an empty array, and since `concat()` is non-modifying, `selectedCards.concat()` at line 28 will always send an array that is just `[card]`. So it's only ever looking at the cost of one card at a time instead of the sum.

The way to do it is to remove the variable here and do this on line 28:

```ts
multiSelectCondition: (card, currentlySelectedCards) => this.costSum(currentlySelectedCards.concat(card)) <= 3
```

The `currentlySelectedCards` array is a parameter of `multiSelectCardCondition` from the target resolver, which passes in the set of currently selected cards to see if `card` is allowed to be added to the list.

I saw you have a test that's intended to validate this, there's an issue there as well - left a separate comment.

test/server/cards/05_LOF/upgrades/QuiGonJinnsLightsaber.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+                    context.cellBlockGuard,
+                    context.kylosTieSilencer,
+                ]);
+                expect(context.player1).toHaveChooseNothingButton();
+
+                // Selects Qui Gon Jinn leader
+                context.player1.clickCard(quigonJinnLeader);
+
+                // Verify only units that keep combined cost ≤ 6 can be selected
+                expect(context.player1).not.toBeAbleToSelectExactly([
```
AMMayberry1, line 99: So this case is a little unintuitive, but the quick version is that we shouldn't use `not` with `toBeAbleToSelectExactly()` specifically. I think what you are trying to ask is "the player should not be allowed to select any of these", but this expectation is saying "the player selection list should not be this exact set of cards."

One thing to know is that selected cards still show up in the "selectable" list. So even though Qui-Gon is selected, the selectable list here is:

- Phase-III Dark Trooper
- Cell Block Guard
- Kylo's TIE Silencer
- Qui-Gon

Since the above list doesn't _exactly_ match the list you provided, this expectation returns true. It's always better for these to provide a positive list of what you expect to be targetable instead of a negative list of things you don't want to be targetable. So in this case this expectation should be:

`expect(context.player1).toBeAbleToSelectExactly([quigonnJinnLeader]);`

AMMayberry1 (approved):

PR #1219, opened 2025-05-15
Eeth Koth implementation

aruizc:
image


server/game/cards/05_LOF/units/EethKothSpiritualWarrior.ts
```diff
@@ -0,0 +1,23 @@
...
+    protected override getImplementationId() {
+        return {
+            id: 'eeth-koth#spiritual-warrior-id',
+            internalName: 'eeth-koth#spiritual-warrior',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.addWhenDefeatedAbility({
+            title: 'Use the Force to put Eeth Koth into play as a resource',
```
AMMayberry1, line 14: nit: should just be "Use the force", the `ifYouDo` title will get stitched on soon when we add the logic for doing it automatically

test/server/cards/05_LOF/units/EethKothSpiritualWarrior.spec.ts
```diff
@@ -0,0 +1,88 @@
+describe('Eeth Koth, Spiritual Warrior', function () {
+    integration(function (contextRef) {
+        describe('Eeth Koth, Spiritual Warrior\'s ability', () => {
+            beforeEach(function () {
+                return contextRef.setupTestAsync({
+
```
AMMayberry1, line 6: nit: extra whitespace

test/server/cards/05_LOF/units/EethKothSpiritualWarrior.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+                expect(context.eethKothSpiritualWarrior.exhausted).toBe(true);
+                expect(context.player1).toBeActivePlayer();
+            });
+
+            it('should allow the controller to not use the force put it into play as a resource, player doesn\'t have the Force token', function () {
+                const { context } = contextRef;
+
+                const resourcesBeforeTrigger = context.player1.resources.length;
+
+                context.player1.clickCard(context.cureWounds);
```
AMMayberry1, line 78: nit: can instead do `context.player1.setHasTheForce(false)`

server/game/cards/05_LOF/units/EethKothSpiritualWarrior.ts
AMMayberry1: Just following up on this - we've been having some discussion about how to handle this `title` issue, I think actually we're going to change how we handle it going forward. I'll write up a discord post explaining the change sometime soon, but basically we want to put the whole ability text in the top-level `title` and then the `ifYouDo` will have its own title. So like this:

```ts
this.addWhenDefeatedAbility({
   title: Use the Force. If you do, put Eeth Koth into play as a resource.
   // ...
   ifYouDo: {
      title: 'Put Eeth Koth into play as a resource'
```

AMMayberry1 (approved):

PR #1223, opened 2025-05-16
Sith Holocron

VJubert:


test/server/cards/05_LOF/upgrades/SithHolocron.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+            context.setDamage(context.p2Base, 0);
+
+            // sith holocron bonus should be only for the current attack
+            context.player1.clickCard(context.karis);
+            context.player1.clickCard(context.p2Base);
+            context.player1.clickPrompt('Pass');
+
+            expect(context.player2).toBeActivePlayer();
+            expect(context.p2Base.damage).toBe(3); // 2+1
+        });
```
AMMayberry1, line 61: This is a weird edge case, but let's add a test that it works when attached to an enemy force unit. The upgrade controller would be choosing to damage one of their own units to give the enemy unit +2, so it makes no sense, but we should verify that it works right

AMMayberry1 (approved):

PR #1230, opened 2025-05-16
Fix SearchDeckSystem effect message

Frugghi:


server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -92,9 +92,9 @@ export class SearchDeckSystem<TContext extends AbilityContext = AbilityContext,
         const searchCountAmount = this.computeSearchCount(properties.searchCount, context);
         const message =
         searchCountAmount > 0
-            ? `look at the top ${ChatHelpers.pluralize(searchCountAmount, 'card', 'cards')} of their deck`
```
Frugghi, line 95: This returns an object so it was showing `[object Object]` 😅

AMMayberry1 (approved):

PR #1237, opened 2025-05-18
Add The Will of the Force and Impossible Escape

Frugghi:
| The Will of the Force | Impossible Escape |
| --- | --- |
| image | image |

### DIFF
--- a/server/game/cards/05_LOF/events/ImpossibleEscape.ts
+++ b/server/game/cards/05_LOF/events/ImpossibleEscape.ts
@@ -13,6 +13,7 @@ export default class ImpossibleEscape extends EventCard {
     public override setupCardAbilities() {
         this.setEventAbility({
             title: 'Exhaust a friendly unit or use the Force. If you do either, exhaust an enemy unit and draw a card',
+            optional: true,
             immediateEffect: AbilityHelper.immediateEffects.conditional({
                 condition: (context) => context.player.hasSomeArenaUnit(),
                 onFalse: AbilityHelper.immediateEffects.useTheForce(),
--- a/server/game/cards/05_LOF/events/ImpossibleEscape.ts
+++ b/server/game/cards/05_LOF/events/ImpossibleEscape.ts
@@ -1,6 +1,6 @@
 import AbilityHelper from '../../../AbilityHelper';
 import { EventCard } from '../../../core/card/EventCard';
-import { EventName, RelativePlayer, WildcardCardType } from '../../../core/Constants';
+import { RelativePlayer, WildcardCardType } from '../../../core/Constants';
 
 export default class ImpossibleEscape extends EventCard {
     protected override getImplementationId() {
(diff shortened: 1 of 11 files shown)


server/game/cards/05_LOF/events/ImpossibleEscape.ts
```diff
@@ -0,0 +1,56 @@
...
+        return {
+            id: '9069308523',
+            internalName: 'impossible-escape',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Exhaust a friendly unit or use the Force. If you do either, exhaust an enemy unit and draw a card',
+            optional: true,
```
Frugghi, line 16: Do we know if `You may either` means that you can choose to do nothing? I would guess so but I'm not really sure.

server/game/gameSystems/OptionalSystem.ts
```diff
@@ -0,0 +1,71 @@
...
+        return [`choose if they want to ${format}`, args];
+    }
+
+    public override canAffectInternal(target: Player | Card, context: TContext, additionalProperties: Partial<IOptionalSystemProperties<TContext>> = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        const properties = this.generatePropertiesFromContext(context, additionalProperties);
+
+        return properties.innerSystem.canAffect(target, context, additionalProperties, mustChangeGameState);
+    }
+
+    public override hasLegalTarget(context: TContext, additionalProperties: Partial<IOptionalSystemProperties<TContext>> = {}): boolean {
```
AMMayberry1, line 34: This also has a `mustChangeGameState` parameter it should pass through

server/game/gameSystems/OptionalSystem.ts
```diff
@@ -0,0 +1,71 @@
...
+                () => {
+                    context.game.queueSimpleStep(() => {
+                        const messageArgs = [context.player, ' uses ', context.source, ' to '];
+                        const [effectMessage, effectArgs] = properties.innerSystem.getEffectMessage(context, additionalProperties);
+                        context.game.addMessage('{0}{1}{2}{3}{4}{5}{6}{7}{8}', ...messageArgs, { message: context.game.gameChat.formatMessage(effectMessage, effectArgs) });
+
+                        properties.innerSystem.queueGenerateEventGameSteps(events, context, additionalProperties);
+                    }, `queue generate event game steps for ${this.name}`);
+                },
+                // eslint-disable-next-line @typescript-eslint/no-empty-function
```
AMMayberry1, line 56: nit: I think you can make this go away with `() => undefined`

test/server/cards/05_LOF/events/ImpossibleEscape.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+                    },
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.impossibleEscape);
+
+                expect(context.player2).toBeActivePlayer();
+                expect(context.player1.handSize).toBe(0);
+            });
```
AMMayberry1, line 178: Just to confirm that the `ifYouDo` is operating correctly with the condition removed, let's add cases where the user chooses an option that has no game effect and confirm that the exhaust trigger doesn't happen

AMMayberry1 (approved):

PR #1244, opened 2025-05-20
purge trooper and crushing blow

VJubert:


scripts/mockdata.js
```diff
@@ -512,6 +512,37 @@ const mockCards = [
...
+        },
+        unique: false,
+        internalName: 'crushing-blow'
+    }),
+    buildMockCard({
+        title: 'Purge Trooper',
+        cost: 3,
+        power: 4,
+        hp: 2,
+        hasNonKeywordAbility: false,
```
AMMayberry1, line 534: This should be true. Not really relevant since we're providing the implementation already, but let's fix it just to be safe (will need to bump card data version)

AMMayberry1 (approved):

PR #1247, opened 2025-05-20
Add Grand Inquisitor, Stories Travel Quickly

Frugghi:
| --- | --- |
| image | image |


test/server/cards/05_LOF/leaders/GrandInquisitorStoriesTravelQuickly.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+                    player2: {
+                        groundArena: ['wampa'],
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCardNonChecking(context.grandInquisitor);
+                expect(context.grandInquisitor).not.toHaveAvailableActionWhenClickedBy(context.player1);
+            });
```
AMMayberry1, line 69: Let's add a test with TWI Darth Maul attacking two units, should apply the debuff to both

AMMayberry1 (approved):

PR #1250, opened 2025-05-21
Fix Contract assertion with Darth Maul and Kragan

Frugghi:


server/game/cards/02_SHD/units/KintanIntimidator.ts
```diff
@@ -11,8 +11,8 @@ export default class KintanIntimidator extends NonLeaderUnitCard {
 
     public override setupCardAbilities() {
         this.addOnAttackAbility({
-            title: 'Exhaust the defender when attacking a unit.',
-            immediateEffect: AbilityHelper.immediateEffects.exhaust((context) => ({ target: context.source.activeAttack?.getSingleTarget() }))
+            title: 'Exhaust the defender when attacking a unit',
+            immediateEffect: AbilityHelper.immediateEffects.exhaust((context) => ({ target: context.source.activeAttack?.getAllTargets() }))
```
Frugghi, line 15: This is just in case they release an upgrade that allow you to attack multiple targets

Issue #895, opened 2025-03-12
Make leader pilots be defeated if attached unit changes control


AMMayberry1:
This could happen with eg attaching your pilot to a unit stolen with Change of Heart. Not 100% how this works right now but we need to add tests and potentially fix.

PR #1260, opened 2025-05-21
Fix issues with Impropriety Among Thieves

### LINKED ISSUE #1165: Issues with Impropriety Among Thieves
Multiple bug reports of Impropriety Among Thieves not working correctly - one report was that units didn't return correctly, another was that the effect may have returned a to the field from being in hand.

These aren't validated but should be investigated and we should add more test cases. One very likely cause is that one of the swapped units is going to discard or some other zone, and that causes some kind of exception when the delayed effects resolve which blocks the other unit's delayed effect from resolving.

moyerr:
Fixes #1165

### Test Cases Added:

- Event fizzles if there is no ready friendly non-leader unit
- Event fizzles if there is no ready enemy non-leader unit
- A unit is defeated after exchange of control
- A unit is returned to hand after exchange of control
- A unit changes control again after initial exchange of control
- Both units are defeated, one of which moves to the resource zone
- Both units are returned to hand

### DIFF
--- a/server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
+++ b/server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
@@ -20,9 +20,22 @@ export default class ImproprietyAmongThieves extends EventCard {
                     cardTypeFilter: WildcardCardType.NonLeaderUnit,
                     zoneFilter: WildcardZoneName.AnyArena,
                     cardCondition: (card) => card.canBeExhausted() && !card.exhausted,
+                    immediateEffect: AbilityHelper.immediateEffects.simultaneous([
+                        AbilityHelper.immediateEffects.takeControlOfUnit((context) => ({
+                            newController: context.player.opponent,
+                        })),
+                        AbilityHelper.immediateEffects.delayedCardEffect((context) => ({
+                            title: 'Owner takes control',
+                            when: {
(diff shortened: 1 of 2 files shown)


moyerr:
This should have one more test case for when a unit becomes a leader unit due to leader pilot upgrade after #895

server/game/cards/03_TWI/events/ImproprietyAmongThieves.ts
```diff
@@ -20,44 +20,41 @@ export default class ImproprietyAmongThieves extends EventCard {
                     cardTypeFilter: WildcardCardType.NonLeaderUnit,
                     zoneFilter: WildcardZoneName.AnyArena,
                     cardCondition: (card) => card.canBeExhausted() && !card.exhausted,
-                },
-                enemyUnit: {
-                    dependsOn: 'friendlyUnit',
```
AMMayberry1, line 25: I think we still need this, otherwise it would work when only one player has a unit. I think we don't have tests for that yet, which we definitely need to add - specifically, test cases that it whiffs if there's only a friendly unit and no enemy unit and vice versa

moyerr: Right right, good call

AMMayberry1 (approved):

PR #1268, opened 2025-05-22
Add explanation to why an event is blank

Frugghi:
A lot of people are confused by Relentless 😅


server/game/core/card/EventCard.ts
```diff
@@ -55,13 +55,25 @@ export class EventCard extends EventCardParent {
 
     /** Ability of event card when played. Will be a "blank" ability with no effect if this card is disabled by an effect. */
     public getEventAbility(): EventAbility {
-        return this.isBlank() || !this.hasImplementationFile
-            ? new EventAbility(this.game, this, {
-                title: this.hasImplementationFile ? 'Unimplemented event card ability' : 'No effect',
+        if (this.isBlank()) {
```
AMMayberry1, line 58: So, possible extension here if you're interested - I've been wanting to do this for a while except also to pop up a prompt being like "this event card will have no effect because [reason], do you want to proceed yes / no", which could be one of the two reasons you have here _or_ that it will just have no effect (e.g. Vanquish when there's only a leader unit out). That would just be evaluated using `meetsRequirements()`.

Obviously that'd be a bit more work, lmk if you'd be interested in doing that in this or a separate PR - np if not, as always. Thanks for taking care of all this!

Frugghi: I'll look this in a separate PR!

PR #1274, opened 2025-05-23
Ask confirmation before playing events with no effect

Frugghi:


test/server/actions/PlayEventFromHand.spec.ts
```diff
@@ -34,45 +32,115 @@ describe('Play event from hand', function() {
...
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.forceChoke);
+                expect(context.player1).toHavePrompt('Playing Force Choke will have no effect because its ability will not change the game state');
+                expect(context.player1).toHaveExactPromptButtons(['Play anyway', 'Cancel']);
+
+                context.player1.clickPrompt('Cancel');
+                expect(context.forceChoke).toBeInZone('hand');
+                expect(context.player1.exhaustedResourceCount).toBe(0);
+                expect(context.player1).toBeActivePlayer();
```
AMMayberry1, line 98: Just to confirm that the cancel didn't leave any weird state behind, let's follow this up by actually playing the card out

server/game/actions/PlayEventAction.ts
```diff
@@ -62,6 +72,38 @@ export class PlayEventAction extends PlayCardAction {
...
+            }
+
+            const requirements = eventAbility.meetsRequirements(eventAbilityContext, ['player', 'zone'], true);
+            if (requirements === '' && !context.source.isBlank() && context.source.isImplemented) {
+                return;
+            }
+
+            let reason = '';
+            if (requirements === 'gameStateChange') {
+                reason = 'because its ability will not change the game state';
```
AMMayberry1, line 87: This is the most technically correct thing to say but it might feel clearer to put it as "it has no legal targets" or something in that vein

server/game/actions/PlayEventAction.ts
```diff
@@ -62,6 +72,38 @@ export class PlayEventAction extends PlayCardAction {
...
+            } else if (context.source.isBlank()) {
+                const blankSource = context.source.getOngoingEffectSources(EffectName.Blank);
+                reason = `due to an ongoing effect of ${blankSource[0].title}`;
+            } else if (!context.source.isImplemented) {
+                reason = 'because the card is not implemented yet';
+            }
+
+            this.game.promptWithHandlerMenu(context.player, {
+                activePromptTitle: `Playing ${context.source.title} will have no effect${reason.length > 0 ? ` ${reason}` : ''}`,
```
AMMayberry1, line 96: I'm trying to find a clearer way to put this - maybe "Playing [card] will have no effect. Are you sure you want to proceed?"

And then in the Relentless or unimplemented cases we could say "will have no effect because ...".

Lmk if that feels good to you, if not maybe we can ask for ideas in chat

AMMayberry1 (approved):

PR #1310, opened 2025-05-28
Morgan Elsbeth, Following the Call

moyerr:
| Leader Side | Unit Side |
| :---: | :---: |
|  |  |


server/game/stateWatchers/AttacksThisPhaseWatcher.ts
```diff
@@ -41,6 +41,10 @@ export class AttacksThisPhaseWatcher extends StateWatcher<IAttacksThisPhase> {
             .map((entry) => entry.attacker);
     }
 
+    public cardDidAttack(card: Card): boolean {
+        return this.getAttackers((entry) => entry.attacker === card).length > 0;
```
AMMayberry1, line 45: It's a hair more complicated than this because of the "copy of a card" concept. I.e., the scenario where a unit attacks, is bounced back to hand, and then played back out. It is now a new "copy" of that card and is no longer considered to have attacked.

We use `inPlayId` to track this, basically it's a counter attached to each card that increments every time the card goes through a zone change that makes it considered to be a new copy. You can see an example usage in `getAttackersInPlay()` we'll just need to do the same check here (and add a test case that covers that scenario)

test/server/cards/05_LOF/leaders/MorganElsbethFollowingTheCall.spec.ts
```diff
@@ -0,0 +1,392 @@
...
+                // It is played as a unit
+                expect(context.player1).not.toHaveExactPromptButtons([
+                    'Cancel',
+                    'Play Independent Smuggler',
+                    'Play Independent Smuggler with Piloting'
+                ]);
+                expect(context.independentSmuggler).toBeInZone('groundArena');
+                expect(context.morganElsbeth.exhausted).toBeTrue();
+                expect(context.player2).toBeActivePlayer();
+            });
```
AMMayberry1, line 159: Couple additional cases it would be good to have (if we don't already):
- Confirm that a matching keyword with an opponent's unit doesn't count
- Have an in-play unit with a gained keyword, through its own ability or through an ability gained from another card

AMMayberry1 (approved):

PR #1314, opened 2025-05-29
make play as type mandatory

VJubert:


server/game/cards/01_SOR/bases/EnergyConversionLab.ts
```diff
@@ -21,7 +21,7 @@ export default class EnergyConversionLab extends BaseCard {
                 zoneFilter: ZoneName.Hand,
                 immediateEffect: AbilityHelper.immediateEffects.simultaneous({
                     gameSystems: [
-                        AbilityHelper.immediateEffects.playCardFromHand(),
+                        AbilityHelper.immediateEffects.playCardFromHand({ playAsType: WildcardCardType.Unit }),
```
AMMayberry1, line 24: Let's remove the `targetResolver.cardTypeFilter` property in all of these cases as well so that the `PlayCardSystem` can manage the filter, just in case there are other complex cases like Piloting in the future that pop up

PR #1332, opened 2025-06-01
shatterpoint

VJubert:

### DIFF
--- a/server/game/cards/05_LOF/events/Shatterpoint.ts
+++ b/server/game/cards/05_LOF/events/Shatterpoint.ts
@@ -1,0 +1,45 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { RelativePlayer, TargetMode, WildcardCardType } from '../../../core/Constants';
+import ForceToken from '../tokens/ForceToken';
+
+export default class Shatterpoint extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: 'shatterpoint-id',
+            internalName: 'shatterpoint',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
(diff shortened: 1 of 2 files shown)


test/server/cards/05_LOF/events/Shatterpoint.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+                expect(context.player1).toBeAbleToSelectExactly([context.wampa, context.atst, context.greenSquadronAwing]);
+                context.player1.clickCard(context.atst);
+
+                expect(context.player2).toBeActivePlayer();
+                expect(context.player1.hasTheForce).toBeFalse();
+                expect(context.atst).toBeInZone('discard', context.player2);
+            });
+
+            it('should use the force to defeat a non-leader unit (without the Force)', function () {
```
adbronico, line 38: Can we update this to better describe how this test shows that nothing will happen if this option is chosen while not having the force?

server/game/cards/05_LOF/events/Shatterpoint.ts
```diff
@@ -0,0 +1,45 @@
...
+                        AbilityHelper.immediateEffects.selectCard({
+                            cardTypeFilter: WildcardCardType.NonLeaderUnit,
+                            cardCondition: (card) => card.isUnit() && card.remainingHp <= 3,
+                            innerSystem: AbilityHelper.immediateEffects.defeat(),
+                        }),
+                    ['Use the Force. If you do, defeat a non-leader unit']:
+                        AbilityHelper.immediateEffects.useTheForce()
+                }),
+            },
+            ifYouDo: (ifYouDoContext) => ({
```
adbronico, line 32: I need to think about if there's any better way to handle this situation with the IfYouDo in one block of a select.

adbronico (changes requested):

PR #1347, opened 2025-06-04
Fix Relentless preventing events with smuggle from Tech from being played

Frugghi:
Relentless is a bit weird and blanks events everywhere, even in hidden zones. This is technically incorrect because they should be blanked only after the cost is paid but it can't be done easily at the moment because of the logic to resolve early targets when playing a card.


moyerr (approved):
Great catch!

PR #1354, opened 2025-06-06
Cal Kestis

VJubert:


test/server/cards/05_LOF/leaders/CalKestisICantKeepHiding.spec.ts
```diff
@@ -0,0 +1,120 @@
+describe('Cal Kestis, I Can\'t Keep Hiding', function () {
+    integration(function (contextRef) {
+        describe('Cal Kestis\'s Leader side ability', function () {
```
Frugghi, line 3: I think we should add one more test to check that the leader side ability can't be used without the force.

adbronico (approved):

PR #1373, opened 2025-06-07
Praetorian Guard, Heavy Missile Gunship, Adept of Anger, Life Wind Sage, Ravening Gundark

VJubert:

### DIFF
--- a/test/server/cards/05_LOF/units/PraetorianGuard.spec.ts
+++ b/test/server/cards/05_LOF/units/PraetorianGuard.spec.ts
@@ -1,0 +1,88 @@
+describe('Praetorian Guard', () => {
+    integration(function (contextRef) {
+        it('Praetorian Guard does not gains Sentinel while you do not control a unit with 4 or more power', async () => {
+            await contextRef.setupTestAsync({
+                phase: 'action',
+                player1: {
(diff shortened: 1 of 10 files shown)


test/server/cards/05_LOF/units/PraetorianGuard.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+            const { context } = contextRef;
+
+            context.player2.clickCard(context.wampa);
+
+            expect(context.player2).toBeAbleToSelectExactly([context.praetorianGuard]);
+
+            context.player2.clickCard(context.praetorianGuard);
+        });
+    });
+});
```
adbronico, line 88: Can we add a check that the Praetorian Guard does not gain Sentinel while an enemy unit has 4 or more power?

adbronico (changes requested):

test/server/cards/05_LOF/units/PraetorianGuard.spec.ts
VJubert: wampa has 4 power, it's handle by every tests 😆

adbronico (approved):

PR #1375, opened 2025-06-07
Kylo Ren, We're Not Done Yet

moyerr:
| Leader Side | Unit Side |
| :---: | :---: |
| image | image |

### DIFF
--- a/server/game/cards/05_LOF/leaders/KyloRenWereNotDoneYet.ts
+++ b/server/game/cards/05_LOF/leaders/KyloRenWereNotDoneYet.ts
@@ -1,6 +1,8 @@
 import AbilityHelper from '../../../AbilityHelper';
+import type { TriggeredAbilityContext } from '../../../core/ability/TriggeredAbilityContext';
 import { LeaderUnitCard } from '../../../core/card/LeaderUnitCard';
-import { CardType, EventName, GameStateChangeRequired, RelativePlayer, TargetMode, WildcardCardType, ZoneName } from '../../../core/Constants';
+import { CardType, EventName, RelativePlayer, WildcardCardType, ZoneName } from '../../../core/Constants';
+import type { IThenAbilityPropsWithSystems } from '../../../Interfaces';
 
 export default class KyloRenWereNotDoneYet extends LeaderUnitCard {
     protected override getImplementationId() {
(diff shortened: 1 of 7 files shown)


server/game/cards/05_LOF/leaders/KyloRenWereNotDoneYet.ts
```diff
@@ -0,0 +1,54 @@
...
+        this.addActionAbility({
+            title: 'Discard a card from your hand. If you discard an Upgrade this way, draw a card',
+            cost: AbilityHelper.costs.exhaustSelf(),
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.discardCardsFromOwnHand((context) => ({
+                    target: context.player,
+                    amount: 1
+                })),
+                AbilityHelper.immediateEffects.conditional((context) => ({
+                    condition: context.events.some(
```
adbronico, line 23: Does context.events ever have more than one event here? I'm wondering if we can do like context.events[0] or something, or if there's any weirdness that can happen here if say Migs is on the field when the discard happens. I'm guessing not since the Kylo ability should fully resolve before any triggers?

test/server/cards/05_LOF/leaders/KyloRenWereNotDoneYet.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+
+                // Use Kylo Ren's ability
+                context.player1.clickCard(context.kyloRen);
+                context.player1.clickPrompt('Discard a card from your hand. If you discard an Upgrade this way, draw a card');
+                context.player1.clickCard(context.fallenLightsaber);
+
+                // Verify the card was discarded and no new card was drawn
+                expect(context.fallenLightsaber).toBeInZone('discard');
+                expect(context.kyloRen.exhausted).toBeTrue();
+                expect(context.player1.hand.length).toBe(0);
```
adbronico, line 108: Can we verify that Kylo base took 3 damage from lack of cards in deck?

test/server/cards/05_LOF/leaders/KyloRenWereNotDoneYet.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+                const { context } = contextRef;
+
+                // Deploy Kylo Ren
+                context.player1.clickCard(context.kyloRen);
+                context.player1.clickPrompt('Deploy Kylo Ren');
+
+                expect(context.kyloRen).toBeInZone('groundArena');
+                expect(context.player2).toBeActivePlayer();
+            });
+        });
```
adbronico, line 261: Can we add one more test where the cost of selected upgrades is greater than the available resources and verify the game limits what upgrades can be played to ones that can be afforded?

adbronico (changes requested):

test/server/cards/05_LOF/leaders/KyloRenWereNotDoneYet.spec.ts
moyerr: Done!

test/server/cards/05_LOF/leaders/KyloRenWereNotDoneYet.spec.ts
moyerr: Done!

server/game/cards/05_LOF/leaders/KyloRenWereNotDoneYet.ts
moyerr: Yeah even with nested triggers, the `some` should find the discard event. So I think this is okay as-is unless you think there could be an issue with it?

adbronico (approved):

PR #1382, opened 2025-06-08
Kylo Ren's Lightsaber & Corrupted Saber

moyerr:
| Kylo Ren's Lightsaber | Corrupted Saber |
| :---: | :---: |
| image | image |

### DIFF
--- a/test/server/cards/05_LOF/upgrades/KyloRensLightsaber.spec.ts
+++ b/test/server/cards/05_LOF/upgrades/KyloRensLightsaber.spec.ts
@@ -52,6 +52,14 @@ describe('Kylo Ren\'s Lightsaber', () => {
 
                 context.moveToNextActionPhase();
 
+                // Kylo Ren is not ready due to the second part of No Good to Me Dead's effect
+                expect(context.kyloRen.exhausted).toBeTrue();
+
+                context.moveToNextActionPhase();
+
+                // Kylo Ren is ready again
+                expect(context.kyloRen.exhausted).toBeFalse();
+
                 // Kylo Ren can be exhausted by friendly card abilities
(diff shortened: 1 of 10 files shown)


test/server/cards/05_LOF/upgrades/KyloRensLightsaber.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+                // It can only target non-Vehicle units (including enemy units)
+                expect(context.player1).toBeAbleToSelectExactly([
+                    context.kyloRen,
+                    context.hylobonEnforcer,
+                    context.battlefieldMarine
+                ]);
+                context.player1.clickCard(context.kyloRen);
+
+                // Player 2 plays No Good to Me Dead to attempt to exhaust Kylo Ren
+                context.player2.clickCard(context.noGoodToMeDead);
```
adbronico, line 41: Can we add a test somewhere (maybe in the No Good to Me Dead test) that if the target cannot be exhausted, it still won't ready during the following regroup phase?

adbronico (changes requested):

test/server/cards/05_LOF/upgrades/KyloRensLightsaber.spec.ts
moyerr: Can you clarify? This ability prevents the unit from being exhausted. It doesn't prevent it from readying.

adbronico: Right so more so want to confirm that the "does not ready" part of the event isn't broken if it's used on a unit that can't be exhausted

adbronico (approved):

PR #1385, opened 2025-06-08
HK87, Jedi In Hiding, Peli Motto

VJubert:

### DIFF
--- a/test/server/cards/05_LOF/units/JediInHiding.spec.ts
+++ b/test/server/cards/05_LOF/units/JediInHiding.spec.ts
@@ -1,0 +1,91 @@
+describe('Jedi In Hiding', () => {
+    integration(function (contextRef) {
+        it('Jedi In Hiding\'s when defeated ability may allow the player to use the force. If they do, opponent discards a card from their hand', async () => {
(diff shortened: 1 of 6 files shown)


test/server/cards/05_LOF/units/JediInHiding.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+            expect(context.player2).not.toHavePassAbilityButton();
+
+            context.player2.clickCard(context.noGloryOnlyResults);
+
+            expect(context.player1).toBeActivePlayer();
+            expect(context.player1.hasTheForce).toBeFalse();
+            expect(context.noGloryOnlyResults).toBeInZone('discard', context.player2);
+        });
+
+        it('Jedi In Hiding\'s when defeated ability may allow the player to use the force. If they do, opponent discards a card from their hand', async () => {
```
moyerr, line 35: I think this test case name needs to be updated to clarify that this case is verifying that the effect cannot be used if the player does not have the force

moyerr (commented):
Just some minor comments about clarifying test cases. Otherwise looks great! Feel free to ping me when this is updated

PR #1387, opened 2025-06-08
Implement The Father, fix The Son's ability

moyerr:
image

### DIFF
--- a/server/game/cards/05_LOF/units/TheFatherMaintainingBalance.ts
+++ b/server/game/cards/05_LOF/units/TheFatherMaintainingBalance.ts
@@ -1,0 +1,30 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+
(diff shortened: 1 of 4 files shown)


server/game/cards/05_LOF/units/TheFatherMaintainingBalance.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public override setupCardAbilities() {
+        this.addTriggeredAbility({
+            title: 'Deal 1 damage to this unit. If you do, the Force is with you.',
+            optional: true,
+            when: {
+                onForceUsed: (event, context) =>
+                    event.player === context.player &&
+                    !context.player.hasTheForce
```
adbronico, line 19: Does this final condition change anything? Technically, you should be able to damage him whenever you use the force, I'm thinking of a case where he has grit

moyerr: @adbronico Yeah, I think this is not necessary. These checks are handled in the `UseTheForceSystem`. I'll remove this!

adbronico (approved):

PR #1393, opened 2025-06-09
Zuckuss

VJubert:

### DIFF
--- a/server/game/cards/05_LOF/units/ZuckussTheFindsman.ts
+++ b/server/game/cards/05_LOF/units/ZuckussTheFindsman.ts
@@ -1,0 +1,41 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { EventName, TargetMode } from '../../../core/Constants';
+
+export default class ZuckussTheFindsman extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '0406487670',
+            internalName: 'zuckuss#the-findsman',
+        };
+    }
+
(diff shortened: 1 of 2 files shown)


server/game/cards/05_LOF/units/ZuckussTheFindsman.ts
```diff
@@ -0,0 +1,41 @@
...
+                mode: TargetMode.DropdownList,
+                options: this.game.playableCardTitles,
+                condition: (context) => context.player.opponent.drawDeck.length > 0   // skip ability if opponent has no cards in deck
+            },
+            then: (thenContext) => ({
+                title: 'Discard the top card of the defending player\'s deck, if a card with that name is discarded, this unit gets +4/+0 for this attack.',
+                thenCondition: (context) => context.player.opponent.drawDeck.length > 0,   // skip ability if opponent has no cards in deck
+                immediateEffect: AbilityHelper.immediateEffects.discardFromDeck((context) => ({
+                    amount: 1,
+                    target: context.source.activeAttack.getDefendingPlayer(),
```
moyerr, line 26: I think `activeAttack` is nullable, so it may be safer to use `context.source.activeAttack?.getDefendingPlayer()`

This might be relevant for cases where there are multiple on-attack triggers, and the first one defeats the defender before this one is resolved. For example, if Zuckuss has the Jedi Lightsaber and attacks a unit with 2HP. Can we add a test case for this?

VJubert: every usage of getDefendingPlayer is not protected, I add this test and I don't crash 

I don't know if you prefer to replace with context.player.opponent or let as it is

moyerr (approved):
Nice work!

PR #1394, opened 2025-06-09
Add Do or Do Not

Frugghi:
image

### DIFF
--- a/server/game/Interfaces.ts
+++ b/server/game/Interfaces.ts
@@ -446,7 +446,7 @@ interface IShieldedKeywordProperties extends IKeywordPropertiesBase {
 }
 
 /** List of keywords that don't have any additional parameters */
-type NonParameterKeywordName =
+export type NonParameterKeywordName =
   | KeywordName.Ambush
   | KeywordName.Grit
   | KeywordName.Hidden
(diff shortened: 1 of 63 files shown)


server/game/cards/05_LOF/events/DoOrDoNot.ts
```diff
@@ -0,0 +1,29 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '2062827036',
+            internalName: 'do-or-do-not',
+        };
+    }
+
+    public override setupCardAbilities() {
+        this.setEventAbility({
+            title: 'Use the Force to draw 2 cards. If you do not, draw 1 card',
```
adbronico, line 14: nit: I think I like having it be "Use the Force. If you do, draw 2 cards. If you do not, draw 1 card."

It's a little verbose but I like calling out the "If You do" parts in the message

adbronico (commented):
I think I'd like @AMMayberry1 to review this due to the engine changes. He's back today-ish so I think we're fine waiting for him to check it

server/game/core/ability/CardAbilityStep.ts
```diff
@@ -158,45 +159,46 @@ export class CardAbilityStep extends PlayerOrCardAbility {
             return null;
         }
 
-        let ifAbility;
-        let effectShouldResolve;
-
-        if (this.properties.ifYouDo) {
-            // if there are no resolved events, we can skip past evaluating "if you do" conditions
+        if (this.properties.ifYouDo || this.properties.ifYouDoNot) {
```
AMMayberry1, line 162: I think this looks good functionally but could maybe be tidied up a bit. I think the way it looks now there's some easy options for setting up a private helper method that creates a sub-ability-step and checks all the conditions

PR #1399, opened 2025-06-09
Disturbance in the force

VJubert:

### DIFF
--- a/server/game/cards/05_LOF/events/APrecariousPredicament.ts
+++ b/server/game/cards/05_LOF/events/APrecariousPredicament.ts
@@ -1,0 +1,47 @@
+import AbilityHelper from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import { PlayType, RelativePlayer, TargetMode, WildcardCardType, ZoneName } from '../../../core/Constants';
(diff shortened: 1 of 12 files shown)


server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -70,7 +70,7 @@ export class CardsLeftPlayThisPhaseWatcher extends StateWatcher<CardLeftPlayEntr
             when: {
                 onCardLeavesPlay: () => true
             },
-            update: (currentState, event) => currentState.concat({ card: event.card, controlledBy: event.card.controller, cardType: event.lastKnownInformation.type })
+            update: (currentState, event) => currentState.concat({ card: event.card, controlledBy: event.lastKnownInformation.controller, cardType: event.lastKnownInformation.type })
```
VJubert, line 73: @AMMayberry1 @adbronico I think we should add an issue to add tests with NGOR for every card who use this watcher

AMMayberry1: Sounds good to me

PR #1407, opened 2025-06-10
Fix Exactly target resolver when there aren't enough targets

### LINKED ISSUE #1401: Better support for "choose exactly N number of targets" effects
Currently there is no way to specify that a target resolver must have exactly N number of targets, and automatically pass resolution if not. If we use `TargetMode.Exactly`, then it will lock up if there aren't enough targets available. Right now we have to work around it using N target resolvers - see Always Two for an example.

This case needs to be able to be handled automatically within the engine. It might be considered just a bug fix for `TargetMode.Exactly` or maybe we need to massage the target modes / properties a bit to have this case fit into the toolkit more organically

Frugghi:
Fixes #1401


moyerr (approved):
This solution was way simpler than I thought it would be. Nice work!

AMMayberry1 (approved):

PR #1420, opened 2025-06-12
Fix damage distribution with replacement effects

Frugghi:


test/server/cards/05_LOF/events/ShienFlurry.spec.ts
```diff
@@ -104,5 +104,32 @@ describe('Shien Flurry', function() {
...
+            context.player1.clickCard(context.shienFlurry);
+            context.player1.clickCard(context.obiwanKenobi);
+
+            context.player2.clickCard(context.noGloryOnlyResults);
+            context.player2.clickCard(context.theLegacyRun);
+
+            expect(context.obiwanKenobi.damage).toBe(4);
+            expect(context.getChatLogs(3)).toContain('player2 uses The Legacy Run to distribute 6 damage among enemy units');
+            expect(context.getChatLogs(3)).toContain('player1 uses Obi-Wan Kenobi\'s gained ability from Shien Flurry to Shien Flurry\'s ability prevents 2 damage to Obi-Wan Kenobi');
+            expect(context.getChatLogs(3)).toContain('player2 uses The Legacy Run to distribute 4 damage to Obi-Wan Kenobi');
```
AMMayberry1, line 132: Would it be feasible to change the wording here to be "deal" instead "distribute"? I think it would be a bit clearer grammatically, I think of distribution as the number of button clicks and dealing damage as how much is actually applied

AMMayberry1 (approved):

Issue #1109, opened 2025-04-18
Add "[X] is defeated" game log


AMMayberry1:
Noticed when trying out stuff like OB that it can be hard to tell exactly what left the board when the chat logs aren't specific enough about the effects. Would be good to have a default "[X] was defeated" log to make it super-clear.

Issue #1272, opened 2025-05-22
Attack log improvements


AMMayberry1:
![Image](https://github.com/user-attachments/assets/bd17d933-8613-4a46-811f-2c8d05b35ede)

Couple random thoughts

- Would be good to still say the player name in logs like this
- I think no one remembers what their base name is (at least I don't), so it might be good to change it to "[player]'s base"

So these would end up looking like `Player 1 attacks Player 2's base with Obi-Wan Kenobi`

PR #1434, opened 2025-06-17
Add chat log when a unit is defeated

### LINKED ISSUE #998: Damage log should say if the unit was defeated
In the case where damage defeats the unit, a log like below should say "...deal 2 damage to Crafty Smuggler, defeating it"

Image

Frugghi:
Fixes #998

### DIFF
--- a/server/game/core/card/propertyMixins/UnitProperties.ts
+++ b/server/game/core/card/propertyMixins/UnitProperties.ts
@@ -858,9 +858,9 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor<TSta
                 this.game.queueSimpleStep(() => {
                     const responsiblePlayer = (defeatEvent as any).defeatSource?.player;
                     if (responsiblePlayer) {
-                        this.game.addMessage('{0} is defeated by {1} due to having no remaining HP', this, responsiblePlayer);
+                        this.game.addMessage('{0}\'s {1} is defeated by {2} due to having no remaining HP', this.controller, this, responsiblePlayer);
                     } else {
-                        this.game.addMessage('{0} is defeated due to having no remaining HP', this);
+                        this.game.addMessage('{0}\'s {1} is defeated due to having no remaining HP', this.controller, this);
                     }
                 }, `Log defeat message for ${this.internalName}`);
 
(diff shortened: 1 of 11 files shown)


server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -838,15 +844,32 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor<TSta
...
+                    this.game.getFrameworkContext(typeof source === 'object' ? source.player : null)
                 );
 
+                // add defeat event to window
+                this.game.addSubwindowEvents(defeatEvent);
+
+                this.game.queueSimpleStep(() => {
+                    const responsiblePlayer = (defeatEvent as any).defeatSource?.player;
+                    if (responsiblePlayer) {
+                        this.game.addMessage('{0} is defeated by {1} due to having no remaining HP', this, responsiblePlayer);
```
AMMayberry1, line 861: This is maybe a little vague - something like "is defeated due to an HP modifier changing" would at least be a little more concrete, and potentially we could add more specific versions for different cases down the road. Or if you have another thought I'm open to alternatives

AMMayberry1:
Resolves #1109

server/game/core/card/propertyMixins/UnitProperties.ts
Frugghi: This is also called in case of adding damage so I kept it a bit generic. It'd be easy to have different versions for damage vs non-damage though.

test/server/core/gameSteps/regroupWindow/RegroupPhase.spec.ts
```diff
@@ -306,6 +307,7 @@ describe('Regroup phase', function() {
             it('can allow the player to play a card, then initiate an attack with ambush', function() {
                 const { context } = contextRef;
                 context.moveToRegroupPhase();
+                expect(context.getChatLogs(1)).toContain('Fireball is defeated by player2 due to having no remaining HP');
```
AMMayberry1, line 310: Sorry for the multiple late comments - this is something I noticed with the attack logs as well (https://github.com/SWU-Karabast/forceteki/issues/1272), but I feel like we should reference the controlling player in any log like this relating to a unit. so e.g. "player1's Fireball is defeated..." etc

Don't need to fix the attack case in this PR, just a high-level comment

PR #1436, opened 2025-06-18
Improve attacks logs and logs with bases

### LINKED ISSUE #1272: Attack log improvements

Frugghi:
Fixes #1272

### DIFF
--- a/test/server/cards/04_JTL/bases/NabatVillage.spec.ts
+++ b/test/server/cards/04_JTL/bases/NabatVillage.spec.ts
@@ -81,8 +81,8 @@ describe('Nabat Village', function() {
                 expect([context.wampa, context.concordDawnInterceptors, context.restock]).toAllBeInBottomOfDeck(context.player1, 3);
                 expect(context.player1).toBeActivePlayer();
                 expect(context.getChatLogs(2)).toEqual([
-                    'player1 uses Nabat Village to choose 3 cards to move to the bottom of their deck',
-                    'player1 uses Nabat Village to move 3 cards to the bottom of their deck',
+                    'player1 uses player1\'s base to choose 3 cards to move to the bottom of their deck',
+                    'player1 uses player1\'s base to move 3 cards to the bottom of their deck',
                 ]);
             });
 
--- a/test/server/cards/04_JTL/bases/NabatVillage.spec.ts
+++ b/test/server/cards/04_JTL/bases/NabatVillage.spec.ts
@@ -81,8 +81,8 @@ describe('Nabat Village', function() {
                 expect([context.wampa, context.concordDawnInterceptors, context.restock]).toAllBeInBottomOfDeck(context.player1, 3);
                 expect(context.player1).toBeActivePlayer();
                 expect(context.getChatLogs(2)).toEqual([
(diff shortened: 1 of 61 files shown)


test/server/cards/04_JTL/bases/NabatVillage.spec.ts
```diff
@@ -81,8 +81,8 @@ describe('Nabat Village', function() {
                 expect([context.wampa, context.concordDawnInterceptors, context.restock]).toAllBeInBottomOfDeck(context.player1, 3);
                 expect(context.player1).toBeActivePlayer();
                 expect(context.getChatLogs(2)).toEqual([
-                    'player1 uses Nabat Village to choose 3 cards to move to the bottom of their deck',
-                    'player1 uses Nabat Village to move 3 cards to the bottom of their deck',
+                    'player1 uses player1\'s base to choose 3 cards to move to the bottom of their deck',
+                    'player1 uses player1\'s base to move 3 cards to the bottom of their deck',
```
moyerr, line 85: Ugh I guess this is an unfortunate side effect of this update. When a base has its own ability, I find it much clearer to call out the base name, as opposed to dealing damage, where I think it is clearer to say "player1's base"

test/server/cards/05_LOF/bases/MysticMonastery.spec.ts
```diff
@@ -43,15 +43,15 @@ describe('Mystic Monastery\'s ability', function() {
...
+            expect(context.getChatLogs(1)).toContain('player1 uses player1\'s base to gain the Force (1 use left)');
 
             context.player2.passAction();
             context.player1.setHasTheForce(false);
 
             context.player1.clickCard(context.mysticMonastery);
             expect(context.player1.hasTheForce).toBe(true);
             expect(context.player2).toBeActivePlayer();
-            expect(context.getChatLogs(1)).toContain('player1 uses Mystic Monastery to gain the Force (0 uses left)');
+            expect(context.getChatLogs(1)).toContain('player1 uses player1\'s base to gain the Force (0 uses left)');
```
moyerr, line 54: > player1 uses player1\'s base

Would it be feasible for these types of phrases to instead say `player1 uses their base`

test/server/cards/04_JTL/bases/NabatVillage.spec.ts
AMMayberry1: +1

AMMayberry1: +1

test/server/cards/05_LOF/bases/MysticMonastery.spec.ts
Frugghi: Were you suggesting this instead of showing the name of the base or would you do it everywhere? I refactor the implementation to keep the name of the base and it would be fairly easy to change `{0}'s base` to `their base`/`the opponent base`.

moyerr (approved):
Great work, love this update! 🙏

PR #1448, opened 2025-06-20
Strip out unneeded data from gamestate updates

AMMayberry1:
Our gamestate updates from the BE were massive and contained a huge amount of unused or redundant data. Stripped out the unused fields to reduce the size to ~1/3 of what it was.

Also consolidated the chat message format a bit to reduce the number of string pieces being sent.

Corresponding FE PR: https://github.com/SWU-Karabast/forceteki-client/pull/399


server/game/core/chat/GameChat.ts
```diff
@@ -47,37 +47,72 @@ export class GameChat {
...
+
+                if (prevFragmentType === 'string' && (thisFragmentType === 'string' || thisFragmentType === 'number')) {
+                    // concatenate simple strings and numbers together to reduce the number of elements sent over the wire
+                    formatOutput[formatOutput.length - 1] += formattedFragment;
+                } else if (thisFragmentType === 'number') {
+                    formatOutput.push(formattedFragment.toString());
+                } else {
+                    formatOutput.push(formattedFragment);
                 }
-            } else if (!argMatch && fragment) {
```
brianries, line 70: We've lost this space separating that we used to do -- is there a reason we did this to begin with?

PR #1450, opened 2025-06-21
Add a chat notification when a player requests a rematch

moyerr:
| Quick Rematch | Regular Rematch |
| :---: | :---: |
| Screenshot 2025-06-21 at 4 13 50 PM | Screenshot 2025-06-21 at 4 15 01 PM |

### DIFF
--- a/server/gamenode/Lobby.ts
+++ b/server/gamenode/Lobby.ts
@@ -376,17 +376,16 @@ export class Lobby {
         const mode = args[0];
         Contract.assertTrue(mode === 'reset' || mode === 'regular', 'Invalid rematch mode, expected reset or regular but receieved: ' + mode);
 
-        const currentUser = this.users.find((u) => u.id === socket.user.getId());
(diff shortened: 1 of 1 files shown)


server/gamenode/Lobby.ts
```diff
@@ -376,13 +376,17 @@ export class Lobby {
         const mode = args[0];
         Contract.assertTrue(mode === 'reset' || mode === 'regular', 'Invalid rematch mode, expected reset or regular but receieved: ' + mode);
 
+        const currentUser = this.users.find((u) => u.id === socket.user.getId());
+        Contract.assertNotNullLike(currentUser, `Unable to find user with id ${socket.user.getId()} in lobby ${this.id}`);
```
AMMayberry1, line 380: nit: can use the `Lobby.getUser()` method instead which handles this check

AMMayberry1 (approved):

PR #1455, opened 2025-06-23
Add controller id to card objects in chat

AMMayberry1:


Beregost:
looks good! Merge whenever you are ready.

PR #1463, opened 2025-06-24
Add setId to card objects in chat messages

AMMayberry1:
This is to facilitate displaying card image on hover in chat messages


danbastin (approved):

AMMayberry1:
Waiting on confirmation from Beregost that this works correctly

Beregost:
looks good, thanks!

PR #1471, opened 2025-06-29
add isBase to shortSummary

Beregost:


server/game/core/card/Card.ts
```diff
@@ -1209,6 +1209,7 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
             controllerId: this.controller.id,
             setId: this.setId,
             type: ChatObjectType.Card,
+            isBase: this.isBase()
```
AMMayberry1, line 1212: If we're adding a new property anyway, might as well make it more flexible and send the specific type instead of just indicating whether it's a base. Probably like `printedType: this.printedType`

AMMayberry1 (approved):

PR #1474, opened 2025-06-30
Logic for bookkeeping of different snapshot types

AMMayberry1:
Refactored the bookkeeping logic for snapshots to use associative data structures so that we can "bookmark" snapshots with different categories such as Action, Phase, etc.

The main goals for this refactor are:

1. Rearrange the bookkeeping for snapshots so they can be stored and retrieved in an associative way, e.g. based on type of snapshots and key values like history offset or ID
2. Ensure that snapshot references are never "leaked" so that we can be sure they are released and garbage collected when no longer used in any snapshot container
3. Enable taking of manual snapshots for test purposes. Eventually this can be extended to supporting player "bookmarks" for playtesting, but this PR doesn't get us all the way there as that's a bit lower-priority than the basic snapshot types.
4. Expose some basic data about what kinds of snapshots are available for each player which would be used in the FE to enable / disable the various Undo buttons depending on how much snapshot history of each type is available

The snapshot creation logic didn't really change much, just refactored into smaller and more specialized components. The key components to look at are:

- **SnapshotManager**: the entrypoint for managing creating snapshots and rolling back to them
- **SnapshotFactory**: creates the new snapshot at each time point and also creates the snapshot containers in a way that they can grab references to the current active snapshot safely. Also facilitates clearing snapshots taken later than a certain point for when we roll back.
- **SnapshotContainerBase**: an abstract base class that manages the relationship with SnapshotFactory to grab snapshots and clear them as needed. All snapshot containers extend from this.
- **GameStateManager**: stripped-down version of what it was before, now it just handles tracking all GameObjects and doing the state updates for a rollback


CDMayberry (commented):
Alright, it looks solid. I'm rather tired so can't deep dive but I don't see anything noticeable. Just had one quick note on parameter properties.

Frugghi (approved):

Frugghi (dismissed):

moyerr (approved):

PR #1488, opened 2025-07-04
Che bato/backend support for sounds

CheBato:
- Added backend support for sounds. Ready for testing.
FE PR: https://github.com/SWU-Karabast/forceteki-client/pull/371


server/gamenode/GameServer.ts
```diff
@@ -370,6 +377,33 @@ export class GameServer {
...
         });
 
+        app.post('/api/save-sound-preferences', authMiddleware(), async (req, res, next) => {
+            try {
+                const { soundPreferences } = req.body;
+                const user = req.user as User;
+
+                // Check if user is authenticated
+                if (user.isAnonymousUser()) {
+                    logger.error(`GameServer (save-sound-preferences): Anonymous user ${user.getId()} attempted to save sound preferences`);
```
AMMayberry1, line 387: nit: let's say "attempted to save sound preferences to dynamodb" just to be extremely explicit what's wrong

server/utils/user/UserFactory.ts
```diff
@@ -224,10 +239,26 @@ export class UserFactory {
      * @param preferences The updated preferences object
      * @returns True if update was successful
      */
-    public async updateUserPreferencesAsync(userId: string, preferences: Record<string, any>): Promise<void> {
+    public async updateUserPreferencesAsync(userId: string, preferences: Record<string, any>): Promise<UserPreferences> {
```
AMMayberry1, line 242: nit: could we rename `preferences` to `updatedPreferences` so its meaning is a little clearer when used inside the method?

server/utils/user/UserFactory.ts
```diff
@@ -224,10 +239,26 @@ export class UserFactory {
...
-            await dbService.saveUserSettingsAsync(userId, preferences);
+
+            // Get existing user preferences
+            const userProfile = await dbService.getUserProfileAsync(userId);
+            const currentPreferences = userProfile?.preferences || {};
+
+            // Merge sound preferences with defaults if they don't exist
+            const mergedPreferences = {
+                ...currentPreferences,
+                ...preferences,
```
AMMayberry1, line 253: Can we go ahead and insert a `...getDefaultPreferences()` here so we don't have to remember to do it later when we add something that needs it

AMMayberry1 (approved):

PR #1489, opened 2025-07-04
Add auth status to lobby user state

danbastin:
Order66 and ThisIsTheWay will present as authenticated locally.


server/gamenode/Lobby.ts
```diff
@@ -134,7 +134,8 @@ export class Lobby {
                 importDeckErrors: u.importDeckValidationErrors,
                 unimplementedCards: this.deckValidator.getUnimplementedCardsInDeck(u.deck?.getDecklist()),
                 minDeckSize: u.deck?.base.id ? this.deckValidator.getMinimumSideboardedDeckSize(u.deck?.base.id) : 50,
-                maxSideBoard: this.deckValidator.getMaxSideboardSize(this.format)
+                maxSideBoard: this.deckValidator.getMaxSideboardSize(this.format),
+                authenticated: u.socket?.user.isDevTestUser() || u.socket?.user.isAuthenticatedUser()
```
AMMayberry1, line 138: We'll need to add something to `sendGameState()` as well, though we'll have to figure out whether to append it here or add it to the Game object itself to include the summary

PR #1515, opened 2025-07-19
fix(undo): moved winners array to game.state

mariotee:


mariotee:
oops. i left the undo flag on in tests. i'll remove that soon

server/game/core/Game.js
```diff
@@ -195,6 +195,7 @@ class Game extends EventEmitter {
             isInitiativeClaimed: false,
             allCards: [],
             actionNumber: -1,
+            winnerNames: []
```
AMMayberry1, line 198: `this.state` is technically `protected`, so whenever a state property needs to be publicly visible we add a property getter for it (and setter if needed). If the state property is an array we also make the getter `readonly` to avoid letting callers modify array contents directly

server/game/core/Game.js
```diff
@@ -733,11 +734,11 @@ class Game extends EventEmitter {
...
      * Display message declaring victory for one player, and record stats for
      * the game
-     * @param {Player | Player[]} winner
+     * @param {Player[]} winnerPlayers
      * @param {String} reason
      */
-    endGame(winner, reason) {
-        if (this.winner) {
+    endGame(winnerPlayers, reason) {
+        if (this.state.winnerNames.length > 0) {
```
AMMayberry1, line 741: nit: where possible I like to use the getter for referencing stuff (once added) so that people don't have to think about or be aware of state that much. So `this.winnerNames` instead of `this.state.winnerNames`, although with arrays that only works when reading them sadly (for modifying we have to use the `state` reference directly)

server/game/core/Game.js
```diff
@@ -733,11 +734,11 @@ class Game extends EventEmitter {
     /**
      * Display message declaring victory for one player, and record stats for
      * the game
-     * @param {Player | Player[]} winner
+     * @param {Player[]} winnerPlayers
```
AMMayberry1, line 737: nit: you can go either way but for something that is optionally an array I find it's a little more convenient to be flexible on the parameter type and then use a helper function we have for array-ifying on the receiving side. That way the caller can use the array or non-array form organically w/out having to think about it. Example:

```js
// winners can be scalar or array
endGame(winners, reason) {
  const winnersAra = Helpers.asArray(winners);
  // ... proceed from here
```

server/game/core/Game.js
AMMayberry1: Sounds good! Yeah you can just resolve comments as you fix them and then ping me when it's ready. I try and check in on PR status as I have time but sometimes I miss stuff, if I see that all comments are resolved though I assume it's ready for another pass

server/game/core/Game.js
```diff
@@ -431,6 +432,14 @@ class Game extends EventEmitter {
         return otherPlayer;
     }
 
+    /**
+     * Gets the winner names of the game
+     * @returns {string[]}
+     */
+    getWinnerNames() {
```
AMMayberry1, line 439: One last nit - to be consistent with the other state properties, could you move this above the constructor definition and convert it to a formal property getter (i.e. `get winnerNames() { ... }`)

mariotee: updated

mariotee: or actually.. let me update the symbol name too to not have "get"

AMMayberry1 (approved):

PR #1546, opened 2025-08-01
Che bato/swu stats api

CheBato:
PR for SWU stats sync
FE PR: https://github.com/SWU-Karabast/forceteki-client/pull/434

- [x] Check that decklist from the opponent isn't being sent to the user because of decklist
- [x] Check if game.getPlayers keeps the players when they leave early
- [x] play the deck once on petranaki so it has some individual card stats, then confirm that playing it on karabast without those stats doesn't break anything


scripts/run-server.js
```diff
@@ -17,6 +17,8 @@ function createEnvFile() {
         // set this to "true" to enable the use of local DynamoDB for account testing
         USE_LOCAL_DYNAMODB: 'false',
         DISCORD_BUG_REPORT_WEBHOOK_URL: '',
+        SWUStatsURL: '',
```
AMMayberry1, line 20: I think we could hardcode this, I'm assuming it's public information?

server/gamenode/GameServer.ts
```diff
@@ -241,6 +243,7 @@ export class GameServer {
         this.testGameBuilder = testGameBuilder;
         this.deckValidator = deckValidator;
         this.bugReportHandler = new BugReportHandler();
+        this.swuStatsHandler = new SWUstatsHandler();
```
AMMayberry1, line 246: nit: the camelCase form of this class name should be `SwuStatsHandler`

server/gamenode/Lobby.ts
```diff
@@ -40,8 +41,11 @@ interface LobbyUser extends LobbySpectator {
 interface PlayerDetails {
     user: User;
     deckID: string;
+    deckLink: string;
+    deckSource: string;
```
AMMayberry1, line 45: We should make this an enum type

server/utils/SWUStats/SWUStatsHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+    apiKey: string;
+    winner: number; // 1 or 2
+    firstPlayer: number; // 1 or 2
+    round: number;
+    winHero: string;
+    loseHero: string;
+    winnerDeck?: string;
+    loserDeck?: string;
+    winnerHealth: number;
+    player1: string; // JSON string with leader/base info
```
AMMayberry1, line 17: Even if the swustats api doesn't, we should have clear interface types for all of these instead of using string representations of json

server/utils/SWUStats/SWUStatsHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+export class SWUstatsHandler {
+    private readonly apiUrl: string;
+    private readonly apiKey: string;
+
+    public constructor() {
+        // Use environment variable for API URL, defaulting to the known endpoint
+        this.apiUrl = process.env.SWUStatsURL;
+        this.apiKey = process.env.SWUStatsAPIKey;
+        if (!this.apiKey) {
+            logger.warn('SWUstatsHandler: No API key configured. Stats may not be accepted by SWUstats.');
```
AMMayberry1, line 50: This should probably throw an error and block the server from starting if we're not in local dev mode

server/utils/SWUStats/SWUStatsHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+    public async sendGameResultAsync(
+        game: Game,
+        player1DeckId: string,
+        player2DeckId: string,
+        player1Deck: IDecklistInternal,
+        player2Deck: IDecklistInternal
+    ): Promise<boolean> {
+        try {
+            // Don't send results for games that ended too early
+            if (game.roundNumber <= 1) {
```
AMMayberry1, line 73: Do we need to do this check here? I thought it was covered in the lobby already before the call to this

server/utils/SWUStats/SWUStatsHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+                logger.info(`Cannot send SWUstats for game with ${players.length} players`);
+                return false;
+            }
+
+            const [player1, player2] = players;
+
+            // Determine winner
+            const winner = this.determineWinner(game, player1, player2);
+            if (winner === 0) {
+                logger.info(`Game ${game.id} ended in a draw or without clear winner, not sending to SWUstats`);
```
AMMayberry1, line 88: Draws don't get sent to swustats?

server/utils/SWUStats/SWUStatsHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+                game,
+                player1,
+                player2,
+                player1DeckId,
+                player2DeckId,
+                player1Deck,
+                player2Deck,
+                winner
+            );
+            // Log the payload for debugging (without sensitive data)
```
AMMayberry1, line 103: What would you qualify as "sensitive" here? I think most of this should be fine to log apart from size concerns. The only stuff we have to be really careful about is PII

server/utils/SWUStats/SWUStatsHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            logger.info(`Successfully sent game result to SWUstats for game ${game.id}`);
+            return true;
+        } catch (error) {
+            logger.error('Failed to send game result to SWUstats', {
+                error: { message: error.message, stack: error.stack },
+                gameId: game?.id
+            });
+            return false;
```
AMMayberry1, line 131: Should be a throw I think

server/utils/SWUStats/SWUStatsHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+    private determineWinner(game: Game, player1: Player, player2: Player): number {
+        if (game.winnerNames.length > 1) {
+            return 0; // Draw
+        }
+        if (game.winnerNames.includes(player1.name)) {
+            return 1;
+        } else if (game.winnerNames.includes(player2.name)) {
+            return 2;
+        }
+        throw new Error('(SWUStats handler): There was an error when determining winner'); // Couldn't determine winner
```
AMMayberry1, line 147: Would be good to log the player names and the winnerNames here so we could debug what happened

server/utils/SWUStats/SWUStatsHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+        player2DeckLink: string,
+        player1Deck: IDecklistInternal,
+        player2Deck: IDecklistInternal,
+        winner: number
+    ): SWUstatsGameResult {
+        // Get leader and base IDs
+        const p1Leader = player1.leader?.setId;
+        const p1Base = player1.base?.setId;
+        const p2Leader = player2.leader?.setId;
+        const p2Base = player2.base?.setId;
```
AMMayberry1, line 167: There's a lot of code that's copy pasted here in this method for p1 and p2, could we create one private method that extracts all data for a player and then just call that twice?

server/utils/SWUStats/SWUStatsHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+            loseHero,
+            winnerDeck: winner === 1 ? JSON.stringify(player1Deck) : JSON.stringify(player2Deck),
+            loserDeck: winner === 1 ? JSON.stringify(player2Deck) : JSON.stringify(player1Deck),
+        };
+    }
+
+    /**
+     * Format a card ID object to string format (e.g., "SOR_001")
+     */
+    private formatCardId(setId: ISetId | undefined): string {
```
AMMayberry1, line 230: Util.setCodeToString() could be used instead

server/utils/SWUStats/SWUStatsHandler.ts
CheBato: yeah i think i know what you mean i'll make it into a single method.

server/utils/SWUStats/SWUStatsHandler.ts
CheBato: The API doesn't indicate it as a possibility.

scripts/run-server.js
```diff
@@ -17,6 +17,8 @@ function createEnvFile() {
         // set this to "true" to enable the use of local DynamoDB for account testing
         USE_LOCAL_DYNAMODB: 'false',
         DISCORD_BUG_REPORT_WEBHOOK_URL: '',
+        SWUStatsURL: 'https://swustats.net/TCGEngine/APIs/SubmitGameResult.php',
```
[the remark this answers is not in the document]
CheBato: ah gotcha

server/gamenode/Lobby.ts
```diff
@@ -625,6 +636,27 @@ export class Lobby {
...
+            if (deckLink.includes('swustats.net')) {
+                return DeckSource.SWUStats;
+            } else if (deckLink.includes('swudb.com')) {
+                return DeckSource.SWUDB;
+            } else if (deckLink.includes('swunlimiteddb.com')) {
+                return DeckSource.SWUnlimitedDB;
+            }
+        }
+        // Default fallback
+        return DeckSource.SWUDB;
```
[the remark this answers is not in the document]
CheBato: yes thats the conclusion regarding the JSON ones

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -0,0 +1,241 @@
...
+                player1,
+                player2,
+                player1DeckId,
+                player2DeckId,
+                player1Deck,
+                player2Deck,
+                winner
+            );
+            // Log the payload for debugging
+            logger.info(`Sending game result to SWUstats for game ${game.id}`, {
```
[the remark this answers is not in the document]
CheBato: sounds good!

server/gamenode/Lobby.ts
```diff
@@ -32,16 +32,25 @@ interface LobbyUser extends LobbySpectator {
     state: 'connected' | 'disconnected';
     ready: boolean;
     deck?: Deck;
+    decklist?: any;
```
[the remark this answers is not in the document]
CheBato: Added to checklist to check before testing and review

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -0,0 +1,241 @@
...
+    public async sendGameResultAsync(
+        game: Game,
+        player1DeckId: string,
+        player2DeckId: string,
+        player1Deck: IDecklistInternal,
+        player2Deck: IDecklistInternal
+    ): Promise<boolean> {
+        try {
+            const players = game.getPlayers();
+            if (players.length !== 2) {
```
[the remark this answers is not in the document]
CheBato: Added to checklist to see if they stay.

server/gamenode/Lobby.ts
```diff
@@ -40,7 +40,8 @@ interface LobbyUser extends LobbySpectator {
 export enum DeckSource {
     SWUStats = 'SWUStats',
     SWUDB = 'SWUDB',
-    SWUnlimitedDB = 'SWUnlimitedDB'
+    SWUnlimitedDB = 'SWUnlimitedDB',
+    Unknown = 'UNKNOWN'
```
AMMayberry1, line 44: nit: our casing for the values is typically also camelcase, could we update these? Should be `'unknown'`, `'swUnlimitedDb'`, `'swuDb'` or `'swudb'`, `'swuStats'`

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -234,8 +223,6 @@ export class SwuStatsHandler {
             gameName: String(game.id),
             winHero,
             loseHero,
-            winnerDeck: winner === 1 ? player1Deck : player2Deck,
-            loserDeck: winner === 1 ? player2Deck : player1Deck,
```
AMMayberry1: Just want to confirm that these were removed intentionally?

scripts/run-server.js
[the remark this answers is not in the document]
AMMayberry1: Just realized this is still here, can we remove it if no longer needed?

AMMayberry1 (approved):

PR #1558, opened 2025-08-06
test(undo): added cases for action and regroup phase undoing

mariotee:


test/helpers/PlayerInteractionWrapper.js
```diff
@@ -769,6 +785,16 @@ class PlayerInteractionWrapper {
         this.clickPrompt('Pass');
     }
 
+    /**
+     * Player clicks Done prompt
+     */
+    clickDone() {
+        if (!this.canAct) {
+            throw new TestSetupError(`${this.name} can't choose Done, because they don't have priority`);
```
AMMayberry1, line 793: This seems wrong - you could be presented with a Done prompt even if it's not your turn, e.g. for indirect damage

test/server/repro.ts
AMMayberry1, line 1: What's this file for?

test/scenarios/undo/Undo.spec.ts
```diff
@@ -1007,7 +1008,155 @@ describe('Undo', function() {
...
+                expect(player2).toBeActivePlayer();
+            });
+
+            it('should allow a player to undo then play a card after passing', function() {
+                const { context } = contextRef;
+                const { player1, player2 } = context;
+                const snapshotId = contextRef.snapshot.takeManualSnapshot(context.player1Object);
+
+                player1.passAction();
+                player2.clickCard(player2.hand[0]);
```
AMMayberry1, line 1062: nit: if you don't I'd prefer we use the convention of naming the specific card (e.g. `context.ardentSympathizer`) throughout these tests to make them easier to read. It's also less likely to break if we change the test setup

mariotee: i tried adding some helpers and refactored other places using the same pattern. i didn't want the test to depend on which card it was unless it was actually testing that card. i feel like the helpers will improve readability and DAMP-ness of the tests

AMMayberry1 (approved):

PR #1562, opened 2025-08-06
CPU stats logging

brianries:
Added cpu stats logging on a 30 second interval using the node.js process utils

Output looks like this next to HeapStats

```
2025-08-06T21:44:06.131Z [info]: [HeapStats] Used: 30.6MB / 33.7MB (0.7% of 4144.0MB limit) | Total physical usage: 74.0MB | System free: 12.71GB
2025-08-06T21:44:06.131Z [info]: [CpuStats] Total Usage: 0.0% (User: 0.0%, System: 0.0%) | System Cores: 16
```

Now with EventLoopStats
```
2025-08-07T01:36:12.396Z [info]: [HeapStats] Used: 30.2MB / 33.7MB (0.7% of 4144.0MB limit) | Total physical usage: 74.3MB | System free: 11.79GB
2025-08-07T01:36:12.397Z [info]: [CpuStats] Total Usage: 0.0% (User: 0.0%, System: 0.0%) | System Cores: 16
2025-08-07T01:36:12.397Z [info]: [EventLoopStats] Event Loop Utilization: 0.2% | Event Loop Duration (ms): min: 13.6, P50: 15.3, P90: 15.9, max: 16.5
```

### DIFF
--- a/server/gamenode/GameServer.ts
+++ b/server/gamenode/GameServer.ts
@@ -7,7 +7,7 @@ import type { DefaultEventsMap, Socket as IOSocket } from 'socket.io';
 import { Server as IOServer } from 'socket.io';
 import { constants as zlibConstants } from 'zlib';
 import { getHeapStatistics } from 'v8';
-import { freemem } from 'os';
+import { freemem, cpus } from 'os';
 
 import { logger } from '../logger';
 
@@ -106,6 +106,8 @@ export class GameServer {
     private readonly deckValidator: DeckValidator;
     private readonly testGameBuilder?: any;
     private readonly queue: QueueHandler = new QueueHandler();
+    private lastCpuUsage: NodeJS.CpuUsage;
+    private lastCpuUsageTime: bigint;
     private readonly userFactory: UserFactory = new UserFactory();
     public readonly deckService: DeckService = new DeckService();
     public readonly bugReportHandler: BugReportHandler;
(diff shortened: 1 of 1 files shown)


server/gamenode/GameServer.ts
```diff
@@ -247,6 +250,11 @@ export class GameServer {
         // set up periodic heap monitoring every 30 seconds
         this.logHeapStats();
         setInterval(() => this.logHeapStats(), 30000);
+
+        // set up periodic CPU monitoring every 30 seconds
+        this.lastCpuUsage = process.cpuUsage();
+        this.lastCpuUsageTime = process.hrtime.bigint();
+        setInterval(() => this.logCpuUsage(), 30000);
```
AMMayberry1, line 257: nit: could we condense the cpu and heap stuff into one method, doesn't feel like they need to be separate. or at least set only one interval that calls both of them.

server/gamenode/GameServer.ts
```diff
@@ -1327,6 +1335,24 @@ export class GameServer {
         }
     }
 
+    private logCpuUsage(): void {
+        const cpuUsageDiff = process.cpuUsage(this.lastCpuUsage);
```
AMMayberry1, line 1339: This is _super_ important and I missed it on the previous PR you did - any code that's called from a timer function like `setInterval()` must **never** throw an uncaught exception, or else it will crash the whole server. Typically we just do a try-catch around the function body and log the error like I did with `logHeapStats()` below

server/gamenode/GameServer.ts
brianries: having the set interval call both

server/gamenode/GameServer.ts
brianries: Thanks, didn't realize this was uncaught exception handler territory, good catch

PR #1575, opened 2025-08-12
feat(undo): added logging around rollbacks

mariotee:


server/game/core/Game.js
```diff
@@ -1767,18 +1772,36 @@ class Game extends EventEmitter {
...
     }
 
-    postRollbackOperations(roundEntryPoint = null) {
+    postRollbackOperations(roundEntryPoint = null, preUndoState = null) {
+        const postUndoState = this.captureGameState('any');
+        const gameStates = {
+            preUndoState,
+            postUndoState,
+        };
+        logger.debug('Rollback completed', {
```
AMMayberry1, line 1800: It doesn't have to happen in this PR, but since we plan to gradually roll out more in-game logging as we have bandwidth, it would be great if we could go ahead and set up some kind helper method on Game that will just automatically insert the lobbyId and gameId metadata.

Basically, stuff would be able to call `game.logDebug()` or `game.logger.debug()` and not have to add these properties. That could also be useful for the thing we were talking about with disabling logging during tests, but that could also be managed elsewhere.

I don't necessarily want to hold up this PR on it since it's a big change relative to the size of this PR, but if you find yourself adding more logging stuff over time it might be good to think about adding that for convenience.

AMMayberry1, line 1800: Leaving a separate comment for this - I do think that we need to figure out how to squelch these logs when running jasmine tests before resolving this PR, kind of separate from the above comment

server/game/core/DiscordDispatcher.ts
```diff
@@ -0,0 +1,298 @@
...
+        const snapshotFileName = `undo-failure-snapshot-settings-${undoFailure.lobbyId}-${new Date().getTime()}.json`;
+        formData.append('files[1]', Buffer.from(snapshotSettingsJson), {
+            filename: snapshotFileName,
+            contentType: 'application/json',
+        });
+
+        return httpPostFormData(this._serverErrorWebhookUrl, formData);
+    }
+
+    public formatAndSendServerErrorAsync(error: Error, lobbyId: string): Promise<EitherPostResponseOrBooolean> {
```
AMMayberry1, line 215: Did you test this one to see how it looks in the server? I think I only see undo failure tests

PR #1583, opened 2025-08-14
Add support for undo at the start of setup phase

Frugghi:
This add support for undoing at the start of the setup phase but not at mulligan and resourcing step yet.


AMMayberry1:
Looks good overall. Do you mind adding a basic test case or two in SnapshotTypes.spec.ts? Even if it's not complete, just having a guard for the current progress would be good

PR #1590, opened 2025-08-16
Feature: State Watcher Rework

CDMayberry:
Reworked the different state watcher derived classes so there is only one instance of each class per game. Reduced the functionality of the StateWatcherRegistrar to simply keep references to the instances and moved the state back into the individual State Watchers.

For Undo, I added a unregister method for cleaning up the events. I also added the @undoMap() decorator. 

Added more asserts that can take a function in the case where the message requires an array mapping.


server/game/core/stateWatcher/StateWatcher.ts
```diff
@@ -41,35 +44,43 @@ export abstract class StateWatcher<TState> {
...
-        card: Card
+        registrar: StateWatcherRegistrar
     ) {
-        this.game = game;
+        super(game);
         this.name = name;
-        this.registrar = registrar;
-        this.registrationKey = card.internalName + '.' + name;
 
-        if (registrar.isRegistered(this.registrationKey)) {
```
AMMayberry1, line 52: I'm having a little trouble finding an equivalent to this - we do have something in the new version that keeps us from registering the same watcher type multiple times, right?

AMMayberry1 (approved):

PR #1594, opened 2025-08-21
Feature: Zone Undo Decorator Conversion

CDMayberry:
Converted all of the Zone classes to use decorators. Added some additional functionality to @undoArray to optionally allow for the use of in-place modification methods. Using it adds another object but might help make conversions easier.

Included a little more code for @registerState to prepare to move away from structuredClone.

### DIFF
--- a/server/game/core/GameObjectBase.ts
+++ b/server/game/core/GameObjectBase.ts
@@ -48,8 +48,9 @@ type UnwrapRefProperty<T> = T extends GameObjectRef<infer U> ?
     (T extends (infer R)[] ? (R extends GameObjectRef<infer U> ? U[] : R[]) :
         T);
 
+// NOTE: We are *temporarily* marking registerState as useFullCopy = true, but in the future this should be removed and moved into the deriving classes as need.
 /** GameObjectBase simply defines this as an object with state, and with a unique identifier. */
-@registerState()
+@registerState(true)
 export abstract class GameObjectBase<T extends IGameObjectBaseState = IGameObjectBaseState> implements IGameObjectBase<T> {
     public readonly game: Game;
 
--- a/server/game/core/GameObjectUtils.ts
+++ b/server/game/core/GameObjectUtils.ts
@@ -14,7 +14,7 @@ const stateClassesStr: Record<string, string> = {};
 
 /**
  * Decorator to capture the names of any accessors flagged as @undoState, @undoObject, or @undoArray for copyState, and then clear the array for the next derived class to use.
- * @param useFullCopy If true, simply uses the bulk copy method rather than using any meta data. This is flagged *per prototype*, and only affects  This is going to be slower, but helps if we have state not easily capturable by the undo decorators.
+ * @param useFullCopy If true, simply uses the bulk copy method rather than using any meta data. This is going to be slower, but helps if we have state not easily capturable by the undo decorators.
  */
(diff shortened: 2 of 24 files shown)


test/helpers/CustomMatchers.js
```diff
@@ -1109,6 +1109,62 @@ var customMatchers = {
             }
         };
     },
+
+    /**
+     * Checks if the actual array contains at least the elements of the expected array. Required for the new UndoArray class.
+     * @param {jasmine.MatchersUtil} matchersUtil
+     */
+    toContainArray: function (matchersUtil) {
```
AMMayberry1, line 1117: Looks like this isn't being used, still needed?

server/game/core/GameObjectUtils.ts
```diff
@@ -334,4 +420,94 @@ class UndoMap<TValue extends GameObjectBase> extends Map<string, TValue> {
         (this.go.state[this.prop] as Map<string, GameObjectRef<TValue>>).clear(key);
         super.clear();
     }
+}
+
+class UndoArray<TValue extends GameObjectBase> extends Array<TValue> {
```
AMMayberry1, line 425: Extending the Array class for this makes me a bit nervous because if we miss overriding a method or if one gets added in an update, then the user will have unexpected results. Is there a way we could mask out any methods we haven't explicitly implemented, or maybe create a new class whole cloth that doesn't even have them?

CDMayberry: All of the above are already caught by the UndoArrayInternal Proxy. If a non-overridden method is called and it happens to change the array, it'll throw an error. Only the overriding functions are setting `this.accessing = true`, if anything else tries to set the array it'll throw an error because that was not flagged.

server/game/core/GameObjectUtils.ts
```diff
@@ -10,18 +10,25 @@ const stateArrayMetadata = Symbol();
...
 
 const stateClassesStr: Record<string, string> = {};
 
 /**
  * Decorator to capture the names of any accessors flagged as @undoState, @undoObject, or @undoArray for copyState, and then clear the array for the next derived class to use.
- * @param fullCopyOnly If true, simply uses the bulk copy method rather than using any meta data.
+ * @param useFullCopy If true, simply uses the bulk copy method rather than using any meta data. This is going to be slower, but helps if we have state not easily capturable by the undo decorators.
  */
-export function registerState<T extends GameObjectBase>() {
+export function registerState<T extends GameObjectBase>(useFullCopy = false) {
```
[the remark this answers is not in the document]
CDMayberry: Changed to a enum

server/game/core/GameObjectBase.ts
```diff
@@ -50,7 +50,7 @@ type UnwrapRefProperty<T> = T extends GameObjectRef<infer U> ?
 
 // NOTE: We are *temporarily* marking registerState as useFullCopy = true, but in the future this should be removed and moved into the deriving classes as need.
 /** GameObjectBase simply defines this as an object with state, and with a unique identifier. */
-@registerState(true)
+@registerState(CopyModeEnum.UseBulkCopy)
```
AMMayberry1, line 53: nit: our convention is to not have `Enum` as the suffix, so just `CopyMode`

AMMayberry1 (approved):

PR #1595, opened 2025-08-21
feat(logging): added max on server error counts per lobby/game

mariotee:


server/game/core/DiscordDispatcher.ts
```diff
@@ -262,7 +258,7 @@ export class DiscordDispatcher implements IDiscordDispatcher {
             },
             {
                 name: 'Stack Trace',
```
AMMayberry1, line 260: I think we can just remove this outright now that we have the text file

server/game/core/snapshot/GameStateManager.ts
```diff
@@ -3,17 +3,17 @@ import type { GameObjectBase, GameObjectRef, IGameObjectBaseState } from '../Gam
...
 import * as Helpers from '../utils/Helpers.js';
-import { DiscordDispatcher } from '../DiscordDispatcher';
 
 export interface IGameObjectRegistrar {
     register(gameObject: GameObjectBase | GameObjectBase[]): void;
     get<T extends GameObjectBase>(gameObjectRef: GameObjectRef<T>): T | null;
 }
 
 export class GameStateManager implements IGameObjectRegistrar {
+    private static readonly MaxServerErrorCount = 5;
```
AMMayberry1, line 13: Ah okay, I think I see the confusion - there's a couple different "limits" involved in this situation.

1. Server error limit: this is for type of error in the lobby where we want to catch an infinite loop of exceptions, so we set some cutoff on how many can happen in a single request and then bail out

3. Discord report limit: we want to avoid flooding the discord channel with messages from a single game that's gotten into a really bad state. For bug reports it's 3, I think that's probably a reasonable limit for this as well since the errors tend to be repetitive. Although it kind of depends on whether we want to have one limit for all server error types or a different limit for each, there's pros and cons.

In this case it looks like you're trying to set up a similar kind of limit as #1 but for the "Game Object UUID" error. This case is different though, we always want to send a discord message if we hit this error even one time.

So what I meant by centralizing on the discord handler was going one step further than what you have now and actually doing the limit tracking for #2 above within the discord handler itself. That should work regardless of whether we want one limit for all server error reports or different limits for each, though obviously the latter involves slightly more setup.

PR #1597, opened 2025-08-21
Feature: Undo V8 Serialization

CDMayberry:
Moved from using structuredClone to using v8.serialization. Also eliminated some additional bulk copy usages.

This includes V2 of @undoMap, which no longer requires the type to be Readonly. Still a bit experimental but it should cut down on creating new objects every time want to update a map.

### DIFF
--- a/server/game/core/GameObjectBase.ts
+++ b/server/game/core/GameObjectBase.ts
@@ -93,9 +93,14 @@ export abstract class GameObjectBase<T extends IGameObjectBaseState = IGameObjec
     /** Sets the state.  */
     public setState(state: T) {
         const oldState = this.state;
-        this.state = structuredClone(state);
-        copyState(this, state);
+        this.state = state;
+        copyState(this, this.state);
         this.afterSetState(oldState);
+    }
+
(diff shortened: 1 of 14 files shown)


server/game/core/GameObjectBase.ts
```diff
@@ -93,11 +93,16 @@ export abstract class GameObjectBase<T extends IGameObjectBaseState = IGameObjec
...
         const oldState = this.state;
-        this.state = structuredClone(state);
-        copyState(this, state);
+        this.state = state;
+        copyState(this, this.state);
         this.afterSetState(oldState);
     }
 
+    /** Be ***very*** careful with this function. This returns a direct reference and should only be used for serialization, never keep this reference stored anywhere. */
+    public getStateUnsafe() {
```
AMMayberry1, line 102: If we're confident this will only ever be used in one place, can we change to a pattern where it's generated as a singleton function object? One example would be the pattern I used for ClearNewerSnapshotsBinding being passed from SnapshotFactory to SnapshotContainerBase

PR #1600, opened 2025-08-22
Take snapshots before mulligan and resourcing

Frugghi:


server/game/core/gameSteps/phases/SetupPhase.ts
```diff
@@ -29,6 +44,26 @@ export class SetupPhase extends Phase {
         );
     }
 
+    private takeSnapshotBeforeMulligan() {
```
AMMayberry1, line 47: nit: could we consolidate down to one function that takes a parameter for the timepoint? It can just be baked into the call within `SimpleStep`

test/scenarios/undo/SnapshotTypes.spec.ts
```diff
@@ -88,6 +175,106 @@ describe('Snapshot types', function() {
...
+                    context.player2.clickDone();
+
+                    expect(context.player1.resources.map((card) => card.internalName)).toEqual([
+                        'armed-to-the-teeth',
+                        'collections-starhopper',
+                    ]);
+                    expect(context.player2.resources.map((card) => card.internalName)).toEqual([
+                        'atst',
+                        'atst',
+                    ]);
```
AMMayberry1, line 276: Awesome job on the tests! Can we also add a couple tests to confirm that transitions within the setup phase work correctly? E.g. from resourcing to mulligan or mulligan to player choice.

Also just a note for future plans - I think an additional behavior we want is that if a player has finished selecting cards to resource and then hits undo, it rolls back to the beginning of the resource step instead of to the previous step. But I'm adding that kind of logic in my current PR for quick undo so I want to get that in first, or maybe I'll just add the setup stuff there as well if it looks like an easy extension. Just wanted to give a heads-up in case you have a similar thought.

Frugghi: I've added a test for going from resourcing to mulligan, going from mulligan to the start of the phase would be equivalent to test that is checking rolling back the phase, right?

AMMayberry1 (approved):

PR #1604, opened 2025-08-23
Added player count metrics

brianries:
- Added stats logging for the player count on the same 30 second interval as the rest of the stats
- Added a simple property override to display stats in dev mode

Here is the format of the outputs:

`2025-08-23T15:45:00.441Z [info]: [PlayerStats] Total User Count: 0 | Player Count: 0 (0.0% anon) | Spectator Count: 0 (0.0% anon)
`

EMF: -- Limited to just total and spectator to not claim all the 10 free custom metrics
image

### DIFF
--- a/server/game/core/snapshot/SnapshotManager.ts
+++ b/server/game/core/snapshot/SnapshotManager.ts
@@ -40,8 +40,8 @@ export class SnapshotManager {
     ]);
 
     private static readonly TestSnapshotLimits = new Map<SnapshotType, number>([
-        [SnapshotType.Action, 2],
-        [SnapshotType.Phase, 1],
+        [SnapshotType.Action, 3],
+        [SnapshotType.Phase, 2],
     ]);
 
     public readonly undoMode: UndoMode;


server/gamenode/GameServer.ts
```diff
@@ -867,6 +868,34 @@ export class GameServer {
         };
     }
 
+    private getGameAndPlayerCounts() {
+        let playerCount = 0;
+        let spectatorCount = 0;
+        let anonymousPlayerCount = 0;
+        let anonymousSpectatorCount = 0;
+
+        const totalGameCount = this.lobbies.size;
```
AMMayberry1, line 877: Ah sorry I should've been more specific - there might be lobbies that don't yet have active games so they aren't counted. You can reuse the logic from here (this is where we generate the game count for the main page): https://github.com/SWU-Karabast/forceteki/blob/6f9ad939366e88e3b78eb9125acd8e77f63dc7ea/server/gamenode/GameServer.ts#L854

AMMayberry1 (approved):

PR #1607, opened 2025-08-24
SwuStats - Added OAuth support

CheBato:
FE PR: https://github.com/SWU-Karabast/forceteki-client/pull/438
- Added OAuth support
- Stats get sent only if Premier
- Adjusted the handler to work with access_tokens.

tests:
Queue
- [x] Both players logged in, both players have swustats deck
- [x] Both players logged in, One player has swustats deck
- [x] Both players logged in, no player has swustats deck
- [x] Both players logged in, one player has unsaved deck one player has swustats deck
- [x] Both players logged in, one player has swudb deck one player has unsaved swustats deck
- [x] Both players logged in, one player has unsaved deck one player has swudb deck
- [x] Both players logged in, one player requeues with swustats deck, one player requeues with swudb deck
- [x] One player logged in, logged player has swustats deck unlogged has swustats
- [x] One player logged in, logged player has swudb deck unlogged has swudb
- [x] Both players not logged in both have swustats deck
- [x] Both players not logged in one has swustats deck one has swudb deck
- [x] Both players not logged in both on swustats decks and both requeue

Lobby
- [x] Both players logged in, both players have swustats deck
- [x] Both players logged in, One player has swustats deck
- [x] Both players logged in, no player has swustats deck
- [x] Both players logged in, one player has unsaved deck one player has swustats deck
- [x] Both players logged in, one player has swudb deck one player has unsaved swustats deck
- [x] One player logged in, logged player has swustats deck unlogged has swustats
- [x] One player logged in, logged player has swudb deck unlogged has swudb
- [x] Both players not logged in both have swustats deck
- [x] Both players not logged in one has swustats deck one has swudb deck

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -24,5 +24,6 @@
   },
   "bookmarks.saveBookmarksInProject": false,
   "files.eol": "\n",
-  "typescript.tsdk": "node_modules\\typescript\\lib"
+  "typescript.tsdk": "node_modules\\typescript\\lib",
+  "typescript.enablePromptUseWorkspaceTsdk": true
 }


server/gamenode/GameServer.ts
```diff
@@ -463,6 +477,51 @@ export class GameServer {
             }
         });
 
+        // This is called from the FE Node.js server hence why we do not need authentication.
```
AMMayberry1, line 480: Not sure what this means - how is it different than other API calls?

server/gamenode/GameServer.ts
```diff
@@ -463,6 +477,51 @@ export class GameServer {
...
+        app.post('/api/link-swustats', async (req, res, next) => {
+            try {
+                const { userId, swuStatsToken, internalApiKey } = req.body;
```
AMMayberry1, line 483: I know it would be hard to actually intercept, but I don't think we should be sending something as sensitive as the NEXTAUTH_SECRET over the wire. We should use some kind of signing method instead - I think we could just create a JWT with some basic data to validate, but there might be something better / simpler

server/gamenode/GameServer.ts
```diff
@@ -463,6 +477,51 @@ export class GameServer {
...
+            } catch (err) {
+                logger.error('GameServer (link-swustats) Server error:', err);
+                next(err);
+            }
+        });
+
+        app.post('/api/unlink-swustats', authMiddleware(), async (req, res, next) => {
+            try {
+                const user = req.user as User;
+                if (user.isAnonymousUser()) {
```
AMMayberry1, line 506: Should the link api have this check as well?

server/gamenode/GameServer.ts
```diff
@@ -463,6 +477,51 @@ export class GameServer {
...
+                        success: false,
+                        message: 'Forbidden'
+                    });
+                }
+                await this.userFactory.linkSwuStatsAsync(userId, swuStatsToken.refreshToken);
+                // add token mapping
+                this.swuStatsTokenMapping.set(userId, swuStatsToken);
+                return res.status(200).json({
+                    success: true,
+                    message: 'Swu stats linked successfully'
```
AMMayberry1, line 495: nit: I think it's either SwuStats or SWUStats. not sure if this will be shown to users or not. same for line 517

server/gamenode/GameServer.ts
```diff
@@ -1471,5 +1530,38 @@ export class GameServer {
...
     }
+
+    /**
+     * Clean up invalid/expired tokens from the user token map
+     */
+    private cleanupInvalidTokens(): void {
+        try {
+            const initialSize = this.swuStatsTokenMapping.size;
+            if (initialSize === 0) {
+                return;
```
AMMayberry1, line 1541: I would say let's remove this shortcut, the log is still informative if there are 0 tokens in the map (since that probably means there's an issue) and it's not really saving any compute time

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -60,18 +64,37 @@ interface PlayerData {
...
         this.apiKey = process.env.SWUSTATS_API_KEY;
+        this.clientId = process.env.SWUSTATS_CLIENT_ID;
+        this.clientSecret = process.env.SWUSTATS_CLIENT_SECRET;
+
         const isDev = process.env.ENVIRONMENT === 'development';
         if (!this.apiKey) {
             logger.warn('SWUStatsHandler: No API key configured. Stats may not be accepted by SWUstats.');
         }
+        if (!this.clientId || !this.clientSecret) {
+            logger.warn('SWUStatsHandler: No OAuth client credentials configured. Token refresh will not be available.');
```
AMMayberry1, line 96: Same as below, this should be an error case if it happens in prod. Also for both cases it would be really useful to explicitly log exactly which value is missing (we should update the error on 99 as well). That way if it fails on server launch we'll be able to quickly debug.

Maybe it's worth setting up `zod` or something at this point to validate the env vars for this?

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -241,15 +264,118 @@ export class SwuStatsHandler {
...
+    private async getAccessTokensAsync(
+        player1Details: PlayerDetails,
+        player2Details: PlayerDetails,
+    ): Promise<{ p1AccessToken: string | null; p2AccessToken: string | null }> {
+        const results = {
+            p1AccessToken: null,
+            p2AccessToken: null
+        };
+        // Handle Player1 swu token
+        if (player1Details.swuStatsToken && this.isTokenValid(player1Details.swuStatsToken)) {
```
AMMayberry1, line 296: We should have one private method that's called once each for p1 and p2 to reduce code duplication

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -241,15 +264,118 @@ export class SwuStatsHandler {
...
+            results.p1AccessToken = player1Details.swuStatsToken.accessToken;
+            logger.info(`SWUStatsHandler: Using existing valid access token for player 1 (${player1Details.user.getId()})`);
+        } else if (player1Details.swuStatsRefreshToken) {
```
AMMayberry1, line 299: What if this is also missing? Error?

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -241,15 +264,118 @@ export class SwuStatsHandler {
...
+     * Check if an access token is still valid (not expired)
+     * @param token The token to check
+     * @returns True if token is valid, false if expired
+     */
+    public isTokenValid(token: ISwuStatsToken): boolean {
+        const now = new Date();
+        const tokenCreationTime = new Date(token.creationDateTime);
+        const tokenExpirationTime = new Date(tokenCreationTime.getTime() + (token.timeToLive * 1000)); // timeToLive is in seconds
+
+        // Add a small buffer (30 seconds) to avoid using tokens that are about to expire
```
AMMayberry1, line 326: Good call. I would honestly expand it to like 5 minutes just to be uber-safe in case there's network delays or processing delays on the swustats side

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -241,15 +264,118 @@ export class SwuStatsHandler {
...
+        const bufferTime = 30000;
```
AMMayberry1, line 327: nit: bufferTimeMs

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -241,15 +264,118 @@ export class SwuStatsHandler {
...
+        player2Details: PlayerDetails,
+    ): Promise<{ p1AccessToken: string | null; p2AccessToken: string | null }> {
+        const results = {
+            p1AccessToken: null,
+            p2AccessToken: null
+        };
+        // Handle Player1 swu token
+        if (player1Details.swuStatsToken && this.isTokenValid(player1Details.swuStatsToken)) {
+            results.p1AccessToken = player1Details.swuStatsToken.accessToken;
+            logger.info(`SWUStatsHandler: Using existing valid access token for player 1 (${player1Details.user.getId()})`);
```
AMMayberry1, line 298: For all of the logs in this file we should be logging username, user id, and lobby id (using the log object schema that Brian set up)

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -241,15 +264,118 @@ export class SwuStatsHandler {
...
+        }
+        return isValid;
+    }
+
+    /**
+     * Refresh an access token using a refresh token
+     * @param refreshToken The refresh token to use
+     * @returns Promise that resolves to the new access token, or null if refresh failed
+     */
+    private async refreshAccessToken(refreshToken: string): Promise<string | null> {
```
AMMayberry1, line 346: nit: refreshTokenAsync

server/utils/user/UserFactory.ts
```diff
@@ -388,4 +389,45 @@ export class UserFactory {
             throw error;
         }
     }
+
+    /**
+     * Add or update the SWUstats refresh token for a user.
+     * - Idempotent: calling again overwrites the old token.
+     * - Does NOT return the token.
+     */
+    public async linkSwuStatsAsync(userId: string, refreshToken: string): Promise<void> {
```
AMMayberry1, line 398: We definitely need to do a check at this step to confirm that the refresh token works and return a failure otherwise. We might as well cache the result if it works, it'll probably be used shortly after this

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -241,15 +264,118 @@ export class SwuStatsHandler {
...
+            if (!response.ok) {
+                const errorText = await response.text();
+                logger.error(`SWUStatsHandler: Token refresh failed: ${response.status} - ${errorText}`);
+                return null;
+            }
+            const tokenResponse = await response.json() as OAuthTokenResponse;
+            logger.info('SWUStatsHandler: Successfully refreshed access token');
+            return tokenResponse.access_token;
+        } catch (error) {
+            logger.error('SWUStatsHandler: Failed to refresh access token', {
```
AMMayberry1, line 375: For these error cases, I think we need to find a way to show an error message for the user at the game-end screen letting them know something went wrong with sending results to swustats. Otherwise it'll be really hard for them or us to realize that something is wrong.

It could just be like an error message within the end-game settings popup.

server/utils/SWUStats/SwuStatsHandler.ts
CheBato: ZOD here actually isn't a bad idea.

server/utils/SWUStats/SwuStatsHandler.ts
CheBato: No at that point its just Null. Because one can have the token and the other doesn't. We return null if the person doesn't have a token/swustats OAuth linked

server/utils/SWUStats/SwuStatsHandler.ts
CheBato: Yessir. I'll do that asap

server/utils/user/UserFactory.ts
AMMayberry1: Oh I just mean actually use the refresh token to generate an access token in order to confirm that it works

server/utils/SWUStats/SwuStatsHandler.ts
AMMayberry1: Yep exactly that, I was going to suggest it

server/gamenode/Lobby.ts
```diff
@@ -1016,20 +1018,30 @@ export class Lobby {
...
                 logger.info(`Lobby ${this.id}: Successfully updated deck stats for game ${game.id}`);
-            }*/
+            }
         } catch (error) {
-            logger.error(`Lobby ${this.id}: Error updating deck stats:`, error);
+            logger.error(`Lobby ${this.id}: Error updating deck stats:`, error, { lobbyId: this.id });
+            // Send error message to client
+            this.playersDetails.forEach((playerDetail) => {
+                const lobbyUser = this.users.find((u) => u.id === playerDetail.user.getId());
+                lobbyUser.socket.send('statsSubmitError', {
```
AMMayberry1, line 1039: When I was testing, I hit an exception here because the lobbyUser's socket was missing - probably because they'd disconnected already? So maybe just needs `lobbyUser.socket?.send`

server/gamenode/Lobby.ts
```diff
@@ -1007,40 +1042,74 @@ export class Lobby {
...
+                        player1User,
+                        player2User,
+                        this.id,
+                        this.server,
+                        lobbyUser1,
+                        lobbyUser2
+                    );
+                    this.playersDetails.forEach((playerDetail) => {
+                        if (playerDetail.deckSource === DeckSource.SWUStats) {
+                            const lobbyUser = this.users.find((u) => u.id === playerDetail.user.getId());
```
AMMayberry1, line 1076: We should be using `this.getUser()` for this every time for safety since it confirms the user was found, unless there's a specific reason we're permitted to not find the correct user. After making that change, we won't need to use the `?` on any user object returned from that method

server/gamenode/Lobby.ts
```diff
@@ -1007,40 +1042,74 @@ export class Lobby {
...
+                            lobbyUser?.socket.send('statsSubmitNotification', {
+                                id: uuid(),
```
AMMayberry1, line 1078: What's the point of this parameter? Why would we need a new UUID for every message?

server/gamenode/Lobby.ts
```diff
@@ -1007,40 +1042,74 @@ export class Lobby {
...
+                                source: 'SWUStats',
+                                message: 'Deck stats successfully updated.'
+                            });
+                        }
+                    });
+                    logger.info(`Lobby ${this.id}: Successfully updated deck stats for game ${game.id}`);
+                } catch (error) {
+                    this.playersDetails.forEach((playerDetail) => {
+                        if (playerDetail.deckSource === DeckSource.SWUStats) {
+                            logger.error(`Lobby ${this.id}: Error updating deck stats to SWUStats:`, error, { lobbyId: this.id, userId: playerDetail.user.getId() });
```
AMMayberry1, line 1090: There should be only two parameters here at most. If both the error and a data object are included, they can be combined like this:

```
logger.error(`Lobby ${this.id}: Error updating deck stats to SWUStats:`, { error: { message: error.message, stack: error.stack }, lobbyId: this.id, userId: playerDetail.user.getId() });
```

Looks like line 1104 is doing it the same way, would appreciate if you could clean that up and any other instances you can find

server/gamenode/Lobby.ts
```diff
@@ -1007,40 +1042,74 @@ export class Lobby {
...
+                            lobbyUser?.socket.send('statsSubmitNotification', {
+                                id: uuid(),
+                                success: false,
+                                type: MessageTypes.Success,
+                                source: 'SWUStats',
+                                message: 'Deck stats successfully updated.'
+                            });
+                        }
+                    });
+                    logger.info(`Lobby ${this.id}: Successfully updated deck stats for game ${game.id}`);
```
AMMayberry1, line 1086: nit: lobbyId parameter needs to be set in the logger data (the second parameter object). Game ID is really not useful right now separate from the lobby ID, but if we do want to include it, it should also be in the logger object.

There's no harm in also including these in the text part, but for log searching and filtering it's critical that they be in the data parameter

server/utils/deck/DeckService.ts
```diff
@@ -254,9 +254,9 @@ export class DeckService {
             const dbService = await this.dbServicePromise;
             // Get the deck using our new flexible lookup method
             const deck = await this.getDeckByIdAsync(userId, deckId);
-
-            // If not found after all lookup methods.
-            Contract.assertNotNullLike(deck, `Deck with ID ${deckId} not found for user ${userId}`);
+            if (!deck) {
```
AMMayberry1, line 257: This feels wrong, it should be the caller's responsibility to confirm that the deck is a saved deck before trying to call this method, since that information should be provided along with the deck data. Otherwise we're just wasting time on a DB call we know will fail

server/gamenode/GameServer.ts
```diff
@@ -365,7 +384,7 @@ export class GameServer {
                 //         logger.error(`GameServer (get-user): Error with syncing Preferences for User ${user.getId()}`, err);
                 //     }
                 // }
-                return res.status(200).json({ success: true, user: { id: user.getId(), username: user.getUsername(), showWelcomeMessage: user.getShowWelcomeMessage(), preferences: user.getPreferences(), needsUsernameChange: user.needsUsernameChange() } });
+                return res.status(200).json({ success: true, user: { id: user.getId(), username: user.getUsername(), showWelcomeMessage: user.getShowWelcomeMessage(), preferences: user.getPreferences(), needsUsernameChange: user.needsUsernameChange(), swuStatsRefreshToken: user.getSwuStatsRefreshToken() } });
```
AMMayberry1, line 387: nit: could we split this into multiple lines, very hard to read at this point

server/gamenode/GameServer.ts
```diff
@@ -493,6 +512,55 @@ export class GameServer {
             }
         });
 
+        // This is called from the FE Node.js server hence why we do not need authentication.
```
AMMayberry1, line 515: nit: just to avoid other folks getting confused the way I did, let's change this comment to clarify that we are authenticating the server but do not need _**user**_ authentication

server/gamenode/GameServer.ts
```diff
@@ -493,6 +512,55 @@ export class GameServer {
...
+                if (process.env.ENVIRONMENT === 'development' && !process.env.INTRASERVICE_SECRET) {
+                    throw new Error('Environment variables INTRASERVICE_SECRET not set');
+                }
+                if (internalApiKey !== process.env.INTRASERVICE_SECRET) {
+                    return res.status(403).json({
+                        success: false,
+                        message: 'Forbidden'
+                    });
+                }
+                const newRefreshToken = await this.swuStatsHandler.refreshTokensAsync(swuStatsToken.refreshToken);
```
AMMayberry1, line 528: Let's maybe add a comment here to explain that we are testing the refresh token to confirm it works

server/gamenode/GameServer.ts
```diff
@@ -493,6 +512,55 @@ export class GameServer {
             }
         });
 
+        // This is called from the FE Node.js server hence why we do not need authentication.
+        app.post('/api/link-swustats', async (req, res, next) => {
+            try {
+                const { userId, swuStatsToken, internalApiKey } = req.body;
```
AMMayberry1, line 518: I'm sorry to be raising this so late, but I realized I'm not super clear here on why we're using a different auth method than any other API. Why would we not require user authentication info here?

server/gamenode/GameServer.ts
```diff
@@ -493,6 +512,55 @@ export class GameServer {
...
+                if (internalApiKey !== process.env.INTRASERVICE_SECRET) {
+                    return res.status(403).json({
+                        success: false,
+                        message: 'Forbidden'
+                    });
+                }
+                const newRefreshToken = await this.swuStatsHandler.refreshTokensAsync(swuStatsToken.refreshToken);
+                await this.userFactory.addSwuStatsRefreshTokenAsync(userId, newRefreshToken.refreshToken);
+                // add token mapping
+                this.swuStatsTokenMapping.set(userId, newRefreshToken);
```
AMMayberry1, line 531: Is this still needed? Looks like the handler is doing this update already

server/gamenode/GameServer.ts
```diff
@@ -1502,6 +1570,36 @@ export class GameServer {
...
+     * Clean up invalid/expired tokens from the user token map
+     */
+    private cleanupInvalidTokens(): void {
+        try {
+            const initialSize = this.swuStatsTokenMapping.size;
+            let removedCount = 0;
+            // Iterate through all tokens and remove invalid ones
+            for (const [userId, token] of this.swuStatsTokenMapping.entries()) {
+                if (!this.swuStatsHandler.isTokenValid(token)) {
+                    this.swuStatsTokenMapping.delete(userId);
```
AMMayberry1, line 1583: Just caught this - it's bad practice to modify the length of a collection you're iterating through, it can create issues. When doing a filter operation like this it's better to create a new map and add the "keep" entries to it

server/gamenode/Lobby.ts
```diff
@@ -944,20 +952,49 @@ export class Lobby {
     /**
      * Private method to update a players stats
      */
-    private async updatePlayerStatsAsync(playerUser: PlayerDetails, opponentPlayerUser: PlayerDetails, score: ScoreType) {
+    private async updatePlayerStatsAsync(playerUser: PlayerDetails, opponentPlayerUser: PlayerDetails, score: ScoreType, lobbyUser: LobbyUser) {
```
AMMayberry1, line 955: nit: now that we have two stats sources, would be good to always be clear which one is being referred to when it could be either. so maybe name this `updateKarabastPlayerStatsAsync` or something in that vein

I don't think the method names for the deckService need to be updated, since it's pretty obvious from context that those can't be related to swustats

server/gamenode/Lobby.ts
```diff
@@ -1005,33 +1042,77 @@ export class Lobby {
...
-                logger.error(`Lobby ${this.id}: Missing deck information (${player1User.deckID}) for player1 ${player1.id}`);
-                return;
+                logger.error(`Lobby ${this.id}: Missing information (${player1User}) for player1 ${player1.id}`, { lobbyId: this.id, userId: player1.id });
+                throw new Error(`Missing information (${player1User}) for player1 ${player1.id}`);
             }
             if (!player2User) {
-                logger.error(`Lobby ${this.id}: Missing information (${player2User.deckID}) for player2 ${player2.id}`);
-                return;
+                logger.error(`Lobby ${this.id}: Missing information (${player2User}) for player2 ${player2.id}`, { lobbyId: this.id, userId: player2.id });
+                throw new Error(`Missing information (${player1User}) for player1 ${player1.id}`);
```
AMMayberry1, line 1050: This is using player1's information

server/gamenode/Lobby.ts
```diff
@@ -1005,33 +1042,77 @@ export class Lobby {
...
             // Send to SWUstats if handler is available
-            if (eitherFromSWUStats) {
-                await this.server.SwuStatsHandler.sendGameResultAsync(
-                    game,
-                    this.playersDetails.find((u) => u.user.getId() === player1.id).deckLink,
-                    this.playersDetails.find((u) => u.user.getId() === player2.id).deckLink,
-                );
-                logger.info(`Lobby ${this.id}: Successfully updated deck stats for game ${game.id}`);
-            }*/
+            if (eitherFromSWUStats && this.format === SwuGameFormat.Premier) {
```
AMMayberry1, line 1063: This if statement is huge, it should be its own method

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -130,16 +177,16 @@ export class SwuStatsHandler {
             });
             if (!response.ok) {
                 const errorText = await response.text();
-                logger.error(`SWUstats API returned error: ${response.status} - ${errorText}`);
-                return false;
+                logger.error(`SWUstats API returned error: ${response.status} - ${errorText}`, { lobbyId });
+                throw new Error(errorText);
```
AMMayberry1, line 181: Will this error message string be shown to the player? If not, then I would remove the `logger.error` line and just use that text for the exception body

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -130,16 +177,16 @@ export class SwuStatsHandler {
...
             }
-
-            logger.info(`Successfully sent game result to SWUstats for game ${game.id}`);
+            logger.info(`Successfully sent game result to SWUstats for game ${game.id}`, { lobbyId });
             return true;
         } catch (error) {
             logger.error('Failed to send game result to SWUstats', {
                 error: { message: error.message, stack: error.stack },
-                gameId: game?.id
+                gameId: game?.id,
```
AMMayberry1, line 188: nit: `game` is not optional, don't need the `?`

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -60,39 +68,61 @@ interface PlayerData {
...
-     * @param player2DeckId Player 2 deck ID
+     * @param player1Details Details about player1
+     * @param player2Details Details about player2
+     * @param lobbyId the lobby id
+     * @param serverObject the server object from where we gain access to the user x accessToken
+     * @param player1LobbyUser player1 lobby object for sending info via socket
+     * @param player2LobbyUser player2 lobby object for sending info via socket
      * @returns Promise that resolves to true if successful, false otherwise
      */
     public async sendGameResultAsync(
```
AMMayberry1, line 113: nit: we should update this name to mention swustats

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -241,15 +291,108 @@ export class SwuStatsHandler {
...
+            logger.info(`SWUStatsHandler: Using existing valid access token for player (${playerDetails.user.getId()})`, { lobbyId, userId: playerDetails.user.getId() });
+        } else if (playerDetails.swuStatsRefreshToken) {
+            // Token is expired or doesn't exist, refresh it
+            logger.info(`SWUStatsHandler: Access token expired or missing for player (${playerDetails.user.getId()}), refreshing...`, { lobbyId, userId: playerDetails.user.getId() });
+            const resultTokens = await this.refreshTokensAsync(playerDetails.swuStatsRefreshToken);
+            serverObject.swuStatsTokenMapping.set(playerDetails.user.getId(), resultTokens);
+            playerAccessToken = resultTokens.accessToken;
+            await this.userFactory.addSwuStatsRefreshTokenAsync(playerDetails.user.getId(), resultTokens.refreshToken);
+        }
+        return playerAccessToken;
```
AMMayberry1, line 333: Looks like we're intentionally returning `null` if there's no swustats registration info present - in that case, we need to (a) update the return type to make it clearer, and (b) add an `@return` to the jsdoc explaining the behavior

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -218,21 +265,24 @@ export class SwuStatsHandler {
     /**
      * Build the game result payload for SWUstats API
      */
-    private buildGameResultPayload(
+    private async buildGameResultPayloadAsync(
```
AMMayberry1, line 268: nit: should also mention swustats in the method name

server/gamenode/Lobby.ts
```diff
@@ -1005,33 +1109,34 @@ export class Lobby {
...
-                logger.error(`Lobby ${this.id}: Missing information (${player2User.deckID}) for player2 ${player2.id}`);
-                return;
+                logger.error(`Lobby ${this.id}: Missing information (${player2User}) for player2 ${player2.id}`, { lobbyId: this.id, userId: player2.id });
+                throw new Error(`Missing information (${player2User}) for player1 ${player2.id}`);
             }
 
-            await this.updatePlayerStatsAsync(player1User, player2User, player1Score);
-            await this.updatePlayerStatsAsync(player2User, player1User, player2Score);
+            await this.updateKarabastPlayerStatsAsync(player1User, player2User, player1Score);
+            await this.updateKarabastPlayerStatsAsync(player2User, player1User, player2Score);
```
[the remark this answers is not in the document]
CheBato: If the first fails without the try/catch it will skip the second and swustats call.

server/gamenode/GameServer.ts
```diff
@@ -492,6 +519,56 @@ export class GameServer {
             }
         });
 
+        // This endpoint is being called by the FE server and not the client which is why we are authenticating the server.
+        app.post('/api/link-swustats', async (req, res, next) => {
```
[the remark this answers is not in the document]
CheBato: Its not somthing we can control unfortunately.

server/gamenode/Lobby.ts
```diff
@@ -1005,33 +1109,34 @@ export class Lobby {
             const player2User = this.playersDetails.find((u) => u.user.getId() === player2.id);
```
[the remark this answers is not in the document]
AMMayberry1: I think 2 is good

server/gamenode/GameServer.ts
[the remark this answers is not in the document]
AMMayberry1: Gotcha - do you know how that would have to be implemented? Trying to figure out how much information I can give Oot

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -114,32 +114,18 @@ export class SwuStatsHandler {
...
-                    type: SwuStatsSubmitStatus.Warning,
-                    source: 'swustats',
-                    message: 'Draws are currently not supported by SWUStats.'
-                });
-                logger.info(`Game ${game.id} ended in a draw or without clear winner, not sending to SWUstats`, { lobbyId });
-                return false;
+                logger.info(`Game ${game.id} ended in a draw or without clear winner, not sending to SWUStats`, { lobbyId });
+                return { type: StatsSaveStatus.Warning,
+                    source: StatsSource.SwuStats,
+                    message: 'Draws are currently not supported by SWUStats.' };
```
AMMayberry1, line 128: nit: remove period at the end

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -170,17 +156,19 @@ export class SwuStatsHandler {
...
+                message: 'Successfully sent game result to SWUStats' };
         } catch (error) {
-            logger.error('Failed to send game result to SWUstats', {
+            logger.error('Failed to send game result to SWUStats', {
                 error: { message: error.message, stack: error.stack },
                 gameId: game.id,
                 lobbyId
             });
-            throw new Error('Failed to send game result to SWUstats');
+            throw new Error(error);
```
AMMayberry1, line 171: nit: should just be `throw error;`

PR #1632, opened 2025-09-04
Budget Scheming, Dismantle the Conspiracy, Taylander Shuttle

sethhelman:
dismantle the conspiracy
budget scheming
taylander shuttle


server/game/cards/06_SEC/events/BudgetScheming.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    public override setupCardAbilities(registrar: IEventAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.setEventAbility({
+            title: 'Give an Experience token to up to 3 Official units',
+            targetResolver: {
+                mode: TargetMode.UpTo,
+                numCards: 3,
+                cardCondition: (card) => card.hasSomeTrait(Trait.Official),
+                immediateEffect: AbilityHelper.immediateEffects.giveExperience()
+            }
```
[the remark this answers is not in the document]
moyerr: Thanks for checking! I think that's good enough. I believe it makes sense because the `GiveExperienceSystem` should only be able to target unit cards 👍

test/server/cards/06_SEC/events/DismantleTheConspiracy.spec.ts
```diff
@@ -0,0 +1,54 @@
+describe('Dismantle the Conspiracy', function() {
+    integration(function(contextRef) {
+        it('Dismantle the Conspiracy\'s ability should allow you to capture any number of enemy non-leader units with 7 or less HP', async function () {
```
moyerr, line 3: I think this is a case where we would want to ensure the user is prompted if there are no valid targets for the capture part of the card. Can we add a test that covers that?

test/server/cards/06_SEC/events/DismantleTheConspiracy.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+                player1: {
+                    hand: ['dismantle-the-conspiracy'],
+                    groundArena: [{ card: 'battlefield-marine', damage: 2 }, 'isb-agent']
+                },
+                player2: {
+                    hand: ['change-of-heart'],
+                    groundArena: ['wampa', 'greedo#slow-on-the-draw', 'death-star-stormtrooper'],
+                    spaceArena: ['wing-leader', 'consortium-starviper', 'avenger#hunting-star-destroyer'],
+                    leader: { card: 'iden-versio#inferno-squad-commander', deployed: true },
+                }
```
moyerr, line 15: can we start with Player 1 having a deployed leader, so we can ensure the leader unit is a valid target for the `friendlyUnit` resolver?

test/server/cards/06_SEC/events/DismantleTheConspiracy.spec.ts
sethhelman: check out the convo with Veld in the engine chat. Sounds like we're leaving this as is for now and there's something being checked deeper in the engine

PR #1634, opened 2025-09-04
[Bugfix] Fix issue where Bo-Katan leader's on-attack ability damages herself

### LINKED ISSUE #1631: Bo-Katan's leader unit ability will damage herself if no targets are chosen
### Description

When Bo-Katan leader is deployed, her on-attack ability will damage herself if another Mandalorian has attacked and no unit is chosen for the second damage prompt.

### Steps to Reproduce

Using the test setup below, follow these steps:

1. P1 attacks base with Koska Reeves
2. P2 attacks base with Death Star Stormtrooper
3. P1 attacks base with Bo-Katan
  a. Choose Death Star Stormtrooper for the first damage prompt
  b. Click "Choose Nothing" for the second damage prompt

**Result:** Bo-Katan deals 1 damage to herself
**Expectation:** No damage is dealt for the second prompt

### Test Setup

```json
{
    "phase": "action",
    "player1": {
        "groundArena": [
            "koska-reeves#loyal-nite-owl"
        ],
        "leader": {
            "card": "bokatan-kryze#princess-in-exile",
            "exhausted": false,
            "deployed": true
        },
        "hasInitiative": true
    },
    "player2": {
        "groundArena": [
            "death-star-stormtrooper"
        ]
    }
}
```

moyerr:
Fixes #1631


test/server/cards/02_SHD/leaders/BoKatanKryzePrincessInExile.spec.ts
```diff
@@ -186,6 +186,51 @@ describe('Bo-Katan Kryze, Princess in Exile', function() {
                 context.player1.clickCard(context.bokatanKryze);
                 expect(context.player2).toBeActivePlayer();
             });
+
+            it('No damage is dealt if no targets are chosen for the second prompt', function () {
```
AMMayberry1, line 190: nit: this test name threw me off a bit, it sounds like it's saying no damage is dealt _at all_ if no second target is chosen. in case we need to come back and do more work, can we clarify that it means no damage is dealt at the second damage step but the first ping still happens

moyerr: Good catch! Yep, that was a result of me actually changing the test implementation itself and not updating the title 😄

AMMayberry1 (approved):

PR #1640, opened 2025-09-05
[Draft] Disclose implemented as Card Ability

moyerr:


server/game/gameSystems/DiscloseAspectsSystem.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    // Private helpers
+
+    private static buildSelectCardProps<TContext extends AbilityContext = AbilityContext>(
+        propertyFactory: PropsFactory<IDiscloseAspectsProperties, TContext>
+    ): PropsFactory<ISelectCardProperties<TContext>, TContext> {
+        const makeProps = (properties: IDiscloseAspectsProperties) => ({
+            zoneFilter: ZoneName.Hand,
+            controller: RelativePlayer.Self,
+            mode: TargetMode.ExactlyVariable as (TargetMode.ExactlyVariable | TargetMode.UpToVariable),
```
AMMayberry1, line 63: Just curious, what caused the need for the typecast here? Feels weird that we're casting into a _less_ specific enum type

test/server/cards/06_SEC/units/MinaBonteriStopThisWar.spec.ts
AMMayberry1, line 1: nit: the "I" in the filename should be capitalized

test/server/cards/06_SEC/units/MinaBonteriStopThisWar.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+                expect(context.player2).toHaveExactViewableDisplayPromptCards([
+                    context.command,
+                    context.c3poProtocolDroid
+                ]);
+                expect(context.player2).toHaveEnabledPromptButton('Done');
+                context.player2.clickDone();
+
+                // Player 1 draws a card
+                expect(context.player1.hand.length).toBe(6);
+            });
```
AMMayberry1, line 74: We should probably add a few more types of combinations of card aspects to make sure everything works. Probably a lot of this should get moved to a specific test suite for the Disclose keyword in the same folder as the other keyword tests.

Here's what I would recommend to start, plus any more you can think of:

- Could we disclose extra by selecting Battlefield Marine to get 1 out of 2 Command, and _then_ add the card "Command" thus ending up unnecessarily disclosing Battlefield Marine? Not sure if this is game legal (I would assume no)
- How does it work with neutral cards right now? Again, I'm assuming it's not legal to include them in the disclose, we should make sure they're excluded by the filter
- We should make sure it doesn't choke if we end up disclosing extra copies of one of the target aspects - e.g., if we disclose two Battlefield Marines (extra Heroism)
- Same as above, except we disclose a card with an aspect that _isn't_ in the filter (e.g. Aggression)

server/game/core/utils/Helpers.ts
```diff
@@ -42,6 +43,37 @@ export function countUniqueAspects(cards: Card | Card[]): number {
...
+                case Aspect.Villainy:
+                    return 'Villainy';
+                default:
+                    return 'Unknown';
+            }
+        })
+        .join(', ');
+}
+
+export function countOccurrences<T>(array: T[]): Map<T, number> {
```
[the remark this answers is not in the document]
moyerr: This is counting the total occurrences though. For example, if `Aspect.Command` occurs twice in the input array, it's resulting entry should be `[Aspect.Command: 2]`

server/game/gameSystems/DiscloseAspectsSystem.ts
moyerr: Yeah, this was a solution to a weird typescript error. Lemme know if there's a better solution, but I believe the error was because `ICardExactlyUpToVariableTargetResolver` requires that `mode` is `TargetMode.ExactlyVariable | TargetMode.UpToVariable`. So something in the relationship between `ISelectCardProperties` and `ICardTargetResolver` must have been telling the compiler that the `mode` property needed to be explicitly typed this way?

moyerr: I think it just evaporates with some of the other changes though, so i can remove this!

test/server/cards/06_SEC/units/MinaBonteriStopThisWar.spec.ts
moyerr: Which one? For clarity, the name + subtitle are `Mina Bonteri, Stop This War`

test/server/cards/06_SEC/units/MinaBonteriStopThisWar.spec.ts
moyerr: Alright, got all these covered except the first one. That one is a little more complex and possibly not worth the effort since we do not know the exact rules yet. I'll reach out in Discord to discuss.

server/game/core/utils/Helpers.ts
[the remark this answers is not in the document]
AMMayberry1: Ohh gotcha, yeah makes sense. Maybe just a jsdoc comment then to explain the use, I just had a hard time wrapping my head around it for some reason lol

test/server/cards/06_SEC/units/MinaBonteriStopThisWar.spec.ts
AMMayberry1: Ope nvm

server/game/core/utils/Helpers.ts
```diff
@@ -49,6 +49,12 @@ export function aspectString(aspects: Aspect[]): string {
         .join(', ');
 }
 
+/**
+ * Counts the occurrences of each item in an array.
```
AMMayberry1, line 53: nit: each _distinct_ item, I think that's what was tripping me up when I first read this

server/game/core/cardSelector/ExactlyVariableXCardSelector.ts
```diff
@@ -19,19 +19,22 @@ export class ExactlyVariableXCardSelector<TContext extends AbilityContext = Abil
     }
 
     public override hasExceededLimit(selectedCards: Card[], context: TContext) {
-        return selectedCards.length > this.numCardsFunc(context);
+        return selectedCards.length > this.numCardsFunc(context, selectedCards);
     }
 
     public override defaultPromptString(context: TContext) {
-        const numCards = this.numCardsFunc(context);
+        const numCards = this.numCardsFunc(context, []);
```
AMMayberry1, line 26: nit: can we remove the `[]` now? Same for other places it was added

server/game/gameSystems/DiscloseAspectsSystem.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public override hasLegalTarget(context: TContext, additionalProperties?: Partial<IDiscloseAspectsProperties>, mustChangeGameState?: GameStateChangeRequired): boolean {
+        return this.generateSelectCardSystem(context, additionalProperties)
+            .hasLegalTarget(context, null, mustChangeGameState);
+    }
+
+    public override canAffectInternal(player: Player, context: TContext, additionalProperties?: Partial<IDiscloseAspectsProperties>, mustChangeGameState?: GameStateChangeRequired): boolean {
+        const selectCardSystem = this.generateSelectCardSystem(context, additionalProperties);
+        return player.hand.some((card) => selectCardSystem.canAffectInternal(card, context));
```
AMMayberry1, line 40: This is a bit fiddly but I think this is slightly off. The most correct meaning the "canAffect" return value here is "can this system affect the target player (the `player` parameter)". I don't think we're currently evaluating that correctly, since the `selectCardSystem` will just implicitly use the `context.player` value as the "acting" player when it evaluates if it has legal targets or not.

I think what we would need to do here for total correctness is:

1. Make a copy of the `context` where `context.player = player`. This can be done with `const newContext = context.copy({ player })`

2. Return the value of `selectCardSystem.hasLegalTarget()` instead of evaluating each individual card, since all we care about is whether there is any combination of cards for the chosen player that will satisfy the selector

server/game/gameSystems/DiscloseAspectsSystem.ts
```diff
@@ -0,0 +1,102 @@
...
+    public eventHandler(): void {}
+
+    public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext, additionalProperties: Partial<IDiscloseAspectsProperties> = {}): void {
+        super.queueGenerateEventGameSteps(events, context, additionalProperties);
+
+        this.generateSelectCardSystem(context, additionalProperties)
+            .queueGenerateEventGameSteps(events, context);
+    }
+
+    public override hasLegalTarget(context: TContext, additionalProperties?: Partial<IDiscloseAspectsProperties>, mustChangeGameState?: GameStateChangeRequired): boolean {
```
AMMayberry1, line 33: I don't think we need this override, the default behavior of `hasLegalTarget` is probably good enough once we address the behavior of `canAffectInternal()` as described below

server/game/core/cardSelector/ExactlyVariableXCardSelector.ts
moyerr: No, the `numCardsFunc` is a property, and properties can't have default args 😭

server/game/gameSystems/DiscloseAspectsSystem.ts
moyerr: Making these changes results in failures for the following two test cases:
- Hand cannot satisfy the aspect requirements (a card is still drawn)
- Player clicks "Choose nothing" (a card is still drawn)

PR #1670, opened 2025-09-13
sly moore, sly moore and chancellor palpatine

VJubert:

### DIFF
--- a/server/game/cards/06_SEC/units/ChancellorPalpatineIAmTheSenate.ts
+++ b/server/game/cards/06_SEC/units/ChancellorPalpatineIAmTheSenate.ts
@@ -1,0 +1,39 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { EventName, KeywordName } from '../../../core/Constants';
+
+export default class ChancellorPalpatineIAmTheSenate extends NonLeaderUnitCard {
+    protected override getImplementationId () {
+        return {
+            id: '7936097828',
+            internalName: 'chancellor-palpatine#i-am-the-senate',
+        };
+    }
+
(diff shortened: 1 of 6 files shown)


server/game/cards/06_SEC/units/ChancellorPalpatineIAmTheSenate.ts
```diff
@@ -0,0 +1,39 @@
...
+    protected override getImplementationId () {
+        return {
+            id: '7936097828',
+            internalName: 'chancellor-palpatine#i-am-the-senate',
+        };
+    }
+
+    public override setupCardAbilities (registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addWhenPlayedAbility({
+            title: 'If you control a leader unit, create 2 Spy tokens and give those tokens Sentinel for this phase',
```
AMMayberry1, line 16: nit: shouldn't have the "if" part

server/game/cards/06_SEC/units/ChancellorPalpatineIAmTheSenate.ts
```diff
@@ -0,0 +1,39 @@
...
+            immediateEffect: abilityHelper.immediateEffects.conditional({
+                condition: (context) => context.player.hasSomeArenaUnit({ condition: (card) => card.isLeaderUnit() }),
```
AMMayberry1, line 18: nit: I think we can filter directly on card type (LeaderUnit) instead of using a filter

server/game/cards/06_SEC/units/ChancellorPalpatineIAmTheSenate.ts
```diff
@@ -0,0 +1,39 @@
...
+                onTrue: abilityHelper.immediateEffects.createSpy({ amount: 2 }),
+            }),
+            then: (thenContext) => ({
```
AMMayberry1, line 21: It's probably simpler to have this as an "ifYouDo", that's basically the effect even though it uses a different word. Then we can just remove the `thenCondition`

PR #1671, opened 2025-09-13
no one ever knew, b2emo, premor

VJubert:

### DIFF
--- a/test/server/cards/06_SEC/events/NoOneEverKnew.spec.ts
+++ b/test/server/cards/06_SEC/events/NoOneEverKnew.spec.ts
@@ -1,0 +1,34 @@
+describe('No One Ever Knew', function() {
+    integration(function(contextRef) {
+        it('For each friendly Official unit, exhaust an enemy unit', async function () {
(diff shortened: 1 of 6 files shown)


test/server/cards/06_SEC/events/NoOneEverKnew.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+            context.player1.clickPrompt('Done');
+
+            // Verify selected enemy units are exhausted and the other is not
+            expect(context.wampa.exhausted).toBeTrue();
+            expect(context.battlefieldMarine.exhausted).toBeTrue();
+            expect(context.pykeSentinel.exhausted).toBeFalse();
+
+            // Priority passes
+            expect(context.player2).toBeActivePlayer();
+        });
```
AMMayberry1, line 32: Can we add a test checking what happens if there are fewer enemy units than officials, I remember this targeting mode had some issues with that sometimes

AMMayberry1 (approved):

PR #1676, opened 2025-09-14
When Has Become Now & It's Not Over Yet

VJubert:


AMMayberry1 (approved):

test/server/cards/06_SEC/events/ItsNotOverYet.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+            // A Spy token should be created for player1, in the ground arena and exhausted
+            const spies = context.player1.findCardsByName('spy');
+            expect(spies.length).toBe(1);
+            expect(spies).toAllBeInZone('groundArena');
+            expect(spies[0].exhausted).toBeTrue();
+
+            expect(context.player2).toBeActivePlayer();
+        });
+
+        it('may pass the optional ability; no unit is readied and no Spy is created', async function () {
```
AMMayberry1, line 55: nit: the test name says that "no Spy is created" which is misleading

PR #1681, opened 2025-09-15
Mandalorian Cleaning up nevarro

sethhelman:
mandalorian cleaning
Check capture logic bug shown in test


scripts/mockdata.js
```diff
@@ -285,6 +285,7 @@ const mockCards = [
         cost: 5,
         hasNonKeywordAbility: true,
         aspects: ['cunning', 'villainy'],
+        aspects: ['cunning', 'villainy'],
```
VJubert, line 288: same line added ?

scripts/mockdata.js
```diff
@@ -314,6 +315,108 @@ const mockCards = [
         arena: 'ground',
         internalName: 'elia-kane#false-convert',
     }),
+    buildMockCard({
```
VJubert, line 318: I think you can revert this file, main should have every cards

server/game/cards/06_SEC/units/TheMandalorianCleaningUpNevarro.ts
```diff
@@ -0,0 +1,33 @@
...
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addTriggeredAbility({
+            title: 'Capture an enemy non-leader unit',
+            when: {
+                onCardDefeated: (event, context) =>
+                    event.isDefeatedByAttacker && DefeatCardSystem.defeatSourceCard(event) === context.source
+            },
+            targetResolver: {
+                activePromptTitle: 'Choose a non-Leader unit',
+                mode: TargetMode.UpTo,
```
VJubert, line 24: the classic mode Exactly should work, you must add optional: true before

scripts/mockdata.js
sethhelman: yeah it did some weird stuff when I merged main into it not really sure what happened in this file

PR #1686, opened 2025-09-15
Add flag to track when confirmation to undo is required

Frugghi:


server/game/core/GameInterfaces.ts
```diff
@@ -23,6 +23,7 @@ export interface GameConfiguration {
     buildSafeTimeout: (callback: () => void, delayMs: number, errorMessage: string) => NodeJS.Timeout;
     userTimeoutDisconnect: (userId: string) => void;
     undoMode?: UndoMode;
+    enableConfirmationToUndo?: boolean;
```
Frugghi, line 26: This is not set anywhere at the moment because I'm not sure how we want to handle it.

server/game/core/snapshot/container/MetaSnapshotArray.ts
```diff
@@ -32,6 +35,24 @@ export class MetaSnapshotArray {
...
+                break;
+            case QuickRollbackPoint.Previous:
+                rollbackEntry = this.entries[this.entries.length - 2];
+                break;
+            default:
+                rollbackEntry = null;
+                break;
+        }
+
+        return rollbackEntry?.snapshotProperties() ?? null;
```
AMMayberry1, line 53: nit: is the `?? null` doing anything?

server/game/core/snapshot/SnapshotManager.ts
```diff
@@ -243,6 +243,21 @@ export class SnapshotManager {
...
     }
 
+    public requiresConfirmationToRollbackTo(settings: IGetSnapshotSettings): boolean {
+        switch (settings.type) {
+            case SnapshotType.Action:
+                return this.actionSnapshots.getSnapshotProperties(settings.playerId, this.checkGetOffset(settings.actionOffset))?.requiresConfirmationToRollback ?? false;
+            case SnapshotType.Manual:
+                return this.manualSnapshots.get(settings.playerId)?.getSnapshotProperties(settings.snapshotId)?.requiresConfirmationToRollback ?? false;
+            case SnapshotType.Phase:
+                return this.phaseSnapshots.getSnapshotProperties(settings.phaseName, this.checkGetOffset(settings.phaseOffset))?.requiresConfirmationToRollback ?? false;
```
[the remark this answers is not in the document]
AMMayberry1: I think technically yes but right now we won't even be showing these buttons in any scenario where we are enforcing the check. So we can do it that way just a safeguard

server/game/core/Game.js
```diff
@@ -991,6 +994,10 @@ class Game extends EventEmitter {
     promptWithMenu(player, contextObj, properties) {
         Contract.assertNotNullLike(player);
 
+        if (this.actionPhaseActivePlayer && this.actionPhaseActivePlayer !== player) {
+            this.snapshotManager.setRequiresConfirmationToRollbackCurrentSnapshot(this.actionPhaseActivePlayer.id);
+        }
```
AMMayberry1, line 999: Would it make sense to just have the logic for this in `UIPrompt` itself so that all the prompt types get it, and in case it's created directly instead of being set up via one of these methods on `Game`?

server/game/core/Game.js
```diff
@@ -1972,9 +1981,39 @@ class Game extends EventEmitter {
...
+            if (!result) {
+                return;
+            }
+
+            this.addAlert(AlertType.Notification, '{0} has rolled back to {1}', this.getPlayerById(playerId), message);
+        };
+
+        if (this.enableConfirmationToUndo && this.snapshotManager.requiresConfirmationToRollbackTo(settings)) {
+            this.promptWithHandlerMenu(this.getPlayerById(playerId).opponent, {
+                activePromptTitle: `You opponent would like to rollback to ${message}. Are you sure you want to allow this?`,
```
AMMayberry1, line 1996: nit: "rollback" is mostly an internal term, I think for all player-facing stuff we should use "undo". Same note for the notification on 2004

server/game/core/GameInterfaces.ts
AMMayberry1: For now it's basically if the Lobby type is anything other than a private lobby

server/game/gameSystems/DrawSystem.ts
```diff
@@ -19,6 +20,11 @@ export class DrawSystem<TContext extends AbilityContext = AbilityContext> extend
     };
 
     public eventHandler(event): void {
+        const gameEvent = event as GameEvent;
+        if (gameEvent.context && gameEvent.context.player && event.player === gameEvent.context.player && event.amount > 0 && event.player.drawDeck.length > 0) {
```
AMMayberry1, line 24: I think we can strip out most of these checks since the event will have fizzled if amount is 0 or the deck is empty. Also I think generally we safely assume that `event.context` and `event.context.player` or set or else there'd be a major issue, though we could definitely add some `assertNotNullLike()` checks

test/helpers/IntegrationHelper.js
```diff
@@ -85,6 +85,10 @@ global.integration = function (definitions, enableUndo = false) {
                 getCurrentSnapshotId: () => gameFlowWrapper.snapshotManager?.currentSnapshotId,
                 getCurrentSnapshottedAction: () => gameFlowWrapper.snapshotManager?.currentSnapshottedAction,
                 rollbackToSnapshot: (settings) => newContext.game.rollbackToSnapshotInternal(settings),
+                quickRollback: (playerId) => {
+                    newContext.game.rollbackToSnapshot(playerId, { type: SnapshotType.Quick, playerId });
+                    newContext.game.continue();
```
AMMayberry1, line 90: Just curious, what made this change necessary?

test/scenarios/undo/Undo.spec.ts
```diff
@@ -1910,5 +1910,483 @@ describe('Undo', function() {
                 ]);
             });
         });
+
+        describe('Rollback confirmation', function() {
```
AMMayberry1, line 1914: Since this file is already massive I think this could merit its own dedicated file

test/scenarios/undo/Undo.spec.ts
```diff
@@ -1910,5 +1910,483 @@ describe('Undo', function() {
...
+                });
+
+                const { context } = contextRef;
+
+                // Generate a quick snapshot
+                context.player2.passAction();
+
+                context.player1.clickCard(context.infernoFour);
+                context.player1.clickDisplayCardPromptButton(context.sabineWren.uuid, 'top');
+                context.player1.clickDisplayCardPromptButton(context.battlefieldMarine.uuid, 'bottom');
```
AMMayberry1, line 2288: One thing I want to confirm about the design - could we roll back here _before_ resolving the prompt? I haven't checked the logic super thoroughly, but we should make sure that if someone tries to e.g. nope out as soon as they see what the two cards are that they can't do it freely

test/scenarios/undo/Undo.spec.ts
```diff
@@ -1910,5 +1910,483 @@ describe('Undo', function() {
...
+
+                expect(context.game.snapshotManager.canQuickRollbackWithoutConfirmation(context.player1.id)).toBeFalse();
+                expect(context.game.snapshotManager.canQuickRollbackWithoutConfirmation(context.player2.id)).toBeTrue();
+
+                contextRef.snapshot.quickRollback(context.player1.id);
+                expect(context.player2).toHaveConfirmUndoPrompt();
+
+                context.player2.clickPrompt('Allow');
+                expect(context.player1).toBeActivePlayer();
+            });
```
AMMayberry1, line 2389: Could we add a test with U-Wing where the opponent gets to do something in the middle of the plays, like they have some trigger that fires based on one of the units. That should switch us into request-undo mode fro the rest of the action

server/game/gameSystems/DrawSystem.ts
Frugghi: `player` is null when drawing cards during the setup/regroup phase. Is that a mistake?

test/helpers/IntegrationHelper.js
Frugghi: The undo confirmation prompt wasn't showing without this.

test/scenarios/undo/Undo.spec.ts
Frugghi: Update the test to include this!

server/game/gameSystems/DrawSystem.ts
Frugghi: I've updated the implementation to always include the player, let me know if you disagree!

test/scenarios/undo/Undo.spec.ts
Frugghi: U-Wing does a search so it'll trigger request-undo from the start. Endless Legion should work for this though!

AMMayberry1 (approved):

PR #1687, opened 2025-09-15
⟡ Elia Kane, False Convert

moyerr:
| Card | Ability prompt | Chat log |
| :---: | :---: | :---: |
|  |  |  |


test/server/cards/06_SEC/units/EliaKaneFalseConvert.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+                // No cards were moved
+                expect(context.restock).toBeInZone('resource', context.player2);
+                expect(context.atst).toBeInZone('resource', context.player2);
+                expect(context.cadBane).toBeInZone('resource', context.player2);
+                expect(context.lomPyke).toBeInZone('resource', context.player2);
+                expect(context.pykeSentinel).toBeInZone('deck', context.player2);
+
+                // Resources are in the correct state
+                expect(context.player2.readyResourceCount).toEqual(2);
+                expect(context.player2.exhaustedResourceCount).toEqual(2);
```
AMMayberry1, line 109: Few additional cases it would be good to have:
- the opponent has 2 resources
- opponent has 0 resources, obviously it's a wildly unlikely scenario but we should make sure nothing breaks at least
- opponent's deck is empty

AMMayberry1 (approved):

PR #1700, opened 2025-09-16
Fix issues with Plot resourcing cards even if the card is unable to b…

adbronico:
…e played


test/server/core/abilities/keyword/Plot.spec.ts
```diff
@@ -126,6 +126,23 @@ describe('Plot keyword', function() {
...
+                        resources: [{ card: 'sneaking-suspicion', exhausted: true }, { card: 'wampa', exhausted: true }, { card: 'wampa', exhausted: true }, { card: 'wampa', exhausted: true }],
+                        deck: ['pyke-sentinel']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.calKestis);
+                context.player1.clickPrompt('Deploy Cal Kestis');
+                expect(context.player2).toBeActivePlayer();
```
AMMayberry1, line 143: Could we also check that the "(No effect)" prefix is added to the UI prompt, since triggering it will do nothing?

moyerr: @AMMayberry1 I think in this case actually, the prompt button is skipped altogether since the triggered ability will have no effect, and there is only a single trigger on the stack

moyerr (approved):
LGTM! Thanks for covering those extra cases 💪

test/server/core/abilities/keyword/Plot.spec.ts
AMMayberry1: Ahh right, all good

PR #1701, opened 2025-09-16
#Moderation - mute

CheBato:
- Added ability to mute users for a X amount of time.


server/gamenode/GameServer.ts
```diff
@@ -388,7 +388,9 @@ export class GameServer {
                     showWelcomeMessage: user.getShowWelcomeMessage(),
                     preferences: user.getPreferences(),
                     needsUsernameChange: user.needsUsernameChange(),
-                    swuStatsRefreshToken: user.getSwuStatsRefreshToken()
+                    swuStatsRefreshToken: user.getSwuStatsRefreshToken(),
+                    isMuted: user.isMuted(),
+                    mutedUntil: user.getMutedUntil()
```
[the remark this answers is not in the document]
CheBato: i have not i'll do a test today.

server/services/DynamoDBInterfaces.ts
```diff
@@ -1,10 +1,17 @@
+export interface IModeration {
```
AMMayberry1, line 1: nit: I think we should use "ModerationAction" as the noun here and everywhere else. So e.g., on line 14, `moderationAction: IModerationAction`

server/utils/user/UserFactory.ts
```diff
@@ -429,4 +433,88 @@ export class UserFactory {
...
+
+                    // Update the user in the database with the new moderation data
+                    await dbService.updateUserProfileAsync(userData.id, {
+                        moderation: userData.moderation
+                    });
+
+                    logger.info(`UserFactory: Initialized moderation start date for user ${userData.id}`, {
+                        userId: userData.id
+                    });
+                } else {
```
AMMayberry1, line 460: nit: to improve readability, we can end the `if (!userData.moderation.startDays)` with a `return` instead of using an `else`. with the nested if blocks that will make it a little easier to follow imo

server/utils/user/UserFactory.ts
```diff
@@ -329,7 +331,9 @@ export class UserFactory {
                 showWelcomeMessage: true,
                 preferences: getDefaultPreferences(),
                 needsUsernameChange: false,
-                swuStatsRefreshToken: null
+                swuStatsRefreshToken: null,
+                mutedUntil: null,
```
AMMayberry1, line 335: nit: I think this isn't needed anymore?

server/services/DynamoDBInterfaces.ts
```diff
@@ -1,10 +1,17 @@
...
+    days: number;
```
AMMayberry1, line 2: nit: maybe `daysRemaining`

server/services/DynamoDBInterfaces.ts
```diff
@@ -1,10 +1,17 @@
...
+    startDays?: Date;
+    hasSeen?: boolean;
```
AMMayberry1, line 4: With setting it up this way we also need a fourth category which is `moderationType` which should be an enum. The only category right now is "Mute" but there could be other things down the road

server/utils/user/UserFactory.ts
```diff
@@ -429,4 +433,88 @@ export class UserFactory {
...
+     * Processes moderation logic for a user
+     * @param userData User data from database
+     * @returns Updated user data with moderation processed
+     */
+    private async processModerationAsync(userData: IUserProfileDataEntity): Promise<IUserProfileDataEntity> {
+        try {
+            const dbService = await this.dbServicePromise;
+
+            // Check if user has moderation field and it's not null
+            if (userData.moderation && userData.moderation.days) {
```
AMMayberry1, line 447: Let's add `Contract.assertNonNegative(userData.moderation.days)` here just to be uber-safe, in case we have a bug somewhere else and this gets set as a negative value somehow

server/utils/user/UserFactory.ts
```diff
@@ -442,12 +448,14 @@ export class UserFactory {
...
             const dbService = await this.dbServicePromise;
-
+            Contract.assertNonNegative(userData.moderation.daysRemaining);
             // Check if user has moderation field and it's not null
-            if (userData.moderation && userData.moderation.days) {
-                if (!userData.moderation.startDays) {
-                    userData.moderation.startDays = new Date();
+            if (userData.moderation && userData.moderation.daysRemaining) {
+                if (!userData.moderation.endDate) {
+                    const MS_PER_DAY = 24 * 60 * 60 * 1000;
```
AMMayberry1, line 455: nit: could we define this at like the file or class level so we're not recomputing it on every call

server/utils/user/UserFactory.ts
```diff
@@ -429,4 +439,87 @@ export class UserFactory {
...
+                    await dbService.updateUserProfileAsync(userData.id, {
+                        moderation: null
+                    });
+
+                    logger.info(`UserFactory: Cleared expired moderation for user ${userData.id}`, {
+                        userId: userData.id,
+                    });
+                }
+            }
+            return userData;
```
AMMayberry1, line 485: nit: our general goal with code layout is to minimize the amount of code that's indented, as it's easier to follow code at the top level instead of inside of an if or for block. in this case we could get an easy win by inverting the logic and doing the return at the top

```ts
if (!userData.moderation || !userData.moderation.daysRemaining) {
  return userData;
}

// function continues from here
```

server/utils/user/UserFactory.ts
```diff
@@ -429,4 +439,87 @@ export class UserFactory {
...
+                    userData.moderation.endDate = new Date(Date.now() + userData.moderation.daysRemaining * MS_PER_DAY);
+                    userData.moderation.hasSeen = false;
+                    userData.moderation.moderationType = ModerationType.Mute;
+
+                    // Update the user in the database with the new moderation data
+                    await dbService.updateUserProfileAsync(userData.id, {
+                        moderation: userData.moderation
+                    });
+
+                    logger.info(`UserFactory: Initialized moderation start date for user ${userData.id}`, {
```
AMMayberry1, line 465: Let's also log the `daysRemaining` and `endDate` values in case we need it for debugging

server/utils/user/UserFactory.ts
```diff
@@ -429,4 +439,87 @@ export class UserFactory {
...
+                        userId: userData.id
+                    });
+                    return userData;
+                }
```
AMMayberry1, line 469: nit: can we add a newline here just to visually separate the if block from what follows, since they're two different major code paths

server/utils/user/UserFactory.ts
```diff
@@ -429,4 +439,87 @@ export class UserFactory {
...
+            const userProfile = await dbService.getUserProfileAsync(userId);
+            Contract.assertNotNullLike(userProfile, `No user profile found for userId ${userId}`);
+
+            if (userProfile.moderation) {
+                userProfile.moderation.hasSeen = true;
+                await dbService.updateUserProfileAsync(userId, {
+                    moderation: userProfile.moderation
+                });
+
+                logger.info(`UserFactory: Set moderation as seen for user ${userId}`);
```
AMMayberry1, line 512: Needs the `{ userId }` argument

PR #1703, opened 2025-09-16
add Bo-Katan Kryze, Alone. source: 'https://media.discordapp.net/atta…

phever:
…chments/1079526596312047636/1414618440466956470/Screenshot_2025-09-08_at_15.07.01.jpg?ex=68cac596&is=68c97416&hm=697b967c7dbb0f0b0969d224dfe051b3a581c22bb1c312f7ae771068f8ed48ba&=&format=webp&width=449&height=633'

### DIFF
--- a/test/server/cards/06_SEC/units/BoKatanKryzeAlone.spec.ts
+++ b/test/server/cards/06_SEC/units/BoKatanKryzeAlone.spec.ts
@@ -1,0 +1,40 @@
+describe('Bo-Katan Kryze, Alone', function () {
+    integration(function (contextRef) {
+        it('Bo-Katan Kryze\'s ability should give all enemy units -3/-3', async function () {
+            await contextRef.setupTestAsync({
+                phase: 'action',
+                player1: {
+                    leader: { card: 'luke-skywalker#faithful-friend', deployed: false },
+                    hand: ['bokatan-kryze#alone'],
+                    groundArena: ['porg'],
+                },
+                player2: {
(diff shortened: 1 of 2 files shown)


test/server/cards/06_SEC/units/BoKatanKryzeAlone.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+                    groundArena: ['wampa', 'rampaging-wampa'],
+                }
+            });
+
+            const { context } = contextRef;
+
+            // play Bo-Katan Kryze, Alone
+            context.player1.clickCard(context.bokatanKryze);
+
+            // select friendly unit to give experience token to (porg)
```
AMMayberry1, line 21: We try to keep cases small and focused to a reasonable degree - could we split out the -3/-3 and "when defeated" abilities into two cases. Totally fine for them to share the same test setup, this can be done using a `describe()` block with a `beforeEach()` as you'll see in a lot of the test suites.

Also, typically we'll do a handful of "negative" cases to ensure the conditions are coded up correctly. In this case both of her abilities are using pretty well-established patterns so we don't need to do too much, the only extra cases I would recommend are:

- Test that if a friendly unit is defeated, her ability doesn't trigger
- We try to always add tests with No Glory, Only Results for any "when defeated" condition because the controller change on defeat can get a little fiddly. Could you add two tests with NGOR, one for each player, and confirm the correct behavior (i.e. her ability should trigger if her controller plays it but not if the opponent does)

PR #1713, opened 2025-09-19
Add Force Throw to the Ban List

moyerr:
Screenshot 2025-09-19 at 11 01 58 AM


adbronico:
I moved this to ready so we can get it out for tomorrow

PR #1716, opened 2025-09-20
Fixes and tests for attack intra-action rollback

CDMayberry:
Fix to clear any active actions if rolled back during an ongoing action


adbronico (approved):

server/game/core/card/propertyMixins/Damage.ts
```diff
@@ -150,5 +150,10 @@ export function WithDamage<TBaseClass extends CardConstructor<TState>, TState ex
 
             this.state.attackEnabled = enabledStatus;
         }
+
+        protected override afterSetState(oldState: any) {
+            // Active Attack is transitive and should always be null during the start of a turn.
```
AMMayberry1, line 155: nit: could we say "during the start of an **action**" just to be precise in case other people end up working on this part

PR #1717, opened 2025-09-20
convene the senate, gnk power droid, figure of unity

VJubert:

### DIFF
--- a/server/game/gameSystems/DistributeAmongTargetsSystem.ts
+++ b/server/game/gameSystems/DistributeAmongTargetsSystem.ts
@@ -86,6 +86,13 @@ export abstract class DistributeAmongTargetsSystem<
             }
         }
 
+        if (targets.length === 0) {
+            targets.push({
+                format: 'no effective {0}',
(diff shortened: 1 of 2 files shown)


test/server/cards/06_SEC/upgrades/FigureOfUnity.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+            expect(context.p2Base.damage).toBe(3);
+            expect(context.p1Base.damage).toBe(9);
+
+            context.player2.clickCard(context.jedhaAgitator);
+            context.player2.clickCard(context.p1Base);
+
+            expect(context.p1Base.damage).toBe(11); // 9 + 2
+            expect(context.p2Base.damage).toBe(3);
+        });
+        it('while the attached unit is ready, each other friendly unit gains Overwhelm, Raid 1, and Restore 1 (test with Oppo Rancisis)', async function () {
```
AMMayberry1, line 36: Let's also test with JTL Ghost with the upgrade attached to eg Ezra

PR #1727, opened 2025-09-22
chuchi and malastare

VJubert:


server/game/cards/06_SEC/units/CongressOfMalastare.ts
```diff
@@ -0,0 +1,46 @@
...
+                limit: AbilityHelper.limit.perRound(1)
+            }),
+        });
+    }
+
+    private isFirstUpgradePlayedByControllerThisPhase(card: Card) {
+        return card.isUpgrade() &&
+          !this.cardsPlayedThisPhaseWatcher.someCardPlayed((playedCardEntry) =>
+              playedCardEntry.playedBy === card.controller &&
+              playedCardEntry.card.isUpgrade() &&
```
AMMayberry1, line 42: I think this instead needs to use `playedCardEntry.playedAsType` to account for pilots, we should add a test to confirm. The test sequence would probably need to be:

1. Play a pilot onto a vehicle
2. Defeat the pilot so that it goes to discard and now `card.isUpgrade()` is false
3. Play another upgrade

PR #1737, opened 2025-09-26
set selectCardMode for displayMultipleForSelection

danbastin:
Ensures selectCardMode gets set for any card using this UIPrompt. Fixes display flicker for cards like Darth Vader Commanding the First Legion. 

Open to suggestions for better places to drop this or for other cards that need to be tested here.


server/game/core/gameSteps/prompts/DisplayCardsForSelectionPrompt.ts
```diff
@@ -18,16 +19,19 @@ export class DisplayCardsForSelectionPrompt extends DisplayCardPrompt<IDisplayCa
     private readonly selectedCardsHandler: (cards: Card[]) => void;
     private readonly showSelectionOrder: boolean;
     private readonly displayTextByCardUuid: Map<string, string>;
+    private readonly selectCardMode: SelectCardMode;
```
AMMayberry1, line 22: It would probably be good to force all of the display prompts to explicitly declare this. Can we move this up to the parent class `DisplayCardPrompt` and make it a constructor param that each class has to pass up via `super()`? If you're not sure how to set it for any particular prompt I can help

bwohldmann:
@AMMayberry1 Made the suggested refactor.  Still learning TypeScript conventions; so definitely welcome to constructive criticism.

PR #1746, opened 2025-09-29
One in a Million

VJubert:


server/game/cards/06_SEC/events/OneInAMillion.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    public override setupCardAbilities (registrar: IEventAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addConstantAbility({
+            title: 'This card can\'t be played from your hand',
+            sourceZoneFilter: WildcardZoneName.Any,
+            ongoingEffect: abilityHelper.ongoingEffects.cardCannot({
+                cannot: AbilityRestriction.Play,
+                restrictedActionCondition: (_, source) => source.zoneName === ZoneName.Hand,
+            })
+        });
```
moyerr, line 22: I wonder if we should add a `addPlayRestrictionAbility()` interface to InPlayCard, similar to what we have for `addDecreaseCostAbility()`. That way we can avoid having to add the `addConstantAbility` interface to `EventCard` where it usually doesn't make sense outside the narrow scope of play restrictions/cost adjusters.

AMMayberry1: Good callout, I agree. @VJubert let me know if you'd like help with that part if you want to give it a shot

AMMayberry1 (approved):

PR #1755, opened 2025-10-01
Cpu and heap performance runs

brianries:
Adds support for capturing cpu and heap performance runs of the server

### DIFF
--- a/server/game/abilities/keyword/exploit/ExploitCostAdjuster.ts
+++ b/server/game/abilities/keyword/exploit/ExploitCostAdjuster.ts
@@ -75,9 +75,9 @@ export class ExploitCostAdjuster extends TargetedCostAdjuster {
             const exploitAdjustment = this.adjustAmountPerTarget * selectedCards.length;
             const otherDiscount = context.costs[this.costPropertyName].otherDiscountsAmount ?? 0;
 
-            const totalCost = (adjustmentProps.totalResourceCost - exploitAdjustment) - otherDiscount;
+            const totalCost = (adjustmentProps.getTotalResourceCost() - exploitAdjustment) - otherDiscount;
 
-            return `Cost: ${adjustmentProps.totalResourceCost} Exploit: ${-exploitAdjustment} Discounts: ${-otherDiscount} Total: ${totalCost}`;
+            return `Cost: ${adjustmentProps.getTotalResourceCost()} Exploit: ${-exploitAdjustment} Discounts: ${-otherDiscount} Total: ${totalCost}`;
         };
 
         return handler;
--- a/server/game/core/cost/CostInterfaces.ts
+++ b/server/game/core/cost/CostInterfaces.ts
@@ -1,9 +1,9 @@
 import type { IUnitCard } from '../card/propertyMixins/UnitProperties';
-import type { Aspect } from '../Constants';
 import type { AdjustedCostEvaluator } from './evaluation/AdjustedCostEvaluator';
 import type { CostAdjuster, CostAdjustResolutionMode } from './CostAdjuster';
 import type { DynamicOpportunityCost } from './evaluation/DynamicOpportunityCost';
 import type { SimpleAdjustedCost } from './evaluation/SimpleAdjustedCost';
+import type { Aspect } from '../Constants';
 
 export enum CostAdjustStage {
     Standard_0 = 'standard_0',
@@ -29,10 +29,10 @@ export interface ICostAdjusterEvaluationTarget {
 }
 
 export interface IAbilityCostAdjustmentProperties {
-    totalResourceCost: number;
+    getTotalResourceCost: (includeAspectPenalties?: boolean) => number;
+    getPenaltyAspects: () => Aspect[];
     matchingAdjusters: Map<CostAdjustStage, CostAdjuster[]>;
     resourceCostType: ResourceCostType;
-    penaltyAspects?: Aspect[];
 }
 
 export interface ICostAdjustmentResolutionProperties extends IAbilityCostAdjustmentProperties {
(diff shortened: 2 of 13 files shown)


server/gamenode/GameServer.ts
```diff
@@ -650,6 +652,70 @@ export class GameServer {
         });
         // *** End of User Object calls ***
 
+        // PROFILING
+        app.get('/api/run-profile', async (req, res, next) => {
+            try {
+                const { secretKey, type, duration } = req.query;
```
AMMayberry1: nit: secrets are typically attached as headers, not query params. Probably not a big deal for us but the biggest difference is that headers aren't logged by default in most applications.

My suggestion would be using header format `Authorization api-key <API_KEY_VALUE>` just to keep aligned with best practices

server/gamenode/GameServer.ts
```diff
@@ -650,6 +652,70 @@ export class GameServer {
...
+                        success: false,
+                        message: 'Profile capture not configured - PROFILE_CAPTURE_SECRET not set'
+                    });
+                }
+
+                if (secretKey !== env.PROFILE_CAPTURE_SECRET) {
+                    logger.warn('[GameServer] Unauthorized profile capture attempt with invalid secret key');
+                    return res.status(403).json({
+                        success: false,
+                        message: 'Forbidden - Invalid secret key'
```
AMMayberry1: nit: it's enough to just say "forbidden", we don't need to provide additional information like saying that there is an api-key in use as opposed to oauth

server/gamenode/GameServer.ts
```diff
@@ -650,6 +652,70 @@ export class GameServer {
...
+                // Parse and validate duration parameter between 1 second and 10 minutes
+                if (!duration) {
+                    return res.status(400).json({
+                        success: false,
+                        message: 'Duration parameter is required'
+                    });
+                }
+
+                const profileDuration = parseInt(duration as string);
+                if (profileDuration < 1 || profileDuration > 600) {
```
AMMayberry1: Enforcing a minimum is good but I'm not sure if we need a maximum, we don't want to have to do a code change if we decide we need a longer run

server/utils/profiler.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+        if (!filename || filename.trim() === '') {
+            throw new Error('Filename is required for S3 upload');
+        }
+        if (buffer.length === 0) {
+            throw new Error('Buffer is empty, cannot upload to S3');
+        }
+
+        // Validate required environment variables
+        if (!env.AWS_REGION || !env.AWS_ACCESS_KEY_ID || !env.AWS_SECRET_ACCESS_KEY || !env.PROFILE_CAPTURE_S3_BUCKET) {
```
AMMayberry1: We should make sure all of these are covered in the `requireEnvVars()` in GameServer. normally we would put it in the constructor (see SwuStatsHandler) but since this object isn't constructed until the api call happens that would be less useful

server/gamenode/GameServer.ts
```diff
@@ -650,6 +652,70 @@ export class GameServer {
         });
         // *** End of User Object calls ***
 
+        // PROFILING
+        app.get('/api/run-profile', async (req, res, next) => {
```
AMMayberry1: This should be a PUT or POST instead of a GET

Dockerfile
```diff
@@ -21,6 +21,7 @@ ENV ENVIRONMENT="production"
 ENV GAME_NODE_HOST="localhost"
 ENV GAME_NODE_NAME="test1"
 ENV GAME_NODE_SOCKET_IO_PORT="9500"
+ENV PROFILE_CAPTURE_S3_BUCKET="karabast-profiling-dumps"
```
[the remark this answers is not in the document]
brianries: I added this directly into the PutObject command under the upload.  Its the only place that really needs it.  No longer has to be configured

server/gamenode/GameServer.ts
```diff
@@ -1817,5 +1883,69 @@ export class GameServer {
...
+            const started = await profiler.startAllocSampling(1024 * 64); // 64KB sampling interval
+            if (!started) {
+                logger.warn('[GameServer] Heap profiling is already active, cannot start new session');
+                return;
+            }
+
+            // Stop profiling after specified duration
+            setTimeout(async () => {
+                try {
+                    const result = await profiler.stopAllocSamplingAsBuffer('gammeserver-allocation');
```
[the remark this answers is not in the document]
brianries: Fixed in the refactor of forceStop

server/utils/profiler.ts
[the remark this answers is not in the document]
brianries: Fixed by hardcoding region.  Makes sense since we're hardcoding the bucket name that exists there too

server/gamenode/GameServer.ts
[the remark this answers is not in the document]
brianries: changed field to durationSeconds

server/gamenode/GameServer.ts
brianries: Changed to POST

server/gamenode/GameServer.ts
brianries: Moved this into the authorization section instead of a param

server/gamenode/GameServer.ts
```diff
@@ -650,6 +652,70 @@ export class GameServer {
         });
         // *** End of User Object calls ***
 
+        // PROFILING
+        app.get('/api/run-profile', async (req, res, next) => {
+            try {
+                const { secretKey, type, duration } = req.query;
+
+                // Verify secret key
+                if (!env.PROFILE_CAPTURE_SECRET) {
```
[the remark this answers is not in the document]
brianries: Made this prod-required to start the game server.
In either or dev or prod, it must exist and match to start the profile session

PR #1758, opened 2025-10-01
Damage Prevention (Super WIP)

adbronico:


server/game/Interfaces.ts
```diff
@@ -38,6 +39,14 @@ import type { Lobby } from '../gamenode/Lobby';
 /** Interface definition for addTriggeredAbility */
 export type ITriggeredAbilityProps<TSource extends Card = Card> = ITriggeredAbilityWhenProps<TSource> | ITriggeredAbilityAggregateWhenProps<TSource>;
 export type IReplacementEffectAbilityProps<TSource extends Card = Card> = IReplacementEffectAbilityWhenProps<TSource> | IReplacementEffectAbilityAggregateWhenProps<TSource>;
+export type IDamagePreventionAbilityProps<TSource extends Card = Card> = Omit<IReplacementEffectAbilityBaseProps<TSource>, 'when'> & {
+    preventionType: DamagePreventionType;
+    preventDamageFromSource?: RelativePlayer; // TSTODO - update to accept an array
+    preventDamageFrom?: DamageSourceType;
```
AMMayberry1, line 45: These two are a bit hard to interpret at first glance. Maybe using suffixes like "FromPlayer" and "DamageOfType".

Also to me having them start with "preventDamage" feels clunky since the property is already being used in the context of an PreventDamageAbility. An alternative would be something like "onlyFromPlayer" and "onlyDamageOfType", though there are probably better options

server/game/Interfaces.ts
```diff
@@ -38,6 +39,14 @@ import type { Lobby } from '../gamenode/Lobby';
...
+    preventionAmount?: number;
+    replaceWithSystem?: GameSystem;
```
AMMayberry1, line 47: nit: we typically use "effect" instead of "system" for things that are surfaced in card implementations, e.g. "immediateEffect". so probably "replaceWithEffect" is more consistent here

server/game/Interfaces.ts
```diff
@@ -38,6 +39,14 @@ import type { Lobby } from '../gamenode/Lobby';
...
+    targetCondition?: (card: Card, context?: TriggeredAbilityContext) => boolean; // This can be used to override the default assumption that the damage is being dealt to context.source
```
AMMayberry1, line 48: "targetCondition" is a standard term for us but I think it's a little vague here because it's not clear what the "target" refers to. Not sure what's a lot clearer though - maybe "protectedCardCondition"?

AMMayberry1, line 48: Can we turn the comment into a jsdoc

server/game/core/ability/DamagePreventionAbility.ts
```diff
@@ -0,0 +1,47 @@
...
+import type Game from '../Game';
+import { DamagePreventionSystem } from '../../gameSystems/DamagePreventionSystem';
+import { AbilityType, RelativePlayer } from '../Constants';
+import TriggeredAbility from './TriggeredAbility';
+
+export default class DamagePreventionAbility extends TriggeredAbility {
+    public constructor(game: Game, card: Card, properties: IDamagePreventionAbilityProps) {
+        const replacementAbilityProps: IReplacementEffectAbilityProps =
+            Object.assign({}, properties, { immediateEffect: new DamagePreventionSystem(properties),
+                when: { onDamageDealt: (event, context) => this.buildDamagePreventionTrigger(event, context, properties) } });
```
AMMayberry1, line 12: nit: we're moving off of `Object.assign()` syntax to object spread syntax with `...`. So this could look like:

```ts
const replacementAbilityProps: IReplacementEffectAbilityProps = {
    ...properties,
    { 
        immediateEffect: new DamagePreventionSystem(properties),
        when: { onDamageDealt: (event, context) => this.buildDamagePreventionTrigger(event, context, properties) }
    } 
};
```

AMMayberry1, line 12: Do we need to be passing everything in `properties` to the DamagePreventionSystem? If not, we should extract the ones we actually need similar to how `cancelProps` is generated in ReplacementEffectAbility

server/game/core/ability/DamagePreventionAbility.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    private buildDamagePreventionTrigger(event, context, properties: IDamagePreventionAbilityProps): boolean {
+        // If a custom targetCondition is provided, this means the damage prevention should apply to the card that meets that condition instead of context.source
+        if (properties.targetCondition) {
+            if (properties.targetCondition(event.card, context) === false) {
+                return false;
+            }
+        } else if (event.card !== context.source) {
+            return false;
+        }
```
AMMayberry1, line 25: nit, streamlining: this could look a little cleaner as

```ts
if (properties.targetCondition && !properties.targetCondition(event.card, context)) {
   return false;
} 
if (event.card !== context.source) {
    return false;
}
```

Although honestly it might be better to just set up a default value of `targetCondition` that just does the `event.card === context.source` check. Then all we have to do is:

```ts
if (!properties.targetCondition(event.card, context)) {
   return false;
}
```

server/game/core/ability/DamagePreventionAbility.ts
```diff
@@ -0,0 +1,47 @@
...
+
+        if (event.isUnpreventable) {
+            return false;
+        }
+
+        if (properties.preventDamageFrom) {
+            if (event.damageSource.type !== properties.preventDamageFrom) {
+                return false;
+            }
+        }
```
AMMayberry1, line 35: nit: can consolidate the nested `if`'s here as well

server/game/core/ability/DamagePreventionAbility.ts
```diff
@@ -0,0 +1,47 @@
...
+
+        if (properties.preventDamageFromSource) {
+            if (properties.preventDamageFromSource === RelativePlayer.Opponent) {
+                return event.damageSource.player !== context.source.controller;
+            } else if (properties.preventDamageFromSource === RelativePlayer.Self) {
+                return event.damageSource.player === context.source.controller;
+            }
+        }
```
AMMayberry1, line 43: We can leverage `EnumHelpers.asConcretePlayer()` to simplify a bit:

```ts
if (properties.preventDamageFromSource) {
    return EnumHelpers.asConcretePlayer(properties.preventDamageFromSource, context.source.controller) === event.damageSource.player;
}
```

server/game/gameSystems/DamagePreventionSystem.ts
```diff
@@ -0,0 +1,65 @@
...
+import { DamageSystem } from './DamageSystem';
+
+export interface IDamagePreventionSystemProperties extends Omit<IReplacementEffectAbilityProps, 'when'> {
+    preventionType: DamagePreventionType;
+    preventDamageFromSource?: RelativePlayer; // TSTODO - update to accept an array
+    preventDamageFrom?: DamageSourceType;
+    preventionAmount?: number;
+    replaceWithSystem?: GameSystem;
+    triggerCondition?: (card: Card, context?: TriggeredAbilityContext) => boolean; // This can be used to override the default trigger condition for special cases
+}
```
AMMayberry1, line 20: This should be extending the parent's properties, `IReplacementEffectSystemProperties`. Also, looks like some of these are only used in the ability and not here in the system, so we can trim down to just the ones that are actually needed

server/game/gameSystems/DamagePreventionSystem.ts
```diff
@@ -0,0 +1,65 @@
...
+        const properties = this.generatePropertiesFromContext(context);
+        switch (properties.preventionType) {
+            case DamagePreventionType.All:
+                return ['prevent all damage to {0}', [context.source]];
+            case DamagePreventionType.Reduce:
+                return ['prevent {0} damage to {1}', [properties.preventionAmount, context.event.card]];
+            case DamagePreventionType.Replace:
+                const replaceWith = properties.replaceWithSystem;
+                const replaceMessage = replaceWith.getEffectMessage(context);
+                return ['{0} instead of {1} taking damage', [replaceMessage, context.event.card]]; // TODO: how the heck do we get the effect description from the replacementImmediateEffect here?
```
AMMayberry1, line 35: Good to check in with Frugghi or Mouse on this, they've been actively updating a lot of this recently anyway so there might be a different approach to use now

server/game/gameSystems/DamagePreventionSystem.ts
```diff
@@ -0,0 +1,65 @@
...
+            default:
+                Contract.fail(`Invalid preventionType ${properties.preventionType} for DamagePreventionSystem`);
+        }
+    }
+
+    protected override getReplacementImmediateEffect(context: TContext, additionalProperties: Partial<IDamagePreventionSystemProperties> = {}): GameSystem<TContext> {
+        const properties = super.generatePropertiesFromContext(context, additionalProperties) as IDamagePreventionSystemProperties;
```
AMMayberry1, line 42: I think it should `this` instead of `super`, that might fix the type issue as well

server/game/gameSystems/ReplacementEffectSystem.ts
```diff
@@ -57,7 +57,7 @@ export class ReplacementEffectSystem<TContext extends TriggeredAbilityContext =
     }
 
     public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext, additionalProperties: Partial<IReplacementEffectSystemProperties<TContext>> = {}) {
-        const event = this.createEvent(null, context, additionalProperties);
+        const event = this.createEvent(null, context, additionalProperties as Partial<TProperties>);
```
AMMayberry1, line 60: I think the issue here is actually the parameter type on the method is wrong, try changing it to be `additionalProperties: Partial<TProperties>`. Every usage of `IReplacementEffectSystemProperties` in the class should be replaced with `TProperties` since we're using a template for it now

server/game/gameSystems/ReplacementEffectSystem.ts
```diff
@@ -77,20 +77,25 @@ export class ReplacementEffectSystem<TContext extends TriggeredAbilityContext =
     }
 
     public override generatePropertiesFromContext(context: TContext, additionalProperties: Partial<IReplacementEffectSystemProperties<TContext>> = {}) {
-        const properties = super.generatePropertiesFromContext(context, additionalProperties);
+        const properties = super.generatePropertiesFromContext(context, additionalProperties as Partial<TProperties>);
```
AMMayberry1, line 80: Yeah I think these should also be using `this` instead of `super`, this is really old code

server/game/gameSystems/DamagePreventionSystem.ts
```diff
@@ -0,0 +1,65 @@
...
+
+        switch (properties.preventionType) {
+            case DamagePreventionType.All:
+                return null; // Returning NoActionSystem here causes an error: Contract assertion failure: Replacement effect 'GameSystem: ' for replacementEffect did not generate any events
+            case DamagePreventionType.Reduce:
+                Contract.assertPositiveNonZero(properties.preventionAmount, 'preventionAmount must be a positive non-zero number for DamagePreventionType.Reduce');
```
AMMayberry1, line 48: nit: should include the value of `properties.preventionAmount` in the error message in case we need to debug here in prod

server/game/gameSystems/DamagePreventionSystem.ts
```diff
@@ -0,0 +1,65 @@
...
+                return new DamageSystem((context) => ({
+                    target: context.event.card,
+                    amount: Math.max(context.event.amount - properties.preventionAmount, 0),
+                    source: context.event.damageSource.type === DamageType.Ability ? context.event.damageSource.card : context.event.damageSource.damageDealtBy, // Copied this from Cassian - why is it capitalized?
```
AMMayberry1, line 52: What's the question about capitalization? Looks pretty standard to me, might be missing something

server/game/gameSystems/DamagePreventionSystem.ts
```diff
@@ -0,0 +1,65 @@
...
+import { ReplacementEffectSystem } from './ReplacementEffectSystem';
+import * as Contract from '../core/utils/Contract';
+import { DamageSystem } from './DamageSystem';
+
+export interface IDamagePreventionSystemProperties extends Omit<IReplacementEffectAbilityProps, 'when'> {
+    preventionType: DamagePreventionType;
+    preventDamageFromSource?: RelativePlayer; // TSTODO - update to accept an array
+    preventDamageFrom?: DamageSourceType;
+    preventionAmount?: number;
+    replaceWithSystem?: GameSystem;
```
AMMayberry1, line 18: Changing this to `GameSystem<TriggeredAbilityContext>` should fix the typecasts you had to do on Maul

server/game/core/card/propertyMixins/TriggeredAbilityRegistration.ts
```diff
@@ -96,6 +116,19 @@ export function WithTriggeredAbilities<TBaseClass extends CardConstructor<TState
             return addedAbility.uuid;
         }
 
+        /**
+             * Adds a dynamically gained damage prevention ability to the card and immediately registers its triggers. Used for "gain ability" effects.
+             *
+             * @returns The uuid of the created triggered ability
+             */
```
AMMayberry1, line 123: nit: spacing is off

server/game/gameSystems/ReplacementEffectSystem.ts
adbronico: This one would cause a loop if we did that

PR #1760, opened 2025-10-01
crucible, the wrong ride

sethhelman:
wrong ride
crucible

### DIFF
--- a/server/game/cards/03_TWI/upgrades/Foresight.ts
+++ b/server/game/cards/03_TWI/upgrades/Foresight.ts
@@ -30,7 +30,7 @@ export default class Foresight extends UpgradeCard {
                 then: {
                     title: 'Reveal and draw the top card of deck',
                     optional: true,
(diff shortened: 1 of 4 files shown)


test/server/cards/06_SEC/events/TheWrongRide.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+                }
+            });
+
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.theWrongRide);
+            expect(context.player2.readyResourceCount).toBe(3);
+        });
+
+        it('The Wrong Ride\'s ability should exhaust 2 enemy resources', async function () {
```
AMMayberry1, line 22: nit: we avoid using identical test case names because then it's hard to distinguish between them if one fails, and generally it makes it confusing to understand what each test is doing differently. In this case we can just append "with Plot" to the end to clarify

PR #1766, opened 2025-10-02
Fix crash when applying a delayed defeat effect to a unit in the discard

Frugghi:
Not sure if there is a smarter way to do this, we can easily check the validity of the target in the `DelayedEffectSystem` because we don't have a triggered ability context to pass to the immediate effect.


server/game/cards/02_SHD/upgrades/UnrefusableOffer.ts
```diff
@@ -31,9 +31,9 @@ export default class UnrefusableOffer extends UpgradeCard {
                 }),
                 ifYouDo: (ifYouDoContext) => ({
                     title: 'At the start of the regroup phase, defeat it',
-                    ifYouDoCondition: (context) => context.events.filter((e) => e.name === 'playCard').length > 0,
+                    ifYouDoCondition: (context) => context.events.filter((e) => e.name === MetaEventName.PlayCard).length > 0 && context.events[0].card.isInPlay(),
```
AMMayberry1, line 34: As a rule of thumb, we shouldn't be referring to `MetaEventName` enums in card code. They're typically related to stuff that we use internally as rules helpers.

In this case, it looks like `PlayCard` is used by the `PlayCardSystem`. It emits a meta-event and then the underlying PlayCardAction should emit an `onCardPlayedEvent`, which is what we most likely should be keying on here. Otherwise we would only catch cards played by `playCard()` effects and not like from hand

PR #1769, opened 2025-10-02
Print chat logs for then/ifYouDo/ifYouDoNot effects

### LINKED ISSUE #1708: Game log text is missing for `then`/`ifYouDo`/`ifYouDoNot` effects
Game log text is missing for certain card ability steps, like `then`, `ifYouDo`, and `ifYouDoNot`.

The reason for this is most of the game effect messages are generated in the `displayMessage()` method of `CardAbility`, but the ones mentioned above are instances of the superclass `CardAbilityStep`, which does not have a `displayMessage()` implementation.

When this is solved, we will have to do a pass to see which game systems log their own message (i.e. in the `eventHandler`) to prevent double-logging for those effects. One example of a system that always logs its own message is the `DiscardSpecificCardSystem`.

### Example

For LOF Kylo Ren, there is no log message for the card draw part of his ability:

Image

Frugghi:
Fixes #1708


moyerr (dismissed):
This is great, awesome work @Frugghi! One small concern about Elia Kane, but I don't see it as super critical, so giving the ✅ anyway

PR #1781, opened 2025-10-03
⟡ Libertine, Under New Ownership

moyerr:


test/server/cards/06_SEC/units/LibertineUnderNewOwnership.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+                expect(context.libertine.getPower()).toBe(7);
+                expect(context.libertine.getHp()).toBe(7);
+                expect(context.libertine.capturedUnits.length).toBe(4);
+
+                // P2 plays Unexpected Escape to free 1 unit from capture
+                context.player2.clickCard(context.unexpectedEscape);
+                context.player2.clickCard(context.libertine);
+                context.player2.clickCard(context.tieBomber);
+
+                // Libertine should now be a 6/7 with 3 captured units
```
AMMayberry1, line 48: Can we add to this test a friendly unit that also has captured units, just to make sure they aren't included in the filter. This should be doable directly in the test setup

PR #1784, opened 2025-10-04
Nimble Prowess

VJubert:


test/server/cards/06_SEC/upgrades/NimbleProwess.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+                        hand: ['nimble-prowess'],
+                        groundArena: ['battlefield-marine'],
+                        spaceArena: ['awing']
+                    },
+                    player2: {
+                        groundArena: ['wampa', 'atst'],
+                        spaceArena: ['green-squadron-awing', 'phoenix-squadron-awing']
+                    }
+                });
+            });
```
moyerr, line 17: We've had issues in the past with upgrades that have "friendly unit" restrictions combined with either taking control and moving the upgrade or playing from opponent's discard. Could you add a test for one of those scenarios just for completeness sake? I'd say just one test with either A FIne Addition or Hondo Ohnaka should work.

moyerr (approved):

PR #1793, opened 2025-10-06
Fix the card mock override

AMMayberry1:


scripts/mockdata.js
```diff
@@ -1694,20 +1694,42 @@ function buildSetStr(card) {
...
+            continue;
+        }
+
+        cardsById.set(buildSetStr(card), card);
     }
 
-    return mockCardNames;
+    const cardsWithMocks = Array.from(cardsById.values()).concat(Array.from(tokensByName.values()));
+
+    return { mockCardNames, cards: cardsWithMocks };
```
moyerr, line 1732: it seems like `mockCardNames` is no longer being modified?

PR #1804, opened 2025-10-07
Diplomatic Immunity, General Grievous

sethhelman:
grievous
diplomatic immunity

### DIFF
--- a/server/game/cards/06_SEC/units/GeneralGrievousScuttlingToSafety.ts
+++ b/server/game/cards/06_SEC/units/GeneralGrievousScuttlingToSafety.ts
@@ -1,6 +1,7 @@
 import type { IAbilityHelper } from '../../../AbilityHelper';
 import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
 import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { ZoneName } from '../../../core/Constants';
 
 export default class GeneralGrievousScuttlingToSafety extends NonLeaderUnitCard {
     protected override getImplementationId() {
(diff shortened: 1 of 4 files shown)


test/server/cards/06_SEC/units/GeneralGrievousScuttlingToSafety.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+                context.player2.clickPrompt('Pass');
+
+                context.player1.clickCard(context.generalGrievousScuttlingToSafety);
+                context.player1.clickCard(context.rebelPathfinder);
+
+                expect(context.generalGrievousScuttlingToSafety.damage).toBe(2);
+                expect(context.rebelPathfinder).toBeInZone('discard');
+            });
+
+            it('should not return a friendly unit to hand when they are attacked', function () {
```
AMMayberry1, line 57: Can we add a test case with an on-attack ability like Sabine1 unit and confirm that we properly get both triggers. It wouldn't actually make a difference in that case, but good to check in case there is some on-attack that could defeat him before he is returned to hand

server/game/cards/06_SEC/units/GeneralGrievousScuttlingToSafety.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addTriggeredAbility({
+            title: 'Return him to his owner\'s hand',
+            when: {
+                onAttackDeclared: (event, context) => event.attack.getAllTargets().includes(context.source),
+            },
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => context.source.zoneName === 'groundArena',
```
AMMayberry1, line 20: nit: we shouldn't use raw strings, use ZoneName.GroundArena instead.

PR #1836, opened 2025-10-15
Exiled from the Force

moyerr:

### DIFF
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -774,11 +774,28 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
 
 
     // *************************************** EFFECT HELPERS ***************************************
+
+    /**
+     * Whether or not this card currently has any blanking effect applied to it. It may still have
+     * some abilities if they are explicitly excluded from a partial blanking effect.
+     *
+     * @returns {boolean} `true` if the card is blanked or partially blanked, `false` otherwise.
+     */
     public isBlank(): boolean {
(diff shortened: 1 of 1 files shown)


test/server/cards/06_SEC/upgrades/ExiledFromTheForce.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+            expect(context.p1Base.damage).toBe(6);
+
+            // Jedi Guardian attacks base
+            context.player1.clickCard(context.jediGuardian);
+            context.player1.clickCard(context.p2Base);
+            expect(context.p2Base.damage).toBe(10);
+            expect(context.p1Base.damage).toBe(6); // No damage restored
+            expect(context.homeOne.upgrades.length).toBe(0); // No experience given from Inspiring Mentor's ability
+            expect(context.player2).toBeActivePlayer();
+        });
```
AMMayberry1: Let's also test with a card that already has natural grit and / or gained grit from another source to make sure nothing weird happens

server/game/core/card/Card.ts
```diff
@@ -775,17 +775,22 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
 
     // *************************************** EFFECT HELPERS ***************************************
     public isBlank(): boolean {
```
AMMayberry1: Since the distinction between this and `isFullyBlanked` is a bit subtle, might be good to add a jsdoc for each explaining how they work

AMMayberry1 (approved):

PR #1840, opened 2025-10-15
Additional checks and improvements for undo confirmation requests

AMMayberry1:
Improved some of the logic for when how we determine when to do a request undo based on game state.

The big change is to block an undo if the opponent has revealed their "intent" for their action on their turn by playing a card or activating an ability in a way that is telegraphed to the non-acting player. To reduce the number of ways this can happen, I've updated many of the "waiting" prompts to no longer give information such as the name of the card being resolved.

Additional changes:

- Fixed a bug where undoing on the start of the opponent's turn with confirmation would not do anything because the confirmation popup changed the rollback point
- Made it so that undo is blocked for cases like PotDS only _after_ the opponent has resolved the prompt
- Rolling back now requires confirmation if you'd be rolling back over one or more opponent actions
- On rollback we will reset the "this snapshot requires confirmation" state so that another rollback on the same action doesn't generate an unnecessary confirmation request

### DIFF
--- a/test/helpers/IntegrationHelper.js
+++ b/test/helpers/IntegrationHelper.js
@@ -84,11 +84,13 @@ global.integration = function (definitions, enableUndo = false) {
             contextRef.snapshot = {
                 getCurrentSnapshotId: () => gameFlowWrapper.snapshotManager?.currentSnapshotId,
                 getCurrentSnapshottedAction: () => gameFlowWrapper.snapshotManager?.currentSnapshottedAction,
+                rollbackToSnapshot: (settings, requestingPlayerId) => {
+                    let playerId = requestingPlayerId;
+                    if (!playerId) {
+                        playerId = settings.playerId ?? '111';
+                    }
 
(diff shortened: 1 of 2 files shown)


test/helpers/IntegrationHelper.js
```diff
@@ -84,7 +84,18 @@ global.integration = function (definitions, enableUndo = false) {
             contextRef.snapshot = {
                 getCurrentSnapshotId: () => gameFlowWrapper.snapshotManager?.currentSnapshotId,
                 getCurrentSnapshottedAction: () => gameFlowWrapper.snapshotManager?.currentSnapshottedAction,
-                rollbackToSnapshot: (settings) => newContext.game.rollbackToSnapshotInternal(settings),
+
+                // TODO THIS PR: add playerId to the below to indicate which player is requesting the rollback
+                // TODO THIS PR: do we still need a separate quickRollback() helper?
+                rollbackToSnapshot: (settings) => {
```
moyerr: Do these TODOs need to be resolved in this PR?

moyerr (approved):

PR #1842, opened 2025-10-16
chebato/swustats accesstoken fix

CheBato:
- Added accessTokens as the main source of truth.
- Cleaned up the flow for refreshing a token.

FE PR: https://github.com/SWU-Karabast/forceteki-client/pull/516

### DIFF
--- a/server/utils/SWUStats/SwuStatsHandler.ts
+++ b/server/utils/SWUStats/SwuStatsHandler.ts
@@ -5,7 +5,7 @@ import type { IDecklistInternal } from '../deck/DeckInterfaces';
 import type { IBaseCard } from '../../game/core/card/BaseCard';
 import { Aspect } from '../../game/core/Constants';
 import { GameCardMetric, type IGameStatisticsTracker } from '../../gameStatistics/GameStatisticsTracker';
-import type { PlayerDetails, IStatsMessageFormat, Lobby } from '../../gamenode/Lobby';
+import type { PlayerDetails, IStatsMessageFormat } from '../../gamenode/Lobby';
 import { StatsSource } from '../../gamenode/Lobby';
 import { StatsSaveStatus } from '../../gamenode/Lobby';
 import type { GameServer, ISwuStatsToken } from '../../gamenode/GameServer';
@@ -105,7 +105,8 @@ export class SwuStatsHandler {
      * @param game The completed game
      * @param player1Details Details about player1
      * @param player2Details Details about player2
-     * @param lobby
+     * @param hasSwuStatsSource the hasSwuStatsSource function from the lobby
+     * @param lobbyId the id of the lobby in string format
      * @param serverObject the server object from where we gain access to the user x accessToken
      * @returns Promise that resolves to true if successful, false otherwise
      */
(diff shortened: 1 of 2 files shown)


server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -343,22 +349,26 @@ export class SwuStatsHandler {
     private async getAccessTokenAsync(
         playerDetails: PlayerDetails,
```
AMMayberry1: Similar note as on the `addSwuStatsRefreshTokenAsync` method below - if all we're using this for is to get the `userId` then let's just take that directly instead. That will make this method more generically reusable so we could call it from somewhere that doesn't have a `PlayerDetails` object

server/gamenode/GameServer.ts
```diff
@@ -401,6 +408,33 @@ export class GameServer {
             }
         });
 
+        app.get('/api/user/:userId/swustatsLink', authMiddleware('swustatsLink'), async (req, res, next) => {
```
AMMayberry1: Is the authMiddleware doing a DB call here? If so then we can just grab the swustats linked status off of that and attach it to `req.user` so we're not doing the same DB get operation twice.

If it only does the DB call in certain situations but not others, then we can just have a conditional check and if `req.user.swustatsLinked` is null then we do the explicit DB check below.

server/gamenode/GameServer.ts
```diff
@@ -386,13 +386,20 @@ export class GameServer {
                 //         logger.error(`GameServer (get-user): Error with syncing Preferences for User ${user.getId()}`, err);
                 //     }
                 // }
+
+                // we check here if the user has a refresh token and a map
+                if (user.isSWUStatsLinked() && !this.swuStatsTokenMapping.get(user.getId())) {
```
AMMayberry1: I think we don't need this code block at all anymore? We don't need to think about the refresh token if it's not being asked for

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -300,7 +300,7 @@ export class SwuStatsHandler {
         player1Details: PlayerDetails,
         player2Details: PlayerDetails,
         winner: number,
-        lobbyId: string,
+        lobby: Lobby,
```
AMMayberry1: Why did we switch to accepting the entire Lobby object here if we just need the ID still?

server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -335,30 +341,34 @@ export class SwuStatsHandler {
...
+        let playerAccessToken = null;
+        const playerTokenData = serverObject.swuStatsTokenMapping.get(userId);
         // Handle Player swu token
-        if (playerDetails.swuStatsToken && this.isTokenValid(playerDetails.swuStatsToken)) {
-            playerAccessToken = playerDetails.swuStatsToken.accessToken;
-            logger.info(`SWUStatsHandler: Using existing valid access token for player (${playerDetails.user.getId()})`, { lobbyId, userId: playerDetails.user.getId() });
-        } else if (playerDetails.swuStatsRefreshToken) {
+        if (playerTokenData && this.isTokenValid(playerTokenData)) {
+            playerAccessToken = playerTokenData.accessToken;
+            logger.info(`SWUStatsHandler: Using existing valid access token for player (${userId})`, lobbyId ? { lobbyId, userId } : { userId });
```
AMMayberry1: nit: it's fine to just do `{ lobbyId, userId }` here, it's fine to log a null value for something to indicate that it's empty. Same for other logs in this method

PR #1848, opened 2025-10-16
diplomatic envoy, escape pod, senate warden

VJubert:


test/server/cards/06_SEC/units/DiplomaticEnvoy.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+                expect(context.player1).toHavePassAbilityPrompt('Ambush');
+                context.player1.clickPrompt('Trigger');
+                context.player1.clickCard(context.specforceSoldier);
+
+                context.player2.passAction();
+
+                // no more Ambush
+                context.player1.clickCard(context.wampa);
+                expect(context.player2).toBeActivePlayer();
+            });
```
AMMayberry1: Can we also check that the ambush doesn't accidentally work for the opponent

AMMayberry1 (approved):

PR #1852, opened 2025-10-17
Update private lobbies to have a setting for enabling request undo

AMMayberry1:


server/game/core/Game.js
```diff
@@ -1972,11 +1971,11 @@ class Game extends EventEmitter {
 
     /** @param {boolean} enabled */
     setUndoConfirmationRequired(enabled) {
-        if (this.enableConfirmationToUndo === enabled) {
+        if (this.snapshotManager.undoMode === UndoMode.Request) {
             return;
         }
```
moyerr: Is this meant to mean that this method can _enable_ it but it cannot _disable it_? Previously, this condition was just hear to return early if the game setting matched the parameter setting, but now the early return ignores the parameter entirely.

PR #1857, opened 2025-10-19
one way out, kreias whispers

sethhelman:
kreias whispers
one way out

### DIFF
--- a/server/game/cards/06_SEC/events/KreiasWhispers.ts
+++ b/server/game/cards/06_SEC/events/KreiasWhispers.ts
@@ -26,7 +26,6 @@ export default class KreiasWhispers extends EventCard {
                     },
                     bottomCard: {
                         activePromptTitle: 'Choose a card to put on the bottom of your deck',
-                        dependsOn: 'topCard',
                         mode: TargetMode.Single,
                         controller: RelativePlayer.Self,
                         zoneFilter: ZoneName.Hand,


server/game/cards/06_SEC/events/KreiasWhispers.ts
```diff
@@ -0,0 +1,42 @@
...
+            internalName: 'kreias-whispers',
+        };
+    }
+
+    public override setupCardAbilities(registrar: IEventAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.setEventAbility({
+            title: 'Draw 3 cards, then put one card from your hand on the top of your deck and another card from your hand on the bottom of your deck',
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
+                AbilityHelper.immediateEffects.draw({ amount: 3 }),
+                AbilityHelper.immediateEffects.selectCard({
```
moyerr: With the wording of this, I don't think they should all be sequential. The draw happens first, then the next two effects are simultaneous. I think we can probably combine a `then:` and a simultaneous system for this. You could even do multiple target resolvers for the top/bottom.

server/game/cards/06_SEC/events/KreiasWhispers.ts
```diff
@@ -0,0 +1,42 @@
...
+                }),
+                AbilityHelper.immediateEffects.selectCard({
+                    activePromptTitle: 'Select 1 card',
+                    mode: TargetMode.Exactly,
+                    controller: RelativePlayer.Self,
+                    zoneFilter: ZoneName.Hand,
+                    numCards: 1,
+                    mustChangeGameState: GameStateChangeRequired.MustFullyResolve,
+                    effect: 'choose 1 card to move to the bottom of their deck',
+                    immediateEffect: AbilityHelper.immediateEffects.moveToBottomOfDeck()
```
moyerr: These two `activePromptTitles` should probably inform the user what is going to happen with the selected card. Also, you can use `TargetMode.Single` instead of `Exactly`/`numCards`.

test/server/cards/06_SEC/events/KreiasWhispers.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+            expect(context.player1).toBeAbleToSelectExactly([
+                context.jediKnight
+            ]);
+            context.player1.clickCard(context.jediKnight);
+            expect(context.jediKnight).toBeInBottomOfDeck(context.player1, 1);
+            expect(context.player1.deck[0]).toBe(context.battlefieldMarine);
+
+            expect(context.player2).toBeActivePlayer();
+        });
```
AMMayberry1: Can we test what happens when there's only one card in hand after drawing? I think it might cause an issue due to the way the target resolvers are set up. If so, just ping me in chat and we can find a different approach

server/game/cards/06_SEC/events/OneWayOut.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    public override setupCardAbilities(registrar: IEventAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.setEventAbility({
+            title: 'Attack with a unit. It gets +1/+0 and gains Overwhelm for this attack. If it is attacking a unit, the defender loses all abilities for this attack',
+            initiateAttack: {
+                attackerLastingEffects: [
+                    { effect: AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Overwhelm) },
+                    { effect: AbilityHelper.ongoingEffects.modifyStats({ power: 1, hp: 0 }) },
+                ],
+                defenderLastingEffects: [
```
AMMayberry1: I think we do actually need to check that the defender is a unit here, from what I'm seeing I think this will activate when a base is attacked (I guess technically we're not using the most correct definition for "defender" for the name defenderLastingEffects). Let's also add a test case where a base is being attacked just to confirm.

server/game/cards/06_SEC/events/KreiasWhispers.ts
```diff
@@ -0,0 +1,42 @@
...
+                targetResolvers: {
+                    topCard: {
+                        activePromptTitle: 'Choose a card to put on top of your deck',
+                        mode: TargetMode.Single,
+                        controller: RelativePlayer.Self,
+                        zoneFilter: ZoneName.Hand,
+                    },
+                    bottomCard: {
+                        activePromptTitle: 'Choose a card to put on the bottom of your deck',
+                        mode: TargetMode.Single,
```
AMMayberry1: Missed this - can we add `dependsOn: 'topCard'`, it helps ensure that these always work right when there's multiple target resolvers

PR #1863, opened 2025-10-20
Feature/sec 130 and 220

mnechromancer:
Adding Ferrix Uprising and Hired Slicer, third time's the charm! Definitely want to try to clean up Hired Slicer a bit more, can't find an exact ability match for it

### DIFF
--- a/server/game/cards/01_SOR/events/BombingRun.ts
+++ b/server/game/cards/01_SOR/events/BombingRun.ts
@@ -19,8 +19,8 @@ export default class BombingRun extends EventCard {
                 mode: TargetMode.Select,
                 activePromptTitle: 'Choose an arena',
                 choices: {
-                    ['Ground']: this.eventEffect(ZoneName.GroundArena, AbilityHelper),
                     ['Space']: this.eventEffect(ZoneName.SpaceArena, AbilityHelper),
+                    ['Ground']: this.eventEffect(ZoneName.GroundArena, AbilityHelper),
                 }
             }
         });
--- a/server/game/cards/01_SOR/events/Outmaneuver.ts
+++ b/server/game/cards/01_SOR/events/Outmaneuver.ts
@@ -19,8 +19,8 @@ export default class Outmaneuver extends EventCard {
                 mode: TargetMode.Select,
                 activePromptTitle: 'Choose an arena',
                 choices: {
-                    ['Ground']: this.eventEffect(ZoneName.GroundArena, AbilityHelper),
                     ['Space']: this.eventEffect(ZoneName.SpaceArena, AbilityHelper),
+                    ['Ground']: this.eventEffect(ZoneName.GroundArena, AbilityHelper),
                 }
             }
         });
--- a/server/game/cards/04_JTL/events/TurbolaserSalvo.ts
+++ b/server/game/cards/04_JTL/events/TurbolaserSalvo.ts
@@ -20,8 +20,8 @@ export default class TurbolaserSalvo extends EventCard {
                 activePromptTitle: 'Choose an arena',
                 showUnresolvable: true,
                 choices: {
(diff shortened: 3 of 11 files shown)


server/game/cards/06_SEC/units/HiredSlicer.ts
```diff
@@ -0,0 +1,93 @@
...
+                title: 'Exhaust a unit that shares a trait with one of those cards',
+                thenCondition: () => {
+                    const revealedCards = ifYouDoContext.events[0].cards || [];
+                    return revealedCards.length > 0;
+                },
+                targetResolver: {
+                    optional: true,
+                    cardTypeFilter: WildcardCardType.Unit,
+                    cardCondition: (card) => {
+                        const revealedCards = ifYouDoContext.events[0].cards || [];
```
AMMayberry1: I think we can remove the `|| []` safety check since we know that we won't get here unless the event resolved, in which case there must be cards

server/game/cards/06_SEC/units/HiredSlicer.ts
```diff
@@ -0,0 +1,93 @@
...
+                        return this.cardSharesTraitWithRevealedCards(card, revealedCards);
+                    },
+                    immediateEffect: AbilityHelper.immediateEffects.exhaust()
+                },
+                then: {
```
AMMayberry1: This doesn't hold 100% of the time but as a rule of thumb we don't use `then` unless the card has that word explicitly (but there are a few common exceptions). Since the way the card is worded implies that the effects resolve simultaneously, probably the best thing to do is put both effects into the target resolver in the `ifYouDo` part like below (it will look like this based on the context of the other comments):

```ts
targetResolver: {
    AbilityHelper.immediateEffects.exhaust(),
    AbilityHelper.immediateEffects.AbilityHelper.immediateEffects.moveToBottomOfDeck({ target: revealedCards, shuffleMovedCards: true })
  ]);
}
```

server/game/cards/06_SEC/units/HiredSlicer.ts
```diff
@@ -0,0 +1,93 @@
...
+                })
+            },
+            ifYouDo: (ifYouDoContext) => ({
+                title: 'Exhaust a unit that shares a trait with one of those cards',
+                thenCondition: () => {
+                    const revealedCards = ifYouDoContext.events[0].cards || [];
+                    return revealedCards.length > 0;
+                },
+                targetResolver: {
+                    optional: true,
```
AMMayberry1: Another case where I'm not sure why it let you do this - we shouldn't put `optional` at the targetResolver level except in cases where there are multiple resolvers. See Coordinate Front as an example.

For cases like this, the `optional` should live at the ability level (in this as a property directly on the `ifYouDo`)

test/server/cards/06_SEC/units/HiredSlicer.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+                // Wampa has Creature trait, Death Star Stormtrooper has Trooper, Enfys Nest has Underworld
+                expect(context.player1).toBeAbleToSelectExactly([context.deathStarStormtrooper, context.enfysNestMarauder]);
+                context.player1.clickCard(context.deathStarStormtrooper);
+
+                expect(context.deathStarStormtrooper.exhausted).toBe(true);
+
+                // Check that cards were put on bottom of deck
+                expect(context.player1.deck.length).toBe(5);
+                const bottomTwoCards = context.player1.deck.slice(-2);
+                expect(bottomTwoCards).toContain(context.battlefieldMarine);
```
AMMayberry1: We have a `.toBeInBottomOfDeck()` matcher, see the original Mon Mothma unit as an example

PR #1869, opened 2025-10-21
Card add - Hired Slicer (SEC 220)

mnechromancer:
## Add Hired Slicer (SEC 220)

Implements Hired Slicer's On Attack ability:
- Player chooses to reveal top 2 cards from their deck or opponent's deck
- If revealed, optionally exhaust a unit that shares a trait with one of those cards
- Then put the revealed cards on the bottom of that deck in random order

### Implementation Notes
- Uses proper `moveToBottomOfDeck()` GameSystem with `shuffleMovedCards: true`
- Trait computation optimized to happen once in `ifYouDo` closure
- Ability is optional at both the reveal and exhaust steps

### Testing
- 6 comprehensive tests covering normal cases and edge cases
- Tests include TODOs for future engine updates regarding `ifYouDo` validation with fewer than 2 cards

All PR feedback addressed and tests passing! Woo woo!

### DIFF
--- a/server/gamenode/Lobby.ts
+++ b/server/gamenode/Lobby.ts
@@ -195,7 +195,10 @@ export class Lobby {
 
         const previousAuthenticatedStatus = this.previousAuthenticatedStatusByUser.get(user.id);
         if (previousAuthenticatedStatus != null && previousAuthenticatedStatus !== authenticatedStatus) {
-            logger.warn(`Lobby: user ${user.username} authentication status changed from ${previousAuthenticatedStatus} to ${authenticatedStatus}`, { lobbyId: this.id, userName: user.username, userId: user.id });
+            const prevAuthProps = { authenticated: previousAuthenticatedStatus, username: user.username };
+            const newAuthProps = { authenticated: authenticatedStatus, username: user.socket.user.getUsername() };
+
+            logger.warn(`Lobby: user ${user.id} authentication status changed. Previous value: ${JSON.stringify(prevAuthProps)}, new value: ${JSON.stringify(newAuthProps)}`, { lobbyId: this.id, userId: user.id });
         }
 
         this.previousAuthenticatedStatusByUser.set(user.id, authenticatedStatus);


server/game/cards/06_SEC/units/HiredSlicer.ts
```diff
@@ -0,0 +1,61 @@
...
+            // Player chooses which deck to reveal from
+            targetResolver: {
+                mode: TargetMode.Select,
+                choices: (context) => ({
+                    'Your deck': AbilityHelper.immediateEffects.reveal(() => ({
+                        target: context.player.getTopCardsOfDeck(2)
+                    })),
+                    'Opponent\'s deck': AbilityHelper.immediateEffects.reveal(() => ({
+                        target: context.player.opponent.getTopCardsOfDeck(2)
+                    }))
```
moyerr: Nit: I think you can pass in the props directly instead of using a generator function with empty parameter list. So basically, each reveal effect would look like:

```typescript
AbilityHelper.immediateEffects.reveal({
    target: context.player.opponent.getTopCardsOfDeck(2)
})
```

server/game/cards/06_SEC/units/HiredSlicer.ts
```diff
@@ -0,0 +1,61 @@
...
+                    }
+                }
+
+                return {
+                    title: 'Exhaust a unit that shares a trait with one of those cards',
+                    optional: true,
+                    targetResolver: {
+                        cardTypeFilter: WildcardCardType.Unit,
+                        // Target must share at least one trait with the revealed cards
+                        cardCondition: (card) => Array.from(card.traits).some((trait: string) => revealedTraits.has(trait)),
```
moyerr: Nit: If you've got two sets already, you can do an intersection to see if they have any in common. I think we have a helper for this in Helpers.ts

test/server/cards/06_SEC/units/HiredSlicer.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+                expect(context.player1).toHavePrompt('Select one');
+                context.player1.clickPrompt('Opponent\'s deck');
+
+                // Top 2 cards are TIE/LN Fighter (Imperial, Vehicle) and Imperial Interceptor (Imperial, Vehicle)
+                // Note: Chat log uses lowercase 'ln', not 'LN'
+                expect(context.getChatLogs(2)).toContain('player1 uses Hired Slicer to reveal TIE/ln Fighter and Imperial Interceptor');
+
+                // Death Star Stormtrooper has Imperial trait, so it CAN be exhausted!
+                expect(context.player1).toBeAbleToSelectExactly([context.deathStarStormtrooper]);
+                context.player1.clickCard(context.deathStarStormtrooper);
```
moyerr: Just to verify the target filter applies to all units in play could we add the following to the test setup:
- A _friendly_ unit that shares a trait (just for one of these cases is okay, doesn't have to be both) with a revealed card
- A deployed leader that shares a trait

test/server/cards/06_SEC/units/HiredSlicer.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+                expect(context.player1).toBeAbleToSelectExactly([context.deathStarStormtrooper]);
+                context.player1.clickPrompt('Pass');
+
+                // Card should be moved to bottom of deck
+                expect(context.player1.deck.length).toBe(1);
+                expect(context.player1.deck[0]).toBe(context.battlefieldMarine);
+            });
+
+            // TODO: Once engine is updated, this may also need to be adjusted
+            it('should currently trigger ifYouDo even when deck is empty', function () {
```
AMMayberry1: Our practice for cases like this is to write up the correct test case and then just leave it commented out with a TODO like you have

AMMayberry1 (approved):

PR #1872, opened 2025-10-22
Add support for Let's Talk

Ravarix:
image

Followed Finalizer for implementation, except you must select all units *who could have a valid capture target*. 
The order of selection is significant and that will be the order of captures.


test/server/cards/06_SEC/events/LetsTalk.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+            context.player1.clickDone();
+
+            context.player1.clickCard(context.atst);
+            context.player1.clickCard(context.rebelPathfinder);
+
+            expect(context.atst).toBeCapturedBy(context.wampa);
+            expect(context.rebelPathfinder).toBeCapturedBy(context.battlefieldMarine);
+
+        });
+    });
```
VJubert: can you add more tests for this cards ?
- one with multiple arena
- one when opponent has more units than player1
- one with multiple bounty to validate trigger happen in the right window

you should add expectations each step with which unit are selectable too

server/game/cards/06_SEC/events/LetsTalk.ts
```diff
@@ -0,0 +1,63 @@
...
+                activePromptTitle: 'Choose the order for each friendly unit to capture an enemy non-leader unit in the same arena',
+                mode: TargetMode.ExactlyVariable,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: WildcardZoneName.AnyArena,
+                controller: RelativePlayer.Self,
+                numCardsFunc: (context) => context.player.getArenaUnits().length,
+            },
+            then: (chosenUnitsContext) => ({
+                title: 'Each friendly unit captures an enemy non-leader unit in the same arena',
+                immediateEffect: AbilityHelper.immediateEffects.simultaneous(
```
VJubert: open question : should it be sequential instead of simultaneous ?

test/server/cards/06_SEC/events/LetsTalk.spec.ts
```diff
@@ -0,0 +1,65 @@
+describe('Let\'s Talk', function () {
+    integration(function (contextRef) {
+        it('Let\'s Talk costs 6 when own unit is waylayed', async function () {
```
VJubert: costs 3 resources less when a friendly unit left play this phase

server/game/cards/06_SEC/events/LetsTalk.ts
```diff
@@ -0,0 +1,63 @@
...
+import { EventCard } from '../../../core/card/EventCard';
+import { EventName, RelativePlayer, TargetMode, WildcardCardType, WildcardZoneName } from '../../../core/Constants';
+import { CardsLeftPlayThisPhaseWatcher } from '../../../stateWatchers/CardsLeftPlayThisPhaseWatcher';
+
+export default class LetsTalk extends EventCard {
+    private cardsLeftPlayThisPhaseWatcher: CardsLeftPlayThisPhaseWatcher;
+
+    protected override getImplementationId() {
+        return {
+            id: 'lets-talk-id',
```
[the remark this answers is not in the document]
AMMayberry1: Just as a note, that means that the id here will need to switch to the actual FFG uuid. The get-cards script will automatically populate it in _cardMap once mockdata is updated

test/server/cards/06_SEC/events/LetsTalk.spec.ts
AMMayberry1: +1 to all of the above. We definitely need to include a leader unit in all of the tests as well just to make sure that the logic for determining when we stop capturing correctly accounts for that.

server/game/cards/06_SEC/events/LetsTalk.ts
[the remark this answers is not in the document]
AMMayberry1: Steps:
1. go to mockdata.js and remove the entry for Let's Talk
2. re-run `npm run get-cards`
3. go to `test/json/_cardMap.json` and find the entry for Let's Talk to get the real ID

mockdata is a script we use to inject definitions for the cards before they appear in the official FFG database

test/server/cards/06_SEC/events/LetsTalk.spec.ts
```diff
@@ -15,14 +16,12 @@ describe('Let\'s Talk', function () {
...
 
-            context.player1.readyResources(10);
-
             context.player1.clickCard(context.waylay);
             context.player1.clickCard(context.battlefieldMarine);
 
             context.player2.passAction();
 
-            context.player1.readyResources(10);
+            context.player1.readyResources(3); // undo waylay cost
```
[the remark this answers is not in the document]
Ravarix: My goal was trying to make the final `expect(context.player1.exhaustedResourceCount).toBe(6);` clear, as we're expecting the event to cost 6. Happy to remove the line and make the final check `9`.

server/game/cards/06_SEC/events/LetsTalk.ts
```diff
@@ -0,0 +1,98 @@
...
+            arena: zoneName,
+            condition: (card) => card.isNonLeaderUnit()
+        }).length;
+    }
+
+    private countSelectedCardsInZone(selectedCards: Card[], zoneName: ZoneName): number {
+        return selectedCards.filter((selectedCard) => selectedCard.zoneName === zoneName).length;
+    }
+
+
```
AMMayberry1: nit: extra newline

AMMayberry1 (approved):

PR #1879, opened 2025-10-23
Fully Armed and Operational

moyerr:


server/game/cards/06_SEC/events/FullyArmedAndOperational.ts
```diff
@@ -0,0 +1,70 @@
...
+                    })
+                })
+            }))
+        });
+    }
+
+    private opponentDidAttackBaseLastAction(context: AbilityContext): boolean {
+        const opponentsLastActionNumber = this.actionsThisPhaseWatcher.previousActionNumberForPlayer(context.player.opponent);
+
+        if (opponentsLastActionNumber === null) {
```
AMMayberry1: nit: should be `==` for any comparison with `null` or `undefined`

server/game/core/gameSteps/ActionWindow.ts
```diff
@@ -223,7 +223,10 @@ export class ActionWindow extends UiPrompt {
     }
 
     public override complete() {
-        this.game.emitEvent(EventName.OnActionTaken, null, { player: this.activePlayer });
+        this.game.emitEvent(EventName.OnActionTaken, null, {
+            player: this.activePlayer,
+            number: this.actionNumber
```
AMMayberry1: nit: can we call this property `actionNumber`, since these events can be kind of anonymous it's helpful to be uber-explicit in my experience

AMMayberry1 (approved):

PR #1881, opened 2025-10-24
Updated cosmetics

mariotee:


server/services/DynamoDBInterfaces.ts
```diff
@@ -46,14 +46,19 @@ export interface IDeckStatsEntity {
...
 export interface UserPreferences {
-    cardback?: string;
     sound?: {
         muteAllSound?: boolean;
         muteCardAndButtonClickSound?: boolean;
         muteYourTurn?: boolean;
         muteChatSound?: boolean;
         muteOpponentFoundSound?: boolean;
     };
+    cosmetics?: {
```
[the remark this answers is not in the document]
mariotee: i did test this locally

docker-compose.yml
```diff
@@ -14,6 +14,20 @@ services:
       interval: 10s
       timeout: 5s
       retries: 3
+  s3-local:
```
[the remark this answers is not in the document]
mariotee: will comment out

server/services/DynamoDBService.ts
```diff
@@ -83,7 +83,11 @@ class DynamoDBService {
         }
 
         const dbClient = new DynamoDBClient(dbClientConfig);
-        this.client = DynamoDBDocumentClient.from(dbClient);
+        this.client = DynamoDBDocumentClient.from(dbClient, {
+            marshallOptions: {
+                removeUndefinedValues: true
```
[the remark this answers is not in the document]
mariotee: actually i'll have to double check that. it might have been an issue that came up during the schema change. i'll tear down the DynamoDB and make a new one

server/gamenode/GameServer.ts
```diff
@@ -965,6 +969,145 @@ export class GameServer {
...
+        // Cosmetics API endpoints
+        app.get('/api/cosmetics', async (req, res, next) => {
+            try {
+                let cosmetics = await this.cosmeticsService.getCosmeticsAsync();
+
+                if (cosmetics.length === 0) {
+                    try {
+                        // Load fallback cosmetics from file
+                        const fallbackCosmetics = await getFallbackCosmetics(logger);
+                        await this.cosmeticsService.initializeCosmeticsAsync(fallbackCosmetics);
```
AMMayberry1: Will this ever be needed after the first run? If not, let's just remove it from this pathway but still check in the method as like a helper utility for dev work. I can run it once manually before we merge the PR once we feel confident in the cosmetics schema

server/gamenode/GameServer.ts
```diff
@@ -965,6 +969,145 @@ export class GameServer {
...
+                    message: 'Cosmetic deleted successfully'
+                });
+            } catch (error) {
+                logger.error('GameServer (delete-cosmetic) Server error:', error);
+                next(error);
+            }
+        });
+
+        // Cleanup endpoints for development use only
+        app.delete('/api/cosmetics/cleanup/all', authMiddleware('admin-cosmetics'), async (req, res, next) => {
```
AMMayberry1: Both to be a smidge safer and also for general code hygiene, let's maybe factor these off into a separate `setupDevAppRoutes()` that lives behind an env var check so that the routes just don't even exist at all in prod

server/utils/adminUtils.ts
```diff
@@ -0,0 +1,30 @@
+import { logger } from '../logger';
+
+// List of allowed user IDs for admin access - must match exactly what NextAuth provides
+const allowedAdminUsers = [
```
AMMayberry1: I still want to migrate these to a DB entry, user IDs are _more_ secure but we don't to be advertising publicly what the admin IDs are just in case someone finds a way to spoof it. If it would be helpful for time purposes we could ask Chebato to do that part as a separate PR

server/gamenode/GameServer.ts
mariotee: what is "it" in "run it once manually"?

AMMayberry1: Initializing the cosmetics. That should be something we only ever need to do one time ever, right?

server/utils/adminUtils.ts
```diff
@@ -1,30 +1,65 @@
...
+
+interface IAuthError {
+    status: number;
+    error: string;
+}
 
-    // Check if user has admin privileges using the user ID
-    const userId = (user as any)?.id || (user as any)?.userId;
-    const isAdmin = isAdminUser(userId);
+export const checkAdminUserPrivilegesAsync = async (userId: string, role: AdminRole): Promise<IAuthError | null> => {
```
AMMayberry1: If there's a failure, somewhere in this flow we'll also want to emit a message with `logger.error()` indicating the user id and the API name

server/utils/adminUtils.ts
```diff
@@ -1,30 +1,65 @@
...
+
+async function isModeratorAsync(userId: string): Promise<boolean> {
+    const db = await getDynamoDbServiceAsync();
+    const adminUsers = await db.getAdminUsersAsync();
+    return adminUsers.mod.some((modUserId) => modUserId === userId);
+}
+
+interface IAuthError {
+    status: number;
+    error: string;
```
AMMayberry1: nit: `errorMessage` is a bit clearer here

server/gamenode/GameServer.ts
```diff
@@ -1032,49 +1017,101 @@ export class GameServer {
...
-        app.delete('/api/cosmetics/cleanup/all', authMiddleware('admin-cosmetics'), async (req, res, next) => {
+        // Dev user check endpoint
+        app.get('/api/user-is-dev', authMiddleware('user-is-dev'), async (req, res, next) => {
             try {
-                if (process.env.ENVIRONMENT !== 'development') {
-                    return res.status(404).json({
+                // Check if user is authenticated and has dev privileges or higher
+                const authError = await checkAdminUserPrivilegesAsync(req.user.getId(), AdminRole.Developer);
+                if (authError) {
+                    return res.status(authError.status).json({
```
AMMayberry1: For these `user-is-[x]` APIs I feel like it's maybe more idiomatic to always return 200 (unless there's a DB failure) and then include a property that's a true / false in the response. imo status code should indicate if the request succeeded in execution or not, but here we're kind of using it as a sort of boolean return value

server/gamenode/GameServer.ts
```diff
@@ -1032,49 +1017,101 @@ export class GameServer {
...
-        app.delete('/api/cosmetics/:cosmeticId', authMiddleware('admin-cosmetics'), async (req, res, next) => {
+        app.delete('/api/cosmetics/:cosmeticId', authMiddleware('delete-cosmetics-by-id'), async (req, res, next) => {
             try {
                 const { cosmeticId } = req.params;
                 const user = req.user;
 
                 // Check if user is authenticated and has admin privileges
-                if (!user || user.isAnonymousUser()) {
-                    return res.status(401).json({
+                const authError = await checkAdminUserPrivilegesAsync(user, AdminRole.Developer);
```
AMMayberry1: Wondering if it would be cleaner to have these `AdminRole` requirements as like an optional parameter on the `authMiddleware` so we won't need this boilerplate code on all the routes.

For the `user-is-[x]` APIs they would stay the way they currently are since the goal there is to return a result indicating the auth status at that level.

server/services/DynamoDBInterfaces.ts
```diff
@@ -112,4 +112,16 @@ export interface IGameRecordEntity {
     player2BaseId: string;
     timestampStart: Date;
     timestampEnd: Date;
+}
+
+export enum AdminRole {
+    Admin = 'admin',
+    Developer = 'dev',
```
AMMayberry1: nit: our convention with the string representations for these is to always have it match the enum name. so in this case `Developer = 'developer'` and `Moderator = 'moderator'`. being consistent about that helps reduce the chance of mismatched values for messages between BE and FE

server/services/DynamoDBInterfaceSchemas.ts
```diff
@@ -41,3 +41,9 @@ export const iDeckDataEntitySchema = z.object({
     stats: iDeckStatsEntitySchema.optional(),
 }) as z.ZodType<IDeckDataEntity>;
 
+// generated by chat-gpt
+export const AdminUserListsSchema = z.object({
+    admins: z.array(z.string().uuid()),
```
AMMayberry1: These names don't match what's currently in `IAdminUserListsEntity`

server/gamenode/GameServer.ts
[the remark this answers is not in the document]
mariotee: i was thinking at first a regular mod wouldn't be able to delete. but i can update if we're okay with that

server/gamenode/GameServer.ts
```diff
@@ -1009,6 +1020,131 @@ export class GameServer {
                 next(err);
             }
         });
+
+        // Cosmetics API endpoints
+        app.get('/api/cosmetics', async (req, res, next) => {
+            try {
+                let cosmetics = defaultCosmetics;
+                if (process.env.ENVIRONMENT === 'development') {
```
AMMayberry1: Ah okay now that I'm looking a bit closer, there's definitely some room for cleanup here. The two code paths (development and production) are pretty much identical, it's definitely possible to reduce duplicated code and have everything be one flow with maybe a slight deviation between the two

server/services/DynamoDBService.ts
```diff
@@ -10,10 +10,11 @@ import {
 } from '@aws-sdk/lib-dynamodb';
 import { logger } from '../logger';
 import * as Contract from '../game/core/utils/Contract';
-import type { IDeckDataEntity, IDeckStatsEntity, IUserProfileDataEntity, UserPreferences } from './DynamoDBInterfaces';
+import { type IDeckDataEntity, type IDeckStatsEntity, type IUserProfileDataEntity, type IUserPreferences, type IServerRoleUsersListsEntity } from './DynamoDBInterfaces';
 import { z } from 'zod';
 import { iDeckDataEntitySchema, iDeckStatsEntitySchema } from './DynamoDBInterfaceSchemas';
```
AMMayberry1: nit: since we're in here anyway, can we fix the capitalization on the first letter on these - should be e.g. `IDeckDataEntitySchema`

seed-dynamo.ts
```diff
@@ -0,0 +1,38 @@
+// scripts/seedCosmetics.ts
```
AMMayberry1: Since this is different than all of our other script files (it's TS), we should add a comment here saying that the way to run this is with `ts-node`. that will also be mentioned in the wiki but good to have it here.

scripts/syncS3.ts
```diff
@@ -0,0 +1,123 @@
+// copyS3ToLocal.ts
+
+// to run this script use ts-node scripts/syncS3.ts
+import {
+    S3Client,
+    ListObjectsV2Command,
+    GetObjectCommand,
+    PutObjectCommand,
+    CreateBucketCommand
+} from '@aws-sdk/client-s3';
```
AMMayberry1: This needs to be added to package.json as a dev dependency. We should also add ts-node as a dev dependency.

AMMayberry1:
General note - if you didn't already, let's test on mobile to make sure everything scales right (particularly the playmats and the customization preferences page)

scripts/syncS3.ts
```diff
@@ -0,0 +1,123 @@
+// copyS3ToLocal.ts
```
AMMayberry1: The "Deathstar II" playmat didn't auto-populate for me for some reason, or at least it's showing up as empty (see below). If it looks right on your end, lmk and I'll debug some to see if I can find the issue on my side. 

Image

scripts/seed-dynamo.ts
```diff
@@ -0,0 +1,50 @@
+// scripts/seedCosmetics.ts
```
AMMayberry1: Would it make sense for this script to just be the same as syncS3? Is there ever a situation where you would want to run one but no the other?

scripts/syncS3.ts
```diff
@@ -0,0 +1,123 @@
...
+import type { Readable } from 'stream';
+import '../server/env';
+
+const BUCKET_NAME = 'karabast-customization';
+
+const prodS3 = new S3Client({
+    region: 'us-east-1',
+    credentials: {
+        accessKeyId: process.env.AWS_ACCESS_KEY_ID ?? '',
+        secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY ?? ''
```
AMMayberry1: Since these are required, we should also have an error here instead of providing the `''` default value

docker-compose.yml
```diff
@@ -15,6 +15,22 @@ services:
       timeout: 5s
       retries: 3
 
+  # Uncomment the following section to enable a local S3 mock server
```
AMMayberry1: We can just remove this now that it's configured on the FE side

CheBato:
> General note - if you didn't already, let's test on mobile to make sure everything scales right (particularly the playmats and the customization preferences page)

I'll take a look tomorow for the mobile version.

scripts/syncS3.ts
```diff
@@ -0,0 +1,174 @@
...
+        secretAccessKey: 'verySecretKey1'
+    }
+});
+
+
+/**
+ * Seeds the local DynamoDB with fallback cosmetics data.
+ * Only runs when ENVIRONMENT=development and USE_LOCAL_DYNAMODB=true
+ */
+async function seedCosmetics(): Promise<void> {
```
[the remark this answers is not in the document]
CheBato: Yes there is. The difference is that we get rid of the checks ENVIRONMENT=development and USE_LOCAL_DYNAMODB=true and ofcourse we need to uncomment the cosmeticsData that is commented since that is for production. After we do that it should work like it did from your first test haha

AMMayberry1 (approved):

PR #1882, opened 2025-10-24
Confidence in Victory

moyerr:

### DIFF
--- a/server/game/gameSystems/WinGameSystem.ts
+++ b/server/game/gameSystems/WinGameSystem.ts
@@ -1,4 +1,5 @@
 import type { AbilityContext } from '../core/ability/AbilityContext.js';
+import type { FormatMessage } from '../core/chat/GameChat.js';
 import { MetaEventName } from '../core/Constants.js';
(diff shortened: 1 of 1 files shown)


server/game/gameSystems/WinGameSystem.ts
```diff
@@ -0,0 +1,30 @@
+import type { AbilityContext } from '../core/ability/AbilityContext.js';
+import { MetaEventName } from '../core/Constants.js';
+import type { IPlayerTargetSystemProperties } from '../core/gameSystem/PlayerTargetSystem.js';
+import { PlayerTargetSystem } from '../core/gameSystem/PlayerTargetSystem.js';
+import type { Player } from '../core/Player.js';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface IWinGameProperties extends IPlayerTargetSystemProperties { }
```
AMMayberry1: Let's maybe add a "winReason" string (or `(context) => string` handler) that will emit a win reason, I expect this will catch people off guard more than once lol. And add at least test case to make sure it's emitted right

PR #1888, opened 2025-10-27
Galen Erso

AMMayberry1:


test/server/cards/06_SEC/units/GalenErsoYoullNeverWin.spec.ts
```diff
@@ -0,0 +1,1325 @@
...
+
+                context.player2.clickCard(context.battlefieldMarine);
+                context.player1.passAction();
+                context.player2.clickCard(context.lukeSkywalker);
+                context.player2.clickPrompt('Give a shield to a heroism unit you played this phase');
+                context.player2.clickCard(context.battlefieldMarine);
+                expect(context.battlefieldMarine).toHaveExactUpgradeNames(['shield']);
+            });
+
+            it('should not be blanked when their title is named', async function () {
```
moyerr: nit: case name should indicate that the leader is deployed or is a "leader unit" to distinguish from previous test case

PR #1893, opened 2025-10-28
Add Vigil

adbronico:

### DIFF
--- a/test/server/cards/06_SEC/units/VigilSecuringTheFuture.spec.ts
+++ b/test/server/cards/06_SEC/units/VigilSecuringTheFuture.spec.ts
@@ -108,6 +108,40 @@ describe('Vigil, Securing the Future', function() {
                 expect(context.vigil.damage).toBe(2);
             });
 
+            it('should not increase damage dealt to Vigil if the damage was reduced to 0 first', async function () {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['finn#on-the-run'],
+                        spaceArena: ['vigil#securing-the-future']
+                    },
+                    player2: {
+                        spaceArena: ['concord-dawn-interceptors']
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.finn);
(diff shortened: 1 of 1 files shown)


moyerr (commented):
Awesome work, dude, this looks really good! I'll let @AMMayberry1 come give final sign-off but I really like this 💪

server/game/core/gameSteps/abilityWindow/ReplacementEffectWindow.ts
```diff
@@ -43,13 +45,33 @@ export class ReplacementEffectWindow extends TriggerWindowBase {
...
+    private getAllReplacedAbilities(context: TriggeredAbilityContext, event: GameEvent): Set<PlayerOrCardAbility> {
+        const replacedAbilities = new Set<PlayerOrCardAbility>();
+
+        if (event.isReplacementEvent) {
+            if (event.context.ability === context.ability) {
+                replacedAbilities.add(event.context.ability);
+            }
+
+            if (event.replacesEvent) {
+                this.getAllReplacedAbilities(context, event.replacesEvent).forEach((ability) => {
```
AMMayberry1: What's the advantage of recursion here? My understanding is that it would just be a straight line of replacement events, so couldn't we just walk up the chain of `event.replacesEvent` with a loop until we hit null? The recursion feels like a ton of overhead for something that simple

test/server/cards/06_SEC/units/VigilSecuringTheFuture.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+                context.player1.passAction();
+                context.player2.clickCard(context.torpedoBarrage);
+                context.player2.clickPrompt('Deal indirect damage to opponent');
+                context.player1.setDistributeIndirectDamagePromptState(new Map([
+                    [context.vigil, 5]
+                ]));
+
+                expect(context.vigil).toBeInZone('discard');
+                expect(context.player1).toBeActivePlayer();
+            });
```
AMMayberry1: Let's do a test with Paige Tico attached to Vigil to confirm the damage isn't increased, since Vigil would technically be dealing the damage to itself

PR #1904, opened 2025-11-03
Always two fix

mnechromancer:
### Problem
The card "Always Two" was implemented with `TargetMode.Exactly` requiring exactly 2 unique Sith units to be selected. This caused the defeat effect to fail when fewer than 2 valid targets were available, making the card unplayable in those situations.

### Solution
Changed `TargetMode.Exactly` to `TargetMode.UpTo` to allow selecting 0-2 unique Sith units. The defeat effect remains in the same structural location (within the `targetResolver`'s `sequential` `immediateEffect`) to maintain consistency with the original implementation.

### Changes
- **Card Implementation** (AlwaysTwo.ts): Changed target mode from `Exactly` to `UpTo`
- **Tests** (AlwaysTwo.spec.ts): Updated and added test cases to cover:
  - Selecting 2 units (original behavior maintained)
  - Selecting 1 unit (new: defeat effect still executes)
  - Selecting 0 units (new: defeat effect still executes)

### Behavior
The "defeat all other friendly units" effect now always executes regardless of how many unique Sith units are selected:
- ✅ 2 units selected → buff those 2, defeat all others
- ✅ 1 unit selected → buff that 1, defeat all others
- ✅ 0 units selected → defeat all friendly units

### Testing
All tests pass, including new test cases that verify the defeat effect executes even when fewer than 2 targets are available.

---

### DIFF
--- a/server/game/core/Player.ts
+++ b/server/game/core/Player.ts
@@ -553,7 +553,6 @@ export class Player extends GameObject<IPlayerState> implements IGameStatisticsT
     public isCardInPlayableZone(card: Card, playingType: PlayType = null) {
         // Check if card can be legally played by this player out of discard from an ongoing effect
         if (
-            playingType === PlayType.PlayFromOutOfPlay &&
             card.zoneName === ZoneName.Discard &&
             card.hasOngoingEffect(EffectName.CanPlayFromDiscard)
         ) {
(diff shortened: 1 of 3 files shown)


test/server/cards/05_LOF/events/AlwaysTwo.spec.ts
```diff
@@ -88,12 +86,65 @@ describe('Always Two', function () {
             // Player 1 plays Always Two
             context.player1.clickCard(context.alwaysTwo);
 
-            // Player 1 is warned that the card will have no effect
-            expect(context.player1).toHavePrompt('Playing Always Two will have no effect. Are you sure you want to play it?');
-            context.player1.clickPrompt('Play anyway');
+            // Player 1 can select the one available unique Sith unit
+            expect(context.player1).toHavePrompt(prompt);
+            expect(context.player1).toBeAbleToSelectExactly([context.scimitar]);
```
AMMayberry1: This is incorrect. The way it's worded, you have to select two unique Sith units. If you don't select exactly two, then the "if you do" part will not happen. The defeat still will though

AMMayberry1 (approved):

PR #1914, opened 2025-11-05
"(No effect)" for modal abilities

mnechromancer:
no_effect_modal

### DIFF
--- a/test/helpers/CustomMatchers.js
+++ b/test/helpers/CustomMatchers.js
@@ -68,16 +68,7 @@ var customMatchers = {
 
                 for (let expected of expecteds) {
                     result.pass = buttons.some(
-                        (button) => {
-                            if (button.disabled) {
-                                return false;
-                            }
-                            // Match exact text or text with "(No effect)" prefix
-                            const buttonText = button.text.toString().toLowerCase();
-                            const expectedText = expected.toString().toLowerCase();
-                            return util.equals(button.text, expected, customEqualityMatchers) ||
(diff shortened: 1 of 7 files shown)


server/game/gameSystems/ChooseModalEffectsSystem.ts
```diff
@@ -35,10 +35,19 @@ export class ChooseModalEffectsSystem<TContext extends AbilityContext = AbilityC
             if (amountOfRemainingChoices === 0) {
                 return;
             }
+
+            // Create a mapping of original labels to annotated display labels
+            const choiceMapping = new Map<string, string>();
+            Object.entries(listOfAvailableEffects).forEach(([label, system]) => {
+                const hasEffect = this.modalChoiceHasAnyLegalEffect(system, context);
+                const displayLabel = hasEffect ? label : `${label} (No effect)`;
```
moyerr: I think this should be a prefix to (a) make it immediately noticeable and (b) match what we have for the triggered ability resolution prompt

AMMayberry1: nit: let's have (No effect) be a prefix instead of a suffix like we do in the trigger window. people are lazy and may not read all the way to the end of the text lol

mnechromancer: That makes total sense, changing now

AMMayberry1 (approved):

PR #1915, opened 2025-11-05
Add support for providing a factory method for ability titles

moyerr:
This update allows us to provide clarity within certain ability titles that may want to reference specific information about the board state or triggering event. These changes include an update to Grand Admiral Thrawn's ability titles, to provide clarity about which unit's **When Defeated** ability is getting repeated:

| Ability resolution choice | Optional ability trigger |
| :---: | :---: |
|  |  |

### DIFF
--- a/server/game/cards/05_LOF/events/AlwaysTwo.ts
+++ b/server/game/cards/05_LOF/events/AlwaysTwo.ts
@@ -20,22 +20,19 @@ export default class AlwaysTwo extends EventCard {
                 cardTypeFilter: WildcardCardType.Unit,
                 controller: RelativePlayer.Self,
                 cardCondition: (card) => card.unique && card.hasSomeTrait(Trait.Sith),
-                immediateEffect: AbilityHelper.immediateEffects.sequential([
-                    AbilityHelper.immediateEffects.simultaneous([
-                        AbilityHelper.immediateEffects.giveShield({
-                            amount: 2
-                        }),
-                        AbilityHelper.immediateEffects.giveExperience({
-                            amount: 2
-                        }),
-                    ]),
-                    AbilityHelper.immediateEffects.defeat((context) => ({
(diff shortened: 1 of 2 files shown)


server/game/cards/04_JTL/leaders/GrandAdmiralThrawnHowUnfortunate.ts
```diff
@@ -13,14 +13,16 @@ export default class GrandAdmiralThrawnHowUnfortunate extends LeaderUnitCard {
 
     protected override setupLeaderSideAbilities(registrar: ILeaderUnitLeaderSideAbilityRegistrar, AbilityHelper: IAbilityHelper) {
         registrar.addTriggeredAbility({
-            title: 'Exhaust this leader to use the When Defeated ability again',
+            title: (context) => (context
+                ? `Exhaust ${context.source.title} to use ${context.event.card.title}'s When Defeated ability again`
```
AMMayberry1: nit: let's maybe put quotes around "When Defeated" for prompts of this type to make it easier to read (e.g., `to use Arihnda Pryce's "When Defeated" ability again`)

server/game/core/ability/PlayerOrCardAbility.ts
```diff
@@ -128,8 +127,14 @@ export abstract class PlayerOrCardAbility<T extends IPlayerOrCardAbilityState =
 
     public override toString() {
         return this.properties.cardName
-            ? `'${this.properties.cardName} ability: ${this.title}'`
-            : `'Ability: ${this.title}'`;
+            ? `'${this.properties.cardName} ability: ${this.getTitle()}'`
+            : `'Ability: ${this.getTitle()}'`;
```
AMMayberry1: How will this work with no context parameter? Might need to use `getFrameworkContext()` just so we have something to pass in and don't get null errors in the debugger when it tries to do this

PR #1918, opened 2025-11-07
Update Let's Call it War to use a then

adbronico:


server/game/cards/06_SEC/events/LetsCallItWar.ts
```diff
@@ -13,22 +13,22 @@ export default class LetsCallItWar extends EventCard {
 
     public override setupCardAbilities(registrar: IEventAbilityRegistrar, AbilityHelper: IAbilityHelper) {
         registrar.setEventAbility({
-            title: 'Deal 3 damage to a unit. If you have the initiative, you may deal 2 damage to another unit in the same arena.',
-            targetResolvers: {
-                firstUnit: {
+            title: 'Deal 3 damage to a unit',
```
AMMayberry1: nit: title should stay the same

AMMayberry1 (approved):

PR #1925, opened 2025-11-11
Improve Savage Opress log

### LINKED ISSUE #1454: LOF Savage Oppress unit doesn't log anything for damaging base
Related issue: https://github.com/SWU-Karabast/forceteki/issues/1462

If the LOF Savage Oppress unit ability is triggered on play (9 damage to base), nothing appears in the chat window specifically:

Image

VJubert:
Currently, logs are :

- player1 uses Savage Opress to use the Force
- player1 uses Savage Opress to deal 9 damage to their own base

I update the first log to 'player1 uses Savage Opress to not take damage by using the Force'

Improve and resolve #1454

### DIFF
--- a/server/game/cards/05_LOF/units/SavageOpressImbuedWithHate.ts
+++ b/server/game/cards/05_LOF/units/SavageOpressImbuedWithHate.ts
@@ -19,7 +19,7 @@ export default class SavageOpressImbuedWithHate extends NonLeaderUnitCard {
             },
             optional: true,
             immediateEffect: AbilityHelper.immediateEffects.useTheForce(),
-            effect: 'not take damage by using the Force',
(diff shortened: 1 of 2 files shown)


server/game/cards/05_LOF/units/SavageOpressImbuedWithHate.ts
```diff
@@ -19,6 +19,7 @@ export default class SavageOpressImbuedWithHate extends NonLeaderUnitCard {
             },
             optional: true,
             immediateEffect: AbilityHelper.immediateEffects.useTheForce(),
+            effect: 'not take damage by using the Force',
```
AMMayberry1: nit, clarity: maybe `'not damage their base by instead using the Force'`

PR #1927, opened 2025-11-11
Refactor cost systems in preparation for Vuutun Palaa

AMMayberry1:
Being able to properly handle the Vuutun Palaa + Exploit interaction requires making some changes to our cost adjustment flow so that we can maintain a sort of "adjuster context" by which different adjuster stages can pass context information so that the Exploit + VP interaction can be handled correctly.

Stage 1 of 2 for the VP implementation is refactoring our current cost system to use this adjuster context data. I've made the following changes in that direction:

- Consolidated the cost adjuster evaluation logic out of `Player` and into `ResourceCost` and `PlayCardResourceCost` to increase readability and have all of the cost-related logic live together
- Broke the cost adjuster logic into: (a) an "evaluation" pass which builds up the initial adjuster context and calculates max possible discount to determine if the cost can be paid at all, and (b) an "apply / trigger" pass which uses that context to apply the adjustments to the cost
- Refactored the cost logic in `AbilityResolver` and `PlayerOrCardAbility` a bit to leverage these changes. generally it's simplifying the logic in those places and pushing it into the Cost and Adjuster classes

### DIFF
--- a/server/game/cards/01_SOR/events/TheEmperorsLegion.ts
+++ b/server/game/cards/01_SOR/events/TheEmperorsLegion.ts
@@ -24,7 +24,7 @@ export default class TheEmperorsLegion extends EventCard {
             title: 'Return each unit in your discard pile that was defeated this phase to your hand.',
             immediateEffect: AbilityHelper.immediateEffects.returnToHand((context) => {
                 const friendlyUnitsDefeatedThisPhaseInDiscard =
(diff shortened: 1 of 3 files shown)


adbronico (dismissed):

moyerr (approved):
Great work with this! Really glad I read through it all. It started to click once I got to the `ResourceCost.ts` file. Also really nice to see some of the complexity around costs just evaporate from `Player.ts`, `AbilityResolver.ts`, and `PlayerOrCardAbility.ts`. Love it 🚀

PR #1947, opened 2025-11-24
[Bugfix] Update cards & tests that check control of leader unit

moyerr:
There are 5 cards that contain the text "if you control a leader unit":

- Outer Rim Headhunter (SOR)
- Jedha Agitator (SOR)
- Rugged Survivors (SOR)
- Steadfast Battalion (SOR)
- Chancellor Palpatine (SEC)

The condition check we were performing wasn't exactly right for these cards, and it was most obvious with pilot leaders. Specifically, the Poe Dameron leader was satisfying the condition when he shouldn't.

### DIFF
--- a/server/game/cards/01_SOR/units/JedhaAgitator.ts
+++ b/server/game/cards/01_SOR/units/JedhaAgitator.ts
@@ -1,7 +1,7 @@
 import type { IAbilityHelper } from '../../../AbilityHelper';
 import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
 import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
-import { ZoneName } from '../../../core/Constants';
+import { CardType, ZoneName } from '../../../core/Constants';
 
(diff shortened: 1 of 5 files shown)


server/game/cards/01_SOR/units/JedhaAgitator.ts
```diff
@@ -15,9 +15,10 @@ export default class JedhaAgitator extends NonLeaderUnitCard {
         registrar.addOnAttackAbility({
             title: 'If you control a leader unit, deal 2 damage to a ground unit or base',
             targetResolver: {
+                activePromptTitle: 'Deal 2 damage to a ground unit or base',
                 cardCondition: (card) => (card.isUnit() && card.zoneName === ZoneName.GroundArena) || card.isBase(),
                 immediateEffect: AbilityHelper.immediateEffects.conditional({
-                    condition: (context) => context.player.leader.isDeployableLeader() && context.player.leader.deployed,
+                    condition: (context) => context.player.hasSomeArenaUnit({ condition: (card) => card.isLeaderUnit() }),
```
AMMayberry1: nit: we could do this in a slightly more compact way with `hasSomeArenaUnit({ type: CardType.LeaderUnit })`

PR #1948, opened 2025-11-24
added animation system

mariotee:
feat: added defeat animation for units and upgrades

feat: added animation for shield breaks

chore: prep for animations prefs

feat: added healing animation

### DIFF
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -20,6 +20,7 @@ const { AbilityResolver } = require('./gameSteps/AbilityResolver.js');
 const { AbilityContext } = require('./ability/AbilityContext.js');
 const Contract = require('./utils/Contract.js');
 const { cards } = require('../cards/Index.js');
+const { ServerAnimationQueue } = require('./animation/ServerAnimationQueue.js');
 
 const { EventName, ZoneName, Trait, WildcardZoneName, TokenUpgradeName, TokenUnitName, PhaseName, TokenCardName, AlertType, SnapshotType, RollbackRoundEntryPoint, RollbackSetupEntryPoint, GameErrorSeverity, GameEndReason } = require('./Constants.js');
 const { StateWatcherRegistrar } = require('./stateWatcher/StateWatcherRegistrar.js');
@@ -302,6 +303,9 @@ class Game extends EventEmitter {
             currentPhase: null,
             prevActionPhasePlayerPassed: null
         };
+
+        /** @private @type {import('./animation/ServerAnimationQueue.js').ServerAnimationQueue} */
+        this.animationQueue = new ServerAnimationQueue();
 
         this.tokenFactories = null;
         this.stateWatcherRegistrar = new StateWatcherRegistrar(this);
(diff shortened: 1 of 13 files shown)


server/game/core/animation/AnimationBuilder.ts
```diff
@@ -0,0 +1,323 @@
...
+     * @param targetId - Card UUID or base identifier (P1BASE/P2BASE)
+     * @param amount - Damage value (positive integer)
+     * @param sourceId - Optional attacker/source UUID
+     * @param metadata - Optional additional metadata
+     * @returns Animation event for damage, or null if amount is invalid
+     */
+    public static damage(
+        targetId: string,
+        amount: number,
+        sourceId?: string,
```
AMMayberry1: What's the point of sourceId? Looks like we aren't really using it for anything

server/game/core/animation/ServerAnimationQueue.ts
```diff
@@ -0,0 +1,141 @@
...
+    public addMultiple(events: AnimationEvent[]): void {
+        this.events.push(...events);
+    }
+
+    /**
+     * Optimizes the animation queue by combining duplicate or similar events.
+     * - Combines damage/heal events on the same target
+     * - Eliminates duplicate state changes
+     */
+    public optimize(): void {
```
AMMayberry1: I'm a little confused about the reasoning for when we would optimize vs when we wouldn't. On the client side it looks like we're playing the events in sequence, with the intent being to show the events one at a time. But here we are (sometimes) collapsing multiple events into one and removing steps. I think we should decide whether our intent is to show all events in sequence or try and collapse into as few animations as possible, and design around that.

server/game/core/event/EventWindow.ts
```diff
@@ -210,6 +211,118 @@ export class EventWindow extends BaseStepWithPipeline {
...
+    }
+
+    /**
+     * Generic animation batching system.
+     * Collects animation data from resolved events and batches events of the same type
+     * that occurred in the same window to display simultaneously.
+     */
+    private batchAnimations() {
+        // Group animation data by event type
+        const animationsByEvent = new Map<string, Record<string, unknown>[]>();
```
AMMayberry1: nit: we prefer Map over Record where possible

server/game/core/Game.js
```diff
@@ -1466,6 +1470,31 @@ class Game extends EventEmitter {
         this.emit(event.name, event);
     }
 
+    /**
+     * Queues a single animation event to be sent to the client with the next game state update.
+     * @param {import('./animation/AnimationTypes').AnimationEvent} animationEvent
+     */
+    queueAnimation(animationEvent) {
```
AMMayberry1: Same note here, we can consolidate these down to one signature

server/game/core/Game.js
```diff
@@ -1466,6 +1470,31 @@ class Game extends EventEmitter {
...
+        this.animationQueue.addMultiple(animationEvents);
+    }
+
+    /**
+     * Clears the animation queue and resets optimization flag. Should be called after
+     * all players have received their game state with animations.
+     */
+    clearAnimations() {
+        this.animationQueue.clear();
+        this._animationsOptimized = false;
```
AMMayberry1: Feels like this should be a property on `AnimationQueue`

PR #1958, opened 2025-11-30
#env-check

CheBato:
- Added an additional check for when env is development and USE_LOCAL_DYNAMODB is set to true while dynamoDB isn't started.


server/services/DynamoDBService.ts
```diff
@@ -39,6 +39,10 @@ export async function getDynamoDbServiceAsync() {
         try {
             await dynamoDbService.ensureLocalTableExistsAsync();
         } catch (err) {
+            if (process.env.ENVIRONMENT === 'development' && process.env.USE_LOCAL_DYNAMODB === 'true') {
```
AMMayberry1: We don't need to check `ENVIRONMENT` here, that would've been checked for `isLocalMode` already

AMMayberry1 (approved):

PR #1964, opened 2025-12-02
Added getBlockedFromPlayReason

mnechromancer:
Added method to determine if card is blocked from play by opponent's effect. This is backend logic for the frontend implementation of the lock icon, visible when an opponent blocks a player's cards from being played (via cards like Trade Route Taxation & Regional Governor).

### DIFF
--- a/server/game/gameSystems/DrawSystem.ts
+++ b/server/game/gameSystems/DrawSystem.ts
@@ -31,6 +31,7 @@ export class DrawSystem<TContext extends AbilityContext = AbilityContext> extend
             gameEvent.context.game.snapshotManager.setRequiresConfirmationToRollbackCurrentSnapshot(gameEvent.context.player.id);
         }
 
+        event.cards = event.player.drawDeck.slice(0, event.amount);
         event.player.drawCardsToHand(event.amount);
     }
 
@@ -64,7 +65,6 @@ export class DrawSystem<TContext extends AbilityContext = AbilityContext> extend
     protected override addPropertiesToEvent(event, player: Player, context: TContext, additionalProperties: Partial<IDrawProperties>): void {
         const { amount } = this.generatePropertiesFromContext(context, additionalProperties);
         super.addPropertiesToEvent(event, player, context, additionalProperties);
-        event.cards = event.player.drawDeck.slice(0, amount);
         event.amount = amount;
     }
 


server/game/core/card/Card.ts
```diff
@@ -1216,6 +1217,72 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
...
     }
 
+    /**
+     * Checks if this card is blocked from being played by an opponent's effect
+     * (e.g., Regional Governor naming this card, or Trade Route Taxation blocking events).
+     * This is used to display the lock icon on cards that could be played but are blocked.
+     * @returns A string describing why the card is blocked (with source card name), or null if not blocked
+     */
+    public getBlockedFromPlayReason(): string | null {
+        // Only check if the card is in a zone from which it could potentially be played:
```
AMMayberry1: Things could also be played from the deck zone with effects like Ezra or Reinforcement Walker, or capture with Dryden Vos. I think probably the easiest and safest check is simply if they're in an arena or not; for that we have `EnumHelpers.isArena()`.

server/game/core/card/Card.ts
```diff
@@ -1216,6 +1217,72 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
...
+        // Get the restriction values and their sources to find the blocking card
+        const restrictions = this.controller.getOngoingEffectValues<Restriction>(EffectName.AbilityRestrictions);
+        const sources = this.controller.getOngoingEffectSources(EffectName.AbilityRestrictions);
+
+        // Check if player has a restriction on playing this specific card type
+        if (playRestriction) {
+            for (let i = 0; i < restrictions.length; i++) {
+                if (restrictions[i].isMatch(playRestriction, context)) {
+                    const sourceName = sources[i]?.title || 'an effect';
+                    return `Blocked by ${sourceName}`;
```
AMMayberry1: What's the reason to return it as a string? Looks like we aren't using that on the FE currently, unless I'm mistaken

server/game/core/card/Card.ts
```diff
@@ -1216,6 +1217,72 @@ export class Card<T extends ICardState = ICardState> extends OngoingEffectSource
         return this.zoneName === ZoneName.Resource;
     }
 
+    /**
+     * Checks if this card is blocked from being played by an opponent's effect
+     * (e.g., Regional Governor naming this card, or Trade Route Taxation blocking events).
+     * This is used to display the lock icon on cards that could be played but are blocked.
+     * @returns A string describing why the card is blocked (with source card name), or null if not blocked
+     */
+    public getBlockedFromPlayReason(): string | null {
```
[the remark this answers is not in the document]
mnechromancer: Done!

server/game/actions/PlayUpgradeAction.ts
```diff
@@ -62,11 +62,32 @@ export class PlayUpgradeAction extends PlayCardAction {
...
+     * Check if playing an upgrade card is restricted for the given player and card.
+     * @param player The player attempting to play the upgrade
+     * @param card The upgrade card being played
+     * @param context Optional context for more detailed restriction checks
+     * @returns true if the play is restricted, false otherwise
+     */
+    public static isPlayRestricted(player: any, card: any, context?: AbilityContext): boolean {
+        // If no context provided, create a minimal one for restriction checks
+        // The mock ability with card property is needed for restrictedActionCondition checks (e.g., Regional Governor)
+        const checkContext = context ?? new AbilityContext({
```
AMMayberry1: I left a separate comment about passing down a `context` parameter so that we might not need this anymore, but if that ends up not being applicable then we have a built-in way to generate a "blank" context with `Game.getFrameworkContext()`

server/game/actions/PlayUpgradeAction.ts
```diff
@@ -62,11 +62,32 @@ export class PlayUpgradeAction extends PlayCardAction {
...
+            game: player.game,
+            player: player,
+            source: card,
+            ability: { card, isPlayCardAbility: () => false } as any
+        });
+
+        return (
+            player.hasRestriction(AbilityRestriction.Play, checkContext) ||
+            player.hasRestriction(AbilityRestriction.PlayUpgrade, checkContext) ||
+            player.hasRestriction(AbilityRestriction.PutIntoPlay, checkContext)
```
AMMayberry1: Checking the ongoing effects (of which restrictions are one) is an expensive operation for us, so it'll be better if we can pass in multiple AbilityRestriction values at once. That isn't supported yet but we have a standard pattern for doing that and I need to do it to support another PR as well, so I'll knock that out myself and ping you when this can be updated

AMMayberry1: PR is merged, should be able to pass these in as one array now: https://github.com/SWU-Karabast/forceteki/pull/2022

server/game/actions/PlayUpgradeAction.ts
```diff
@@ -66,28 +68,24 @@ export class PlayUpgradeAction extends PlayCardAction {
...
+        }
+        if (player.hasRestriction(AbilityRestriction.PutIntoPlay, context)) {
+            return AbilityRestriction.PutIntoPlay;
+        }
+        return null;
     }
 
     public override meetsRequirements(context = this.createContext(), ignoredRequirements: string[] = []): string {
-        if (PlayUpgradeAction.isPlayRestricted(context.player, context.source, context)) {
+        if (PlayUpgradeAction.getPlayRestriction(context.player, context.source, context) !== null) {
```
AMMayberry1: It's really important to always use `!= null` instead of `!== null`, since the former will catch `undefined` and the latter won't

server/game/actions/PlayUnitAction.ts
```diff
@@ -62,12 +64,31 @@ export class PlayUnitAction extends PlayCardAction {
...
 
+    /**
+     * Check if playing a unit card is restricted for the given player and card.
+     * @param player The player attempting to play the unit
+     * @param card The unit card being played
+     * @param context The context for restriction checks
+     * @returns The AbilityRestriction blocking play, or null if not restricted
+     */
+    public static getPlayRestriction(player: Player, card: Card, context: AbilityContext): PlayRestriction | null {
+        if (player.hasRestriction(AbilityRestriction.Play, context)) {
```
AMMayberry1: These should be folded into one like this:

```ts
if (player.hasRestriction([AbilityRestriction.Play, AbilityRestriction.PlayUnit, AbilityRestriction.PutIntoPlay], context)) {
    ...
}
```

Same note for the other files.

AMMayberry1: I think we crossed the streams a little bit with our restrictions, looks like `AbilityRestriction.Play` is being used on both players and cards. We'll straighten it out later but for now that just means that we need to check that one on both the player and the card. Same for the other files.

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -234,6 +237,52 @@ export class PlayableOrDeployableCard<T extends IPlayableOrDeployableCardState =
...
+        }
+
+        // Use the subclass implementation to check which restriction is blocking play
+        const playRestriction = this.getPlayRestriction(this.controller, context);
+        if (playRestriction === null) {
+            return null;
+        }
+
+        // Find the source of the blocking restriction
+        const restrictions = this.controller.getOngoingEffectValues<Restriction>(EffectName.AbilityRestrictions);
```
AMMayberry1: Sorry for not catching this on an earlier pass - instead of having to do a second search here to get the actual effect value, we can just update `getPlayRestriction()` to return the whole Restriction object instead of just the effect name.

To facilitate that, I'm adding a new helper method `getMatchingRestrictions()` in this PR which will make it so we can grab and return all of the matching effects. Then we won't need these `getOngoingEffectValues()` and `getOngoingEffectSources()` calls, we can just read the value right off of the objects: https://github.com/SWU-Karabast/forceteki/pull/2116

PR #1968, opened 2025-12-02
Introduce Matchmaking Rules

moyerr:
## Description

This PR introduces Matchmaking Rules to the `QueueHandler`, starting with a single `RematchCooldownRule`, which prevents players from being matched after a specified cooldown period if they faced each other in their previous match.

The interface makes the feature extensible, so we can easily add more matchmaking rules in the future. For example, if the players have a block-list, we could create a new Matchmaking Rule that prevents players from matching against somebody in their blocklist.

### DIFF
--- a/server/utils/user/UserFactory.ts
+++ b/server/utils/user/UserFactory.ts
@@ -86,6 +86,9 @@ export class UserFactory {
                 queryUser = query.user;
             }
             if (queryUser) {
+                if (queryUser.username && !queryUser.username.toLowerCase().startsWith('anonymous')) {
+                    logger.info(`Auth Anon-creation: creating user with id ${queryUser.id} with non-anonymous name ${queryUser.username}`, { userId: queryUser.id });
+                }
                 return new AnonymousUser(queryUser.id, queryUser.username);
             }
         }


server/gamenode/GameServer.ts
```diff
@@ -1904,6 +1905,20 @@ export class GameServer {
...
+     * Called near lobby end-of-life when users are disconnecting. Records matchmaking info
+     * (such as opponents and game end time) for future matchmaking before lobby tear-down.
+     */
+    public recordExpiringMatchmakingEntry(game: Game, lobby: Lobby): void {
+        Contract.assertNotNullLike(game.finishedAt, 'Finished game must have a finishedAt timestamp');
+
+        if (lobby.gameType === MatchType.Quick) {
+            // Update queue handler with finished game info
+            const [player1, player2] = game.getPlayers();
+            this.queue.setPreviousMatchEntry(player1.user.id, player2.user.id, game.finishedAt.getTime());
```
AMMayberry1: We should be careful that this doesn't bloat too much and eat up memory - some strategy for deleting stale entries would be good. For other similar caches we use `setInterval()` to establish a cleanup timer that will periodically scrape out expired entries, probably a good idea to do that here as well

server/gamenode/QueueHandler.ts
```diff
@@ -18,19 +20,28 @@ export enum QueuedPlayerState {
...
+export interface PreviousMatchEntry {
+    opponentUserId: string;
+    endTimestamp: number;
+}
+
 export class QueueHandler {
     private queues: Map<SwuGameFormat, QueuedPlayer[]>;
     private playersWaitingToConnect: QueuedPlayerEntry[] = [];
+    private playerPreviousMatch: Map<string, PreviousMatchEntry>;
+    private readonly cooldownSeconds = 60;
```
AMMayberry1: I'll raise this in chat for discussion, but 60s seems a bit long

server/gamenode/QueueHandler.ts
```diff
@@ -125,6 +137,29 @@ export class QueueHandler {
...
+            endTimestamp: endTimestamp
+        });
+
+        this.playerPreviousMatch.set(player2UserId, {
+            opponentUserId: player1UserId,
+            endTimestamp: endTimestamp
+        });
+
+        // After the cooldown period, remove the previous match entries
+        setTimeout(() => {
```
AMMayberry1: Individual timers do put a small amount of load on the server, so ideally we want to avoid any solution where the number of timers could scale up kind of high. So an interval timer that checks for all expired entries is probably better overall

server/gamenode/QueueHandler.ts
```diff
@@ -25,16 +27,29 @@ interface QueuedPlayerEntry {
...
     public constructor() {
         this.queues = new Map<SwuGameFormat, QueuedPlayer[]>();
+        this.playerPreviousMatch = new Map<string, PreviousMatchEntry>();
 
         Object.values(SwuGameFormat).forEach((format) => {
             this.queues.set(format, []);
         });
+
+        // Cleanup previous match entries periodically
+        setInterval(() => this.cleanupPreviousMatchEntries(), QueueHandler.COOLDOWN_INTERVAL_SECONDS * 1000);
```
AMMayberry1: Every 15 seconds is definitely more frequent than we need, it's not worth interrupting the server that much. We really just need to make sure we aren't keeping days of records at a time, so like once an hour is probably fine

PR #1975, opened 2025-12-04
han solo & zeb orellios

VJubert:


scripts/mockdata.js
```diff
@@ -1,6 +1,43 @@
...
+        setId: {
+            set: 'LAW',
+            number: 37,
+        },
+        unique: true,
+        arena: 'ground',
+        internalName: 'han-solo#hibernation-sick',
+    }),
+    buildMockCard({
+        title: 'Zeb Orellios',
```
AMMayberry1: Let's go ahead and just put in the correct spelling now (Orrellios), on the assumption it'll get fixed

AMMayberry1 (approved):

PR #1982, opened 2025-12-06
#dev-cardbacks

CheBato:
FE: https://github.com/SWU-Karabast/forceteki-client/pull/585

- Added Contributor role to the list of roles
- Added check if a user is a contributor he is sent with the cosmetics data.

### DIFF
--- a/test/server/cards/02_SHD/leaders/BobaFettDaimyo.spec.ts
+++ b/test/server/cards/02_SHD/leaders/BobaFettDaimyo.spec.ts
@@ -15,9 +15,6 @@ describe('Boba Fett, Daimyo', function () {
                         hand: ['outer-rim-headhunter'],
                         groundArena: ['wampa'],
                     },
-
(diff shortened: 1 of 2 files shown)


server/utils/ServerRoleUsersCache.ts
```diff
@@ -76,4 +76,22 @@ export class ServerRoleUsersCache {
...
+     * Checks if a user is a contributor.
+     * @param userId - The user ID to check
+     * @returns true if the user is a contributor, false otherwise
+     */
+    public isContributor(userId: string): boolean {
+        const data = this.cache.getValue();
+        if (!data) {
+            return false;
+        }
+        return (
```
AMMayberry1: We shouldn't be doing this check here, we should follow the pattern we use for all other cases and have this check only the `contributors` list and then if we want to give the same permission to other roles we can check them explicitly

AMMayberry1 (approved):

PR #1990, opened 2025-12-11
Sarlacc and Pit

bwohldmann:
image

image

### DIFF
--- a/test/server/cards/04_JTL/events/BarrelRoll.spec.ts
+++ b/test/server/cards/04_JTL/events/BarrelRoll.spec.ts
@@ -79,8 +79,45 @@ describe('Barrel Roll', function () {
                 expect(context.greenSquadronAwing.exhausted).toBe(true);
                 expect(context.player2).toBeActivePlayer();
             });
+        });
 
-            // TODO: Make sure Blue Leader when deployed as ground unit can't be selected by Barrel Roll for the attack/exhaust ability
+        it('Barrel Roll\'s ability cannot target a space unit which moves to ground arena', async function () {
+            await contextRef.setupTestAsync({
+                phase: 'action',
+                player1: {
+                    hand: ['barrel-roll', 'blue-leader#scarif-air-support'],
+                    spaceArena: ['awing']
+                },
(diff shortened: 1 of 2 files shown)


server/game/cards/07_LAW/units/TheSarlaccOfCarkoonHorrorOfTheDuneSea.ts
```diff
@@ -0,0 +1,40 @@
...
+        return {
+            id: 'the-sarlacc-of-carkoon#horror-of-the-dune-sea-id',
+            internalName: 'the-sarlacc-of-carkoon#horror-of-the-dune-sea',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addOnAttackAbility({
+            title: 'Put a unit from your discard pile on the bottom of your deck. Deal damage equal to that unit\'s power to an enemy ground unit',
+            targetResolvers: {
```
VJubert: I think there is 2 solutions for your problem here, maybe you could use an implicit then in the implementation or instead of using simultaneous on the second part, you should put the discarded card to the bottom of deck on the first part and deal damage on the second

server/game/cards/07_LAW/bases/GreatPitOfCarkoon.ts
```diff
@@ -0,0 +1,30 @@
...
+    protected override getImplementationId() {
+        return {
+            id: 'great-pit-of-carkoon-id',
+            internalName: 'great-pit-of-carkoon',
+        };
+    }
+
+    public override setupCardAbilities(registrar: IBaseAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.setEpicActionAbility({
+            title: '[Discard a unit from your hand]: Search your deck for a card named The Sarlacc of Carkoon, reveal it, and draw it',
```
AMMayberry1: nit: our convention is not to include the cost in these titles

PR #1994, opened 2025-12-13
jabba's rancor

VJubert:

### DIFF
--- a/test/server/cards/04_JTL/units/FleetInterdictor.spec.ts
+++ b/test/server/cards/04_JTL/units/FleetInterdictor.spec.ts
@@ -28,8 +28,43 @@ describe('Fleet Interdictor', function() {
 
             expect(context.republicArc170).toBeInZone('discard', context.player2);
             expect(context.player1).toBeActivePlayer();
+        });
 
-            // TODO: Add a test when a space unit is upgraded with a leader unit and is considered a leader unit should be able to defeat it
+        it('Fleet Interdictor\'s ability can target a space leader unit that cost 3 or less', async function () {
+            await contextRef.setupTestAsync({
+                phase: 'action',
+                player1: {
+                    hand: ['waylay'],
+                    spaceArena: ['fleet-interdictor'],
+                },
+                player2: {
+                    hand: ['vanquish'],
+                    spaceArena: ['awing', 'seventh-fleet-defender'],
+                    leader: 'asajj-ventress#i-work-alone',
+                    hasInitiative: true,
+                }
+            });
+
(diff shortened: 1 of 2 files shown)


server/game/cards/07_LAW/units/JabbasRancorSnackTime.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addOnAttackAbility({
+            title: 'An opponent chooses a ground unit they control. You may deal 7 damage to that unit',
+            targetResolver: {
+                controller: RelativePlayer.Opponent,
+                choosingPlayer: RelativePlayer.Opponent,
+                cardTypeFilter: WildcardCardType.Unit,
+                zoneFilter: ZoneName.GroundArena,
+            },
```
moyerr: Since the opponent is the one being prompted, we can add an `activePromptTitle:` to the target resolver that reads better from their perspective.

server/game/cards/07_LAW/units/JabbasRancorSnackTime.ts
```diff
@@ -0,0 +1,33 @@
...
+            then: (context) => ({
+                title: `Deal 7 damage to ${context.target.title}`,
+                optional: true,
+                immediateEffect: abilityHelper.immediateEffects.damage({
+                    target: context.target,
+                    amount: 7,
+                }),
+            })
```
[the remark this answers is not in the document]
VJubert: Yes because optional is after must be after the targetResolver (you can look grey squadron y-wing for reference)

server/game/cards/07_LAW/units/JabbasRancorSnackTime.ts
```diff
@@ -0,0 +1,34 @@
...
+            id: 'jabbas-rancor#snack-time-id',
+            internalName: 'jabbas-rancor#snack-time',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addOnAttackAbility({
+            title: 'An opponent chooses a ground unit they control. You may deal 7 damage to that unit',
+            targetResolver: {
+                activePromptTitle: 'Choose a ground unit. Your opponent may deal it 7 damage',
```
moyerr: Nit: for consistent phrasing, we should probably say "may deal 7 damage to it"

PR #2004, opened 2025-12-18
Doctor Aphra

bwohldmann:
image


test/server/cards/07_LAW/units/DoctorAphraDiggingForAnswers.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+                }
+            });
+
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.doctorAphraDiggingForAnswers);
+            context.player1.clickCard(context.p2Base);
+
+            expect(context.player2).toBeActivePlayer();
+        });
```
AMMayberry1: Let's add a test where the deck is empty just to make sure nothing breaks

Issue #1993, opened 2025-12-13
entireDeckSearch doesn't shuffle when no valid targets in deck.


bwohldmann:
When using entireDeckSearch when the target deck has no cards meeting cardCondition, the whole effect "fizzles" and no shuffle occurs.  See: https://github.com/SWU-Karabast/forceteki/blob/eda9bcc3920a6a4397ef304643911a0263ebf4ba/server/game/cards/07_LAW/bases/GreatPitOfCarkoon.ts#L19-L26

PR #2014, opened 2025-12-20
Shuffle Fix when using entireDeckSearch

bwohldmann:
`entireDeckSearch` will now shuffle deck when `shuffleWhenDone` is true when there are no valid targets to search for.

Addresses #1993


test/server/cards/02_SHD/events/BountyPosting.spec.ts
```diff
@@ -93,8 +93,8 @@ describe('Bounty Posting', function() {
                 // expect(context.player1).toHaveExactDisplayPromptCards({
                 //     invalid: [context.tielnFighter, context.cellBlockGuard, context.pykeSentinel, context.hylobonEnforcer]
                 // });
-                // expect(context.player1).toHaveEnabledPromptButton('Take nothing');
-                // context.player1.clickPrompt('Take nothing');
+                expect(context.player1).toHaveEnabledPromptButton('Take nothing');
+                context.player1.clickPrompt('Take nothing');
```
AMMayberry1: I think we do eventually want a solution that will avoid giving away information in this situation, but it would probably be a different solution than this. We've talked about adding a dummy UI prompt that will automatically time itself out after a second or two to create delay without requiring user input, specifically in the context of Disclose since that one comes up a lot.

All that to say, I'd rather build a dedicated solution that will work for all of these scenarios instead of hacking it in here, especially since the UI doesn't make it super clear what's going on as currently set up.

In order for this to auto-pass when there's no target, the System needs to tell the engine that there will be no effect so that the engine doesn't attempt to trigger it. That's accomplished most easily here by making it so that `SearchEntireDeckSystem.hasLegalTarget()` returns false. That could be done either as a change to the parent method or an overload, not sure which would be best offhand.

Feel free to follow up in chat if you have any questions or want to suggest a different plan

bwohldmann: So if we modify the existing `hasLegalTarget` method to return false if the prompt would be empty (`SearchEntireDeckSearch`) or populated but with no valid targets (`SearchDeckSystem`), we get this flow:

image

If we select 'Use it anyway', the shuffles don't trigger because those trigger with the `onSearchComplete` handler that isn't going to get fired.

The bigger issue though is we can tip to the player if their search is going to whiff with certain implementations, and then they could just back out of it.  Example:  https://github.com/SWU-Karabast/forceteki/blob/22870dbec3ebb14dd76f97becd92ef5ab7f5f6fd/test/server/cards/01_SOR/units/GrandMoffTarkinDeathStarOverseer.spec.ts#L101-L116

PR #2017, opened 2025-12-22
Adds Salacious Crumb, Cackling Companion

scapegoat-code:
small_SWH_07en_0210_Salacious_Crumb_First_Look_7d4cfc0dc9

Adds the ts file for Salacious Crumb, Cackling Companion, along with the spec.ts and mock card data.
At the time, the testing does contain "Jabba the Hutt, Crime Boss" which hasn't been implemented, so those tests will show up as failed.

### DIFF
--- a/server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
+++ b/server/game/cards/01_SOR/units/GuardianOfTheWhills.ts
@@ -37,7 +37,7 @@ export default class GuardianOfTheWhills extends NonLeaderUnitCard {
             ongoingEffect: AbilityHelper.ongoingEffects.decreaseCost({
                 amount: 1,
                 match: (card, adjusterSource) => card.isUpgrade() && this.isFirstUpgradePlayedOnThisCopy(card, adjusterSource),
-                attachTargetCondition: (attachTarget, adjusterSource) => attachTarget === adjusterSource,
+                attachTargetCondition: (attachTarget, _, adjusterSource) => attachTarget === adjusterSource,
                 limit: AbilityHelper.limit.perRound(1),
             }),
         });
--- a/server/game/cards/03_TWI/units/GeneralGrievousTrophyCollector.ts
+++ b/server/game/cards/03_TWI/units/GeneralGrievousTrophyCollector.ts
@@ -16,7 +16,7 @@ export default class GeneralGrievousTrophyCollector extends NonLeaderUnitCard {
             title: 'Ignore the aspect penalty on each Lightsaber upgrade you play on this unit',
             ongoingEffect: AbilityHelper.ongoingEffects.ignoreAllAspectPenalties({
                 cardTypeFilter: WildcardCardType.Playable,
-                attachTargetCondition: (target, source) => target === source,
+                attachTargetCondition: (target, _, source) => target === source,
                 match: (card) => card.hasSomeTrait(Trait.Lightsaber)
             })
         });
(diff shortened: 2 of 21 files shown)


scripts/mockdata.js
```diff
@@ -205,6 +205,25 @@ const mockCards = [
...
         internalName: 'jabbas-rancor#snack-time',
     }),
+    buildMockCard({
+        title: 'Salacious Crumb',
+        subtitle: 'Cackling Companion',
+        power: 0,
+        hp: 2,
+        cost: 1,
+        hasNonKeywordAbility: true,
+        keywords: ['raid 2'],
```
bwohldmann: Don't believe you need to declare keywords like raid, sentinel, restore, etc.

test/server/cards/07_LAW/units/SalaciousCrumbCacklingCompanion.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+
+            it('should be ready when controlling LAW Jabba the Hutt leader', function() {
+                const { context } = contextRef;
+                
+                context.player1.clickCard(context.salaciousCrumbCacklingCompanion);
+                expect(context.salaciousCrumbCacklingCompanion.exhausted).toBeFalse();
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should not be ready when controlling a LAW Jabba the Hutt leader unit', function() {
```
bwohldmann: think you want to remove the 'not' from the it assertion here

bwohldmann:
For your unit test errors, I THINK you need to commit a new version of card-data-version.txt that increments the version.  Think you're ok to just set it to the date of your commit and not worry about the minor version number (will presumably get sorted on PR merge if there's a conflict); but that's somewhat of a guess as I'm rather new myself.

For the linting errors, recommend installing the ESLint extension in VISCode to see those called out pre-commit.

ev-bwohldmann:
> The two failures are coming from the test on "Jabba the Hutt, Crime Boss" which has yet to be implemented. Wasn't sure if I should remove it.

Gotcha.  I think when I looked at a previous run, there was a similar error shown on the Crumb card.  Guessing a follow-up commit addressed that.

test/server/cards/07_LAW/units/SalaciousCrumbCacklingCompanion.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+            expect(context.salaciousCrumbCacklingCompanion.exhausted).toBeFalse();
+            expect(context.player2).toBeActivePlayer();
+        });
+
+        it('should NOT be ready when not controlling and Jabba leaders or units ', async function() {
+            await contextRef.setupTestAsync({
+                phase: 'action',
+                player1: {
+                    hand: ['salacious-crumb#cackling-companion'],
+                    leader: 'the-mandalorian#sworn-to-the-creed'
```
AMMayberry1: nit: let's add a Jabba of some sort under the opponent's control just to confirm that the filter works

test/server/cards/07_LAW/units/SalaciousCrumbCacklingCompanion.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+                return contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        hand: ['salacious-crumb#cackling-companion'],
+                        leader: 'jabba-the-hutt#crime-boss',
+                    },
+                });
+            });
+
+            it('should be ready when controlling LAW Jabba the Hutt leader', function() {
```
AMMayberry1: We can just remove these LAW ones to unblock the tests, we don't usually bother testing more than one variant of the leader anyway

PR #2018, opened 2025-12-23
Qui-Gon Jinn + SearchDeckSystem Enhancement

bwohldmann:
image

Also modified `DeckSearchSystem` to support `remainingCardsImmediateEffect` similar to the existing `selectedCardsImmediateEffect`

### DIFF
--- a/server/game/cards/04_JTL/units/TheInvisibleHandCrawlingWithVultures.ts
+++ b/server/game/cards/04_JTL/units/TheInvisibleHandCrawlingWithVultures.ts
@@ -3,7 +3,6 @@ import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityR
 import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
 import { Trait, WildcardCardType } from '../../../core/Constants';
 import { CostAdjustType } from '../../../core/cost/CostAdjuster';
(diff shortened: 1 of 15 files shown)


scripts/mockdata.js
```diff
@@ -205,6 +205,25 @@ const mockCards = [
         arena: 'ground',
         internalName: 'jabbas-rancor#snack-time',
     }),
+    buildMockCard({
+        title: 'Qui-Gon Jinn',
+        subtitle: 'Influence Chance',
```
AMMayberry1: nit: should be "Influencing"

test/server/cards/07_LAW/units/QuiGonJinnInfluenceChance.spec.ts
```diff
@@ -0,0 +1,181 @@
+
+describe('Qui-Gon Jinn Influence Chance', function () {
+    integration(function (contextRef) {
+        it('on attack, should look at the top 3 cards of your deck, discard 1, put the rest on top in any order', async function () {
```
AMMayberry1: nit: when you're reusing the same test setup multiple times, you can use a `describe()` block with a `beforeEach()` to declare a test setup that will be used for all of the `it()` cases inside the block. If you just do a quick text search for `beforeEach` you can find some examples for reference

PR #2019, opened 2025-12-26
chopper spectre three, boba fett for a price

sethhelman:
boba fett
chopper

### DIFF
--- a/server/game/core/gameSteps/phases/RegroupPhase.ts
+++ b/server/game/core/gameSteps/phases/RegroupPhase.ts
@@ -42,14 +42,19 @@ export class RegroupPhase extends Phase {
             this.takeActionSnapshotsForPromptedPlayers();
         }
 
-        for (const player of this.game.getPlayers()) {
-            // create a single event for drawing cards step
-            new DrawSystem({ amount: 2 }).resolve(
-                player,
-                this.game.getFrameworkContext(player),
-                TriggerHandlingMode.ResolvesTriggers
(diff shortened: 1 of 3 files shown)


server/game/cards/07_LAW/units/ChopperSpectreThree.ts
```diff
@@ -0,0 +1,24 @@
...
+            id: 'chopper#spectre-three-id',
+            internalName: 'chopper#spectre-three',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addWhenPlayedAbility({
+            title: 'Give an experience token to this unit. If you control a Cunning or Vigilance unit, give two experience tokens to him instead',
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) => context.player.isAspectInPlay(Aspect.Cunning, context.source) || context.player.isAspectInPlay(Aspect.Vigilance, context.source),
```
AMMayberry1: nit: a small optimization we could do would be to combine the aspect filters like `context.player.isAspectInPlay([Aspect.Cunning, Aspect.Vigilance], context.source)`. That _should_ work but if not just ping me and I'll do a small fix. It's worth it to avoid having to scan through the entire list of units twice.

AMMayberry1 (approved):

PR #2026, opened 2026-01-02
unit test optimizations

brianries:
Multiple major things changed:
 - moveCard to moveCardQuiet to stop the game.continue() calls during setup
 - batch add cards function for SimpleZone which is used by moveAllNonBaseZonesToRemoved
 - memoized isDevelopment (funny enough it was 7% of the time in unit tests because of getRef gamestate manager calls)
 - UnitTestCardDataGetter,getCardInternalSyn/Async added caching for the file reading
 - EnumHelpers.checkConvertToEnum now caches a map of the enum strings (the keys are all constants)

Overall on my home PC the unit tests are running 2.7x faster

```
main:
4856 specs, 0 failures, 2 pending specs
Finished in 259.059 seconds

new branch:
4855 specs, 0 failures, 2 pending specs
Finished in 95.794 seconds
```


server/game/core/utils/EnumHelpers.ts
```diff
@@ -3,15 +3,31 @@ import { CardType, ZoneName, DeckZoneDestination, RelativePlayer, WildcardCardTy
...
+export function checkConvertToEnum<T extends object>(values: string | string[], enumObj: T): T[keyof T][] {
     const result: T[keyof T][] = [];
+    const lookupMap = getEnumLookupMap(enumObj);
 
-    const enumValues = Object.values(enumObj);
     for (const value of Helpers.asArray(values)) {
-        const matchingValue = enumValues.find((enumValue) => enumValue.toLowerCase() === value.toLowerCase());
-        if (matchingValue) {
-            result.push(matchingValue as T[keyof T]);
+        const matchingValue = lookupMap.get(value.toLowerCase());
```
AMMayberry1: The way this is set up now, will it return lowercase versions of the enum string values? It's important that the case be returned as-is. If we're just using lowercase for string matching purposes that's fine

server/game/core/zone/SimpleZone.ts
```diff
@@ -51,13 +51,37 @@ export abstract class SimpleZone<TCard extends Card = Card> extends ZoneAbstract
...
+    /**
+     * Adds multiple cards to the zone in a single operation.
+     * More efficient than calling addCard multiple times as it only triggers state tracking once.
+     */
+    public addCards(cards: TCard[]) {
+        if (cards.length === 0) {
+            return;
+        }
+        for (const card of cards) {
+            Contract.assertFalse(this.cards.includes(card), `Attempting to add card ${card.internalName} to ${this} twice`);
```
AMMayberry1: Probably more efficient to make a Set here for faster search

test/helpers/PlayerInteractionWrapper.js
```diff
@@ -844,6 +865,23 @@ class PlayerInteractionWrapper {
...
 
+    /**
+     * Moves cards between zones WITHOUT calling game.continue().
+     * Use this for batch operations during test setup to avoid pipeline overhead.
+     * Call Util.refreshGameState() once after all moves are complete.
+     * @param {String|DrawCard} card - card to be moved
+     * @param {String} targetZone - zone where the card should be moved
+     * @param {String | String[]} searchZones - zones where to find the card object
+     */
+    moveCardQuiet(card, targetZone, searchZones = 'any') {
```
AMMayberry1: nit: I don't think "quiet" clearly expresses what the difference is or when this should be used. Since the intent is that it's only called during setup, maybe a name that makes that clear?

server/game/core/utils/EnumHelpers.ts
brianries: The lookupMap for enum Ability would store this:

Map {
  'action' => 'action',
  'constant' => 'constant',
  'damagemodification' => 'damageModification',
  'delayedeffect' => 'delayedEffect',
  'event' => 'event',
  'replacementeffect' => 'replacementEffect',
  'triggered' => 'triggered'
}

So we're storing a lowercase key, but keeping the original camel cased string as the value, so we should be returning a non modified enum string

AMMayberry1 (approved):

PR #2028, opened 2026-01-02
# Added Player-report-feature

CheBato:
FE PR: https://github.com/SWU-Karabast/forceteki-client/pull/607

- Added player report feature for in-game and lobbies
- Three categories
- automatic disable chat when reporting.
- Chat messages are contained.
- Fixed issue where bug reports didn't always have Player1 and Player2

For it to work you also need to set the DISCORD_PLAYER_REPORT_WEBHOOK_URL to a webhook url

### DIFF
--- a/server/gamenode/Lobby.ts
+++ b/server/gamenode/Lobby.ts
@@ -1189,6 +1189,13 @@ export class Lobby {
         );
 
         this.game = game;
+
+        for (const player of this.game.getPlayers()) {
+            const userWrapper = this.getUser(player.user.id);
+            userWrapper.deck = player.lobbyDeck;
+
+            logger.info(`Test deck synchronized for user: ${userWrapper.username}`);
+        }
     }
 
     private async startGameAsync() {


server/game/core/DiscordDispatcher.ts
```diff
@@ -71,128 +75,154 @@ export class DiscordDispatcher implements IDiscordDispatcher {
...
             }
+            if (!this._playerReportWebhookUrl) {
+                throw new Error('No Discord webhook URL configured for player reports. Player reports cannot be sent to Discord.');
+            }
         }
     }
 
-    public async formatAndSendBugReportAsync(bugReport: ISerializedReportState): Promise<EitherPostResponseOrBoolean> {
-        // Always log the bug report
+    public async formatAndSendReportAsync(report: ISerializedReportState, isBugReport: boolean): Promise<EitherPostResponseOrBoolean> {
```
AMMayberry1: A name like `isBugReport` is not super informative in this case because it doesn't clearly say what the alternative is. so if someone was reading this, they would have to dig around in the code to find the context to answer that question.

In this case it's almost always better to use an enum that clearly names each option. That also decreases the overhead if we ever add additional cases.

server/game/core/DiscordDispatcher.ts
```diff
@@ -98,12 +98,28 @@ export class DiscordDispatcher implements IDiscordDispatcher {
...
+            case ReportType.BugReport:
+                reportTypeLabel = 'Bug report';
+                webhookLink = this._bugReportWebhookUrl;
+                break;
+            case ReportType.PlayerReport:
+                reportTypeLabel = 'Player report';
+                webhookLink = this._playerReportWebhookUrl;
+                break;
+            default:
+                throw new Error('Unsupported reportType');
```
AMMayberry1: nit: for debugging purposes it's always good to include as much information as possible so we can reconstruct what happened from the logs. TS gets a little unhappy here about typing so it usually needs to look something like `Unsupported reportType: ${(reportType as any)}`

PR #2029, opened 2026-01-03
Fix/#799 null leader fix

HelloMyNameIsSeth:
Test games now properly display in lobby preview. This prevents the leader null error that was occurring when lobby is viewed. 

image

### DIFF
--- a/server/utils/profanityFilter/termLists/BasicProfanityList.ts
+++ b/server/utils/profanityFilter/termLists/BasicProfanityList.ts
@@ -16,6 +16,7 @@ type EnglishProfaneWord =
   | 'bestiality'
   | 'bitch'
   | 'blowjob'
+  | 'blowme'
   | 'bollocks'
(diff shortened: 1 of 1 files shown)


server/gamenode/Lobby.ts
```diff
@@ -1027,6 +1028,16 @@ export class Lobby {
             const player1 = this.users[0];
             const player2 = this.users[1];
 
+            if (player1.id === 'exe66' || player2.id === 'th3w4y') {
```
AMMayberry1: Could we refactor so that the data will be correctly populated for test users and we don't have to do this workaround? Otherwise we'll have to put in effort to maintain this if we make changes to how player data works

HelloMyNameIsSeth: Sorry, I entirely misinterpreted what you were asking to be updated. My last update should be more in line with what you were asking for.

Issue #1932, opened 2025-11-13
DiscardCardsFromHandSystem and other PlayerTarget systems don't correctly report legal target


AMMayberry1:
There is a bug with Syril Karn we had to work around where the "unless they discard" part didn't work correctly. If the opponent had no cards in hand, the damage should be forced, but it was allowing the opponent to still choose to discard.

The reason for this is that PlayerTargetSystems by default consider themselves to always have a legal target since there is always a player that exists to be targeted. We need to make some adjustments to the philosophy behind this reasoning for player target systems in particular, so that cases like Syril will behave in an intuitive way.

Issue #1931, opened 2025-11-13
Explicit support for building "unless" abilities


AMMayberry1:
Abilities like Syril Karn that use "unless" can be tricky to set up, since the "unless" ability (e.g. discard for Syril) needs to be evaluated differently. I.e., if the unless part _can't_ happen, then we default to first part of the ability. Whether that first part (Syril damage) can happen or not is irrelevant.

We need to add a special target mode like "SelectUnless" that handles the evaluation for this correctly, and update the existing "Unless" cards to use it.

Note that Syril in particular also has a separate bug that needs to be fixed before this will work for him: https://github.com/SWU-Karabast/forceteki/issues/1932

bwohldmann:
Resolved in PR #2030

PR #2030, opened 2026-01-04
Implement Conditional 'Unless' Targeting

bwohldmann:
- Implemented functionality to better support 'unless' conditions on cards such as 'Look The Other Way' and 'Syril Karn' so we don't have to implement checks on if the unless condition can be met or not.  Addresses Issues #1931 and mitigates #1932 but still want to look at the PlayerTargetSystems to potentially shore those up.

Introduces the following interfaces to the SelectTargetResolver code:

```ts
export interface ISelectUnlessTargetResolver<TContext extends AbilityContext> extends ISelectTargetResolverBase<TContext> {
    mode: TargetMode.SelectUnless;
    /** The effect the choosing player can pay to avoid the default effect (SelectUnless mode only) */
    unlessEffect: ISelectUnlessEffect<TContext>;
    /** The effect to resolve if the choosing player declines or cannot pay the unless effect (SelectUnless mode only) */
    defaultEffect: ISelectUnlessEffect<TContext>;
}

/** Represents an effect paired with its prompt button for SelectUnless mode */
export interface ISelectUnlessEffect<TContext extends AbilityContext> {
    effect: GameSystem<TContext> | ((context: TContext) => GameSystem<TContext>);
    promptButtonText: NamedAction;
}
```

An Example Implementation:

```ts
    public override setupCardAbilities(registrar: IEventAbilityRegistrar, AbilityHelper: IAbilityHelper) {
        registrar.setEventAbility({
            title: 'Exhaust a unit unless its controller pays 2 resources.',
            targetResolvers: {
                targetUnit: {
                    cardTypeFilter: WildcardCardType.Unit
                },
                opponentsChoice: {
                    mode: TargetMode.SelectUnless,
                    dependsOn: 'targetUnit',
                    unlessEffect: {
                        effect: (context) => AbilityHelper.immediateEffects.payResourceCost({
                            target: context.targets.targetUnit.controller,
                            amount: 2
                        }),
                        promptButtonText: NamedAction.Pay
                    },
                    defaultEffect: {
                        effect: (context) => AbilityHelper.immediateEffects.exhaust({
                            target: context.targets.targetUnit,
                        }),
                        promptButtonText: NamedAction.Exhaust
                    },
                    choosingPlayer: (context) => EnumHelpers.asRelativePlayer(context.player, context.targets.targetUnit.controller),
                    activePromptTitle: (context) => `[Exhaust] ${context.targets.targetUnit.title} or [Pay] 2 resources`,
                    highlightCards: (context) => context.targets.targetUnit,
                }
            }
        });
    }
```
- Refactored existing cards to use this new pattern.
- Also addressed an issue with 'Hired Slicer' where the 'reveal from deck' mechanic was incorrectly marked as optional.

### DIFF
--- a/server/game/core/ability/abilityTargets/SelectTargetResolver.ts
+++ b/server/game/core/ability/abilityTargets/SelectTargetResolver.ts
@@ -9,6 +9,7 @@ import { GameStateChangeRequired, Stage, TargetMode } from '../../Constants';
 import type { Player } from '../../Player';
 import type { IPassAbilityHandler } from '../../gameSteps/AbilityResolver';
 import * as Helpers from '../../utils/Helpers';
+import * as Contract from '../../utils/Contract';
 
 /** Target resolver for selecting between multiple prompted choices due to an effect */
 export class SelectTargetResolver extends TargetResolver<ISelectTargetResolver<AbilityContext> | ISelectUnlessTargetResolver<AbilityContext>> {
@@ -29,6 +30,8 @@ export class SelectTargetResolver extends TargetResolver<ISelectTargetResolver<A
                     : this.properties.choices;
             case TargetMode.SelectUnless:
                 return this.buildSelectUnlessChoices(context);
+            default:
+                Contract.fail(`Unknown TargetMode value '${(this.properties as any).mode}'`);
         }
     }
 


server/game/core/ability/abilityTargets/SelectTargetResolver.ts
```diff
@@ -60,6 +100,20 @@ export class SelectTargetResolver extends TargetResolver<ISelectTargetResolver<A
...
     protected override resolveInternal(player: Player, context: AbilityContext, targetResults: ITargetResult, passPrompt?: IPassAbilityHandler) {
+        // For SelectUnless mode, check if unlessEffect can be resolved
+        // If not, automatically resolve defaultEffect without prompting
+        if (this.properties.mode === TargetMode.SelectUnless) {
+            const unlessEffect = this.getUnlessEffect(context);
+            if (unlessEffect && !unlessEffect.hasLegalTarget(context)) {
+                // unlessEffect cannot be resolved, automatically apply defaultEffect
+                const defaultEffect = this.getDefaultEffect(context);
+                if (defaultEffect) {
+                    context.selects[this.name] = new SelectChoice('__default__');
```
AMMayberry1: Do we even need this special-case handling - could we just set `choices` to only be the default effect and then let everything else proceed normally from there? Then we wouldn't need a case in `getGameSystems()` to check for it

AMMayberry1 (approved):

PR #2037, opened 2026-01-08
Update card data with reprint set codes & use them in deck validation logic

moyerr:
## Description

This PR updates the card data to include a `setCodes` property for all non-token cards. This property is an array of the card's non-promo set codes. In other words, it includes all core-set printings of a card, but excludes things like OP promos, Convention-exclusive promos, and Gamegenic promos. The `setId` property is still in tact, and it now represents the most recent core-set printing of a card (this was _mostly_ true already).

Deck validation logic is also updated in this PR to determine format legality of a card based on whether or not that card has a printing in a legal rotation block for the format. 

#### Individual Card Data

- New `setCodes` property added
- `setId` now represents the latest printing 
- Most cards' `setId` property is unchanged, with the only exceptions being the LOF and SEC reprints

#### Set Code Map

- No longer includes set codes for OP promo sets (`JTLP`, `LOFP`, `SECP`) or Gamegenic promo bases (`GG`)
- Now correctly maps every `IBH` set code to a single card id (that set has different set codes for each copy of the card included in the Intro: Battle Hoth product)

#### No further changes

As expected, there are no changes to the following card data files:

- `_allNonLeaderCardTitles.json`
- `_cardMap.json`
- `_leaderNames.json`
- `_playableCardTitles.json`
- `_mockCardNames.json`


server/utils/deck/DeckValidator.ts
```diff
@@ -19,16 +19,32 @@ enum SwuSet {
     LAW = 'law'
 }
 
-const legalSets = [SwuSet.SOR, SwuSet.SHD, SwuSet.TWI, SwuSet.JTL, SwuSet.LOF, SwuSet.IBH, SwuSet.SEC];
+const rotationBlocks = new Map<string, SwuSet[]>([
```
AMMayberry1: Would it be possible to use sets instead of arrays for efficient checking, since this gets looped over a lot?

AMMayberry1 (dismissed):

PR #2038, opened 2026-01-09
Dengar & Han's Golden Dice

moyerr:
| ⟡ Dengar, Take Your Shot | ⟡ Han's Golden Dice |
| :---: | :---: |
| image | image |

### DIFF
--- a/server/game/core/Game.js
+++ b/server/game/core/Game.js
@@ -303,12 +303,12 @@ class Game extends EventEmitter {
             winnerNames: [],
             lastGameEventId: 0,
             currentPhase: null,
-            prevActionPhasePlayerPassed: null
+            prevActionPhasePlayerPassed: null,
+            movedCards: []
         };
 
         this.tokenFactories = null;
         this.stateWatcherRegistrar = new StateWatcherRegistrar(this);
-        this.movedCards = [];
         this.cardDataGetter = details.cardDataGetter;
         this.playableCardTitles = this.cardDataGetter.playableCardTitles;
         this.allNonLeaderCardTitles = this.cardDataGetter.allNonLeaderCardTitles;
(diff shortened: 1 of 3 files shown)


test/server/cards/07_LAW/units/DengarTakeYourShot.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+                // All units should be defeated
+                expect(context.dengar).toBeInZone('discard');
+                expect(context.hiddenSharpshooter).toBeInZone('discard');
+                expect(context.battlefieldMarine).toBeInZone('discard');
+
+                // Dengar should get a Credit token since Hidden Sharpshooter was the highest cost enemy unit
+                expect(context.player1.credits).toBe(1);
+            });
+
+            it('only resolves once if multiple units with the same highest cost are defeated simultaneously', async function () {
```
AMMayberry1: Just to be thorough, it would be good to add a test where there are multiple defeats and none of them are the highest cost

test/server/cards/07_LAW/units/DengarTakeYourShot.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+                ]);
+
+                // Arbitrarily pick the first one to resolve
+                context.player1.clickPrompt('Create a Credit token: Battlefield Marine');
+                expect(context.player1.credits).toBe(1);
+
+                // No further action to resolve for the second trigger since the limit is once per round
+                expect(context.player2).toBeActivePlayer();
+            });
+        });
```
AMMayberry1: Late addition - would also be good to set up a situation where the highest-cost thing defeated is actually an upgrade attached to one of the defeated units, just to make sure that is correctly filtered out from consideration

PR #2042, opened 2026-01-13
Allow Credit tokens to be used for game effect resource payments

moyerr:
## Description

This PR makes a handful of updates that allow credit tokens to be used when making resource payments for game effects, such as Blue Leader's **When Played** ability. The high level changes are:

- Introduce a new `GameEffectResourcePaymentSystem` to handle these specific effects. This system taps into the cost logic by leveraging a new `ResourceCost` subclass called `GameEffectResourcePayment`. I specifically avoided the "cost" naming for this subclass because these effects are not "costs" from a rules perspective, but much of the same logic does apply.
- Add a `matchGameEffectResourcePayments` property to the cost adjuster properties, which allows specific cost adjusters to opt-in to making cost adjustments for these game effects
- Update both ResourceCost and CostAdjuster to avoid assuming `context.player` is the player for whom the cost is being paid or adjusted, respectively. Because these effects can target an opponent, we need to take a different strategy:
    - For ResourceCost, we now accept a function to provide the player responsible for paying the costs.
    - For CostAdjuster, we now use `this.sourcePlayer` to get the player for whom the cost is being adjusted. This makes sense because each adjuster is attached to a single player and should only be able to adjust costs (or payments) for that player

### DIFF
--- a/test/scenarios/cost/TargetedCostAdjusterCombinations.spec.ts
+++ b/test/scenarios/cost/TargetedCostAdjusterCombinations.spec.ts
@@ -60,7 +60,7 @@ describe('Cost adjuster combinations', function() {
                 context.player1.clickCardInDisplayCardPrompt(context.captainTypho);
 
                 // Prompt for credits should appear, as Captain Typho costs 1 after the discount
-                expect(context.player1).toHavePrompt('Use Credit tokens for Captain Typho');
+                expect(context.player1).toHavePrompt('Use Credit tokens to pay for Captain Typho');
                 expect(context.player1).toHaveExactPromptButtons(['Use 1 Credit']);
                 context.player1.clickPrompt('Use 1 Credit');
 
@@ -213,7 +213,7 @@ describe('Cost adjuster combinations', function() {
                 context.player1.clickPrompt('Done');
 
                 // Prompt for credits should appear
-                expect(context.player1).toHavePrompt('Use Credit tokens for Asajj Ventress');
+                expect(context.player1).toHavePrompt('Use Credit tokens to pay for Asajj Ventress');
                 expect(context.player1).toHaveExactPromptButtons(['Select amount', 'Pay costs without Credit tokens']);
                 context.player1.clickPrompt('Select amount');
 


server/game/cards/01_SOR/events/ForceLightning.ts
```diff
@@ -30,9 +30,10 @@ export default class ForceLightning extends EventCard {
                     mode: TargetMode.DropdownList,
                     options: (context) => Array.from({ length: context.player.readyResourceCount + 1 }, (_x, i) => `${i}`),
                     immediateEffect: AbilityHelper.immediateEffects.simultaneous([
-                        AbilityHelper.immediateEffects.payResourceCost((context) => ({
+                        AbilityHelper.immediateEffects.exhaustResources((context) => ({
                             amount: parseInt(context.select),
                             target: context.player,
+                            isCost: true
```
AMMayberry1: Having to do it this way feels pretty unintuitive to me, I feel like someone working to write a similar effect would struggle to understand if they need to do this vs just using `payResources`.

Could we potentially keep using `payResources` but then add a parameter like `doNotAllowCredits` or something, with a comment explaining the reasoning?

server/game/core/cost/CostAdjuster.ts
```diff
@@ -52,6 +52,9 @@ export interface ICostAdjusterPropertiesBase {
     /** Whether the cost adjuster should adjust activation costs for abilities. Defaults to false. */
     matchAbilityCosts?: boolean;
 
+    /** Whether the cost adjuster should adjust game effect resource payments (e.g. Blue Leader). Defaults to false. */
+    matchGameEffectResourcePayments?: boolean;
```
AMMayberry1: nit: the word "game" in here makes this feel confusing to me, at first I assume this meant standard game mechanics (as opposed to card effects)

server/game/costs/ResourceCost.ts
```diff
@@ -5,26 +5,32 @@ import { GameEvent } from '../core/event/GameEvent';
...
 
-    public abstract readonly isPlayCost: boolean;
-
-    public constructor(resourceCostAmount: number) {
+    public constructor(
+        resourceCostAmount: number,
+        payer: (context: AbilityContext) => Player = (context) => context.player
+    ) {
         this.resourceCostAmount = resourceCostAmount;
+        this.payer = payer;
```
AMMayberry1: nit: `payer` feels a bit too close to `player` and might be misread, maybe `payingPlayer`?

AMMayberry1 (approved):

AMMayberry1 (approved):

PR #2044, opened 2026-01-13
Allow Galen Erso to blank the Credit token

moyerr:


test/server/cards/06_SEC/units/GalenErsoYoullNeverWin.spec.ts
```diff
@@ -1219,144 +1219,272 @@ describe('Galen Erso - You\'ll Never Win', function() {
...
-                    phase: 'action',
-                    player1: {
-                        hand: ['galen-erso#youll-never-win'],
-                        groundArena: ['wampa']
-                    },
-                    player2: {
-                        groundArena: [{ card: 'atst', upgrades: ['shield', 'shield'] }]
-                    }
+        describe('Galen Erso - You\'ll Never Win\'s ability should name a card. While he is in play, named tokens', function() {
+            describe('Shield', function() {
```
moyerr: Oof this diff is miserable. Nothing actually changed with the Shield tests, I just tucked them all inside this new `describe` to separate them from the Credit token tests

server/game/core/cost/DefeatCreditTokensCostAdjuster.ts
```diff
@@ -38,6 +38,17 @@ export class DefeatCreditTokensCostAdjuster extends CostAdjusterWithGameSteps {
             return false;
         }
 
+        // TODO: If there is ever an effect that can selectively blank Credit tokens,
+        // this class will need to account for which Credits can actually be used to
+        // adjust costs. For now, it's all or nothing.
+        const creditsAreBlanked = this.sourcePlayer.baseZone.credits.every((credit) =>
```
AMMayberry1: Since we're already assuming that it's all or nothing anyway, could we just check a single credit token and use that to make the decision? This will probably be a very slow operation due to the ongoing effect checks involved, so if we can cut down on redundant checks that'd be very good

PR #2045, opened 2026-01-13
Putting a Team Together & C3PO

VJubert:


test/server/cards/07_LAW/units/C3POTranslationProtocol.spec.ts
```diff
@@ -0,0 +1,30 @@
+describe('C-3PO, Translation Protocol', function () {
+    integration(function (contextRef) {
+        it('C-3PO\'s ability should give an Experience token to another non-leader unit that shares a Trait with a friendly leader', async function () {
```
moyerr: Some other test cases to consider:
- A unit whose only shared trait with the leader is a trait gained from an effect
- A unit whose only shared trait with the leader is removed by an effect
- A leader with a gained trait (Mythosaur)

moyerr (approved):

PR #2051, opened 2026-01-15
fix phantom choose nothing

VJubert:


server/game/cards/07_LAW/units/PhantomSpectreShuttle.ts
```diff
@@ -18,13 +18,15 @@ export default class PhantomSpectreShuttle extends NonLeaderUnitCard {
...
-            }
+                immediateEffect: abilityHelper.immediateEffects.playCardFromHand({
+                    playAsType: WildcardCardType.Unit,
+                }),
+            },
+            then: (thenContext) => ({
+                title: 'Give an Experience token to the played unit',
+                thenCondition: () => thenContext.target != null,
+                immediateEffect: abilityHelper.immediateEffects.giveExperience({ target: thenContext.target })
+            })
```
moyerr: I think the only issue with the original implementation was that `giveExperience()`'s default target is `context.source`. If you manually specify `context.target` I think it should work the way we expected.

PR #2058, opened 2026-01-16
[Bugfix] Bail on Credit adjuster when adjusted cost reaches 0

moyerr:
Noticed a bug when using Kelleran Beq to play a card free (after discount) while having credit tokens. 

The credit token adjuster was still being triggered, which should not happen if the adjustment amount has already reached 0.

### DIFF
--- a/server/game/cards/06_IBH/events/IveFoundThem.ts
+++ b/server/game/cards/06_IBH/events/IveFoundThem.ts
@@ -35,11 +35,11 @@ export default class IveFoundThem extends EventCard {
                             target: topCards,
                             cardCondition: (card) => card.isUnit(),
                             canChooseFewer: false,
-                            immediateEffect: AbilityHelper.immediateEffects.sequential([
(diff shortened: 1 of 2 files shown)


server/game/core/cost/DefeatCreditTokensCostAdjuster.ts
```diff
@@ -65,6 +65,10 @@ export class DefeatCreditTokensCostAdjuster extends CostAdjusterWithGameSteps {
         costAdjustTriggerResult: ICostAdjustTriggerResult,
         abilityCostResult?: ICostResult
     ) {
+        if (this.isCancelled || !this.canAdjust(context.source, context, costAdjustTriggerResult)) {
```
AMMayberry1: Per the comment above, instead of calling `canAdjust()` we can just do an explicit check on the remaining adjusted value here. Since this is only called when the payment is actually happening, we can concretely know for sure whether or not there is any cost remaining that needs to be covered.

PR #2065, opened 2026-01-19
Swu stats deck link server PR

OotTheMonk:
Must be merged first before the client PR https://github.com/SWU-Karabast/forceteki-client/pull/606


server/utils/SWUStats/SwuStatsHandler.ts
```diff
@@ -439,4 +460,69 @@ export class SwuStatsHandler {
...
+            const decksData = await response.json() as ISwuStatsDecksResponse;
+            logger.info(`SWUStatsHandler: Successfully fetched ${decksData.decks.length} decks for user ${userId}`);
+            return decksData;
+        } catch (error) {
+            logger.error('SWUStatsHandler: Failed to fetch user decks', {
+                error: { message: error.message, stack: error.stack },
+                userId
+            });
+            return null;
+        }
```
CheBato: we probably should avoid returning null here since when an error occures we should handle it in the parent function, so i'd replace this with a throw

server/gamenode/GameServer.ts
```diff
@@ -480,6 +480,68 @@ export class GameServer {
...
+                    return res.status(404).json({
+                        success: false,
+                        message: 'SWU Stats account not linked or token expired'
+                    });
+                }
+
+                // Transform SWU Stats decks to our format
+                const transformedDecks = decksData.decks.map((deck) => ({
+                    id: deck.id,
+                    name: deck.name || 'Untitled Deck',
```
AMMayberry1: Is it expected that some swustats decks don't have names?

PR #2071, opened 2026-01-21
Swubase integration - link accounts and game stat tracking

the-medo:
This feature links SWUBase account to Karabast account, similar to what SWUStats does. Then, if user is linked and uses swubase decks, his statistics are tracked. These stats are tracked strictly for user and not public (team feature on swubase is planned to share multiple decklists and track stats together, but still, not public).

A lot of the stuff is copied / heavily inspired to what CheBato did with swustats integration. I tried to simplify/reuse some stuff, but didn't want to touch the swustats handler so OotTheMonk doesn't have to think much in case he would like to change some part of swustats integration.

---------------

### Notes for release
1. Should be merged together with client-side PR: https://github.com/SWU-Karabast/forceteki-client/pull/613
2. Needs two new ENV variables: SWUBASE_CLIENT_ID and SWUBASE_CLIENT_SECRET

---------------

### Simplified flow of account linking:
1. Karabast - Link Swubase button redirects to Swubase approval page
2. Swubase generates code and redirects back to the karabast, where swubase endpoint is called to confirm the link.

This flow is described in more detail in swubase repo: https://github.com/the-medo/swu-collection/blob/main/docs/karabast-integration/integration-workflow.md

### Stat tracking:
- Only matches with at least one linked user are sent (+using swubase deck link). Having swubase decklink without linked account is not enough. 
- Deck urls / ids of not linked users are not sent

Example of stat tracking payload: https://github.com/the-medo/swu-collection/blob/main/server/routes/integration/karabast/game-result/karabast-payload.example.json

## How to test:
Two new env variables will be needed in `forceteki` repo:

```
SWUBASE_CLIENT_ID=""
SWUBASE_CLIENT_SECRET=""
```
I'll provide testing ones to anyone, just don't want to post it here. New client id / secret will be generated upon release (by you or me, doesn't really matter that much).

1. Login with normal user (when i try it with available testing users, i get `[SWUBase] Callback failed: No valid user session` so I'm not sure if that is possible currently) - so google/discord providers will be needed
2. Login on [Swubase](https://swubase.com/) too
3. Go to Preferences tab - on the bottom, click "Link Swubase" button and go through the flow. You should end up back in Karabast with Green button saying account is linked.

For game result submission test, just simply play a test game, but linked account needs to use swubase deck link, for example:
`https://swubase.com/decks/45068753-3ac5-4c0f-b3f8-a1c3921bfe25`
`https://swubase.com/decks/96c9e579-13ed-49ea-ba2f-daaccae9bb48`

This game should be logged for you in the [currently hidden statistics page](https://swubase.com/statistics) on swubase - you will just see payload for now, not any real stats. First I'd like to have karabast - swubase integration ready, I'll work on how to actually display these statistics in the following days/weeks.

Also recorded a 1min video of linking + stat tracking:
https://www.youtube.com/watch?v=zRf9R7O4D-Q


server/gamenode/GameServer.ts
```diff
@@ -61,13 +67,17 @@ interface ILobbyMapping {
...
+
+export interface IToken {
     accessToken: string;
     refreshToken: string;
     creationDateTime: Date;
     timeToLiveSeconds: number;
 }
 
+export type ISwuStatsToken = IToken;
+export type ISwuBaseToken = IToken;
```
CheBato: Is there a reason we need to differentiate between the two?

server/utils/SWUBase/SwuBaseHandler.ts
```diff
@@ -0,0 +1,379 @@
...
+                gameId: game.id,
+                lobbyId: lobbyId
+            });
+            throw error;
+        }
+    }
+
+    private buildPlayerData(player: Player, accessToken: string | null) {
+        const d = player.lobbyDeck;
+        return {
```
CheBato: Where possible lets avoid using single char variable names since it might be difficult to infer what the variable is. Something like playerLobbyDeck would be alright or you could just use the player.lobbyDeck

server/utils/SWUBase/SwuBaseHandler.ts
```diff
@@ -0,0 +1,379 @@
...
+import { StatsMessageKey } from '../stats/statsMessages';
+import type { IGameStatisticsTracker } from '../../gameStatistics/GameStatisticsTracker';
+import { GameCardMetric } from '../../gameStatistics/GameStatisticsTracker';
+
+interface OAuthTokenResponse {
+    access_token: string;
+    refresh_token: string;
+    expires_in?: number;
+    token_type?: string;
+    scope?: string;
```
CheBato: Since SwuBaseHandler and SwuStatsHandler use similiar interfaces we could consolidate them into a single file.

CheBato (changes requested):

AMMayberry1 (approved):

PR #2089, opened 2026-01-24
Chat messages delta sending

brianries:
This branch ships the game state chat messages in batches instead of resending the entire state on every action.  Effectively we keep an array of messages on the client 'in sync' with the server.  It uses a user's lastMessageOffset to track which messages need to be sent on every game state update.  The client side than simply 'places' these messages into an indexed array, and does some simple gap detection if *somehow* messages were desynced or dropped.  

Every reconnect / new join sets the users lastMessageOffset to 0

**Note: this has to be merged in tandem with the client side branch!!**

### DIFF
--- a/server/game/cards/01_SOR/units/ChewbaccaLoyalCompanion.ts
+++ b/server/game/cards/01_SOR/units/ChewbaccaLoyalCompanion.ts
@@ -11,11 +11,8 @@ export default class ChewbaccaLoyalCompanion extends NonLeaderUnitCard {
     }
 
     public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
-        registrar.addTriggeredAbility({
+        registrar.addOnDefenseAbility({
             title: 'Ready Chewbacca',
-            when: {
-                onAttackDeclared: (event, context) => event.attack.getAllTargets().includes(context.source)
-            },
             immediateEffect: AbilityHelper.immediateEffects.ready()
         });
     }
--- a/server/game/cards/02_SHD/units/ClanSaxonGauntlet.ts
+++ b/server/game/cards/02_SHD/units/ClanSaxonGauntlet.ts
@@ -12,11 +12,8 @@ export default class ClanSaxonGauntlet extends NonLeaderUnitCard {
     }
 
     public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
-        registrar.addTriggeredAbility({
+        registrar.addOnDefenseAbility({
             title: 'Give an experience token to a unit',
-            when: {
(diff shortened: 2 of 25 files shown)


server/gamenode/Lobby.ts
```diff
@@ -1203,6 +1209,15 @@ export class Lobby {
             this.bo3LobbyReadyTimer?.stop();
             this.rematchRequest = null;
             this.statsUpdateStatus.clear();
+
+            // Reset message tracking for all clients on new game
+            for (const user of this.users) {
```
AMMayberry1: Since we need to kind of separately manage the lobby and game messages, instead of resetting like this could we have a message counter on the `Game` object and then some logic in this class that will return that value if it exists? And if not, it will return the count property on the `Lobby` object

server/gamenode/Lobby.ts
```diff
@@ -2072,16 +2101,45 @@ export class Lobby {
...
             }
         }
     }
 
+    /**
+     * Handle client request for message retransmit when gaps are detected.
+     * Client calls this with (startIndex, endIndex) to get messages in that range.
+     */
+    private retransmitGameMessages(socket: Socket, startIndex: number, endIndex: number): void {
+        if (!this.game) {
```
AMMayberry1: The `Lobby` has its own chat manager that will also need to be accounted for. Long-term we want to consolidate the chat a little more between game and lobby but that's probably out of scope for this PR

server/gamenode/Lobby.ts
```diff
@@ -1019,6 +1024,7 @@ export class Lobby {
             }
 
             spectator.state = 'disconnected';
+            spectator.lastMessageOffset = 0;
```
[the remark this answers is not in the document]
brianries: I think you're right.  The lastMessageOffset isn't updated if the client isn't connected on sendGameState, so it'll track where they left off anyway.  And if for some reason its wrong, they'll retransmit.

Removed these from the code.

server/gamenode/Lobby.ts
brianries: Should I rename this variable to its obvious its the game state chat? Do we think its worth managing the lobby state chat in the same fashion? I imagine that one is small enough that it wouldn't win us much?  Let me know!

server/gamenode/Lobby.ts
```diff
@@ -1403,6 +1418,12 @@ export class Lobby {
                 return;
             }
 
+            // Handle message retransmit requests from clients
+            if (command === 'retransmitGameMessages') {
```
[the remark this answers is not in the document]
brianries: Made this a lobby message.   Also had to add a lobby message handler for spectators so they could do the retransmitMessages calls

server/gamenode/Lobby.ts
brianries: Moved the tracking of this into the Game object using a Map of id->offset which cleaned up a fair bit of code

PR #2091, opened 2026-01-25
Add the ability to have Bo1 and Bo3 Limited Lobbies

supercodepoet:
A potential change to allow lobbies (public/private) Bo1 or Bo3 Limited games. I wanted to test how far Claude Code could get in adding a new feature. Now that we have Limited PQs, Sectors, and Regionals, I have seen folks wanting an easier way to test their Limited builds.

### DIFF
--- a/server/gamenode/GameServer.ts
+++ b/server/gamenode/GameServer.ts
@@ -974,8 +974,12 @@ export class GameServer {
         app.post('/api/create-lobby', this.buildAuthMiddleware(), async (req, res, next) => {
             try {
                 const { deck, format, isPrivate, gamesToWinMode, lobbyName } = req.body;
+                const user = req.user as User;
+
                 let { allow30CardsInMainBoard } = req.body;
-                const user = req.user;
+
+                // Limited format always uses 30-card minimum
+                allow30CardsInMainBoard = allow30CardsInMainBoard || (format === SwuGameFormat.Limited);
 
                 // Check if the user is already in a lobby
                 if (!this.canUserJoinNewLobby(user.getId())) {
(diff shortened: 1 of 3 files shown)


server/gamenode/GameServer.ts
```diff
@@ -973,7 +973,8 @@ export class GameServer {
 
         app.post('/api/create-lobby', this.buildAuthMiddleware(), async (req, res, next) => {
             try {
-                const { deck, format, isPrivate, gamesToWinMode, lobbyName, allow30CardsInMainBoard } = req.body;
+                const { deck, format, isPrivate, gamesToWinMode, lobbyName } = req.body;
+                let { allow30CardsInMainBoard } = req.body;
```
CheBato: I'm just thinking do we need to supply the allow30CardsInMainBoard in the req.body at all since we later check the format and set it to true?

server/gamenode/Lobby.ts
```diff
@@ -352,6 +352,7 @@ export class Lobby {
             rematchRequest: this.rematchRequest,
             matchingCountdownText: this.matchingCountdownText,
             allow30CardsInMainBoard: this.allow30CardsInMainBoard,
+            formatUses30CardDeck: DeckValidator.formatUses30CardMinimum(this.gameFormat),
```
CheBato: I'm curious is **formatUses30CardDeck** really necesary since we basically check for the same thing in allow30CardsinMainBoard? Consequently we probably won't need the DeckValidator function **formatUses30CardMinimum**.

server/utils/deck/DeckValidator.ts
```diff
@@ -20,6 +20,9 @@ enum SwuSet {
     LAW = 'law'
 }
 
+// Configurable: the set allowed in Limited format (update when new sets release)
+const limitedLegalSet: SwuSet = SwuSet.SEC;
```
AMMayberry1: I know we said that the current-set format would be what we'd choose by default, but that gets a little weird for times like right now when both Premier and Preview would have traffic. Potentially we'd need both concepts for limited as well...

server/utils/deck/DeckValidator.ts
```diff
@@ -38,6 +41,11 @@ const legalBlocksForFormat = new Map<SwuGameFormat, Set<SwuRotationBlock>>([
     [SwuGameFormat.NextSetPreview, new Set([SwuRotationBlock.BlockA, SwuRotationBlock.BlockB])]
 ]);
 
+// For formats that restrict to specific sets (not rotation blocks)
+const legalSetsForFormat = new Map<SwuGameFormat, Set<SwuSet>>([
```
AMMayberry1: nit: if there's only one entry here, do we need a whole `Map` for it?

PR #2103, opened 2026-01-28
Max Rebo and his band 🥁 🎷 🎹 

moyerr:
| ⟡ Max Rebo | ⟡ The Max Rebo Band |
| :---: | :---: |
| Encore! | Jatz-Wailers |
|  |  |

### DIFF
--- a/server/game/core/GameObject.ts
+++ b/server/game/core/GameObject.ts
@@ -112,9 +112,17 @@ export abstract class GameObject<T extends IGameObjectState = IGameObjectState> 
      * can be a value of {@link AbilityRestriction} or an arbitrary string such as a card name.
      */
     public hasRestriction(actionType: (AbilityRestriction | EffectName) | (AbilityRestriction | EffectName)[], context?: AbilityContext) {
-        return this.getOngoingEffectValues<Restriction>(EffectName.AbilityRestrictions).some((restriction) =>
-            restriction.isMatch(actionType, context)
-        );
+        return this.getOngoingEffectValues<Restriction>(EffectName.AbilityRestrictions)
+            .some((restriction) => restriction.isMatch(actionType, context));
+    }
+
+    /**
+     * Returns true if the card has any ability restriction matching the given name. Restriction names
+     * can be a value of {@link AbilityRestriction} or an arbitrary string such as a card name.
+     */
+    public getMatchingRestrictions(actionType: (AbilityRestriction | EffectName) | (AbilityRestriction | EffectName)[], context?: AbilityContext) {
+        return this.getOngoingEffectValues<Restriction>(EffectName.AbilityRestrictions)
+            .filter((restriction) => restriction.isMatch(actionType, context));
     }
 
     public getShortSummary() {
--- a/server/game/cards/02_SHD/events/DetentionBlockRescue.ts
+++ b/server/game/cards/02_SHD/events/DetentionBlockRescue.ts
@@ -14,7 +14,7 @@ export default class DetentionBlockRescue extends EventCard {
 
     public override setupCardAbilities(registrar: IEventAbilityRegistrar, AbilityHelper: IAbilityHelper) {
         registrar.setEventAbility({
-            title: 'Deal 2 damage to a unit or base',
+            title: 'Deal 3 damage to a unit. If that unit is guarding any captured cards, deal 6 damage instead.',
             targetResolver: {
                 cardTypeFilter: WildcardCardType.Unit,
                 immediateEffect: AbilityHelper.immediateEffects.conditional({
--- a/server/game/cards/07_LAW/events/AttackFromAllSides.ts
+++ b/server/game/cards/07_LAW/events/AttackFromAllSides.ts
@@ -1,0 +1,34 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import type { IEventAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { WildcardCardType } from '../../../core/Constants';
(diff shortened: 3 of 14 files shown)


server/game/core/GameObject.ts
```diff
@@ -89,6 +95,29 @@ export abstract class GameObject<T extends IGameObjectState = IGameObjectState>
         return result;
     }
 
+    public filterOngoingEffects<Value = any>(filter: IOngoingEffectFilters): { source: Card; type: EffectName; value: Value }[] {
+        const effects = this.state.ongoingEffects;
+        const result: { source: Card; type: EffectName; value: Value }[] = [];
+        // eslint-disable-next-line @typescript-eslint/prefer-for-of
+        for (let i = 0; i < effects.length; i++) {
```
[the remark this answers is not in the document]
moyerr: Yeah, valid question. I was really just following the pattern set by existing `ongoingEffects` methods in this class, so not totally sure if this was for optimization reasons initially.

server/game/core/gameSteps/phases/Phase.ts
```diff
@@ -20,16 +23,20 @@ export enum PhaseInitializeMode {
 export abstract class Phase extends BaseStepWithPipeline {
     protected readonly name: PhaseName;
     protected readonly snapshotManager: SnapshotManager;
+    protected readonly onPhaseStarted: (Phase) => void = null;
+    protected readonly additionalPhaseEffect: GameObjectRef<AdditionalPhase>;
```
AMMayberry1: There are some utilities now that make it so we don't have to explicitly manage the `GameObjectRef` here. Specifically, look at usages of the `@undoObject()` decorator.

For here we could just do something like:

```ts
@undoObject()
protected readonly additionalPhaseEffect: AdditionalPhase;
```

Then it manages the mapping back and forth under the hood

server/game/core/Game.js
```diff
@@ -1389,7 +1409,88 @@ class Game extends EventEmitter {
...
+        }
+    }
+
+    /**
+     * Creates additional regroup phases as needed based on ongoing effects.
+     * @param {PhaseInitializeMode} regroupInitializeMode
+     */
+    checkCreateAdditionalRegroupPhases(regroupInitializeMode) {
+        /** @type {AdditionalPhase[]} */
+        const additionalRegroupPhaseEffects = this.getPlayers()
```
AMMayberry1: We already build this list in `buildRegroupPhaseStep`, could we pass that in here to avoid doing this twice (it's a potentially expensive operation)

server/game/core/Game.js
```diff
@@ -1389,7 +1409,88 @@ class Game extends EventEmitter {
...
+            const regroupPhase = new RegroupPhase(
+                this,
+                this._snapshotManager,
+                usedInitializeMode  // All initialize modes after the first should be Normal
+                    ? PhaseInitializeMode.Normal
+                    : regroupInitializeMode,
+                effect.getRef()
+            );
+
+            this.pipeline.queueStep(regroupPhase);
```
AMMayberry1: This seems like it might be double-adding the phases to the pipeline. `buildRegroupPhaseStep()` returns a list of phases which are then queued, if we're also doing it here then I'm guessing it's getting added twice? Would the current tests catch that?

test/scenarios/undo/AdditionalPhases.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+            const assertSecondRegroupPhaseStartState = (context) => {
+                // Additional Regroup Phase: Fireball is defeated by taking 1 more damage, each player draws C,D
+                expect(context.game.currentPhase).toBe('regroup');
+                expect(context.fireball).toBeInZone('discard', context.player1);
+                expect([context.p1Draws.c, context.p1Draws.d]).toAllBeInZone('hand', context.player1);
+                expect([context.p2Draws.c, context.p2Draws.d]).toAllBeInZone('hand', context.player2);
+            };
+
+            undoIt('Fireball takes damage in each regroup phase', function() {
```
AMMayberry1: I'm actually not sure if wrapping this with `undoIt()` gives any benefit here since that will reset the whole test and run it a second time. That's more for wrapping a basic card unit test in undo to confirm nothing breaks.

Since we're doing the rollbacks within the test we can just use a standard `it()` like we do in `PhaseStartAndEnd.spec.ts`.

server/game/core/gameSteps/phases/Phase.ts
moyerr: I don't think I can use these because the `Phase` object doesn't inherit from `GameObjectBase` (it is a `BaseStep` subclass)

server/game/core/Game.js
moyerr: Similar to the response below, we can't share this filtered list because one gets called when we build the Round pipeline, and one gets called when the first Regroup phase ends. Game state will not be the same at these two time points.

server/game/core/gameSteps/phases/Phase.ts
AMMayberry1: Oh got it, I was thinking about this wrong - so then yeah I think we just shouldn't be using `GameObjectRef` at all here. That's exclusively to be used for undo-able state properties that are attached to a `GameObjectBase`, which is the base class for all undo-able objects.

Its purpose is to be handle that we can use to refer to objects before and after an undo has occurred, assuming that the object holding the Ref is a `GameObjectBase` whose state will be updated as part of the undo. I think in an earlier conversation I misspoke and said that we can't hold complex objects as undo state - we can, we just have to use the `GameObjectRef` wrapper.

Since a `Phase` will always be wiped out completely during an undo, we don't need to worry about the `GameObjectRef` here and should just use the `AdditionalPhase` directly.

AMMayberry1 (approved):

PR #2104, opened 2026-01-29
bt1, 000, shadow of stygeon prime

sethhelman:
bt1
shadow of stygeon prime

### DIFF
--- a/scripts/mockdata.js
+++ b/scripts/mockdata.js
@@ -332,7 +332,7 @@ const mockCards = [
         unique: true,
         arena: 'ground',
         internalName: 'bt1#blastomech',
-      }),
+    }),
     buildMockCard({
         title: 'Latts Razzi',
         subtitle: 'Deadly Whipmaster',


server/game/cards/07_LAW/units/0-0-0TranslationAndTorture.ts
```diff
@@ -0,0 +1,33 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '000#translation-and-torture-id',
+            internalName: '000#translation-and-torture',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addOnAttackAbility({
+            title: 'Put an Aggression card from your discard pile on the bottom of your deck. If you do, deal 1 damage to each enemy base',
```
moyerr: We eventaully want to be able to display aspect icons inline in the FE anywhere they are mentioned in prompts and logs. When we get there, the fewer places we need to replace the aspect text, the easier it will be. To that end, can you use the `aspectString` function from `EnumHelpers.ts` to write out the Aggression aspect name?

test/server/cards/07_LAW/upgrades/ShadowOfStygeonPrime.spec.ts
```diff
@@ -0,0 +1,56 @@
+describe('Frozen in Carbonite', function() {
+    integration(function(contextRef) {
+        describe('Frozen in Carbonite\'s ability', function() {
```
moyerr: Copypasta

PR #2109, opened 2026-01-31
The Client LAW leader

Frugghi:
| ⟡ The Client | Please Lower Your Blaster | 
| :---: | :---: |
| image | image |

### DIFF
--- a/server/game/gameSystems/CreateForceTokenSystem.ts
+++ b/server/game/gameSystems/CreateForceTokenSystem.ts
@@ -14,7 +14,6 @@ export class CreateForceTokenSystem<TContext extends AbilityContext = AbilityCon
     public override readonly effectDescription = 'gain the Force';
 
     public override eventHandler(event, additionalProperties: Partial<ICreateForceTokenProperties> = {}): void {
-        const context = event.context;
         for (const token of event.generatedTokens) {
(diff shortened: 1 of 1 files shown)


test/server/cards/07_LAW/leaders/TheClientPleaseLowerYourBlaster.spec.ts
```diff
@@ -0,0 +1,111 @@
+describe('The Client, Please Lower Your Blaster', function() {
+    integration(function(contextRef) {
+        describe('The Client\'s leader undeployed ability', function() {
```
moyerr: Hey, welcome back @Frugghi! 😄 

The ability implementation looks great! But this unit will really be the first stress test for the `TokensCreatedThisPhaseWatcher`. For that reason, I'd like to have all our bases covered as far as token creation:

- Token upgrades (already included)
- Credit tokens (already included)
- Token units
- Force token

I can already tell you that the Force token won't be picked up by the watcher. That should be an easy fix though: in `CreateForceTokenSystem.ts` we'll need to set the `generatedTokens` property on the event like we do with all the other token types.

I wonder if we can also add a case where a friendly ability makes an opponent create tokens, to verify that it doesn't satisfy The Client's condition. Something like Jesse from TWI.

PR #2111, opened 2026-01-31
Conveyex & KT9

bwohldmann:
image

image

### DIFF
--- a/server/game/cards/07_LAW/leaders/TheClientPleaseLowerYourBlaster.ts
+++ b/server/game/cards/07_LAW/leaders/TheClientPleaseLowerYourBlaster.ts
@@ -1,0 +1,53 @@
+import type {
+    ILeaderUnitAbilityRegistrar,
+    ILeaderUnitLeaderSideAbilityRegistrar
+} from '../../../core/card/AbilityRegistrationInterfaces';
+import { LeaderUnitCard } from '../../../core/card/LeaderUnitCard';
+import type { IAbilityHelper } from '../../../AbilityHelper';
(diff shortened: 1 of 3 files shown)


scripts/mockdata.js
```diff
@@ -400,6 +400,40 @@ const mockCards = [
         unique: false,
         internalName: 'combat-exercise',
     }),
+    buildMockCard({
```
moyerr: Looks like the mockdata for these was added to FFG's data today, so we won't need these mocks after all

server/game/cards/07_LAW/units/ConveyexSecurityCaptain.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addConstantAbility({
+            title: 'Enemy Credit tokens lose all abilities',
+            targetController: RelativePlayer.Opponent,
+            targetZoneFilter: WildcardZoneName.Any,
+            targetCardTypeFilter: WildcardCardType.Token,
+            matchTarget: (card) => card.isCreditToken(),
+            ongoingEffect: abilityHelper.ongoingEffects.blankCard()
```
moyerr: As an odd technicality (for our engine, not per game rules), Credit tokens live in the base zone. So we could specifically scope the zone filter to the Base zone.

Also, we should be able to use our `.loseAllAbilities()` ongoing effect. It's the same one we use for Kazuda, Force Lighting, Brain Invaders, etc.

PR #2117, opened 2026-02-02
Implementing Aerie, Weazel and Enfys Helmet

aruizc:
image
image
image


scripts/mockdata.js
```diff
@@ -400,6 +400,60 @@ const mockCards = [
...
+        aspects: ['aggression', 'heroism'],
+        types: ['unit'],
+        traits: ['underworld', 'vehicle', 'transport'],
+        setId: {
+            set: 'LAW',
+            number: 184,
+        },
+        unique: true,
+        arena: 'space',
+        internalName: 'aerie#cloud-rider-dropship',
```
moyerr: Hyphens get eliminated when we convert to `internalName`, so it would be `aerie#cloudrider-dropship`

moyerr (approved):

PR #2133, opened 2026-02-06
Betrayed Trust

moyerr:
| Betrayed Trust |
| :---: |
|  |


server/game/core/attack/Attack.ts
```diff
@@ -50,6 +50,10 @@ export class Attack {
     }
 
     public getAttackerCombatDamage(context: AbilityContext): number | null {
+        if (this.attacker.hasRestriction(AbilityRestriction.DealCombatDamage)) {
+            return 0;
```
AMMayberry1: Instead of doing it this way it might be slightly better to just prevent the damage event from being emitted at all. I'm wondering if there could be a case where some replacement effect increases the damage value, and then the event will go live again since the value would be nonzero.

I think Vigil could do that, might be good to add a test for it

PR #2137, opened 2026-02-06
update Mockdata

sethhelman:


scripts/mockdata.js
```diff
@@ -17,6 +17,76 @@ const mockCards = [
...
+    }),
+    buildMockCard({
+        title: 'Honor-Bound Partisan',
+        power: 2,
+        hp: 2,
+        cost: 2,
+        hasNonKeywordAbility: true,
+        aspects: ['Command', 'Aggression'],
+        types: ['unit'],
+        traits: ['rebel', 'twilek'],
```
moyerr: I think you had it right before `twi\'lek`

PR #2140, opened 2026-02-07
Capture Fix for HK-47 and Corellian Hounds

adbronico:


test/server/cards/05_LOF/units/HK47ExclamationDieMeatbag.spec.ts
```diff
@@ -154,5 +154,34 @@ describe('HK-47, Exclamation: Die Meatbag', function() {
             expect(context.player2).toBeActivePlayer();
             expect(context.p2Base.damage).toBe(0);
         });
+
+
+        it('HK-47 should not be triggered when units die while he is captured', async function () {
```
AMMayberry1: There's a little bit of timing nuance for being defeated by a card effect vs being defeated by damage, let's add a second test where the captor is defeated by damage to ensure that works as well

test/server/cards/05_LOF/units/HK47ExclamationDieMeatbag.spec.ts
```diff
@@ -154,5 +154,34 @@ describe('HK-47, Exclamation: Die Meatbag', function() {
...
+                player2: {
+                    hand: ['take-captive'],
+                    groundArena: ['wampa'],
+                    hasInitiative: true,
+                }
+            });
+
+            const { context } = contextRef;
+
+            context.player2.clickCard(context.takeCaptive);
```
AMMayberry1: We have a test setup thing for setting up capture now - `capturedUnits: ['hk47#exclamation-die-meatbag']`, similar format as upgrades

PR #2155, opened 2026-02-10
[Bugfix] SEC Cassian leader interaction with Condemn

moyerr:
### Description

SEC Cassian leader was dealing damage, when he should have been immediately defeated when declaring an attack, due to Condemn removing all abilities.


moyerr:
Subsumed by #2252

PR #2168, opened 2026-02-13
Savage, sidious, brothers

sethhelman:
sidious
brothers
savage back
savage front

### DIFF
--- a/test/server/cards/07_TS26/leaders/SavageOpressYouMustHaveYourRevenge.spec.ts
+++ b/test/server/cards/07_TS26/leaders/SavageOpressYouMustHaveYourRevenge.spec.ts
@@ -14,8 +14,11 @@ describe('Savage Opress, You Must Have Your Revenge', function() {
                 });
                 const { context } = contextRef;
 
+                // Only highest power unit gets Overwhelm
                 expect(context.echoBaseDefender.hasSomeKeyword('overwhelm')).toBeTrue();
                 expect(context.battlefieldMarine.hasSomeKeyword('overwhelm')).toBeFalse();
+
+                // Overwhelm not given to opponent's units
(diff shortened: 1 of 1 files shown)


server/game/cards/07_TS26/events/Brothers.ts
```diff
@@ -0,0 +1,45 @@
...
+                        type: AbilityType.DamageModification,
+                        modificationType: DamageModificationType.PreventAll,
+                        damageOfType: DamageSourceType.Attack
+                    }),
+                }
+            },
+            then: {
+                title: 'Attack with another unique unit. Prevent all combat damage that would be dealt to it for this attack',
+                initiateAttack: {
+                    attackerCondition: (card) => card.unique,
```
moyerr: I do think we still need to verify that it isn't the same as the first chosen unit. In case there is an effect that allows the first unit to ready during/after its attack.

test/server/cards/07_TS26/events/Brothers.spec.ts
```diff
@@ -0,0 +1,55 @@
+describe('Brothers', function () {
+    integration(function (contextRef) {
+        describe('Brothers\' ability', function () {
```
moyerr: Since the wording is "up to 2", we should have cases where there are 2 valid targets, but the player chooses 0 and 1.

test/server/cards/07_TS26/leaders/SavageOpressYouMustHaveYourRevenge.spec.ts
```diff
@@ -0,0 +1,71 @@
+describe('Savage Opress, You Must Have Your Revenge', function() {
+    integration(function(contextRef) {
+        describe('Savage Opress\' undeployed ability ', function() {
```
moyerr: Let's at least verify some of the cases with overwhelm damage via an attack

PR #2184, opened 2026-02-15
Liberty

VJubert:

### DIFF
--- a/server/utils/deck/DeckValidator.ts
+++ b/server/utils/deck/DeckValidator.ts
@@ -167,6 +167,9 @@ export class DeckValidator {
         const startingDeckSizeValue = allow30CardsInMainBoard ? 30 : 50;
 
         const baseData = this.getCardCheckData(baseId);
+        if (!baseData) {
+            return startingDeckSizeValue;
+        }
         return startingDeckSizeValue + (baseData.minDeckSizeModifier ?? 0);
     }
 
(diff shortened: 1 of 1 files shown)


server/game/cards/07_LAW/units/LibertyDrawTheirFire.ts
```diff
@@ -0,0 +1,36 @@
...
+                onAttack: true,
+            },
+            // cannotTargetFirst: true,
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                controller: RelativePlayer.Opponent,
+                immediateEffect: abilityHelper.immediateEffects.simultaneous([
+                    abilityHelper.immediateEffects.exhaust(),
+                    abilityHelper.immediateEffects.returnToHand((context) => {
+                        return ({
```
AMMayberry1: nit: we don't need `return` here, we can just build the object directly

test/server/cards/07_LAW/units/LibertyDrawTheirFire.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+            context.player2.clickCard(context.awing);
+
+            context.player1.clickCard(context.liberty);
+            context.player1.clickCard(context.awing);
+
+            expect(context.player2).toBeActivePlayer();
+            expect(context.awing).toHaveExactUpgradeNames([]);
+            expect(context.awing.exhausted).toBeTrue();
+            expect(context.majorVonreg).toBeInZone('base', context.player2);
+        });
```
AMMayberry1: Let's add one upgrade that is controlled by the other player and confirm it goes back to their hand

PR #2196, opened 2026-02-16
Implementing Broken Horn, Vizago's Pride

aruizc:
image

### DIFF
--- a/server/game/cards/07_LAW/events/ChokeOnAspirations.ts
+++ b/server/game/cards/07_LAW/events/ChokeOnAspirations.ts
@@ -1,0 +1,34 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import type { IEventAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { EventCard } from '../../../core/card/EventCard';
+import { RelativePlayer, WildcardCardType } from '../../../core/Constants';
+import { Trait } from '../../../core/Constants';
+
+export default class ChokeOnAspirations extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: 'choke-on-aspirations-id',
+            internalName: 'choke-on-aspirations',
+        };
+    }
+
+    public override setupCardAbilities(registrar: IEventAbilityRegistrar, AbilityHelper: IAbilityHelper): void {
+        registrar.setEventAbility({
+            title: 'Deal up to 5 damage to a friendly unit',
+            immediateEffect: AbilityHelper.immediateEffects.distributeDamageAmong({
+                amountToDistribute: 5,
+                canDistributeLess: true,
+                controller: RelativePlayer.Self,
+                canChooseNoTargets: true,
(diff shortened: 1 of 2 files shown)


test/server/cards/07_LAW/units/BrokenHornVizagosPride.spec.ts
```diff
@@ -0,0 +1,36 @@
+describe('Broken Horn, Vizago\'s Pride', function () {
+    integration(function (contextRef) {
+        describe('Broken Horn, Vizago\'s Pride when played ability', function () {
+            it('should let the player to draw a card and resource the top card of their deck', async function () {
```
VJubert: can you add a test with an empty deck. It should take 3 damage from draw and not resource anything

server/game/cards/07_LAW/units/BrokenHornVizagosPride.ts
```diff
@@ -0,0 +1,30 @@
...
+        return {
+            id: 'broken-horn#vizagos-pride-id',
+            internalName: 'broken-horn#vizagos-pride'
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addWhenPlayedAbility({
+            title: 'If you have fewer cards in hand than your opponent, draw a card. If you control fewer resources than your opponent, resource the top card of your deck',
+            immediateEffect: AbilityHelper.immediateEffects.sequential([
```
AMMayberry1: This should be `simultaneous()`

test/server/cards/07_LAW/units/BrokenHornVizagosPride.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+                expect(context.greenSquadronAwing).toBeInZone('deck');
+                context.player1.clickCard(context.brokenHornVizagosPride);
+
+                // Player 1 should draw a card and resource the top card of their deck since they have fewer cards in hand and fewer resources than Player 2
+                expect(context.player1.hand.length).toBe(startingHandSize); // As player drawed a card but also played a card
+                expect(context.player1.resources.length).toBe(startingResourcesP1 + 1);
+                expect(context.greenSquadronAwing).toBeInZone('hand', context.player1);
+                expect(context.wampa).toBeInZone('resource', context.player1);
+                expect(context.player2).toBeActivePlayer();
+            });
```
AMMayberry1: Let's add a test where there is one card left in deck, then confirm that it goes to hand and nothing is resourced

server/game/cards/07_LAW/units/BrokenHornVizagosPride.ts
aruizc: I tried using simultaneous(), but it looks like draw and resource are affecting the same card at the same time. I think this should be sequential: first draw the card, then resource it. Let me know your thoughts

PR #2203, opened 2026-02-18
⟡ Vermillion, Qi'ra's Auction House

moyerr:
| ⟡ Vermillion |
| :---: |
| Qi'ra's Auction House |
|  |


server/game/cards/07_LAW/units/VermillionQirasAuctionHouse.ts
```diff
@@ -0,0 +1,81 @@
...
+                        target: context.player.getTopCardOfDeck(),
+                    }),
+                    ['Opponent\'s deck']: AbilityHelper.immediateEffects.reveal({
+                        useDisplayPrompt: true,
+                        target: context.player.opponent.getTopCardOfDeck(),
+                    })
+                })
+            },
+            ifYouDo: (outerContext) => ({
+                title: 'Choose a player',
```
AMMayberry1: This might not be worth it, but one small edge case optimization we could do would be to bypass the player selection completely if neither player can legally play the card. The most frequent example would probably be like an upgrade with attach restrictions or something in that vein.

We'd basically have to write a method that builds a System object using the same `playCardFromOutOfPlay()` as on line 59 and then call `canAffect()` or `hasLegalTarget()` on it with a context for each player. Doable, but maybe not worth the effort since it's a major edge case. Totally your call.

PR #2222, opened 2026-02-19
Update pre-LAW "on attack completed" abilities to match new template

AMMayberry1:


server/game/core/ability/TriggeredAbility.ts
```diff
@@ -97,6 +99,12 @@ export default class TriggeredAbility extends CardAbility<ITriggeredAbillityStat
         this.mustChangeGameState = !!this.properties.ifYouDo || !!this.properties.ifYouDoNot
             ? GameStateChangeRequired.MustFullyResolve
             : GameStateChangeRequired.MustFullyOrPartiallyResolve;
+
+        if ('attackerMustSurvive' in properties) {
```
moyerr: This string-based check is new to me. This just checks for the existence of the property? Should we also check that the value itself is `true`?

AMMayberry1: Yeah admittedly this is me being kind of lazy, it's not great practice to use but we don't really have a way to do type discrimination on these properties and I don't want to add a whole system for that just to do this slightly more cleanly. But yes, it does just check that the property exists. And yeah it should also be checking that it's `true`, very good call

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -209,6 +210,7 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor<TSta
         private _tokensCreatedThisPhaseWatcher: TokensCreatedThisPhaseWatcher;
         private _cardsPlayedThisWatcher: CardsPlayedThisPhaseWatcher;
         private _leadersDeployedThisPhaseWatcher: LeadersDeployedThisPhaseWatcher;
+        private _unitsDefeatedThisPhaseWatcher: UnitsDefeatedThisPhaseWatcher;
```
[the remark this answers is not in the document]
AMMayberry1: Yeah the watcher stuff could use a pretty serious cleanup; the entire "registration" concept is actually totally unnecessary. And the `StateWatcherLibrary` really ought to be a static for a couple reasons. I almost started in on some of that in this PR but thankfully stopped myself from indulging in the scope creep 😆

moyerr (approved):

PR #2225, opened 2026-02-19
commence the festivities, salvaged materials, salvaged blaster

VJubert:


server/game/cards/07_LAW/events/SalvagedMaterials.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public override setupCardAbilities (registrar: IEventAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.setEventAbility({
+            title: 'Play an Item upgrade from your discard pile. It costs 3 resources less. At the start of the next regroup phase, defeat it.',
+            targetResolver: {
+                zoneFilter: ZoneName.Discard,
+                controller: RelativePlayer.Self,
+                cardTypeFilter: WildcardCardType.Upgrade,
+                cardCondition: (card) => card.isUpgrade() && card.hasSomeTrait(Trait.Item),
```
AMMayberry1: I think we can remove the `card.isUpgrade()` check here and let it be handled by `playAsType: WildcardCardType.Upgrade`. That way if there's ever a new rules thing like e.g. events that can be played as upgrades, this should still function right

PR #2229, opened 2026-02-20
collateral damage & daring delve

VJubert:

### DIFF
--- a/server/game/cards/07_LAW/units/IndustriousTeam.ts
+++ b/server/game/cards/07_LAW/units/IndustriousTeam.ts
@@ -1,0 +1,25 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { WildcardCardType } from '../../../core/Constants';
+
+export default class IndustriousTeam extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '5713950795',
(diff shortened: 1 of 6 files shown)


server/game/cards/07_LAW/events/DaringDelve.ts
```diff
@@ -0,0 +1,40 @@
...
+            then: (thenContext) => {
+                const discardedAggressionCards = thenContext.events
+                    .filter((event) => event.name === EventName.OnCardDiscarded)
+                    .map((event) => event.card)
+                    .filter((card) => card.hasSomeAspect(Aspect.Aggression));
+
+                return {
+                    title: 'Return a discarded Aggression card to your hand',
+                    optional: true,
+                    targetResolver: {
```
AMMayberry1: This will open up the entire discard pile popup. I wonder if it would be simpler UX to use a DisplayCard modal since we only care about at most two cards

server/game/cards/07_LAW/events/CollateralDamage.ts
```diff
@@ -0,0 +1,42 @@
...
+                cardTypeFilter: WildcardCardType.Unit,
+                immediateEffect: AbilityHelper.immediateEffects.damage({
+                    amount: 2
+                })
+            },
+            then: (thenContext) => ({
+                title: 'Deal 2 damage to a base or another unit in the same arena',
+                targetResolver: {
+                    activePromptTitle: () => `Deal 2 damage to a base or a ${thenContext.target.zoneName === ZoneName.GroundArena ? 'ground' : 'space'} unit`,
+                    cardTypeFilter: [WildcardCardType.Unit, CardType.Base],
```
AMMayberry1: nit: we can probably just remove this since the condition is already filtering by type

PR #2236, opened 2026-02-20
Flash the Vents

Frugghi:
image

### DIFF
--- a/test/server/cards/07_LAW/events/FlashTheVents.spec.ts
+++ b/test/server/cards/07_LAW/events/FlashTheVents.spec.ts
@@ -134,7 +134,7 @@ describe('Flash the Vents', function () {
 
             expect(context.p1Base.damage).toBe(1);
             expect(context.p2Base.damage).toBe(0);
-            expect(context.jediStarfighter).toBeInZone('discard');
+            expect(context.rebellionYwing).toBeInZone('discard');
             expect(context.jediStarfighter).toBeInZone('discard');
         });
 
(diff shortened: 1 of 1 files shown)


server/game/gameSystems/AttackStepsSystem.ts
```diff
@@ -377,7 +377,8 @@ export class AttackStepsSystem<TContext extends AbilityContext = AbilityContext>
             properties.attacker as IUnitCard,
             [attackTarget],
             properties.isAmbush,
-            properties.attackerCombatDamageOverride
+            properties.attackerCombatDamageOverride,
+            context.game.getLastAttackId() // reuse the last attack id instead of generating a new one
```
AMMayberry1: What's the reason to do it this way?

test/server/cards/07_LAW/events/FlashTheVents.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+
+            expect(context.player1).toBeAbleToSelectExactly([context.p1Base, context.p2Base]);
+            context.player1.clickCard(context.p1Base);
+
+            expect(context.p1Base.damage).toBe(1);
+            expect(context.p2Base.damage).toBe(0);
+            expect(context.jediStarfighter).toBeInZone('discard');
+            expect(context.jediStarfighter).toBeInZone('discard');
+        });
+    });
```
AMMayberry1: I'm not 100% confident of the timing on this, but I think if a "when attack ends" ability deals damage to a base that should also be counted. We can test that using the new Cassian unit if he's the one attacking.

server/game/gameSystems/AttackStepsSystem.ts
Frugghi: It was to ensure that `thenContext.game.getLastAttackId()` returned the correct attack id but as you pointed out in the previous comment it doesn't work for nested attacks!

PR #2237, opened 2026-02-21
LAW Maul

AMMayberry1:
image

### DIFF
--- a/server/game/cards/07_LAW/units/MaulMasterOfTheShadowCollective.ts
+++ b/server/game/cards/07_LAW/units/MaulMasterOfTheShadowCollective.ts
@@ -1,7 +1,7 @@
 import type { IAbilityHelper } from '../../../AbilityHelper';
 import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
 import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
-import { DamageType, RelativePlayer, WildcardCardType } from '../../../core/Constants';
+import { RelativePlayer, WildcardCardType } from '../../../core/Constants';
 import type { StateWatcherRegistrar } from '../../../core/stateWatcher/StateWatcherRegistrar';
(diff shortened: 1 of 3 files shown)


server/game/cards/07_LAW/units/MaulMasterOfTheShadowCollective.ts
```diff
@@ -0,0 +1,57 @@
...
+            title: 'If Maul dealt combat damage to a base, take control of an enemy non-leader unit until Maul leaves the arena',
+            attackerMustSurvive: true,  // this is an optimization we can do for now since there isn't any effect that cares about momentarily taking control
+            immediateEffect: AbilityHelper.immediateEffects.conditional({
+                condition: (context) =>
+                    this.damageDealtThisPhaseWatcher.unitHasDealtDamage(
+                        context.source,
+                        (entry) =>
+                            entry.combatDamageAttackId === context.event.attack.id &&
+                            ((entry.damageType === DamageType.Combat && entry.targets.some((target) => target.isBase())) || entry.damageType === DamageType.Overwhelm)
+                    ),
```
moyerr: Between this and Chirrut, I wonder if it's worth encapsulating this logic in its own method on the watcher?

server/game/cards/07_LAW/units/MaulMasterOfTheShadowCollective.ts
```diff
@@ -0,0 +1,57 @@
...
+                            newController: context.player,
+                            target: context.target
+                        })),
+                        AbilityHelper.immediateEffects.whenSourceLeavesPlayDelayedCardEffect((context) => ({
+                            title: 'Return the stolen unit to its owner',
+                            target: context.target,
+                            immediateEffect: AbilityHelper.immediateEffects.takeControlOfUnit((context) => ({
+                                newController: selectCardContext.target.owner,
+                                target: selectCardContext.target
+                            })),
```
moyerr: Similar to Change of Heart and Tarkin, I think we need to explicitly say leader units shouldn't be excluded from targeting when the delayed effect resolves. 

This is for the case where the stolen unit becomes a leader unit due to a pilot leader

PR #2240, opened 2026-02-23
Update SearchDeckSystem to leverage other game systems for effects & messages

### LINKED ISSUE #1657: SearchDeckSystem does not fire an `OnCardRevealed` event for cases where the searched card is revealed
Now that we have cards that trigger when a card is revealed (SEC Padmé Amidala leader), we should ensure that all "reveal" effects actually fire an `OnCardRevealed` event. 

For the specific case of Padmé, her ability would not trigger in these deck search scenarios because the card is being revealed from the deck and not the hand, but for completeness, this game system should still send an `OnCardRevealed` event.

moyerr:
The investigation initiated by [this player report](https://discord.com/channels/1220057752961814568/1345468050568380568/1475408516968878174) ultimately revealed that the `SearchDeckSystem` was pretty outdated. It was performing game actions and emitting messages manually, the latter of which were not accurate for all scenarios. 

The fix for this ended up being to leverage our extensive `GameSystem` architecture, since those already perform game actions and emit game messages. This did have the unfortunate side effect of changing log messages that were covered by unit tests, so there are a lot of changes to update those tests.

Incidentally, this fixes #1657

### DIFF
--- a/server/game/cards/06_SEC/units/ChairmanPapanoidaUndauntedDiplomat.ts
+++ b/server/game/cards/06_SEC/units/ChairmanPapanoidaUndauntedDiplomat.ts
@@ -15,7 +15,7 @@ export default class ChairmanPapanoidaUndauntedDiplomat extends NonLeaderUnitCar
     public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
         const aspects = [Aspect.Aggression, Aspect.Aggression];
         registrar.addTriggeredAbility({
-            title: `Disclose ${EnumHelpers.aspectString(aspects)}. If you do, create a Spy token`,
+            title: `Disclose ${EnumHelpers.aspectString(aspects)} to create a Spy token`,
             collectiveTrigger: true,
             when: {
                 onCardsDrawn: (_event, context) => context.game.currentPhase === PhaseName.Action,
--- a/server/game/core/gameSteps/abilityWindow/TriggerWindowBase.ts
+++ b/server/game/core/gameSteps/abilityWindow/TriggerWindowBase.ts
@@ -170,7 +170,7 @@ export abstract class TriggerWindowBase extends BaseStep {
         }
 
         // Check to if we're dealing with a multi-selection of the 'same' ability
-        const repeatedAbilities = this.getRepeatedAbilityTriggers(abilitiesToResolve, this.resolved.map((resolved) => resolved.ability));
+        const repeatedAbilities = this.getRepeatedAbilityTriggers(abilitiesToResolve);
 
         for (const repeatedAbility of repeatedAbilities) {
             // if an ability is triggered multiple times and uses a collective trigger, filter down to one instance of it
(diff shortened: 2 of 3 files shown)


test/server/cards/05_LOF/events/Psychometry.spec.ts
```diff
@@ -56,7 +56,10 @@ describe('Psychometry', function() {
                 });
                 context.player1.clickCardInDisplayCardPrompt(context.jediLightCruiser);
 
-                expect(context.getChatLogs(2)).toContain('player1 takes Jedi Light Cruiser');
+                expect(context.getChatLog()).toEqual(
+                    'player1 uses Psychometry to reveal Jedi Light Cruiser, to draw a card, and to move 4 cards to the bottom of their deck'
```
VJubert: On all these texts, it could be : to draw this/these card(s) ?

server/game/core/Player.ts
```diff
@@ -727,8 +727,11 @@ export class Player extends GameObject<IPlayerState> implements IGameStatisticsT
      * Shuffles the deck, displaying a message in chat
      * @param {AbilityContext} context
      */
-    public shuffleDeck(context: AbilityContext = null) {
-        this.game.addMessage('{0} is shuffling their deck', this);
+    public shuffleDeck(context: AbilityContext = null, emitMessage = false) {
```
AMMayberry1: Could we just remove `context` since it's not used?

AMMayberry1: I'm wondering if `emitMessage` should default to true, feels like emitting the log should be the norm

server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -334,4 +362,57 @@ export class SearchDeckSystem<TContext extends AbilityContext = AbilityContext,
...
+        effectMessages.push({ format: effectMessage, args: effectArgs });
+
+        const moveEvents = [];
+        moveSystem.queueGenerateEventGameSteps(moveEvents, context);
+        context.game.queueSimpleStep(() => {
+            context.game.openEventWindow(moveEvents);
+        }, 'resolve move for remaining cards');
+    }
+
+    private revealHandler(properties: ISearchDeckProperties, context: TContext, event: any, selectedCards: Set<Card>, effectMessages: FormatMessage[]) {
```
AMMayberry1: nit: when we call something a "handler" I usually think of it as a discrete object as opposed to a method, feels a little off for these to be named this way. not a blocker though if you find this convention intuitive

test/server/cards/05_LOF/events/Psychometry.spec.ts
moyerr: Discussed in discord: there will be a follow-up to add a `RevealAndDrawSystem`, which will have clearer messages for cases like this

server/game/gameSystems/SearchDeckSystem.ts
moyerr: That's fair. Now that I think of it, I do prefer more imperative naming for instance methods, e.g. `handleCardReveal` or something.

server/game/gameSystems/SearchEntireDeckSystem.ts
```diff
@@ -22,7 +22,12 @@ export class SearchEntireDeckSystem<TContext extends AbilityContext = AbilityCon
     // constructor needs to do some extra work to ensure that the passed props object ends up as valid for the parent class
     public constructor(propertiesOrPropertyFactory: ISearchEntireDeckProperties<TContext> | ((context?: AbilityContext) => ISearchEntireDeckProperties<TContext>)) {
         const propertyWithSearchCount = GameSystem.appendToPropertiesOrPropertyFactory<ISearchDeckProperties<TContext>, 'searchCount'>(propertiesOrPropertyFactory, { searchCount: -1 });
-        super(propertyWithSearchCount);
+        const propertiesWithSearchCountAndRemainingCardsHandler = GameSystem.appendToPropertiesOrPropertyFactory<ISearchDeckProperties<TContext>, 'remainingCardsHandler'>(propertyWithSearchCount, {
+            remainingCardsHandler: () => {
+                // No op since we always shuffle after an entire deck search
```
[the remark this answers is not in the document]
moyerr: Good call. I think I'll actually use an override of the method that handles the remaining cards, since this should be the case for all entire deck searches.

PR #2252, opened 2026-02-28
Fix timing bug with SEC Cassian leader + Condemn

AMMayberry1:


moyerr (dismissed):

moyerr:
Superseded by #2562, which includes this work plus a follow-up fix for Anakin's Podracer's condition + the resulting chat-log timing.

PR #2266, opened 2026-03-03
Add chat logs for delayed effects

### LINKED ISSUE #2193: Delayed effects produce no game log message when they resolve
Delayed effects like the defeat part of Sneak Attack produce no message in the game log. This can cause some confusion, especially if players miss the logs from when the delayed effect was created.

We should emit game logs when a delayed effect resolves, so the game state is clearer in the moment.

Frugghi:
Fixes #2193


moyerr (approved):
Great work! 🚀

PR #2288, opened 2026-03-12
Chebato/mod tools

CheBato:
User Management allows moderators to search for players by username or ID, view their account details and moderation history, and apply actions: Mute (timed or permanent, preventing chat usage), Warning (logged note on the player's record), and Rename (forces a username change). Active actions can be cancelled, and all actions are tracked with timestamps, duration, and the issuing moderator.

Auto-cleanup: The ModActionCache extends TimedCache and refreshes once every 24 hours in production. On each refresh it re-fetches all active mod actions from DynamoDB. Any action whose expiresAt is in the past gets dropped from the in-memory cache, silently un-muting the user on their next refresh.

Testing the cache cleanup locally:

1. Start the local DynamoDB docker container (docker compose up dynamodb)
2. Run the seedAdmin script to activate yourself as a moderator
3. Set REFRESH_INTERVAL_MINUTES=5 in modActionCache
4. In the seedModAction script, set MY_USER_ID = <your_user_id> and EXPIRES_IN_MINUTES = 2, then run it
5. [optional] Set the ENV variable on the BE DEBUG_MOD_ACTION_CACHE=true
6. Start the server. The seeded mute will be active. After 2 minutes it expires, and the next cache refresh (within 1 min) 
7. cleans it up. Confirm on the Mod Page that the action shows as "Expired".

FE PR: https://github.com/SWU-Karabast/forceteki-client/pull/654


server/utils/ModActionCache.ts
```diff
@@ -0,0 +1,425 @@
...
+
+        return cacheMap.get(playerId) ?? null;
+    }
+
+    /**
+     * Check if a Mute entry has expired, and remove it if so.
+     * Pending mutes (no expiresAt) are never expired.
+     */
+    private validateMuteExpiry(playerId: string, entry: IActiveModActionCacheEntry): boolean {
+        if (entry.actionType !== ModActionType.Mute) {
```
AMMayberry1: nit: honestly this should probably be a contract check that will throw, this method shouldn't be called with a non-mute action

server/gamenode/GameServer.ts
```diff
@@ -438,35 +451,51 @@ export class GameServer {
...
+                // Start with legacy values (backwards compatibility)
+                let moderation = user.getModeration();
+                let needsUsernameChange = user.needsUsernameChange();
+
+                if (user.isAuthenticatedUser()) {
+                    const userId = user.getId();
+                    const activeActions = this.modActionCache.getActiveActionsForPlayer(userId);
+
+                    if (activeActions) {
+                        if (this.modActionCache.playerNeedsRename(userId)) {
```
AMMayberry1: Can we read these directly from `activeActions` since we have that already?

server/utils/ModActionCache.ts
```diff
@@ -0,0 +1,425 @@
...
+
+        if (cleanedCount > 0) {
+            logger.info(`ModActionCache: Cleaned up ${cleanedCount} expired mod actions from DB`);
+        }
+
+        return newCache;
+    }
+
+    /**
+     * Compare which mute should take priority.
```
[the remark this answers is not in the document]
CheBato: correct!

server/utils/ModActionCache.ts
```diff
@@ -0,0 +1,425 @@
...
+        }
+
+        // Incoming is active, existing is pending: compare potential expiresAt
+        if (incoming.expiresAt && !existing.expiresAt) {
+            const existingPotentialExpiry = Date.now() + (existing.durationDays ?? 0) * ModActionCache.MS_PER_DAY;
+            return new Date(incoming.expiresAt).getTime() > existingPotentialExpiry;
+        }
+
+        // Both pending: compare durationDays
+        return (incoming.durationDays ?? 0) > (existing.durationDays ?? 0);
```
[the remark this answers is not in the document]
CheBato: sounds good i'll do that :)

server/gamenode/GameServer.ts
CheBato: yes we can. Was just thinking if we should then keep the playerNeedsRename at all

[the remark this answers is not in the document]
CheBato: Basically if we add a new rename it will take it from the modActionCache but if it doesn't have a rename in the modActionCache we use the one on the profile object. I was thinking we could do it this way for like 1 month and then clean up the legacy db values. What do you think?

server/utils/ModActionService.ts
```diff
@@ -0,0 +1,425 @@
...
+                    cleanedCount++;
+                } catch (error) {
+                    logger.error(`ModActionCache: Failed to clean up expired action ${action.id}`, {
+                        error: { message: error.message, stack: error.stack }
+                    });
+                }
+                continue;
+            }
+
+            if (!newCache.has(action.playerId)) {
```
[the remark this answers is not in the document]
CheBato: I've misread originally. We actually do a full DB sync and remove the previous cache.

scripts/seedModAction.ts
```diff
@@ -1,17 +1,28 @@
-// scripts/seedModAction.ts
+/**
+ * Seed script for inserting moderation actions directly into local DynamoDB.
+ * Useful for testing mod features (mutes, warnings, renames) without going through the mod UI.
+ *
+ * Prerequisites:
+ *   - ENVIRONMENT=development and USE_LOCAL_DYNAMODB=true in your env
+ *   - Local DynamoDB Docker container running
```
AMMayberry1: nit: would be good to give the docker-compose command here in case people don't know

server/utils/ModActionService.ts
```diff
@@ -319,13 +320,13 @@ export class ModActionService {
     /**
      * Checks if a player has an active force rename.
      */
-    public playerNeedsRename(playerId: string): boolean {
+    public playerActiveRename(playerId: string): IActiveModActionCacheEntry | null {
         const playerActions = this.getPlayerActions(playerId);
         if (!playerActions) {
-            return false;
+            return null;
```
AMMayberry1: nit: if we're now returning null in this case, we can remove this if block completely and just do:

```ts
return playerActions?.get(ModActionType.Rename);
```

AMMayberry1 (approved):

PR #2344, opened 2026-04-09
Arms Deal & Fearless Attack

VJubert:

### DIFF
--- a/server/game/cards/07_TS26/events/FearlessAttack.ts
+++ b/server/game/cards/07_TS26/events/FearlessAttack.ts
@@ -1,0 +1,26 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import { EventCard } from '../../../core/card/EventCard';
+import type { IEventAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+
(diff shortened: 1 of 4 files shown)


server/game/cards/07_TS26/events/FearlessAttack.ts
```diff
@@ -0,0 +1,26 @@
...
+        };
+    }
+
+    public override setupCardAbilities (registrar: IEventAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.setEventAbility({
+            title: 'Attack with a unit. It gets +1/+0 for this attack for each unit controlled by the defending player.',
+            initiateAttack: {
+                attackerLastingEffects: (context) => ({
+                    effect: abilityHelper.ongoingEffects.modifyStats({
+                        power: context.player.opponent.getArenaUnits().length,
```
moyerr: I think this actually is straightforward to make TS-accurate, even if it doesn't matter for us right now. `attackerLastingEffects` property factory has a second parameter (after `context`) for the attack itself, so you can do `attack.getDefendingPlayer().getArenaUnits().length`

moyerr (approved):

PR #2347, opened 2026-04-09
Luke Skywalker & Emperor Palpatine

VJubert:

### DIFF
--- a/server/game/cards/08_ASH/leaders/LukeSkywalkerICanSaveHim.ts
+++ b/server/game/cards/08_ASH/leaders/LukeSkywalkerICanSaveHim.ts
@@ -4,8 +4,13 @@ import type {
     ILeaderUnitLeaderSideAbilityRegistrar
 } from '../../../core/card/AbilityRegistrationInterfaces';
 import { LeaderUnitCard } from '../../../core/card/LeaderUnitCard';
+import * as AttackHelpers from '../../../core/attack/AttackHelpers';
+import type { StateWatcherRegistrar } from '../../../core/stateWatcher/StateWatcherRegistrar';
+import type { UnitsDefeatedThisPhaseWatcher } from '../../../stateWatchers/UnitsDefeatedThisPhaseWatcher';
 
 export default class LukeSkywalkerICanSaveHim extends LeaderUnitCard {
+    private unitsDefeatedThisPhaseWatcher: UnitsDefeatedThisPhaseWatcher;
+
     protected override getImplementationId() {
         return {
             id: 'luke-skywalker#i-can-save-him-id',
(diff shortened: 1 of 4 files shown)


server/game/cards/08_ASH/leaders/LukeSkywalkerICanSaveHim.ts
```diff
@@ -0,0 +1,46 @@
...
+    protected override getImplementationId() {
+        return {
+            id: 'luke-skywalker#i-can-save-him-id',
+            internalName: 'luke-skywalker#i-can-save-him',
+        };
+    }
+
+    protected override setupLeaderSideAbilities(registrar: ILeaderUnitLeaderSideAbilityRegistrar, abilityHelper: IAbilityHelper): void {
+        registrar.addTriggeredAbility({
+            title: 'Exhaust your leader. If you do, heal 1 damage from that unit',
```
moyerr: I think Thrawn is a good comparison for this leader ability for a couple reasons:
- We can name the leader himself in the ability title, and use the format "Exhaust Luke Skywalker to heal 1 damage from that unit"
- We can actually add a `contextTitle` to call out the name of the unit that attacked

server/game/cards/08_ASH/leaders/LukeSkywalkerICanSaveHim.ts
```diff
@@ -0,0 +1,46 @@
...
+            when: {
+                onAttackEnd: (event, context) =>
+                    event.attack.attacker.controller === context.player
+            },
+            optional: true,
+            immediateEffect: abilityHelper.immediateEffects.exhaust(),
+            ifYouDo: (ifYouDoContext) => ({
+                title: 'Heal 1 damage from that unit',
```
moyerr: We can name the unit here since we have a context

test/server/cards/08_ASH/leaders/EmperorPalpatineAccordingToMyDesign.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+                    }
+                });
+
+                const { context } = contextRef;
+
+                // No exhausted friendly units, so clicking the leader will show a confirmation prompt
+                context.player1.clickCard(context.emperorPalpatine);
+
+                // When no valid targets, a confirmation prompt appears asking if the player wants to proceed
+                expect(context.player1).toHavePrompt('The ability "Give Advantage tokens to an exhausted friendly unit for each other friendly unit" will have no effect. Are you sure you want to use it?');
```
moyerr: We have `toHaveNoEffectAbilityPrompt` for this

test/server/cards/08_ASH/leaders/LukeSkywalkerICanSaveHim.spec.ts
```diff
@@ -0,0 +1,197 @@
+describe('Luke Skywalker, I Can Save Him', function() {
+    integration(function(contextRef) {
+        describe('Leader side triggered ability', function() {
```
moyerr, line 3: Leader side should have a case where the attacking unit dies. I would expect the ability to fizzle (not trigger/prompt) since the unit is no longer a valid target for the effect?

test/server/cards/08_ASH/leaders/LukeSkywalkerICanSaveHim.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+            });
+
+            it('heals 2 damage from itself when attacking (trigger multiple times)', function() {
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.atst);
+                context.player1.clickCard(context.p2Base);
+
+                expect(context.player1).toBeAbleToSelectExactly([context.atst, context.p1Base]);
+                context.player1.clickCard(context.p1Base);
```
moyerr, line 185: Can we put the expectation here where the ability resolves? That way we know no other affect provided the healing.

test/server/cards/08_ASH/leaders/LukeSkywalkerICanSaveHim.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+
+                context.player2.passAction();
+
+                context.player1.clickCard(context.battlefieldMarine);
+                context.player1.clickCard(context.p2Base);
+
+                expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine, context.p1Base]);
+                context.player1.clickCard(context.p1Base);
```
moyerr, line 194: Same here

server/game/cards/08_ASH/leaders/LukeSkywalkerICanSaveHim.ts
```diff
@@ -4,26 +4,40 @@ import type {
...
                 onAttackEnd: (event, context) =>
                     event.attack.attacker.controller === context.player
             },
             optional: true,
-            immediateEffect: abilityHelper.immediateEffects.exhaust(),
+            immediateEffect: abilityHelper.immediateEffects.conditional({
+                condition: (context) =>
+                    AttackHelpers.attackerSurvived(context.event.attack, this.unitsDefeatedThisPhaseWatcher),
+                onTrue: abilityHelper.immediateEffects.exhaust(),
+            }),
```
moyerr: Okay, I guess my previous comment about the whole ability fizzling is incorrect. Luke should be able to exhaust here, but then there should be no effect for the `ifYouDo`, right?

VJubert: Luke should be able to exhaust but ability should do nothing. I didn't test ifYouDo but this is not the first time we cheat a bit for activation, SOR Boba was in the same case

moyerr (approved):

PR #2348, opened 2026-04-10
Executor

VJubert:


server/game/cards/08_ASH/units/ExecutorFinalDestructionOfTheAlliance.ts
```diff
@@ -0,0 +1,35 @@
...
+        };
+    }
+
+    public override setupCardAbilities (registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addConstantAbility({
+            title: 'This unit gets +1/+0 for each upgrade on other friendly units',
+            ongoingEffect: abilityHelper.ongoingEffects.modifyStats((target) => {
+                const upgradeCountOnOtherFriendlyUnits = target.controller.getArenaUpgrades({
+                    condition: (card) => card.parentCard !== target,
+                }).length;
```
moyerr: Unfortunately, I think this only counts upgrades _controlled_ by the player. So this would incorrectly count upgrades controlled by the player on enemy units. Similarly, it would _not_ count upgrades on friendly units controlled by the opponent.

Can we update the strategy here, and also add tests for these scenarios?

moyerr (approved):

PR #2351, opened 2026-04-11
Support keyword and ASH Ezra Bridger unit

moyerr:
| ⟡ Ezra Bridger |
| :---: |
| The Force Is All I Need |
|  |


server/game/cards/08_ASH/units/EzraBridgerTheForceIsAllINeed.ts
```diff
@@ -0,0 +1,30 @@
...
+        return {
+            id: 'ezra-bridger#the-force-is-all-i-need-id',
+            internalName: 'ezra-bridger#the-force-is-all-i-need',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper): void {
+        registrar.addOnAttackAbility({
+            title: 'Give a unit -3/-0 for the phase',
+            optional: true,
```
VJubert, line 17: To reduce prompt I thing you should do targetResolver with conditional inside immediateAbility. The result should be pretty close but with a choose nothing button instead of a first prompt

server/game/cards/04_JTL/units/TheGhostHeartOfTheFamily.ts
```diff
@@ -15,11 +15,9 @@ export default class TheGhostHeartOfTheFamily extends NonLeaderUnitCard {
         registrar.addConstantAbility({
             title: 'Each other friendly Spectre unit gains this unit\'s keywords',
             matchTarget: (card, context) => card !== context.source && card.isUnit() && card.hasSomeTrait(Trait.Spectre),
-            ongoingEffect: AbilityHelper.ongoingEffects.gainKeywords((target, context) => {
-                const blankedKeywords = context.source.getOngoingEffectValues(EffectName.LoseKeyword);
-                return context.source.getOngoingEffectValues(EffectName.GainKeyword)
-                    .filter((keywordProps) => !blankedKeywords.includes(keywordProps.keyword));
-            })
+            ongoingEffect: AbilityHelper.ongoingEffects.gainKeywords((_, context) =>
```
AMMayberry1, line 18: What's the reason for the change here?

moyerr: Yeah, that's a valid question. Essentially the original implementation was a bit of a hack, but it wasn't really clear that it was hacky, or _why_ we needed the hack in the first place. But now that we can get the `KeywordNameOrProperties` directly from a card's `keywords`, we no longer need workarounds. 

Also, this is still the only card impl that shares one unit's keywords with other units, and I'd like for it to be the best example of how to implement this type of effect.

AMMayberry1 (approved):

PR #2357, opened 2026-04-14
Blade of Talzin & Han Solo

VJubert:


server/game/cards/08_ASH/units/HanSoloItllWork.ts
```diff
@@ -0,0 +1,27 @@
...
+        registrar.addWhenPlayedAbility({
+            title: 'Deal 3 damage to this unit. Give 3 Advantage tokens to a unit',
+            targetResolver: {
+                activePromptTitle: 'Give 3 Advantage tokens to a unit',
+                cardTypeFilter: WildcardCardType.Unit,
+                immediateEffect: abilityHelper.immediateEffects.simultaneous([
+                    abilityHelper.immediateEffects.damage((context) => ({ amount: 3, target: context.source })),
+                    abilityHelper.immediateEffects.giveAdvantage({ amount: 3 })
+                ])
+            }
```
moyerr: I think technically only one of these abilities depends on the target resolver. Game-state wise, I can't think of a scenario where it would matter right now, but I do think it would be more correct to put the damage in a `simultaneous` system with a `selectCard` system to choose a target for the tokens.

moyerr (approved):

PR #2362, opened 2026-04-16
Dooku's Solar Sailer

VJubert:


test/server/cards/07_TS26/units/DookusSolarSailerBeautifulAndExpensive.spec.ts
```diff
@@ -0,0 +1,143 @@
+describe('Dooku\'s Solar Sailer, Beautiful and Expensive', function() {
+    integration(function(contextRef) {
+        describe('Dooku\'s Solar Sailer\'s when played ability', function() {
```
moyerr, line 3: We should probably verify that it works when the opponent's base is the one healed too.

And since this is a new watcher, let's also verify that the ability doesn't work if a base was healed in the _previous_ phase, but not in this one.

moyerr (approved):

PR #2363, opened 2026-04-16
remove mockdata for twin suns

VJubert:


scripts/fetchdata.js
```diff
@@ -137,6 +137,10 @@ function populateMissingData(attributes, id) {
         case '2157679168':
             attributes.title = 'Zeb Orrelios'; // Fix spelling
             break;
+        case '9349017358':
+            attributes.title = 'C-3PO';
+            attributes.internalName = 'c3po#die-jedi-dogs';
```
moyerr: I don't think it's necessary to add the `internalName` here. Internal name is computed and added after this step, so it will have the correct spelling at that point

scripts/mockdata.js
```diff
@@ -1,711 +1,5 @@
...
-        internalName: 'jedi-general',
-    }),
-    buildMockCard({
-        title: 'Secret Marriage',
-        cost: 2,
-        hasNonKeywordAbility: true,
-        aspects: ['Vigilance'],
-        types: ['event'],
-        traits: ['plan'],
-        keywords: ['plot'],
```
moyerr, line 504: Did we check that this still come through the data with plot? All the SEC cards were missing it, so we had to add it manually in `fetchdata.js`

moyerr (approved):

PR #2366, opened 2026-04-18
Add Keyboard Shortcuts to User Preferences

sethhelman:
Adding code to help support saving new user preferences for keyboard shortcuts (help please IDK if these actually do what we want them to do) FE PR: https://github.com/SWU-Karabast/forceteki-client/pull/667


server/utils/user/UserFactory.ts
```diff
@@ -27,6 +27,10 @@ const getDefaultCosmeticsPreferences = () => ({
 export const getDefaultPreferences = (): IUserPreferences => ({
     sound: getDefaultSoundPreferences(),
     cosmetics: getDefaultCosmeticsPreferences(),
+    keyboardShortcuts: {
+        passTurn: 'SPACE',
+        undo: 'U',
```
CheBato: Would it be possible to set the keyboardShortcuts here the same way we do with getDefaultCosmeticsPreferences() and getDefaultSoundPreferences()? The reason we do this is so we have a function available if we need the default Preferences anywhere else and to make it easier to edit in one spot.

CheBato (changes requested):

CheBato (approved):

PR #2387, opened 2026-05-10
Opt-in opponent-archetype matchmaking filter (BE)

pmossman:
# Opt-in opponent-archetype matchmaking filter (BE)

Adds an opt-in matchmaking-queue filter so a player can restrict the public-queue opponents they're willing to face by leader, leader+aspect, or leader+specific-base archetype. Default behavior is unchanged: players without preferences match against anyone.

Pairs with SWU-Karabast/forceteki-client#674 (Manage page + queue-form panel).

## What's added

**`server/gamenode/MatchmakingRules.ts`** — new `MatchmakingRule.leaderArchetypeFilter()` factory + `LeaderArchetypeFilterRule` class. A match passes the rule iff each player's filter accepts the other's leader+base. Players with `matchPreferences` undefined / disabled / empty accept anyone.

Wire format:

```ts
type BaseConstraint =
    | { kind: 'aspect'; aspect: Aspect }
    | { kind: 'baseType'; baseIds: string[]; label?: string };

interface OpponentArchetype {
    leaderId: string;
    baseConstraint?: BaseConstraint;  // omit for "any base of this leader"
    enabled?: boolean;                 // defaults true
}

interface MatchPreferences {
    enabled: boolean;
    allowedArchetypes: OpponentArchetype[];
}
```

`baseType.baseIds` is the inline list of acceptable card ids, so the BE doesn't need a separate base-type registry — the FE resolves a user-friendly category (e.g. "Aggression - Force") to ids and sends them on the wire.

**`server/gamenode/QueueHandler.ts`** — adds `matchPreferences` and `baseAspects` (pre-resolved at queue-entry time) to `QueuedPlayerToAdd`; both flow through `IMatchmakingPlayerEntry` to the rule. Empty/unset values preserve the current "match anyone" behavior.

**`scripts/fetchdata.js`** — emits a `_baseTypes.json` artifact alongside the existing card data. Each entry groups bases by `(aspect, hp, traits)` so the FE picker can offer "Aggression - Force - 28hp" / "Aggression - Vanilla - 30hp" / etc. as functional types, plus one entry per unique-named rare base. Per-base `rarity` is captured so the FE can mark rare bases.

## Tests

`test/server/gamenode/MatchmakingRules.spec.ts` — 22 Jasmine specs covering:

- Both players unfiltered → any match passes (default behavior)
- One player filtered, other unfiltered → only the filtered player's allowlist applies
- Leader-only constraint
- Leader + aspect base constraint
- Leader + specific baseType (single id, multiple ids in a group)
- Per-archetype `enabled: false` excluded from the active set
- Empty `allowedArchetypes` or all-disabled → matches anyone
- Missing leader/base on opponent's deck → no match

All 6310 backend specs pass.

## Notes for review

- Wire format is forward-compatible: a future client that knows nothing about `baseConstraint` still works (omit it for leader-only filters), and a future client that adds new constraint kinds can extend the union.
- The `baseAspects` precomputation in `QueueHandler` avoids doing a card-data lookup at match-time.
- No matchmaker timeout / fallback behavior added — that's a separate consideration if narrow filters create long queue waits.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,7 +4,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "forceteki",
       "dependencies": {
         "@aws-sdk/client-dynamodb": "^3.758.0",
         "@aws-sdk/client-s3": "^3.933.0",
--- a/scripts/fetchdata.js
+++ b/scripts/fetchdata.js
@@ -553,9 +553,11 @@ function labelForGroup(group) {
     const hp = group.hp ? `${group.hp}hp` : '';
     const text = group.text || '';
     if (text === '') {
-        // Vanilla bases — no rules text. Tag is just the HP (e.g.,
-        // "Aggression - 30hp"), since vanilla bases differ only in stats.
-        return `${aspect} - ${hp}`;
+        // Vanilla bases — no rules text. The tag is "Vanilla" so when an
+        // aspect icon is rendered alongside the label and the leading
+        // aspect word is stripped, the FE still has a meaningful descriptor
+        // (e.g. "Vanilla - 30hp") rather than just "30hp".
+        return `${aspect} - Vanilla - ${hp}`;
     }
     if ((/force token/i).test(text) || (/force is with you/i).test(text)) {
         return `${aspect} - Force - ${hp}`;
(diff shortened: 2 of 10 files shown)


pmossman:
Moving to a fork-internal PR for review iteration before re-opening against upstream.

PR #2390, opened 2026-05-11
Backend support for 1v1 "Faux Suns"

adbronico:
Changes include:
- Two-leader support for 1v1 Twin Suns
- Blast and Plan* token
- New tests

*still some edge cases WIP


server/utils/deck/DeckValidator.ts
```diff
@@ -199,11 +206,11 @@ export class DeckValidator {
...
     public getMaxSideboardSize(format: SwuGameFormat, cardPool: CardPool): number {
-        // Sideboard is only restricted in Premier and Eternal. We relax the restriction in Next Set mode.
-        if (format === SwuGameFormat.Open || format === SwuGameFormat.Limited || cardPool === CardPool.NextSet) {
-            return -1;
+        // Sideboard is only restricted in Premier and Eternal. We relax the restriction in other modes.
+        if (format === SwuGameFormat.Premier || format === SwuGameFormat.Eternal) {
+            return 10;
         }
-        return 10;
+        return -1;
```
moyerr: Just want to flag that this appears to be a regression. Sideboard is not always enforced for Premier/Eternal. It should be relaxed if those modes are using the `NextSet` card pool

server/game/core/Game.ts
```diff
@@ -1327,6 +1363,86 @@ export class Game extends EventEmitter {
         this.resolveGameState();
     }
 
+    public claimPlanToken(player: Player): void {
```
moyerr: It would be great if both Plan and Blast abilities were encapsulated in their own game systems.

server/game/core/Game.ts
```diff
@@ -145,6 +148,29 @@ export class Game extends EventEmitter {
...
+        return this.state.isBlastTokenClaimed;
+    }
+
+    public set isBlastTokenClaimed(value: boolean) {
+        this.state.isBlastTokenClaimed = value;
+    }
+
+    public allClaimTokensClaimed(): boolean {
+        if (this.format === SwuGameFormat.FauxSuns) {
+            return this.isInitiativeClaimed && this.isPlanTokenClaimed && this.isBlastTokenClaimed;
```
AMMayberry1: This condition is incorrect - since there's only two players, all three tokens will never be claimed

server/game/core/gameSteps/ActionWindow.ts
```diff
@@ -195,13 +211,23 @@ export class ActionWindow extends UiPrompt {
             this.game.addMessage('{0} passes', this.activePlayer);
         }
 
+        const allTokensClaimed = this.game.allClaimTokensClaimed();
+
         if (this.prevPlayerPassed) {
-            // in the (unusual) case that both players pass without claiming initiative, phase ends and initiative stays where it is
-            this.activePlayer.passedActionPhase = true;
-            this.activePlayer.opponent.passedActionPhase = true;
+            if (allTokensClaimed) {
```
AMMayberry1: This logic is definitely incorrect, `prevPlayerPassed` specifically means that the opponent just took the Pass action in which case it's not possible that they claimed

server/game/core/gameSteps/ActionWindow.ts
```diff
@@ -195,13 +211,23 @@ export class ActionWindow extends UiPrompt {
...
+                this.activePlayer.opponent.passedActionPhase = true;
+            } else {
+                // Tokens are still unclaimed; treat this pass as a token-claim pass and continue the phase
+                this.setPassStatus(true);
+            }
         } else if (this.activePlayer.opponent.passedActionPhase) {
-            // if opponent already claimed initiative, we're done
-            this.activePlayer.passedActionPhase = true;
+            if (allTokensClaimed) {
+                // Opponent has already passed (e.g. claimed initiative) and all tokens are claimed → active player is done too
```
AMMayberry1: Also here. Even apart from the issue that only two tokens total can be claimed, I'm not sure what this is intended to handle

server/utils/deck/DeckValidator.ts
```diff
@@ -54,11 +55,16 @@ export class DeckValidator {
...
+
     private static parseSets(cardData: ICardDataJson): SwuSetId[] {
-        if (cardData.setCodes) {
-            return cardData.setCodes.map((code) => EnumHelpers.checkConvertToEnum(code.set, SwuSetId)[0]);
-        }
-        return [EnumHelpers.checkConvertToEnum(cardData.setId.set, SwuSetId)[0]];
+        const setCodes = cardData.setCodes ?? [cardData.setId];
+        // Use tryConvertToEnum so cards with unrecognized set codes (e.g. test fixtures) produce an
+        // empty array rather than throwing. An empty array means the card matches no legal set and
+        // will always fail checkFormatLegality, which is the correct behaviour.
```
AMMayberry1, line 67: What's the reason for this behavior change - is there a reason we want to start silently dropping invalid cards instead of throwing an error?

server/game/core/utils/EnumHelpers.ts
```diff
@@ -22,6 +22,19 @@ function getEnumLookupMap<T extends object>(enumObj: T): Map<string, T[keyof T]>
 }
 
 export namespace EnumHelpers {
+    // convert a set of strings to an enum type, silently dropping values not present in the enum
+    export function tryConvertToEnum<T extends object>(values: string | string[], enumObj: T): T[keyof T][] {
```
AMMayberry1, line 26: See comment on the deck validation changes - if my logic in that comment makes sense, we don't have a use case for  this

server/game/core/zone/BaseZone.ts
```diff
@@ -54,35 +58,36 @@ export class BaseZone extends ZoneAbstract<IBaseZoneCard> {
...
+            leader.initializeZone(this);
+        }
     }
 
     public setLeader(leader: ILeaderCard) {
         Contract.assertEqual(leader.controller, this.owner, `Attempting to add card ${leader.internalName} to ${this} as leader but its controller is ${leader.controller}`);
-        Contract.assertIsNullLike(this._leader, `Attempting to add leader ${leader.internalName} to ${this} but a leader is already there`);
 
-        this._leader = leader;
+        this._leaders = [...this._leaders, leader];
```
AMMayberry1, line 84: nit: much better to do `this._leaders.add(leader)` since it avoids creating a temp array that has to be immediately garbage collected

server/game/core/Game.ts
```diff
@@ -145,6 +148,29 @@ export class Game extends EventEmitter {
...
+        this.state.isPlanTokenClaimed = value;
+    }
+
+    public get isBlastTokenClaimed() {
+        return this.state.isBlastTokenClaimed;
+    }
+
+    public set isBlastTokenClaimed(value: boolean) {
+        this.state.isBlastTokenClaimed = value;
+    }
```
AMMayberry1, line 169: nit: we have some decorators now that can be used so that we no longer have to explicitly write out these get / set operations. I don't remember the details offhand but copilot should be able to figure it out

server/game/core/Game.ts
```diff
@@ -1324,6 +1354,29 @@ export class Game extends EventEmitter {
         this.resolveGameState();
     }
 
+    public claimPlanToken(player: Player): void {
+        this.isPlanTokenClaimed = true;
+        player.passedActionPhase = true;
+
+        new ClaimPlanTokenSystem({}).resolve(
```
AMMayberry1: This is a little bit backwards - typically the system is what sets the state on the Game object as part of its resolution.

So like if we were to fix the initiative token to follow this pattern, there would be a `ClaimInitiativeSystem` which is triggered by the button press in `ActionWindow` and then as part of its `eventHandler` it would call the `claimInitiative()` method on Game which would just look like this:

```ts
    public claimInitiative(player: Player): void {
        this.initiativePlayer = player;
        this.isInitiativeClaimed = true;
        player.passedActionPhase = true;
    }
```

server/game/core/Player.ts
```diff
@@ -136,6 +136,15 @@ export class Player extends GameObject implements IGameStatisticsTrackable {
         return this._leader;
     }
 
+    @stateRef() private accessor _secondLeader: ILeaderCard | null = null;
+    public get secondLeader(): ILeaderCard | null {
```
AMMayberry1: I don't love this leader vs secondLeader distinction since there's not an obvious logical way to separate the two leaders and declare which one is "second". We'll probably just need to do what we did with TWI Maul and the attack targets - turn the accessor into an array get and make it so that everything can handle the case of it being an array.

server/game/gameSystems/ClaimPlanTokenSystem.ts
```diff
@@ -0,0 +1,56 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    public override eventHandler(event: any): void {
+        const player = event.player as Player;
+        const context = event.context;
+        const game = context.game;
+
+        // We use PassesTriggersToParentWindow so that triggered abilities (Rey, Seasoned Fleet Admiral, etc.)
+        // are collected by the outer OnPlanTokenClaimed window and fire AFTER the put-on-bottom prompt —
+        // i.e. the full Plan token effect resolves before any triggers respond.
+        new DrawSystem({ amount: 1 }).resolve(player, context, TriggerHandlingMode.PassesTriggersToParentWindow);
```
AMMayberry1: This should be handled as a contingent event

server/game/gameSystems/ClaimPlanTokenSystem.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        // Queue the put-on-bottom prompt as a simple step so it runs after the draw (and any
+        // contingent empty-deck damage) has fully resolved.
+        game.queueSimpleStep(() => {
```
AMMayberry1: Pretty sure this also should be handled as a contingent event

server/gamenode/Lobby.ts
```diff
@@ -1373,7 +1373,7 @@ export class Lobby {
             id: uuidv4(),
             allowSpectators: false,
             owner: 'Order66',
-            gameMode: GameMode.Premier,
+            gameMode: GameMode.OneVsOne,
```
AMMayberry1: nit: what's the reason for the name change here?

server/utils/deck/DeckValidator.ts
```diff
@@ -328,6 +346,28 @@ export class DeckValidator {
                 this.checkFormatLegality(leaderData, format, legalSets, failures);
             }
 
+            // Validate second leader (TwinSuns formats only).
+            const rules = formatRules.get(format);
+            if (rules?.leaderCount === 2) {
```
AMMayberry1, line 380: Feels like there ought to be an `else` branch that's validating that the non-TS formats have exactly one leader

test/server/cards/06_SEC/units/DarthTrayaLordOfBetrayal.spec.ts
```diff
@@ -129,5 +130,82 @@ describe('Darth Traya, Lord of Betrayal', function() {
...
+                    phase: 'action',
+                    format: 'fauxSuns',
+                    player1: {
+                        leader: { card: 'grand-inquisitor#hunting-the-jedi', exhausted: true },
+                        secondLeader: { card: 'saw-gerrera#bring-down-the-empire', exhausted: true },
+                        groundArena: ['darth-traya#lord-of-betrayal'],
+                        base: 'kestro-city',
+                    },
+                    player2: {
+                        leader: { card: 'yoda#sensing-darkness', exhausted: true },
```
AMMayberry1, line 146: Just noticed that p2 only has one leader here. We'll need to update the schema for the TS format to require that both players have two leaders

server/game/core/zone/BaseZone.ts
adbronico: leaders are an array not a set so i'm not sure if we can do that. Even using array's .push would bypass the state setting logic I think? Ditto for removing using filter vs. splice

server/game/core/Game.ts
adbronico: Claude says: Game doesn't extend GameObjectBase, so the @statePrimitive/@stateRef decorators aren't available here. Its state is backed by a plain IGameState object, so explicit get/set passthroughs to this.state are the correct pattern in this class.

I don't know much about the state stuff so I'm not sure what to even look into to adapt Game to use that?

server/utils/deck/DeckValidator.ts
adbronico: Yeah becaause I had to create a "test-only" set that needed removal. Cards without a set ID would also fail validation

server/game/core/utils/EnumHelpers.ts
adbronico: See above

server/game/core/gameSteps/phases/ActionPhase.ts
```diff
@@ -95,6 +95,8 @@ export class ActionPhase extends Phase {
             player.cleanupFromActionPhase();
         }
         this.game.isInitiativeClaimed = false;
+        this.game.isPlanCounterClaimed = false;
+        this.game.isBlastCounterClaimed = false;
```
AMMayberry1: Let's gate these on a check that the game mode is TS, to make sure that nothing is referring to them in non-TS game modes. That's pretty easy to do, just add a property getter that throws if the game mode isn't TS

server/game/core/gameSteps/ActionWindow.ts
```diff
@@ -137,13 +137,22 @@ export class ActionWindow extends UiPrompt {
 
     public override activePromptInternal(player: Player): IPlayerPromptStateProperties {
         const { mustTakeCardAction, overrideActionPromptTitle } = this.getSelectableCards();
+        const isFauxSuns = this.game.format === SwuGameFormat.FauxSuns;
+        const mustClaimToken = isFauxSuns && !this.game.allClaimTokensClaimed();
 
         const buttons: IButton[] = [
-            { text: 'Pass', arg: 'pass', disabled: mustTakeCardAction },
+            { text: 'Pass', arg: 'pass', disabled: mustTakeCardAction || mustClaimToken },
```
AMMayberry1, line 144: Is this saying that you're not allowed to pass your action if there are available tokens to claim? That seems wrong

server/game/core/snapshot/SnapshotInterfaces.ts
```diff
@@ -139,6 +139,8 @@ export interface IGameState {
     initiativePlayer: GameObjectId<Player> | null;
     actionPhaseActivePlayer: GameObjectId<Player> | null;
     isInitiativeClaimed: boolean;
+    isPlanCounterClaimed: boolean;
+    isBlastCounterClaimed: boolean;
```
AMMayberry1: Let's make these optional with `?`

server/game/core/Game.ts
```diff
@@ -151,6 +154,33 @@ export class Game extends EventEmitter {
         this.state.isInitiativeClaimed = value;
     }
 
+    public get isPlanCounterClaimed() {
+        return this.state.isPlanCounterClaimed;
```
AMMayberry1, line 157: Let's gate these on a check that the game mode is TS, to make sure that nothing is reading / writing to them in non-TS game modes. That's pretty easy to do, just add a check on the getters / setters that throws if the game mode isn't TS

server/game/core/Game.ts
```diff
@@ -1357,13 +1389,28 @@ export class Game extends EventEmitter {
...
+        new ClaimInitiativeSystem({}).resolve(
+            player,
+            this.getFrameworkContext(player),
+            TriggerHandlingMode.ResolvesTriggers
+        );
+    }
 
-        // update game state for the sake of constant abilities that check initiative
-        this.resolveGameState();
+    public claimPlanCounter(player: Player): void {
```
AMMayberry1: In all of the places where we now have three signatures, could we just collapse down to one that accepts an enum for the token type

server/game/gameSystems/ClaimBlastCounterSystem.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    // TSTODO: damage all opponents' bases
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    protected override updateEvent(event: any, player: Player, context: TContext, additionalProperties: Partial<IClaimBlastCounterProperties>): void {
+        super.updateEvent(event, player, context, additionalProperties);
+
+        // The base damage fires as a contingent of OnBlastCounterClaimed, so prevention effects
+        // (e.g. Close the Shield Gate) and triggers (e.g. Boba Fett) see it as a distinct damage event.
+        event.setContingentEventsGenerator(() => [
+            new DamageSystem({ type: DamageType.Ability, amount: 1 }).generateRetargetedEvent(player.opponent.base, context),
```
AMMayberry1: These should be using `generateEvent` instead of `generateRetargetedEvent`

server/game/gameSystems/ClaimInitiativeSystem.ts
```diff
@@ -0,0 +1,27 @@
...
+    public override eventHandler(event: any): void {
+        const player = event.player as Player;
+        const game = event.context.game;
+
+        game.initiativePlayer = player;
+        game.isInitiativeClaimed = true;
+        player.passedActionPhase = true;
+
+        // Update game state for the sake of constant abilities that check initiative.
+        game.resolveGameState();
```
AMMayberry1: This could cause issues, we shouldn't be force-resolving game state inside of events

server/game/gameSystems/ClaimPlanCounterPutOnBottomSystem.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Prompts the target player to choose a card from their hand and put it on the
+ * bottom of their deck. Used as a contingent event of {@link ClaimPlanCounterSystem}.
+ * If the player's hand is empty the prompt is skipped.
+ */
+export class PutOnBottomFromHandSystem<TContext extends AbilityContext = AbilityContext>
+    extends PlayerTargetSystem<TContext, IPutOnBottomFromHandProperties> {
+    public override readonly name = 'putOnBottomFromHand';
+    public override readonly eventName = EventName.OnPlanCounterPutOnBottom;
```
AMMayberry1, line 29: If this is going to be welded to the plan counter then let's also name the class that way

AMMayberry1:
Here's some stuff that came out of claude review, I think all of these are valid:

**Major**
1. Duplicate leaders pass deck validation — confirmed empirically

Rules 12.2.2 and 12.3.1 require two different leaders that aren't copies of each other. I built a FauxSuns deck with the same leader in both slots and validateInternalDeck returned {} — zero failures. The game then happily builds both.

checkMaxCopiesOfCard only iterates deck.deck + sideboard, so the maxCardCopies: 1 rule never reaches the leader slots, and the new second-leader branch checks aspects but never compares the two leader ids. This also creates two copies of a unique card in play, breaking the uniqueness rule once both deploy. The fix belongs next to checkTwinSunsLeaderAspectConflict.

Related: MissingSecondLeader / MixedAlignmentLeaders are correctly not filtered by filterOutSideboardingErrors so they block at import, but they're absent from Lobby.startBlockingDeckSizeFailures, so game-start re-validation won't catch them.

2. Player.deckLeader silently truncates

deckLeader now returns _deckLeaders[0] with no assertion. getSingleLeader() has the assert but is used only by tests. This is the single biggest long-term correctness risk: every future caller that reaches for deckLeader is a silent Twin Suns bug with no test or type signal. Two live callers already are — SwuStatsHandler and SwuBaseHandler will record only the first leader. I'd delete deckLeader and force callers to pick getAllDeckLeaders() or getSingleLeader().

3. BaseZone.leader has a lying type

The getter changed from ILeaderCard | null to non-nullable ILeaderCard, but returns _leaders[0], which is undefined once leaders deploy. I confirmed baseZone.leaders.length === 0 with both deployed. It currently has zero callers, so either delete it or type it honestly — as written it's a crash waiting for its first consumer.

4. setLeader lost its duplicate guard (affects standard format)

setLeader dropped Contract.assertIsNullLike(this._leader, ...) for an unconditional append. Any double-add now silently produces a duplicated leader — wrong count, duplicate entry in cards — instead of failing loudly. An assertFalse(this._leaders.includes(leader)) restores the invariant without constraining Twin Suns.

**Moderate**
5. allClaimTokensClaimed() is misnamed and hardcodes 2 players. It returns claimedCount >= 2, which is right for 1v1 but reads as the opposite of what it computes. The actual rule (12.6.1.A) is "you may pass only if no counter is available or you already took one". Deriving the threshold from getPlayers().length and naming it for the rule would make 3–4 player support a data change.

6. Claim state is now applied asynchronously. claimInitiative used to set initiativePlayer / isInitiativeClaimed / passedActionPhase synchronously; those now land in the system's event handler on a queued step, while ActionWindow.pass(false) runs immediately after and reads opponent.passedActionPhase. rotateActivePlayer() reads it too. It works today, but it's an undocumented ordering dependency on an already-subtle pipeline.

7. PutOnBottomFromHandSystem hand-rolls prompting and movement. It calls game.promptForSelect inside eventHandler and then target.moveTo(...) directly, bypassing MoveCardSystem entirely — no move event, no replacement-effect handling. The codebase already has this exact pattern: PrivateManufacturing uses selectCard({ zoneFilter: Hand, controller: Self, immediateEffect: moveToBottomOfDeck() }). Reusing it would delete this file. It also doesn't override canAffect for the empty-hand case.

8. New systems default to targeting the opponent. PlayerTargetSystem.defaultTargets returns [context.player.opponent]. claimBlastCounter / claimPlanCounter are exported in GameSystemLibrary, so a card author calling claimBlastCounter({}) would claim for the opponent — and ClaimBlastCounterSystem then damages player.opponent.base, i.e. the caller's own base. Worth overriding defaultTargets to [context.player].

**Minor**
9. checkTwinSunsLeaderAspectConflict compares raw 'heroism' / 'villainy' strings instead of the Aspect enum (it does correctly use front-side aspects per 12.2.1.A). The dual casing secondleader (SWUDB) vs secondLeader (internal) is a trap — the non-Twin-Suns TooManyLeaders branch only checks the camelCase key, and the test DeckBuilder writes the lowercase one into an IDecklistInternal. Live paths are covered by validateSwuDbDeck, but the acknowledged TODO: audit other deck builders before shipping is real. getSingleLeader() is production-dead code.

server/game/core/gameSteps/ActionWindow.ts
adbronico: That's correct, you can't pass in Twin Suns unless there are no tokens left to claim so I'm mirroring that for Faux Suns

server/game/gameSystems/ClaimBlastCounterSystem.ts
adbronico: This will work for the blast but not for the plan counter

server/game/core/gameSteps/ActionWindow.ts
```diff
@@ -236,9 +236,13 @@ export class ActionWindow extends UiPrompt {
...
     }
 
+    // NOTE: claimInitiative/claimPlan/claimBlast all share this shape — claimCounter(...) only queues the
+    // claim event window (it doesn't set passedActionPhase synchronously), and pass(false) immediately after
+    // relies on that flag already being set by the time it (via Game.rotateActivePlayer()) reads it. See the
+    // ordering note in ClaimCounterSystem.eventHandler for why this is safe today.
     public claimInitiative() {
         this.game.addMessage('{0} claims initiative and passes', this.activePlayer);
-        this.game.claimInitiative(this.activePlayer);
+        this.game.claimCounter(this.activePlayer, ClaimCounterType.Initiative);
```
AMMayberry1: We can also collapse these into a single method that takes a `ClaimCounterType`

server/game/core/Game.ts
```diff
@@ -682,6 +696,9 @@ export class Game extends EventEmitter {
     /**
      * Checks who the next legal active player for the action phase should be and updates activePlayer. If none available, sets it to null.
      */
+    // NOTE: when called via ActionWindow.pass() right after a claimCounter() call (see ActionWindow.claim*()),
+    // this read of passedActionPhase depends on ClaimCounterSystem's event window having already run and set
+    // that flag, even though claimCounter() itself only queues that window rather than running it synchronously.
```
AMMayberry1: Ditto here, don't need this comment

server/game/core/gameSteps/ActionWindow.ts
```diff
@@ -236,9 +236,13 @@ export class ActionWindow extends UiPrompt {
         // }
     }
 
+    // NOTE: claimInitiative/claimPlan/claimBlast all share this shape — claimCounter(...) only queues the
+    // claim event window (it doesn't set passedActionPhase synchronously), and pass(false) immediately after
+    // relies on that flag already being set by the time it (via Game.rotateActivePlayer()) reads it. See the
+    // ordering note in ClaimCounterSystem.eventHandler for why this is safe today.
```
AMMayberry1: We can remove this comment, this is a pretty standard practice throughout the engine

server/game/core/Player.ts
```diff
@@ -137,6 +137,13 @@ export class Player extends GameObject implements IGameStatisticsTrackable {
 
     @stateRefArray() private accessor _deckLeaders: readonly ILeaderCard[] = [];
 
+    /**
+     * Returns only the player's primary (first) deck leader, silently ignoring a second leader if
+     * one is present (e.g. Twin Suns). Only reach for this when the caller genuinely wants "just the
+     * first leader" (e.g. a stats/analytics path that isn't Twin-Suns-aware and can tolerate recording
+     * one leader) — for anything gameplay-facing, use {@link getAllDeckLeaders} or, if the code path is
+     * guaranteed to run only in single-leader formats, {@link getSingleLeader} (which asserts that).
+     */
```
AMMayberry1: This feels like a hack to get around the fact that we haven't sorted out the stats tracking story yet. I'd rather remove this method and just disable stats tracking for TS modes for now.

We can reach out to Oot and Medo and ask about integration, I'm sure they'd both be very interested

server/game/gameSystems/ClaimCounterSystem.ts
```diff
@@ -0,0 +1,98 @@
...
+            case ClaimCounterType.Initiative:
+                game.initiativePlayer = player;
+                game.isInitiativeClaimed = true;
+                break;
+            case ClaimCounterType.Plan:
+                game.isPlanCounterClaimed = true;
+                break;
+            case ClaimCounterType.Blast:
+                game.isBlastCounterClaimed = true;
+                break;
```
AMMayberry1, line 46: Let's a add a default case with `Contract.fail()` just as an escape valve

server/game/gameSystems/ClaimCounterSystem.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        // NOTE ON ORDERING: this handler runs when the queued event window opens, not synchronously when
+        // resolve() is called (see Game.claimCounter). ActionWindow.claim*() calls Game.claimCounter(...)
+        // and then immediately calls pass(false), which — via Game.rotateActivePlayer() — reads this same
+        // player.passedActionPhase flag. That read only sees the value set here because GamePipeline drains
+        // steps queued during the current step (including this event window) before the pipeline advances
+        // past the ActionWindow's own completion. This isn't independently enforced or unit-tested at the
+        // pipeline level — if that step-ordering ever changes, this flag could be read stale.
```
AMMayberry1: Can also remove this comment

server/game/gameSystems/ClaimPlanCounterPutOnBottomSystem.ts
```diff
@@ -0,0 +1,63 @@
...
+ * the put-on-bottom choice happens before any "a card was drawn" triggers fire.
+ */
+export class ClaimPlanCounterPutOnBottomSystem<TContext extends AbilityContext = AbilityContext>
+    extends PlayerTargetSystem<TContext, IClaimPlanCounterPutOnBottomProperties> {
+    public override readonly name = 'claimPlanCounterPutOnBottom';
+    public override readonly eventName = EventName.OnPlanCounterPutOnBottom;
+
+    public override defaultTargets(context: TContext): Player[] {
+        return context.player ? [context.player] : [];
+    }
```
AMMayberry1, line 33: Let's add an override of `queueGenerateEventGameSteps()` that will throw if this created in a non-TS game

AMMayberry1 (approved):

PR #2394, opened 2026-05-14
Add user preference for timer visibility

moyerr:
Client PR: SWU-Karabast/forceteki-client#680

### DIFF
--- a/server/game/core/Player.ts
+++ b/server/game/core/Player.ts
@@ -265,8 +265,8 @@ export class Player extends GameObject implements IGameStatisticsTrackable {
 
     /**
      * Called by ByoyomiTimer when this player's main timer crosses a warning threshold.
-     * Emits a chat alert, but only when at least one player has the timer fully hidden —
-     * players with a visible clock already have their own warning signal.
+     * Emits a chat alert, so that the player and their opponent are aware of the time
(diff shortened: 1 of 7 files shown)


AMMayberry1:
I'll mention this in chat as well, but one concern (I think Caz raised it) is that when the timer is hidden players will have no feedback or warning about the main time running down currently. I might be misremembering but I think we did remove the 20s / 10s chat log warnings.

I do think it's worth adding a little bit of extra UX that will kick in only when the main timer is hidden. Maybe we bring back in the yellow / red clock icon from the old timer design for certain warning intervals and also emit chat logs, but only when the main timer is hidden by at least one of the players?

server/game/core/Player.ts
```diff
@@ -260,6 +263,24 @@ export class Player extends GameObject implements IGameStatisticsTrackable {
           !this.game.isEnded;
     }
 
+    /**
+     * Called by ByoyomiTimer when this player's main timer crosses a warning threshold.
+     * Emits a chat alert, but only when at least one player has the timer fully hidden —
```
AMMayberry1: > only when at least one player has the timer fully hidden

Is this still true?

AMMayberry1 (approved):

PR #2395, opened 2026-05-18
Address Double-Cross Rule Clarification

bwohldmann:
Per the 5/6/2026 rule clarification around Double-Cross; if the lower cost selected unit doesn't actually change controllers (i.e. Rey), no credits should be generated.  


server/game/cards/07_LAW/events/DoubleCross.ts
```diff
@@ -50,15 +50,23 @@ export default class DoubleCross extends EventCard {
...
                     ? lowerCostUnit.controller
-                    : thenContext.events.find((event) => event.card === lowerCostUnit)?.newController;
+                    : lowerCostUnitEvent?.newController;
 
                 return {
                     title: 'The player who takes control of the lower-cost unit creates Credit tokens equal to the difference between those units\' costs.',
                     thenCondition: (_) => costDifference > 0 && creditRecipient,
                     immediateEffect: abilityHelper.immediateEffects.createCreditToken({
-                        amount: costDifference,
+                        amount: effectiveAmount,
```
moyerr, line 69: Should the `thenCondition` now check `effectiveAmount` too? Seems like the whole then block can be skipped if the effective amount is 0

moyerr (approved):

PR #2400, opened 2026-05-19
Chimaera

VJubert:


server/game/cards/08_ASH/units/ChimaeraAFrighteningReality.ts
```diff
@@ -0,0 +1,45 @@
...
+            targetResolvers: {
+                friendlyUnits: {
+                    cardTypeFilter: WildcardCardType.Unit,
+                    controller: RelativePlayer.Self,
+                },
+                enemyUnits: {
+                    cardTypeFilter: WildcardCardType.Unit,
+                    controller: RelativePlayer.Opponent,
+                    cardCondition: (card) => !card.isLeaderUnit(),
+                },
```
moyerr, line 29: I think we have a `.NonLeaderUnit` filter, don't we?

server/game/cards/08_ASH/units/ChimaeraAFrighteningReality.ts
```diff
@@ -0,0 +1,45 @@
...
+            },
+            then: (thenContext) => ({
+                title: 'This unit gets +2/+0 for this attack',
+                immediateEffect: abilityHelper.immediateEffects.defeat(() => {
+                    return { target: [thenContext.targets.friendlyUnits, thenContext.targets.enemyUnits] };
+                })
```
moyerr: Wrong title. Also, we can just have the props for the `defeat` system, we don't need a function `() => {}`

server/game/cards/08_ASH/units/ChimaeraAFrighteningReality.ts
```diff
@@ -0,0 +1,45 @@
...
+            })
+        });
+
+        registrar.addTriggeredAbility({
+            title: 'Heal 2 damage from your base',
+            when: {
+                onCardDefeated: (event, context) => event.card.isUnit() && event.card.controller !== context.player,
+            },
+            immediateEffect: abilityHelper.immediateEffects.heal((context) => ({ amount: 2, target: context.player.base }))
```
moyerr, line 38: I think we need to use LKI here. Check out Bo-Katan, Alone as a reference.

test/server/cards/08_ASH/units/ChimaeraAFrighteningReality.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+
+                const { context } = contextRef;
+
+                context.player1.clickCard(context.chimaera);
+                context.player1.clickCard(context.awing);
+
+                expect(context.player2).toBeActivePlayer();
+                expect(context.awing).toBeInZone('discard');
+                expect(context.sabineWren).toBeInZone('groundArena');
+            });
```
moyerr, line 105: I think this is incorrect. The "if you do" means you _must_ choose a friendly unit _and_ and enemy non-leader unit. Right? So in this case A-Wing would not be defeated since you were unable to choose both.

test/server/cards/08_ASH/units/ChimaeraAFrighteningReality.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+
+                context.player1.clickCard(context.chimaera);
+                context.player1.clickCard(context.awing);
+
+                expect(context.player2).toBeActivePlayer();
+                expect(context.awing).toBeInZone('discard');
+            });
+        });
+
+        describe('Triggered ability', function() {
```
moyerr, line 126: I think we should cover some of the more nuanced cases where LKI would come into play:
- Defeating a piloting upgrade should not trigger because it was not a unit when it was in play
- Defeating a unit the enemy stole _should_ trigger because it it was an enemy unit when in play

test/server/cards/08_ASH/units/ChimaeraAFrighteningReality.spec.ts
```diff
@@ -118,10 +117,9 @@ describe('Chimaera, A Frightening Reality', function() {
                 const { context } = contextRef;
 
                 context.player1.clickCard(context.chimaera);
-                context.player1.clickCard(context.awing);
 
                 expect(context.player2).toBeActivePlayer();
-                expect(context.awing).toBeInZone('discard');
+                expect(context.awing).toBeInZone('spaceArena');
```
moyerr, line 122: Title of this test case need to be updated now that we have this working correctly

moyerr (approved):

PR #2404, opened 2026-05-20
Add stats to get decks response

santiagovazquez:
Work needed for "Sort by Most Played" on the /DeckPage. Context [here](https://github.com/SWU-Karabast/forceteki-client/issues/532).

This PR just adds the stats on the `/get-decks` endpoint so we could sort the list client-side


server/utils/deck/DeckService.ts
```diff
@@ -75,6 +75,7 @@ export class DeckService {
                     // Extract just the inner deck object which matches ILocalStorageDeckData
                     ...deckData.deck,
                     deckID: deckData.id,
+                    stats: deckData.stats,
```
AMMayberry1: Do you know if this is also including the matchup stats? If so that could end up being a lot of extra data, would be good if we can make sure we're just sending the basic win / loss / draw counts over the wire

santiagovazquez: sure, let me make the change

server/utils/deck/DeckService.ts
```diff
@@ -71,11 +71,16 @@ export class DeckService {
...
 
             return decks
-                .map((deckData) => ({
-                    // Extract just the inner deck object which matches ILocalStorageDeckData
-                    ...deckData.deck,
-                    deckID: deckData.id,
-                }));
+                .map((deckData) => {
+                    // just pick results data to avoid sending anything else from the stats object
+                    const stats = deckData.stats ? Util.pick(deckData.stats, 'wins', 'losses', 'draws') : undefined;
```
santiagovazquez, line 76: I used `pick` over `omit` since someone could add new values into the stats object and we'd be including those here

AMMayberry1 (approved):

PR #2407, opened 2026-05-20
Remove support for pre-CR7 attack rules

moyerr:


server/game/core/attack/AttackFlow.ts
```diff
@@ -12,14 +12,12 @@ import type { IAttackableCard } from '../card/CardInterfaces';
 import { Contract } from '../utils/Contract';
 
 export enum AttackRulesVersion {
```
AMMayberry1: nit: I'd vote to just remove this enum completely at this point

moyerr: Yeah, I was split on that, so totally good either way. I'll remove it!

AMMayberry1 (approved):

PR #2418, opened 2026-05-21
Wistling Birds

moyerr:
| Whistling Birds |
| :----: |
|  |

### DIFF
--- a/server/game/core/attack/AttackFlow.ts
+++ b/server/game/core/attack/AttackFlow.ts
@@ -63,7 +63,8 @@ export class AttackFlow extends BaseStepWithPipeline {
 
         // The OnAttackDamageResolved event currently does not carry an `attackerLastKnownInformation`.
         // If a future ability triggers on this event and needs the attacker's pre-damage state,
-        // attach LKI here in the same shape as OnAttackDeclared / OnAttackEnd.
+        // attach LKI here in the same shape as OnAttackDeclared / OnAttackEnd (capture occurs
+        // for OnAttackEnd inside dealDamage just before the damage window opens).
         this.context.game.createEventAndOpenWindow(
             EventName.OnAttackDamageResolved,
             this.context,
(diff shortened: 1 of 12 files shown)


test/server/cards/08_ASH/upgrades/WhistlingBirds.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+                expect(context.p2Base.damage).toBe(5);
+
+                // Whistling Birds triggers — each enemy ground unit takes 2 damage
+                expect(context.wampa.damage).toBe(2);
+                expect(context.mandalorianWarrior.damage).toBe(2);
+
+                // Enemy space unit is unaffected (different arena)
+                expect(context.cartelSpacer.damage).toBe(0);
+
+                expect(context.player2).toBeActivePlayer();
```
VJubert, line 64: can you check if battlefieldMarine has damage ?

VJubert (approved):

server/game/core/event/LastKnownInformation.ts
```diff
@@ -0,0 +1,86 @@
...
+    Contract.fail(`Unexpected card type: ${card.type}`);
+}
+
+/**
+ * Registers a pre-resolution hook on `event` that captures `card`'s last known information
+ * onto `event.lastKnownInformation`. The snapshot is taken right before the event's
+ * EventWindow resolves its events, so the captured state reflects the card immediately
+ * before any in-window changes (e.g. defeat from damage events sharing the same window).
+ */
+export function addLastKnownInformationToEvent(event: GameEvent, card: Card): void {
```
AMMayberry1, line 82: Bit of a nit tbh, but I feel like this might be the better approach for the attack events rather than statically constructing the LKI when the event is created.

The major difference between the two timing-wise in this case is replacement effects. A pre-resolution effect would happen after any replacement effects have resolved. So for now there's no case where it would matter, but if there were something in the future that swaps out the attacker at the replacement stage, the current version wouldn't work correctly.

server/game/core/attack/AttackFlow.ts
```diff
@@ -55,6 +61,10 @@ export class AttackFlow extends BaseStepWithPipeline {
         // ensure that this resolves after the damage events
         attackCompleteEvent.order = 1;
 
+        // The OnAttackDamageResolved event currently does not carry an `attackerLastKnownInformation`.
+        // If a future ability triggers on this event and needs the attacker's pre-damage state,
+        // attach LKI here in the same shape as OnAttackDeclared / OnAttackEnd (capture occurs
+        // for OnAttackEnd inside dealDamage just before the damage window opens).
```
AMMayberry1, line 67: I could go either way on this. I'd be totally fine adding the LKI here now (and maybe even adding defender LKI on all three), it seems safe and straightforward and could reduce potential confusion later if someone's expecting to see it on the event.

That said, we wouldn't have any testing in place for it currently. So I'm good either way

PR #2419, opened 2026-05-22
Add GH action for validating card implementation and test file names

AMMayberry1:
Added a script that will check all of the implementation files and test files to confirm that each card has a matching test in the right place with the right name. Made updates to files that did not pass the check.

Cards which are expected not to have a test file (e.g., Data Vault) can add a special comment which will mark them for exemption.

This script will be run as a GH action on PRs to catch typos or missed test files.


VJubert (dismissed):

.github/workflows/pullrequest.yml
```diff
@@ -9,33 +9,46 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-        - uses: actions/checkout@v4
-        - uses: actions/setup-node@v4
-          with:
-            node-version: '22.x'
-            cache: 'npm'
-        - run: npm install
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
```
moyerr: As a bit of bookkeeping, can we bump these to their current versions? Should be `v6` for both

moyerr (approved):

PR #2423, opened 2026-05-22
When using Scramble Fighters and Moff Jerjerrod, my TIE fighters did not ready

santiagovazquez:
## Bug report

**Message:** When using Scramble Fighters and Moff Jerjerrod, my TIE fighters did not ready
**File:**
[bug-report-0ce29c51-0b06-46d4-ac35-e5a0c8c20414-1779462375855.json](https://github.com/user-attachments/files/28157987/bug-report-0ce29c51-0b06-46d4-ac35-e5a0c8c20414-1779462375855.json)

Screenshot 2026-05-22 at 2 54 53 PM
Screenshot 2026-05-22 at 2 55 08 PM

## Scenario
Use the follow game setup file to check `Scramble Fighters` do not ready them
```json
{
    "phase": "action",
    "player1": {
        "hand": ["scramble-fighters"],
        "groundArena": ["moff-jerjerrod#we-shall-redouble-our-efforts"],
        "resources": 10,
        "base": "data-vault",
        "leader": "governor-pryce#tyrant-of-lothal",
        "hasInitiative": true
    },
    "player2": {
        "base": "city-in-the-clouds",
        "leader": "ahsoka-tano#trust-in-the-force"
    }
}

```

Testing with Covering the Wing

```json
{
    "phase": "action",
    "player1": {
        "hand": ["covering-the-wing"],
        "groundArena": ["wampa", "moff-jerjerrod#we-shall-redouble-our-efforts"],
        "spaceArena": ["alliance-xwing"],
        "resources": 10,
        "base": "data-vault",
        "leader": "governor-pryce#tyrant-of-lothal",
        "hasInitiative": true
    },
    "player2": {
        "groundArena": ["battlefield-marine"],
        "spaceArena": ["red-three#unstoppable"],
        "base": "city-in-the-clouds",
        "leader": "ahsoka-tano#trust-in-the-force"
    }
}
```

Testing with Jetpack (should create 2 shield tokens and both shields should be removed at the end of the phase)

```json
{
    "phase": "action",
    "player1": {
        "hand": ["jetpack"],
        "groundArena": ["battlefield-marine", "moff-jerjerrod#we-shall-redouble-our-efforts"],
        "resources": 10,
        "base": "data-vault",
        "leader": "governor-pryce#tyrant-of-lothal",
        "hasInitiative": true
    },
    "player2": {
        "base": "city-in-the-clouds",
        "leader": "ahsoka-tano#trust-in-the-force"
    }
}
```

## Proposed solution
Moff Jerjerrod was replacing the original event which I think it's not right since it could have multiple implications (in this case TIE fighters are ready and can't attack bases but not the ones created by Moff). I went for a solution to avoid replacing the original event and just "add the same amount of tokens with their default behavior"

### DIFF
--- a/server/game/cards/08_ASH/units/MoffJerjerrodWeShallRedoubleOurEfforts.ts
+++ b/server/game/cards/08_ASH/units/MoffJerjerrodWeShallRedoubleOurEfforts.ts
@@ -5,8 +5,10 @@ import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
 import type { TokenName } from '../../../core/Constants';
 import { TokenCardName, TokenUnitName, TokenUpgradeName } from '../../../core/Constants';
 import type { GameSystem } from '../../../core/gameSystem/GameSystem';
+import type { IPlayerTargetSystemProperties } from '../../../core/gameSystem/PlayerTargetSystem';
 import { Contract } from '../../../core/utils/Contract';
 import { EnumHelpers } from '../../../core/utils/EnumHelpers';
+import type { ICreateTokenUnitRequiredProperties } from '../../../gameSystems/CreateTokenUnitSystem';
 
 export default class MoffJerjerrodWeShallRedoubleOurEfforts extends NonLeaderUnitCard {
     protected override getImplementationId() {
@@ -48,7 +50,11 @@ export default class MoffJerjerrodWeShallRedoubleOurEfforts extends NonLeaderUni
         AbilityHelper: IAbilityHelper
     ): GameSystem<TriggeredAbilityContext<NonLeaderUnitCard>> {
         const { tokenType, amount, player, card } = context.event;
-        const doubledUnitTokenProperties = { amount: amount * 2, target: player, entersReady: context.event.entersReady };
+        const doubledUnitTokenProperties: ICreateTokenUnitRequiredProperties & Pick<IPlayerTargetSystemProperties, 'target'> = {
+            amount: amount * 2,
+            entersReady: context.event.entersReady,
+            target: player
+        };
 
         const systemForToken: Record<TokenName, GameSystem<TriggeredAbilityContext<NonLeaderUnitCard>>> = {
             // Units
--- a/server/game/cards/08_ASH/units/MoffJerjerrodWeShallRedoubleOurEfforts.ts
+++ b/server/game/cards/08_ASH/units/MoffJerjerrodWeShallRedoubleOurEfforts.ts
@@ -37,13 +37,14 @@ export default class MoffJerjerrodWeShallRedoubleOurEfforts extends NonLeaderUni
             // Replacement only happens if Jerjerrod is defeated
             onlyIfYouDoEffect: AbilityHelper.immediateEffects.defeat(),
             replaceWith: (context) => ({
-                replacementImmediateEffect: this.buildDoubledTokenSystem(context, AbilityHelper),
+                replacementImmediateEffect: this.buildAdditionalTokenSystem(context, AbilityHelper),
+                cancelOriginalEvent: false,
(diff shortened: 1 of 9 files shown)


server/game/cards/08_ASH/units/MoffJerjerrodWeShallRedoubleOurEfforts.ts
```diff
@@ -48,15 +48,16 @@ export default class MoffJerjerrodWeShallRedoubleOurEfforts extends NonLeaderUni
         AbilityHelper: IAbilityHelper
     ): GameSystem<TriggeredAbilityContext<NonLeaderUnitCard>> {
         const { tokenType, amount, player, card } = context.event;
+        const doubledUnitTokenProperties = { amount: amount * 2, target: player, entersReady: context.event.entersReady };
```
AMMayberry1: My concern with this approach is that if we add other properties to the token configuration later for other types of effects, we'll need to remember to come back here and add it (which we probably won't).

We need to make this something that will generate a compile-time error if it needs to be updated. If you have any thoughts on how to make that happen feel free, I have a couple ideas as well that might work

santiagovazquez: If you have any directions on how I should procede with this one, let me know! I agree with your feedback but at the same time I don't quite have the same knowledge of the code to be able to produce different approaches here

AMMayberry1: Of course! I think what we need to land on here is using some sort of concrete interface definition with required properties. So here it would end up looking like e.g.:

```ts
const doubledUnitTokenProperties: ICreateTokenUnitRequiredProperties = { amount: ... };
```

then if this hypothetical ICreateTokenUnitRequiredProperties is defined in a way where any new props are always marked as required, this will fail compilation whenever we add anything new.

The trick is figuring out how to set up that interface. We have `ICreateTokenUnitProperties` which exists in `CreateTokenUnitSystem`, but we can't use that directly for this purpose since (a) it extends `IPlayerTargetSystemProperties` and therefore has a lot of extra props we don't need here, and (b) all of its props are optional so they wouldn't fail at compile time if we added a new one here.

I'm wondering if we could maybe do some Typescript magic like this:

```ts
// properties are marked as non-optional here so anything that implements it will have to be updated if it changes later
export interface ICreateTokenUnitRequiredProperties {
  amount: number;
  entersReady: boolean;
}

export interface ICreateTokenUnitProperties extends IPlayerTargetSystemProperties {
  // find some way to inherit props of ICreateTokenUnitRequiredProperties but make them optional
}
```

That might be the wrong approach but I feel confident there's way to accomplish this general approach with TS. If you're not seeing anything let me know and I can carve out a bit more time to poke at it!

moyerr (commented):
Great work on this! I love where you and @AMMayberry1 landed with the required properties.

I wanted to double-check with you on something. Is the `context.resolvedEvents` required? That will be hard to remember for any future abilities that need to reference `generatedTokens`. And on that note: there are 2 other cards that reference generated tokens that we probably need to update and add tests for:

- Jetpack: should create 2 shield tokens and _both_ shields should be removed at the end of the phase
- Covering the Wing: should create 2 X-Wing tokens, and neither should be a valid target for the Shield

santiagovazquez:
@moyerr I believe it is. Since resolvedEvents is after the replacement effects are applied. For Scramble Fighters, the first sequential step queues "create 8 TIEs". Moff can replace that with "create 16 TIEs instead". If the second step uses `context.events[0]?.generatedTokens`, it points at the original replaced event's generated tokens, not the 16 tokens that entered play. So the "cannot attack bases this phase" effect won't be applied to all.
I added new tests for the old cards you mentioned and code has to be updated to work correctly with Moff.

AMMayberry1 (approved):

PR #2428, opened 2026-05-24
add exportable game format, schema, & validator

Ravarix:
The purpose of this PR is to provide a schema of game state & action history to be used for export/replay/initialState. This purposefully does not provide any hooks into the game life cycle yet to provide recording functionality.

I considered building the action history as deltas to preserve size, but the savings really aren't worth the increased complexity and potential desync. Recording the full game state after every action gives us full reliability and an average game would be still be far less than the size of a card image.

The design looks very similar to our current test schema, but uses set codes instead of name slugs. Ex:

```
                phase: 'action',
                player1: {
                    leader: { card: 'SOR_001', deployed: true, damage: 3, exhausted: true, upgrades: ['shield'] },
                    base: { card: 'SOR_229', damage: 10 },
                    groundArena: [
                        {
                            card: 'SOR_010',
                            damage: 2,
                            exhausted: true,
                            upgrades: [{ card: 'SOR_050', owner: 2, controller: 1 }],
                            capturedUnits: [{ card: 'SOR_020', owner: 2 }],
                        },
                    ],
                    spaceArena: [{ card: 'JTL_042b', owner: 2 }],
                    hand: ['SOR_011', 'SOR_012'],
                    deck: ['SOR_013', 'SOR_014'],
                    discard: ['SOR_015'],
                    resources: [{ card: 'SOR_016', exhausted: true }, 'SOR_017'],
                    hasInitiative: true,
                    hasForceToken: true,
                    credits: 2,
                },
                player2: {
                    leader: 'SOR_002',
                    base: 'SOR_230',
                    resources: [],
                },
            });
```

### DIFF
--- a/server/game/core/gameStateFormat/GameStateFormatExporter.ts
+++ b/server/game/core/gameStateFormat/GameStateFormatExporter.ts
@@ -144,17 +144,23 @@ function exportPlayerState(player: Player, game: Game, players: [Player, Player]
     }
 
     // Arenas — controlled by this player, excluding attached upgrades and the
-    // deployed leader (which is serialised under `leader` above).
+    // leader card itself (serialised under `leader` above).
+    //
+    // We identify the leader card by object identity rather than `isLeaderUnit()`.
+    // A vehicle hosting a pilot leader also returns `isLeaderUnit() === true` and
+    // overrides `getType()` to `NonTokenLeaderUnit`, so a type-based check would
+    // incorrectly drop the vehicle from the output.
+    const leaderCard = player.leader as unknown as Card;
     const groundCards = game.groundArena
         .getCards({ controller: player })
-        .filter((c) => !c.isAttached() && !c.isLeaderUnit());
+        .filter((c) => !c.isAttached() && c !== leaderCard);
     if (groundCards.length > 0) {
         state.groundArena = groundCards.map((c) => exportCardEntry(c, player, players));
     }
 
     const spaceCards = game.spaceArena
         .getCards({ controller: player })
-        .filter((c) => !c.isAttached() && !c.isLeaderUnit());
+        .filter((c) => !c.isAttached() && c !== leaderCard);
     if (spaceCards.length > 0) {
         state.spaceArena = spaceCards.map((c) => exportCardEntry(c, player, players));
     }
@@ -207,9 +213,17 @@ function exportLeaderState(leader: ILeaderCard, players: [Player, Player]): Card
     const state: IExportedLeaderState = { card: id };
 
     if (leader.isLeaderUnit()) {
+        // Regular deployed leader: in the arena as a standalone unit.
         state.deployed = true;
         exportDeployedLeaderState(leader, state, players);
+    } else if ((leader as any).deployed === true) {
+        // Pilot leader: deployed as an upgrade attached to a vehicle.
+        // `isLeaderUnit()` returns false here because `isAttached()` is true, but
+        // `_deployed` is still set. Damage and upgrades are tracked on the host
+        // vehicle (via `exportCardEntry`), so we only record `deployed: true`.
+        state.deployed = true;
     } else {
+        // Undeployed leader.
         if (leader.exhausted) {
             state.exhausted = true;
         }
--- a/server/game/core/gameStateFormat/GameStateFormatExporter.ts
+++ b/server/game/core/gameStateFormat/GameStateFormatExporter.ts
@@ -1,0 +1,395 @@
+import type { ISetId } from '../../../game/Interfaces';
+import type { Game } from '../Game';
+import type { Player } from '../Player';
+import type { Card } from '../card/Card';
+import type { IBaseCard } from '../card/BaseCard';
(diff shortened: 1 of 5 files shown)


server/game/core/gameStateFormat/GameStateFormatExporter.ts
```diff
@@ -0,0 +1,409 @@
...
+
+// ── Public API ────────────────────────────────────────────────────────────────
+
+/**
+ * Exports the current live game position to the portable `IExportedGameState`
+ * format. The result is safe to `JSON.stringify` and can be re-imported (after
+ * schema validation via `validateGameState`) to recreate the board position.
+ *
+ * `player1` in the output corresponds to `game.getPlayers()[0]` (the lobby
+ * owner); `player2` corresponds to `game.getPlayers()[1]` (the opponent).
```
AMMayberry1, line 37: Worth calling out that this is a bit different than how we do it for e.g. bug reports, in which player1 is the _reporting_ player. I feel like it might make sense to use the same convention here where it's from the "frame of reference" of whoever actually took the snapshot, depending on the specific use cases and how people will be thinking about it

server/game/core/gameStateFormat/GameStateFormatExporter.ts
```diff
@@ -0,0 +1,409 @@
...
+
+// ── Helpers ───────────────────────────────────────────────────────────────────
+
+const TOKEN_NAME_SET = new Set<string>(TOKEN_NAMES);
+
+/**
+ * Returns the canonical `CardId` for a card instance.
+ *
+ * - Token cards have no set codes and are returned as their `internalName`
+ *   slug (e.g. `"shield"`, `"clone-trooper"`).
```
AMMayberry1, line 381: Tokens do have set codes, they just use a bit of a different convention. E.g., Clone Trooper is TWI_T02. I think we don't use them internally though, which is something we might need to fix for this to work. Also, Shield and Experience get repeated every set pretty much.

server/game/core/gameStateFormat/GameStateFormatExporter.ts
```diff
@@ -0,0 +1,409 @@
...
+ * Internally calls {@link exportGameState} and augments the resulting state
+ * with `round` and `actionNumber` fields (which are omitted from standalone
+ * state exports but required in history entries).
+ *
+ * @param game   The live game to snapshot.
+ * @param seq    Monotonically increasing sequence number across the game (1-based, caller-managed).
+ * @param action The action that produced this state. Omit for the initial
+ *               post-setup entry (seq 1).
+ */
+export function exportHistoryEntry(
```
AMMayberry1, line 109: Is this the format that will be used for history tracking? If so, can you give a summary of the basic design. It looks like it will be recording individual actions, but I'm not totally sure I follow

server/game/core/gameStateFormat/GameStateFormatExporter.ts
```diff
@@ -0,0 +1,409 @@
...
+    if (spaceCards.length > 0) {
+        state.spaceArena = spaceCards.map((c) => exportCardEntry(c, player, players));
+    }
+
+    // Discard
+    if (player.discardZone.count > 0) {
+        state.discard = player.discardZone.cards.map(cardToId);
+    }
+
+    // Full deck, top → bottom order
```
AMMayberry1, line 173: One thing we have to be considerate of is not making it possible to either cheat mid-game by exporting a game state with hidden information or to steal an opponent's decklist by exporting in a way that shows everything.

We probably will end up needing separate rules for separate situations, good to start thinking that through now.

server/game/core/gameStateFormat/GameStateFormatExporter.ts
```diff
@@ -0,0 +1,409 @@
...
+    };
+    if (action !== undefined) {
+        entry.action = action;
+    }
+    return entry;
+}
+
+// ── Player ────────────────────────────────────────────────────────────────────
+
+function exportPlayerState(player: Player, game: Game, players: [Player, Player]): IExportedPlayerState {
```
AMMayberry1, line 134: Do you think it's better to maintain a duplicate set of logic for exporting the new format, or potentially would it be better to use the existing export logic for the bug report / test format and then have a conversion layer to convert from that format to this one?

I think either could work, just different sets of tradeoffs worth considering. Especially once versioning for this schema comes into the picture

server/game/core/gameStateFormat/GameStateFormatExporter.ts
```diff
@@ -0,0 +1,409 @@
...
+    if (player.creditTokenCount > 0) {
+        state.credits = player.creditTokenCount;
+    }
+
+    return state;
+}
+
+// ── Leader ────────────────────────────────────────────────────────────────────
+
+function exportLeaderState(leader: ILeaderCard, players: [Player, Player]): CardId | IExportedLeaderState {
```
AMMayberry1, line 211: Do the tests already have a flow for generating a summary from a game state? We'll want to make sure we have tests that fail if we make engine changes that break any of this

PR #2431, opened 2026-05-25
BE changes for supporting card language preference

AMMayberry1:
Adds two properties to the DB related to supporting card languages:
- A new preference option for the selected card language (defaults to EN)
- A tracking date for having seen the timer tutorial popup, same as we do for the undo tutorial popup

### DIFF
--- a/server/utils/user/UserFactory.ts
+++ b/server/utils/user/UserFactory.ts
@@ -134,6 +134,21 @@ export class UserFactory {
         }
     }
 
+    public async setTimerPopupSeenStatus(userId: string): Promise<boolean> {
+        try {
+            const dbService = await this.dbServicePromise;
+            const userProfile = await dbService.getUserProfileAsync(userId);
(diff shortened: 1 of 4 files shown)


server/utils/user/UserFactory.ts
```diff
@@ -26,6 +26,7 @@ const getDefaultCosmeticsPreferences = () => ({
 
 const getDefaultGameOptionsPreferences = () => ({
```
VJubert, line 27: maybe this return could be type with Pick<IUserPreferences, 'gameOptions'> ?

VJubert (approved):

moyerr (approved):

server/utils/user/UserFactory.ts
AMMayberry1: Yeah good call, I think that could be applied for all of these though so I'll do that in a separate PR to keep it consistent

PR #2438, opened 2026-05-25
fix error with sabine masterpiece & kanan jarrus

VJubert:
We have received several reports from users which says Sabine's Masterpiece does not exhaust resource for opponent. HandlerMenuPrompt crashes with an array containing a null value. I’m not entirely sure why, but I’ve added a filter to remove that value.


test/server/cards/04_JTL/units/SabinesMasterpieceCrazyColorful.spec.ts
```diff
@@ -132,6 +132,43 @@ describe('Sabine\'s Masterpiece, Crazy Colorful', function() {
...
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['kanan-jarrus#spectre-one'],
+                        spaceArena: ['sabines-masterpiece#crazy-colorful'],
+                        base: { card: 'echo-base', damage: 9 }
+                    },
+                    player2: {
+                        base: { card: 'jabbas-palace', damage: 5 }
+                    }
```
moyerr: Let's remove the base damage for the repro case

moyerr (approved):

PR #2440, opened 2026-05-26
Improve resourcing prompts

santiagovazquez:
Issue was [reported on the FE](https://github.com/SWU-Karabast/forceteki-client/issues/448) but I think this could be an easy solution for it.

Initial sourcing phase:
Screenshot 2026-05-26 at 10 22 35 AM
Screenshot 2026-05-26 at 10 22 46 AM

Later on

Screenshot 2026-05-26 at 10 21 04 AM
Screenshot 2026-05-26 at 10 21 25 AM

### DIFF
--- a/server/game/core/gameSteps/prompts/ResourcePrompt.ts
+++ b/server/game/core/gameSteps/prompts/ResourcePrompt.ts
@@ -57,6 +57,9 @@ export class ResourcePrompt extends AllPlayerPrompt {
 
     public override activePromptInternal(player: Player): IPlayerPromptStateProperties {
         let promptText = null;
+        const selectedCount = this.selectedCards[player.name].length;
+        const selectCardMode = this.nCardsToResource === 1 ? SelectCardMode.Single : SelectCardMode.Multiple;
+
         if (this.nCardsToResource !== 1) {
             promptText = `Select ${this.nCardsToResource} cards to resource`;
         } else {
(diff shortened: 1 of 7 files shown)


test/helpers/PlayerInteractionWrapper.ts
```diff
@@ -914,10 +918,17 @@ export class PlayerInteractionWrapper {
      * Player clicks Done prompt
      */
     public clickDone() {
-        if (!this.currentButtons.includes('Done')) {
-            throw new TestSetupError(`${this.name} can't click Done, because it is not present in the prompt`);
+        const currentPrompt = this.player.currentPrompt();
+        const doneButton = currentPrompt.buttons.find((button: { arg: string; text: { toString: () => string } }) =>
+            button.arg === 'done' || button.text.toString().toLowerCase() === 'done'
```
santiagovazquez, line 923: we no longer rely only on the text `done` but also in the `button.arg`

server/game/core/gameSteps/prompts/ResourcePrompt.ts
```diff
@@ -66,9 +69,13 @@ export class ResourcePrompt extends AllPlayerPrompt {
...
         return {
-            selectCardMode: this.nCardsToResource === 1 ? SelectCardMode.Single : SelectCardMode.Multiple,
+            selectCardMode,
             menuTitle: promptText,
-            buttons: [{ text: 'Done', arg: 'done', disabled: !hasEnoughSelected }],
+            buttons: [{
+                text: selectedCount === 0 && selectCardMode === SelectCardMode.Single ? 'Skip Resourcing' : 'Confirm Resources',
+                arg: 'done',
+                disabled: !hasEnoughSelected,
+            }],
```
moyerr, line 78: I think "Done" still makes sense to confirm, as long as the "Skip Resources" one is clear.

I don't know off the top of my head how simple it is, but would it be possible to make the "Skip Resources" variant red? Since it's somewhat of a destructive action, letting the color be an indication could also help avoid any confusion

santiagovazquez: The change in done comes from https://github.com/SWU-Karabast/forceteki-client/issues/448 

For the different color when skipping, I was thinking the same thing when testing it. Is there a way to pass the FE maybe a "warn" or "danger" to actions so we know the color it should be used?

moyerr (approved):

PR #2446, opened 2026-05-27
fix zeb orrelios

VJubert:

### DIFF
--- a/test/server/cards/08_ASH/units/ZebOrreliosFistsWorkEveryTime.spec.ts
+++ b/test/server/cards/08_ASH/units/ZebOrreliosFistsWorkEveryTime.spec.ts
@@ -368,6 +368,55 @@ describe('Zeb Orrelios, Fists Work Every Time', function () {
                 expect(context.player1).toBeActivePlayer();
                 expect(context.p2Base.damage).toBe(2);
             });
+
+            it('should trigger when enemy upgrade which had been took control is defeated', async function () {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        hand: ['evidence-of-the-crime'],
+                        groundArena: ['zeb-orrelios#fists-work-every-time']
+                    },
+                    player2: {
+                        hand: ['outer-rim-constable'],
+                        groundArena: [{ card: 'sabine-wren#explosives-artist', upgrades: ['sabines-lightsaber#not-alone'] }]
+                    }
+                });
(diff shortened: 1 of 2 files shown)


test/server/cards/08_ASH/units/ZebOrreliosFistsWorkEveryTime.spec.ts
```diff
@@ -49,7 +49,7 @@ describe('Zeb Orrelios, Fists Work Every Time', function () {
         });
 
         describe('Zeb Orrelios\'s triggered ability', function () {
-            it('should deal 1 damage to a base when a unit you control with upgrades is defeated', async function () {
+            it('should deal 1 damage to a base when a unit you control with token upgrades is defeated', async function () {
```
moyerr: This case doesn't involve token upgrades -- it's Academy Training

test/server/cards/08_ASH/units/ZebOrreliosFistsWorkEveryTime.spec.ts
```diff
@@ -136,7 +161,212 @@ describe('Zeb Orrelios, Fists Work Every Time', function () {
...
+
+                // Should prompt to deal 1 damage to a base
+                expect(context.player1).toBeAbleToSelectExactly([context.p1Base, context.p2Base]);
+                context.player1.clickCard(context.p2Base);
+
+                expect(context.player1).toBeActivePlayer();
+                expect(context.p2Base.damage).toBe(2);
+            });
+
+            it('should trigger when a multiple friendly upgrades are defeated (from friendly unit)', async function () {
```
moyerr, line 285: Should say "(from enemy unit)"

test/server/cards/08_ASH/units/ZebOrreliosFistsWorkEveryTime.spec.ts
moyerr, line 1: Solid test cases! Could we also add one more where Zeb's controller takes control of an enemy upgrade, and then it gets defeated? That should verify the controller piece of the LKI, since it has a different owner.

test/server/cards/08_ASH/units/ZebOrreliosFistsWorkEveryTime.spec.ts
VJubert: Wampa is a friendly for Zeb Orrelios ?

moyerr (approved):

PR #2454, opened 2026-05-27
Alphabet Squadron U-Wing & Emperor's Messenger

VJubert:


server/game/cards/08_ASH/units/EmperorsMessenger.ts
```diff
@@ -0,0 +1,19 @@
...
+            internalName: 'emperors-messenger',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addOnAttackAbility({
+            title: 'Ready a resource',
+            immediateEffect: abilityHelper.immediateEffects.readyResources({ amount: 1 })
+        });
+    }
```
moyerr, line 22: It doesn't say "friendly" resource. I think we need to put the effect inside a player target resolver

moyerr (approved):

PR #2459, opened 2026-05-29
Add support for Melee.gg decks

santiagovazquez:
Add support for deck URLs from melee.gg. 

[FE part](https://github.com/SWU-Karabast/forceteki-client/pull/718)


server/utils/deck/MeleeDeckFetcher.ts
```diff
@@ -0,0 +1,226 @@
...
+
+        logger.warn(`MeleeDeckFetcher: melee.gg card not found: ${cardName}`);
+        return null;
+    }
+
+    private normalizeMeleeCardName(value: string): string {
+        return value.trim()
+            .replace(/"/g, '\'')
+            .replace(/\s+/g, ' ')
+            .replace(/\bOrellios\b/g, 'Orrelios');
```
VJubert: do we have to do the same for C3PO (on TS26 there is C3P0) ?

server/utils/deck/MeleeDeckFetcher.ts
```diff
@@ -0,0 +1,227 @@
...
+    public async fetchAsync(deckLink: string): Promise<ISwuDbFormatDecklist> {
+        if (typeof deckLink !== 'string' || deckLink.trim().length === 0) {
+            throw new MeleeFetchError(400, 'Invalid deckLink format');
+        }
+
+        if (!deckLink.includes('melee.gg')) {
+            throw new MeleeFetchError(400, 'Invalid deckLink format');
+        }
+
+        const match = deckLink.match(/\/Decklist\/View\/([^/?#]+)\/?/i);
```
AMMayberry1: nit: we could remove some of the explicit error checks above if we make this regex larger to include the melee.gg part, so that an invalid string will just fail to match

server/utils/deck/MeleeDeckFetcher.ts
```diff
@@ -0,0 +1,227 @@
...
+        const deck = this.parseDeckHtml(html, deckId, deckLink);
+
+        if (!deck.leader?.id || !deck.base?.id) {
+            throw new MeleeFetchError(502, 'Melee deck import failed: leader or base was not found');
+        }
+
+        return deck;
+    }
+
+    private parseDeckHtml(html: string, deckId: string, deckLink: string): ISwuDbFormatDecklist {
```
AMMayberry1: nit: I wouldn't put it past melee to occasionally have formatting issues, so I would say let's explicitly handle parsing failures here and have an error path to handle them (e.g., being able to return null on parsing failure and throwing a specific error at the call site if that happens or something like that)

server/gamenode/GameServer.ts
```diff
@@ -1267,15 +1272,15 @@ export class GameServer {
                 if (typeof deckLink !== 'string' || deckLink.trim().length === 0) {
                     return res.status(400).json({ success: false, error: 'Missing deckLink' });
                 }
-                if (!deckLink.includes('swudb.com')) {
-                    return res.status(400).json({ success: false, error: 'Only swudb.com links are supported by this endpoint' });
+                if (!deckLink.includes('swudb.com') && !deckLink.includes('melee.gg')) {
```
AMMayberry1: Just to address it now in case this pattern ever extends any further - can we go ahead and like a DeckLinkResolver static class or something that can wrap checks like this and also do the dispatching stuff currently in `fetchDeckLinkAsync()`. That just reduces the amount of code sites we have to hunt down for adding a third option in the future.

AMMayberry1 (approved):

PR #2460, opened 2026-05-29
conflict within, long live the empire, bwing rearguard

sethhelman:
bwing rearguard
long live the empire
the conflict within


server/game/cards/08_ASH/events/LongLiveTheEmpire.ts
```diff
@@ -0,0 +1,30 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '9020748123',
+            internalName: 'long-live-the-empire'
+        };
+    }
+
+    public override setupCardAbilities(registrar: IEventAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.setEventAbility({
+            title: 'Defeat a friendly unit. If you do, resource the top card of your deck',
```
VJubert:
```suggestion
            title: 'Defeat a friendly Imperial unit. If you do, resource the top card of your deck',
```

test/server/cards/08_ASH/events/LongLiveTheEmpire.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+                },
+                player2: {
+                    groundArena: [{ card: 'atst', upgrades: ['pointless-to-resist'] }],
+                    leader: { card: 'sabine-wren#galvanized-revolutionary', deployed: true, upgrades: ['fulcrum'] }
+                }
+            });
+
+            const { context } = contextRef;
+
+            const startingResources = context.player1.readyResourceCount;
```
VJubert: why use readyResourceCount instead resources.length ? In fact, it's the same value but semanticaly it's not the same

test/server/cards/08_ASH/units/B-WingRearguard.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+
+            context.player2.clickCard(context.awing);
+            expect(context.player2).toBeAbleToSelectExactly([context.bwingRearguard, context.p1Base]);
+            expect(context.bwingRearguard.hasSomeKeyword('sentinel')).toBe(false);
+            context.player2.clickCard(context.p1Base);
+            expect(context.p1Base.damage).toBe(2);
+            expect(context.awing).toBeInZone('spaceArena', context.player2);
+        });
+    });
+});
```
VJubert, line 130: Another test could be play blue leader and use his ability to move it to ground arena

test/server/cards/08_ASH/upgrades/TheConflictWithin.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+                context.player2.clickCard(context.p1Base);
+                expect(context.frontierAtrt.exhausted).toBeTrue();
+
+                context.player1.claimInitiative();
+                context.player2.passAction();
+
+                context.player1.clickDone();
+                context.player2.clickDone();
+
+                expect(context.player2).toHaveEnabledPromptButtons(['Pay', 'Exhaust']);
```
VJubert, line 41: can you add a test for prompt message ?

test/server/cards/08_ASH/upgrades/TheConflictWithin.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+                context.player2.clickPrompt('Pay');
+
+                expect(context.frontierAtrt.exhausted).toBeFalse();
+                expect(context.player2.readyResourceCount).toBe(1);
```
VJubert: why not use exhaustedResourceCount ?

server/game/cards/08_ASH/upgrades/TheConflictWithin.ts
```diff
@@ -0,0 +1,42 @@
...
+        };
+    }
+
+    public override setupCardAbilities(registrar: IUpgradeAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addGainTriggeredAbilityTargetingAttached({
+            title: `Exhaust it unless its controller pays ${TextHelper.resource(3)}`,
+            when: {
+                onCardReadied: (event, context) => context.source === event.card
+            },
+            targetResolver: {
```
AMMayberry1, line 22: For cases like this there's a little bit of extra wiring we have to do to get the UI stuff to show up correctly. I spun it up and local dev and here's how the current version looks:

Image

Compare with how the Syril unit ability looks, where target unit is highlighted and we get colored buttons in the bottom-right:

Image

I think the fix is easy, just refer to how Syril's ability is wired up.

Looking into this I also noticed a few issues that I'll bring up in chat to fix (outside of this PR):
- Dedra leader and In Debt to Crimson Dawn seem to be wired up wrong, in different ways
- Feels like we really should have some kind of check that would catch this

AMMayberry1 (approved):

PR #2465, opened 2026-05-29
Add aspect penalties to cards in hand

santiagovazquez:
Add aspect penalties to cards in active player's cards in hand.

Needed for https://github.com/SWU-Karabast/forceteki-client/issues/154

### DIFF
--- a/server/game/core/card/Card.ts
+++ b/server/game/core/card/Card.ts
@@ -1391,6 +1391,9 @@ export class Card extends OngoingEffectSourceBase implements IGameStatisticsTrac
         const selectionState = activePlayer.getCardSelectionState(this);
         const shouldBeHidden = this.zone.hiddenForPlayers === WildcardRelativePlayer.Any ||
           (!isActivePlayer && this.zone.hiddenForPlayers === RelativePlayer.Opponent);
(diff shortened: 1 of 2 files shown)


server/game/core/card/Card.ts
```diff
@@ -1386,6 +1386,9 @@ export class Card extends OngoingEffectSourceBase implements IGameStatisticsTrac
         const selectionState = activePlayer.getCardSelectionState(this);
         const shouldBeHidden = this.zone.hiddenForPlayers === WildcardRelativePlayer.Any ||
           (!isActivePlayer && this.zone.hiddenForPlayers === RelativePlayer.Opponent);
+        const aspectPenaltyCost = this.zoneName === ZoneName.Hand && isActivePlayer
```
AMMayberry1: I don't think this is doing what we would need to for the FE work item. This is just sending a numerical amount, whereas I was envisioning on the FE side that we would actually render the specific aspects

santiagovazquez: Oh, ic. I was mimicking what I saw in force table (which I liked) and I tried to implement it. Unluckily since we're scaling everything (instead of having fixed versions of a card), it's almost impossible to try to render and display a "+2" after the cost number in card.

AMMayberry1 (approved):

PR #2480, opened 2026-06-02
Bugfix: TS Anakin and Padme leader abilities

moyerr:
The abilities should be able to work if 2 friendly units entered play in the phase, but 1 was defeated.

Previous interpretation was that at least 2 units needed to have entered play and survived, but current consensus is that "2 units entered play as friendly units" is sufficient. So token-creating units will count if the token unit is removed.

### DIFF
--- a/server/game/cards/07_TS26/leaders/AnakinSkywalkerProtectHerAtAllCosts.ts
+++ b/server/game/cards/07_TS26/leaders/AnakinSkywalkerProtectHerAtAllCosts.ts
@@ -1,7 +1,6 @@
 import type { IAbilityHelper } from '../../../AbilityHelper';
 import type { AbilityContext } from '../../../core/ability/AbilityContext';
 import type { ILeaderUnitAbilityRegistrar, ILeaderUnitLeaderSideAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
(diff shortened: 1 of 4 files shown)


server/game/cards/07_TS26/leaders/AnakinSkywalkerProtectHerAtAllCosts.ts
```diff
@@ -31,9 +30,13 @@ export default class AnakinSkywalkerProtectHerAtAllCosts extends LeaderUnitCard
                 controller: RelativePlayer.Self,
                 cardCondition: (card, context) =>
                     card.canBeInPlay() &&
-                    this.friendlyUnitsThatEnteredPlayThisPhase(context).includes(card),
+                    this.unitsEnteredPlayThisPhaseWatcher.getCardsEnteredPlay((entry) =>
+                        entry.card.controller === context.player &&
```
AMMayberry1, line 34: Just want to confirm - this will be checking the card's _current_ controller, not the controller it had at the time it entered play. Is that intended?

moyerr: Yeah, it checks both. We use `entry.playedBy` below

AMMayberry1 (approved):

Issue #2326, opened 2026-04-02
Timing of modified play card action


VJubert:
**Describe the bug**
Following this clarification : https://discord.com/channels/1079526508688847078/1079527291283067020/1488946751032528959, a modified play action will resolve entirely before checking stat. That's matter when playing against a SHD Snoke or when playing a 0/0

**To Reproduce**
Against SHD Snoke, play a 2/2 with LAW Phantom
Against SHD Snoke, play a 2/2 with LOF Burden of Master
Against SHD Snoke, play TS26 Jedi General with a Republic Leader. (xit on #2321)

moyerr:
Easy Prey from Homeworlds is another example

moyerr:
Adding context and repro/test cases, and distinguishing this from the simultaneous-*damage* + HP-buff bug (#2808, fixed by #2481).

**Root causes (engine)**

This is the "modified play / enter-play + HP buff" timing bug. Two independent gaps make a simultaneous enter-play HP buff impossible today:

1. There's no hook on the game object created by the first effect — e.g. Jedi General's `createCloneTrooper()` must fully resolve before the `ifYouDo` `giveExperience()` can target the token, so the buff is necessarily applied in a later step.
2. Even if (1) were solved, there's no built-in way to attach an upgrade simultaneously as a unit enters play. The game-state defeat check (`resolveGameState` → `checkDefeatedByOngoingEffect`) runs first and defeats the unit under SHD Snoke's −2/−2 before the buff lands.

**Confirmed failing test**

`test/server/cards/07_TS26/units/JediGeneral.spec.ts:25` (currently `xit`) — create a 2/2 Clone Trooper + Experience with SHD Snoke in play. I enabled it on the #2481 branch and ran it: the Clone is defeated (→ `outsideTheGame`) instead of surviving at 1/1. #2481's damage-path fix does not (and cannot) address this, since Snoke is an ongoing HP reduction and the buff is applied sequentially.

**Additional repro**

- LOF Three Lessons + SHD Snoke: `sequential([playCardFromHand, …, giveExperience])` — a 2-HP unit played into Snoke's −2/−2 is defeated before the Experience step, same mechanism as Jedi General.

**Not this bug**

#2808 (simultaneous *damage* + HP buff resolved in a single event window, e.g. Maul, Collective Ambition) is a distinct issue with a distinct fix (#2481) and should not be conflated with the enter-play timing gap here.

[...]

PR #2481, opened 2026-06-02
Bugfix: Defer damage-based defeat checks to end of event resolution

### LINKED ISSUE #2808: Simultaneous Experience token + damage defeats a 1-HP unit that should survive
### Summary

When a `simultaneous` game system gives an Experience token (+1/+1) **and** deals damage to the same unit, a 1-HP unit is defeated even though the Experience token's +1 HP should keep it alive. The HP buff from the token is not accounted for when the simultaneous damage's defeat check runs.

Per the rules, the token attach and the damage happen simultaneously, so a 1/1 unit becomes 2/2 with 1 damage and should survive (see the SWU Admiral "Three Lessons" ruling — a unit given an Experience token as it takes/enters simultaneous effects survives). Instead it goes to the discard pile.

Surfaced by **Maul, Collective Ambition** (`server/game/cards/07_TS26/leaders/MaulCollectiveAmbition.ts`), whose ability does `simultaneous([giveExperience(), damage({ amount: 1 })])`.

### Minimal reproduction

```ts
it('should not defeat a 1-HP unit given an Experience token + 1 damage simultaneously', async function() {
    await contextRef.setupTestAsync({
        phase: 'action',
        player1: {
            leader: 'maul#collective-ambition',
            resources: 6,
            groundArena: ['jawa-scavenger'], // 1 HP, 1 keyword (Saboteur)
        }
    });

    context.player1.clickCard(context.maul);
    context.player1.clickCard(context.jawaScavenger);

    // Expected: survives with +1/+1 from Experience and 1 damage
    expect(context.jawaScavenger.damage).toBe(1);
    expect(context.jawaScavenger).toHaveExactUpgradeNames(['experience']);
});
```

### Actual

```
Error: Contract assertion failure: Attempting to read property 'damage' on
'jawa-scavenger' but it is in zone 'discard' where the property does not apply
```

The unit is defeated (moved to discard) instead of surviving.

### Expected

The Experience token's +1 HP is applied simultaneously with the 1 damage; the unit ends at 2 HP with 1 damage and stays in play.

### Notes

Likely in how `SimultaneousSystem` (or the shared event window) orders the HP-modifying upgrade attach vs. the damage defeat check — the defeat check appears to evaluate remaining HP before the Experience upgrade's HP modifier is in effect.

moyerr:
This fixes a bug where simultaneous damage and HP buffs could still result in a unit being defeated when the HP buff should have saved it from defeat.

Fixes #2808

### DIFF
--- a/server/game/core/card/propertyMixins/UnitProperties.ts
+++ b/server/game/core/card/propertyMixins/UnitProperties.ts
@@ -935,7 +935,16 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
                 this._expiredLastingEffectChangedRemainingHp = false;
             }
 
-            this.checkDefeated(source);
+            // Defer the defeat check so simultaneous effects in the same event window (e.g. an
+            // HP-buffing upgrade being attached at the same time) can resolve first. Running the
+            // check inline would set _pendingDefeat mid-window, which CardTargetSystem treats as
+            // an untargetable card — cancelling the very upgrade attachment that would have saved
+            // the unit. The check is invoked at the end of the window's resolveEvents step.
+            if (damageAdded > 0 && this.game.currentEventWindow) {
+                this.game.currentEventWindow.addPostEventResolutionCallback(() => this.checkDefeated(source));
+            } else {
+                this.checkDefeated(source);
+            }
 
             return damageAdded;
         }
--- a/test/server/cards/07_TS26/leaders/MaulCollectiveAmbition.spec.ts
+++ b/test/server/cards/07_TS26/leaders/MaulCollectiveAmbition.spec.ts
@@ -210,6 +210,33 @@ describe('Maul, Collective Ambition', function() {
                 expect(context.player2).toBeActivePlayer();
             });
 
+            it('should not defeat a unit at 1 remaining HP since the Experience token and damage are simultaneous', async function() {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        leader: 'maul#collective-ambition',
+                        resources: 6,
+                        groundArena: [{ card: 'wampa', damage: 4 }],
+                    }
+                });
+
+                const { context } = contextRef;
+
+                // Wampa is a 4/5 with Overwhelm, at 4 damage it has 1 HP remaining
+                context.player1.clickCard(context.maul);
+
+                expect(context.player1).toHavePrompt(abilityPrompt);
+                expect(context.player1).toBeAbleToSelectExactly([context.wampa]);
+                context.player1.clickCard(context.wampa);
+
+                // Wampa gets +1/+1 from Experience and 1 damage simultaneously, ending at 6 HP with 5 damage (1 HP remaining)
+                expect(context.maul.exhausted).toBeTrue();
+                expect(context.wampa).toBeInZone('groundArena');
+                expect(context.wampa.damage).toBe(5);
+                expect(context.wampa).toHaveExactUpgradeNames(['experience']);
+                expect(context.player2).toBeActivePlayer();
+            });
+
             it('should exhaust Maul with no effect when there are no units in play', async function() {
                 await contextRef.setupTestAsync({
                     phase: 'action',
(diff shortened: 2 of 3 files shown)


server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -935,7 +935,16 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
...
             }
 
-            this.checkDefeated(source);
+            // Defer the defeat check so simultaneous effects in the same event window (e.g. an
+            // HP-buffing upgrade being attached at the same time) can resolve first. Running the
+            // check inline would set _pendingDefeat mid-window, which CardTargetSystem treats as
+            // an untargetable card — cancelling the very upgrade attachment that would have saved
+            // the unit. The check is invoked at the end of the window's resolveEvents step.
+            if (damageAdded > 0 && this.game.currentEventWindow) {
+                this.game.currentEventWindow.addPostEventResolutionCallback(() => this.checkDefeated(source));
```
AMMayberry1, line 944: Instead of registering a callback could we just integrate this directly into the `resolveGameState` checks that already happen as part of the event window? We already have a basically the same check to see if a unit was defeated by an ongoing effect: https://github.com/SWU-Karabast/forceteki/blob/5d56e622f68b0dac49fadc76b336bf2061eabcb1/server/game/core/Game.ts#L1541

If that works I'd prefer that route since it would make this a little needlessly complex and also this is injecting a new step into the event window which always creates more complexity around timing stuff.

test/server/cards/07_TS26/leaders/MaulCollectiveAmbition.spec.ts
```diff
@@ -210,6 +210,33 @@ describe('Maul, Collective Ambition', function() {
                 expect(context.player2).toBeActivePlayer();
             });
 
+            it('should not defeat a unit at 1 remaining HP since the Experience token and damage are simultaneous', async function() {
```
AMMayberry1, line 213: Let's also add the tests in that Val mentioned for the "play a unit with an experience" + Snoke case. Can't remember if they don't exist or if they exist but are just commented out

server/game/core/card/propertyMixins/UnitProperties.ts
moyerr: Two things from digging into this:

**On reusing `resolveGameState`:** the existing sweep calls `checkDefeatedByOngoingEffect()`, which passes `DefeatSourceType.FrameworkEffect`. `FrameworkDefeatCardSystem.buildDefeatSource` only preserves the real damage source when it's handed an `IDamageSource` object — so routing damage defeats through the sweep as-is would strip source attribution from *every* damage-driven defeat (bounty collection, responsibility per CR 1.18.1, Bravado). The post-event callback exists specifically to carry the real `source` into the deferred check. I'm open to folding it into `resolveGameState` if we thread the source through, but that ends up re-deriving the same mechanism — happy to go that route if you'd prefer it despite that.

**On the Snoke test:** that's the `xit` in `JediGeneral.spec.ts` (create Clone + Experience under Snoke's −2/−2). I confirmed it still fails on this branch, but it's a *different* bug: Snoke is an ongoing −2/−2 (defeat runs through the `resolveGameState` sweep, not `addDamage`), and the Experience is applied in a sequential `ifYouDo` step, so the unit is already defeated before the buff. This PR is scoped to simultaneous damage + HP buff, covered by the Maul spec (including the 1-remaining-HP simultaneity case). The Snoke / enter-play timing bug is tracked in #2326.

test/server/cards/07_TS26/leaders/MaulCollectiveAmbition.spec.ts
moyerr: The Snoke test Val referenced is the `xit` at `JediGeneral.spec.ts:25` — I verified it still fails here, but it's the separate enter-play timing bug (ongoing −2/−2 + a sequential buff), not something this damage-path fix can address. Tracking it in #2326. This PR's simultaneous damage + HP-buff behavior is covered by the Maul spec, including the 1-remaining-HP case at line 213.

VJubert (approved):

PR #2484, opened 2026-06-02
Fix inconsistencies with certain SelectUnless cards

AMMayberry1:
Fixed some inconsistencies with how a couple cards with SelectUnless effects were implemented:

- SEC Dedra leader was implemented as a standard "Select" resolver, it is now "SelectUnless". There probably isn't ever a game situation where this could actually make a difference, but now it's using the correct pattern in case others refer to it as an example. Also added a test case with Dedra + LTP to confirm the correct behavior there.

- In Debt to Crimson Dawn was missing the target unit property for highlighting, so it wasn't correctly rendering the selection halo in the UI to indicate which unit is being considered for the choice


VJubert (commented):
As I said on discord, maybe it could be good to add a test with Dedra & LTP

VJubert (approved):

PR #2485, opened 2026-06-02
shin hati

VJubert:

### DIFF
--- a/server/game/cards/08_ASH/leaders/ShinHatiEagerAdversary.ts
+++ b/server/game/cards/08_ASH/leaders/ShinHatiEagerAdversary.ts
@@ -26,7 +26,7 @@ export default class ShinHatiEagerAdversary extends LeaderUnitCard {
             contextTitle: (context) => this.getContextTitle(context),
             when: {
                 onAttackEnd: (event, context) =>
-                    event.attack.attacker.controller === context.player &&
+                    event.attack.attackingPlayer === context.player &&
                     this.damageDealtThisPhaseWatcher.unitHasDealtCombatDamageToBaseThisAttack(event.attack.attacker, context)
             },
             optional: true,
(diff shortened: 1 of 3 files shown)


server/game/cards/08_ASH/leaders/ShinHatiEagerAdversary.ts
```diff
@@ -0,0 +1,76 @@
...
+        this.damageDealtThisPhaseWatcher = abilityHelper.stateWatchers.damageDealtThisPhase();
+    }
+
+    protected override setupLeaderSideAbilities(registrar: ILeaderUnitLeaderSideAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addTriggeredAbility({
+            title: 'Exhaust a unit that costs less than the amount of combat damage dealt to a base this attack',
+            contextTitle: (context) => this.getContextTitle(context),
+            when: {
+                onAttackEnd: (event, context) =>
+                    event.attack.attacker.controller === context.player &&
```
AMMayberry1: It's possible that the unit could change controller after the attack has ended (I think if it was a stolen unit and was defeated during the attack that would be one case). So let's use `event.attack.attackingPlayer` instead, and maybe add a test if you can think of a situation that would definitely cover this.

server/game/cards/08_ASH/leaders/ShinHatiEagerAdversary.ts
```diff
@@ -0,0 +1,76 @@
...
+                    this.damageDealtThisPhaseWatcher.unitHasDealtCombatDamageToBaseThisAttack(event.attack.attacker, context)
+            },
+            optional: true,
+            immediateEffect: abilityHelper.immediateEffects.exhaust(),
+            ifYouDo: {
+                title: 'Exhaust a unit that costs less than the amount of combat damage dealt to a base this attack',
```
AMMayberry1, line 35: nit: we could generate a title here as well using an `ifYouDoContext`

VJubert: It's already done by activePromptTitle so I didn't make contextTitle

server/game/cards/08_ASH/leaders/ShinHatiEagerAdversary.ts
VJubert: I've made the change and add a test but they are OK in both condition. I don't know if unit have not change at this moment or if it's snapshot information

AMMayberry1 (dismissed):

moyerr (approved):

AMMayberry1 (approved):

PR #2486, opened 2026-06-03
Choose Your Path

VJubert:


server/game/cards/08_ASH/events/ChooseYourPath.ts
```diff
@@ -0,0 +1,38 @@
...
+        return {
+            id: 'choose-your-path-id',
+            internalName: 'choose-your-path',
+        };
+    }
+
+    public override setupCardAbilities(registrar: IEventAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.setEventAbility({
+            title: 'Choose one: If you control a Force Unit, heal 5 damage from your base. If you control a Mandalorian unit, create a Mandalorian token and give an Advantage token to it.',
+            targetResolver: {
```
AMMayberry1, line 18: I think we should convert this to `chooseModalEffects()` like with Vigilance. Functionality-wise they're basically the same except that modalEffects will show the `(No effect)` helper text when it's relevant, which I think will be very helpful here

VJubert: I made the change but sequential does not work anymore with chooseModal. I use the same trick than Shatterpoint last year to give advantage to the created mandalorian

AMMayberry1: Huh, I wasn't aware that was an issue. I'll make a note to investigate and do a fix but I think the current solution is a bit too much of a hack, I'd prefer that we roll back to what you originally had and then when the fix is in we can switch it to modal. Not having the "no effect" text is a small price to pay for a short while

AMMayberry1 (approved):

PR #2496, opened 2026-06-04
vane

sethhelman:
vane front
vane back

### DIFF
--- a/test/server/cards/08_ASH/leaders/VaneQuarrelsomePirate.spec.ts
+++ b/test/server/cards/08_ASH/leaders/VaneQuarrelsomePirate.spec.ts
@@ -1,0 +1,347 @@
+describe('Vane, Quarrelsome Pirate', function() {
+    integration(function(contextRef) {
+        describe('Leader side triggered ability', function() {
+            it('should deal 2 damage to enemy base after defeating a friendly upgrade', async function() {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        leader: 'vane#quarrelsome-pirate',
+                        groundArena: [{ card: 'battlefield-marine', upgrades: ['advantage', 'nimble-prowess'] }]
+                    },
+                    player2: {
+                        groundArena: ['atst'],
+                        spaceArena: [{ card: 'wing-leader', upgrades: ['experience'] }],
+                        leader: { card: 'grand-inquisitor#hunting-the-jedi', deployed: true }
+                    }
+                });
+
+                const { context } = contextRef;
+
+                context.p1Tokens = {
+                    advantage: context.player1.findCardByName('advantage'),
+                };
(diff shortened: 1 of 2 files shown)


test/server/cards/08_ASH/leaders/VaneQuarrelsomePirate.spec.ts
```diff
@@ -0,0 +1,347 @@
...
+
+                expect(context.p1Base.damage).toBe(0);
+                expect(context.p2Base.damage).toBe(0);
+                expect(context.bobaFett.damage).toBe(3);
+
+                expect(context.player2).toBeActivePlayer();
+            });
+        });
+    });
+});
```
VJubert, line 433: can you add a test with enemy upgrade on friendly unit and another taking control of a unit with token upgrade. First shouldn't be selectable but second one it should

test/server/cards/08_ASH/leaders/VaneQuarrelsomePirate.spec.ts
```diff
@@ -0,0 +1,347 @@
...
+                    advantage: context.player1.findCardByName('advantage'),
+                };
+                context.p2Tokens = {
+                    experience: context.player2.findCardByName('experience'),
+                };
+
+                context.player1.clickCard(context.vane);
+                context.player1.clickPrompt('Deal 2 damage to a base');
+                expect(context.player1).not.toHavePassAbilityButton();
+                expect(context.player1).toBeAbleToSelectExactly([context.p1Base, context.p2Base]);
```
VJubert, line 30: it's a bit strange to have to choose target before discarding but maybe it's the commonw way for these costs

test/server/cards/08_ASH/leaders/VaneQuarrelsomePirate.spec.ts
```diff
@@ -0,0 +1,347 @@
...
+                context.player1.clickCard(context.p2Base);
+
+                expect(context.player1).toHavePrompt('Choose a friendly upgrade to defeat');
+                expect(context.player1).toBeAbleToSelectExactly([context.p1Tokens.advantage, context.nimbleProwess]);
+                context.player1.clickCard(context.p1Tokens.advantage);
+
+                expect(context.p2Base.damage).toBe(2);
+                expect(context.p1Base.damage).toBe(0);
+
+                expect(context.player2).toBeActivePlayer();
```
moyerr, line 78: We should assert that the upgrade was defeated.

test/server/cards/08_ASH/leaders/VaneQuarrelsomePirate.spec.ts
sethhelman: It is how the LAW Han leader worked so I assumed it was normal. I agree it is strange

moyerr (approved):

PR #2497, opened 2026-06-04
⟡ The Darksaber, Icon of Leadership

moyerr:
| ⟡ The Darksaber |
| :---: |
| Icon of Leadership |
|  |


adbronico (dismissed):
I think this looks good. If you want @AMMayberry1 to take a look, there's a lot of lot potential test cases (clone, support, etc)

adbronico (approved):
lgtm! well done

AMMayberry1 (approved):

server/game/core/Player.ts
```diff
@@ -822,10 +822,17 @@ export class Player extends GameObject implements IGameStatisticsTrackable {
     }
 
     /**
-     * Returns the aspects for this player (derived from base and leader)
+     * Returns the aspects for this player (derived from base, leader, and any
+     * active `ProvidesAspects` ongoing effects on this player — e.g. The
+     * Darksaber, Icon of Leadership).
      */
     public getAspects() {
```
AMMayberry1, line 829: It looks like this is only used for costs so it might be good to rename it in case checking aspects becomes relevant for something other than costs someday. In that case, Darksaber's aspects wouldn't be counted

server/game/cards/08_ASH/upgrades/TheDarksaberIconOfLeadership.ts
```diff
@@ -0,0 +1,32 @@
...
+            title: 'Attached unit is a leader unit and gains the Mandalorian trait',
+            ongoingEffect: [
+                AbilityHelper.ongoingEffects.isLeader(),
+                AbilityHelper.ongoingEffects.gainTrait(Trait.Mandalorian),
+            ],
+        });
+
+        registrar.addGainConstantAbilityTargetingAttached({
+            title: 'While you are paying costs, this unit provides its aspect icons',
+            ongoingEffect: AbilityHelper.ongoingEffects.providesAspectIcons(),
```
AMMayberry1, line 29: nit: might be good to call it `providesAspectIconsForCosts()` or something just to be super-explicit about what it does

PR #2504, opened 2026-06-04
Reckoning

bwohldmann:
image

### DIFF
--- a/server/game/cards/08_ASH/events/Reckoning.ts
+++ b/server/game/cards/08_ASH/events/Reckoning.ts
@@ -22,6 +22,7 @@ export default class Reckoning extends EventCard {
             targetResolver: {
                 activePromptTitle: (context) => `Choose a unit to deal ${this.getDamageAmount(context)} damage to`,
                 cardTypeFilter: WildcardCardType.Unit,
+                cardCondition: (_card, context) => this.getDamageAmount(context) > 0,
                 immediateEffect: AbilityHelper.immediateEffects.damage(
                     (context) => ({ amount: this.getDamageAmount(context) })
                 ),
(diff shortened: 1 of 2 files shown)


test/server/cards/08_ASH/events/Reckoning.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+                context.player1.clickCard(context.reckoning);
+
+                // 0 damage total — ability fizzles without prompting to select a target
+                expect(context.reinforcementWalker.damage).toBe(0);
+                expect(context.getChatLogs()).toEqual([
+                    'player1 plays Reckoning'
+                ]);
+                expect(context.player2).toBeActivePlayer();
+            });
```
VJubert, line 95: it's a bit strange than engine does not add a prompt warning, telling this will not have any effect. Can you add a TODO to improve this later ?

moyerr: Maybe if it's in a conditional system it will give the warning?

server/game/cards/08_ASH/events/Reckoning.ts
```diff
@@ -0,0 +1,32 @@
...
+        return context.player.getArenaUnits().reduce((total, unit) => total + unit.damage, 0);
+    }
+
+    public override setupCardAbilities(registrar: IEventAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.setEventAbility({
+            title: 'Deal damage to a unit equal to total damage on all your units',
+            targetResolver: {
+                activePromptTitle: (context) => `Choose a unit to deal ${this.getDamageAmount(context)} damage to`,
+                cardTypeFilter: WildcardCardType.Unit,
+                cardCondition: (_card, context) => this.getDamageAmount(context) > 0,
```
VJubert: cardCondition should be use to filter targets, not for disabling ability because a condition is not met. I think you should use abilityHelper.immediateEffets.conditonal()

moyerr (approved):

PR #2512, opened 2026-06-04
Adding The Armorer leader from Ashes

adbronico:
| ⟡ The Armorer|
| :---: |
| Steel Shapes Us |
|  |
|  |

### DIFF
--- a/server/game/core/Player.ts
+++ b/server/game/core/Player.ts
@@ -752,15 +752,14 @@ export class Player extends GameObject implements IGameStatisticsTrackable {
         // STATE: This _is_ OK for now, as the gameObject references are still kept, but ideally these would also be changed to Refs in the future.
         /** @type {PlayableZone[]} */
         this.playableZones = [
-            new PlayableZone(PlayType.Piloting, this.deckZone),
-            new PlayableZone(PlayType.Piloting, this.discardZone),
-            new PlayableZone(PlayType.Piloting, this.handZone),
             new PlayableZone(PlayType.PlayFromHand, this.handZone),
-            new PlayableZone(PlayType.PlayFromOutOfPlay, this.deckZone),
-            new PlayableZone(PlayType.PlayFromOutOfPlay, this.discardZone),
-            new PlayableZone(PlayType.PlayFromOutOfPlay, this.resourceZone),
+            new PlayableZone(PlayType.Piloting, this.handZone),
             new PlayableZone(PlayType.Plot, this.resourceZone),
             new PlayableZone(PlayType.Smuggle, this.resourceZone),
+            new PlayableZone(PlayType.Piloting, this.deckZone),
+            new PlayableZone(PlayType.PlayFromOutOfPlay, this.deckZone),
+            new PlayableZone(PlayType.Piloting, this.discardZone),
+            new PlayableZone(PlayType.PlayFromOutOfPlay, this.discardZone),
         ];
 
         this._baseZone = new BaseZone(this.game, this, this.base, this.leader);
--- a/server/game/core/Player.ts
+++ b/server/game/core/Player.ts
@@ -754,8 +754,8 @@ export class Player extends GameObject implements IGameStatisticsTrackable {
         this.playableZones = [
             new PlayableZone(PlayType.PlayFromHand, this.handZone),
             new PlayableZone(PlayType.Piloting, this.handZone),
-            new PlayableZone(PlayType.Plot, this.resourceZone),
             new PlayableZone(PlayType.Smuggle, this.resourceZone),
+            new PlayableZone(PlayType.Plot, this.resourceZone),
             new PlayableZone(PlayType.Piloting, this.deckZone),
             new PlayableZone(PlayType.PlayFromOutOfPlay, this.deckZone),
             new PlayableZone(PlayType.Piloting, this.discardZone),
--- a/server/game/core/Player.ts
+++ b/server/game/core/Player.ts
@@ -752,14 +752,15 @@ export class Player extends GameObject implements IGameStatisticsTrackable {
         // STATE: This _is_ OK for now, as the gameObject references are still kept, but ideally these would also be changed to Refs in the future.
         /** @type {PlayableZone[]} */
         this.playableZones = [
-            new PlayableZone(PlayType.PlayFromHand, this.handZone),
+            new PlayableZone(PlayType.Piloting, this.deckZone),
+            new PlayableZone(PlayType.Piloting, this.discardZone),
             new PlayableZone(PlayType.Piloting, this.handZone),
-            new PlayableZone(PlayType.Smuggle, this.resourceZone),
-            new PlayableZone(PlayType.Plot, this.resourceZone),
-            new PlayableZone(PlayType.Piloting, this.deckZone), // TODO: interaction with Ezra
+            new PlayableZone(PlayType.PlayFromHand, this.handZone),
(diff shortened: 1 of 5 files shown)


server/game/core/Player.ts
```diff
@@ -752,14 +752,15 @@ export class Player extends GameObject implements IGameStatisticsTrackable {
...
-            new PlayableZone(PlayType.Smuggle, this.resourceZone),
-            new PlayableZone(PlayType.Plot, this.resourceZone),
-            new PlayableZone(PlayType.Piloting, this.deckZone), // TODO: interaction with Ezra
+            new PlayableZone(PlayType.PlayFromHand, this.handZone),
             new PlayableZone(PlayType.PlayFromOutOfPlay, this.deckZone),
-            new PlayableZone(PlayType.Piloting, this.discardZone), // TODO: interactions with Fine Addition
             new PlayableZone(PlayType.PlayFromOutOfPlay, this.discardZone),
+            new PlayableZone(PlayType.PlayFromOutOfPlay, this.resourceZone),
+            new PlayableZone(PlayType.Plot, this.resourceZone),
+            new PlayableZone(PlayType.Smuggle, this.resourceZone),
```
VJubert, line 757: why change order ? before it was order by zone, now by PlayType ?

test/server/cards/08_ASH/leaders/TheArmorerSteelShapesUs.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+
+                // only the perimeter-atrt (a vehicle, entered play this phase) is a valid attach target
+                expect(context.player1).toBeAbleToSelectExactly([context.perimeterAtrt]);
+                context.player1.clickCard(context.perimeterAtrt);
+
+                expect(context.perimeterAtrt).toHaveExactUpgradeNames(['academy-graduate']);
+                expect(context.moistureFarmer).toBeInZone('resource', context.player1);
+            });
+
+            it('cannot be activated if no units entered play this phase', async function () {
```
VJubert: it can be activated but as soft pass

test/server/cards/08_ASH/leaders/TheArmorerSteelShapesUs.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+
+                // no units played this phase — ability activates (has cost) but upgrade cannot attach
+                context.player1.clickCard(context.theArmorer);
+                context.player1.clickPrompt('Use it anyway');
+
+                // ability resolves with no effect — no upgrade selection prompt appears
+                expect(context.theArmorer.exhausted).toBeTrue();
+                expect(context.player2).toBeActivePlayer();
+            });
+        });
```
VJubert, line 159: can you add an upgrade on discard and an unplayable upgrade (like smuggling compartment)

VJubert (dismissed):

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -55,6 +56,12 @@ export class PlayCardSystem<TContext extends AbilityContext = AbilityContext> ex
         canPlayFromAnyZone: false,
     };
 
+    public override canAffect(target: Card | GameObject, context: TContext, additionalProperties: Partial<IPlayCardProperties> = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        // If context.target is a card, use it instead - this means this is being called inside a card selector
+        const cardToPlay = context.target?.isCard?.() ? context.target : target;
+        return super.canAffect(cardToPlay, context, additionalProperties, mustChangeGameState);
+    }
```
moyerr: I'm surprised this is needed. Don't we already have effects that play cards from resources? Tear This Ship Apart does it for the opponent's resources.

test/server/cards/08_ASH/leaders/TheArmorerSteelShapesUs.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+
+                // play a unit so it enters play this phase
+                context.player1.clickCard(context.porg);
+                context.player2.passAction();
+
+                context.player1.clickCard(context.theArmorer);
+
+                // select upgrade from resources (only protector is an upgrade)
+                expect(context.player1).toBeAbleToSelectExactly([context.protector]);
+                context.player1.clickCard(context.protector);
```
moyerr, line 29: For at least the first test we should assert prompt text, so the player UX is clear. We should also assert a choose nothing button since they're selecting from a hidden zone.

server/game/gameSystems/PlayCardSystem.ts
```diff
@@ -55,6 +56,12 @@ export class PlayCardSystem<TContext extends AbilityContext = AbilityContext> ex
         canPlayFromAnyZone: false,
     };
 
+    public override canAffect(target: Card | GameObject, context: TContext, additionalProperties: Partial<IPlayCardProperties> = {}, mustChangeGameState = GameStateChangeRequired.None): boolean {
+        // If context.target is a card, use it instead - this means this is being called inside a card selector
+        const cardToPlay = context.target?.isCard?.() ? context.target : target;
```
AMMayberry1: Does card selector not pass in the target parameter? I'm worried that this could have unintended side effects since there are other situations where context.target would be set

test/server/cards/08_ASH/leaders/TheArmorerSteelShapesUs.spec.ts
moyerr: Nit: should clearly state we expect it to fizzle or that it can be used to soft pass with no effect

server/game/gameSystems/PlayCardSystem.ts
adbronico: Yeah, I was surprised too, but this would wind up with it having No effect because canAffect was being called with the Armorer or the unit being chosen (I forget which) instead of the upgrade.

server/game/gameSystems/PlayCardSystem.ts
adbronico: So specifically what was happening was that the canAffect logic in GameSystem was calling this for the wrong card (I think the Armorer instead of the upgrade to be played)

AMMayberry1: Huh, do you know why that was happening? It's really weird that the Armorer would be getting targeted here since it isn't the card getting played

adbronico: I removed this and everything worked now. I have no idea. I'll chalk it up to me having broken something and fixed it unknowingly

test/server/cards/08_ASH/leaders/TheArmorerSteelShapesUs.spec.ts
adbronico: I have an attach condition check with Moral Authority

server/game/costs/ResourceCost.ts
```diff
@@ -264,6 +264,11 @@ export abstract class ResourceCost<TCard extends Card = Card> implements ICost<A
                 }
             }
 
+            // When playing a card from the resource row, we want to prioritize paying with itself
+            if (context.playType === PlayType.PlayFromOutOfPlay && context.source.zone.name === ZoneName.Resource) {
```
AMMayberry1: nit: can we make this an `else if` just to guard against putting the card on the exhaust list twice

AMMayberry1 (approved):

PR #2514, opened 2026-06-04
⟡ Red Leader, Strike the Reactor

moyerr:
### Dev notes

Includes some small tweaks to the `AttackStepsSystem` to ensure unconditional non-Attack-dependent abilities are granted in time for targeting, so the supported unit can also choose targets in either arena. This actually ended up cleaning up some Saboteur-specific logic and now works in the general case.

| ⟡ Red Leader |
| :---: |
| Strike the Reactor |
|  |

### DIFF
--- a/test/server/cards/08_ASH/units/RedLeaderStrikeTheReactor.spec.ts
+++ b/test/server/cards/08_ASH/units/RedLeaderStrikeTheReactor.spec.ts
@@ -91,6 +91,189 @@ describe('Red Leader, Strike the Reactor', function() {
                 expect(context.battlefieldMarine).toBeInZone('discard');
             });
 
+            it('should not restrict Red Leader\'s targeting when a Sentinel is only in the other arena', async function() {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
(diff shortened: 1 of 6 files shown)


test/server/cards/08_ASH/units/RedLeaderStrikeTheReactor.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+                // Advance to next action phase so Wampa readies
+                context.moveToNextActionPhase();
+
+                // Wampa (ground) can only attack ground targets — cross-arena was for the Support attack only
+                context.player1.clickCard(context.wampa);
+                expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine, context.p2Base]);
+
+                context.player1.clickCard(context.p2Base);
+            });
+        });
```
adbronico, line 309: Can we also add tests verifying Sentinel targeting and Saboteur interacting with sentinel correctly in both arena's simultaneously?

adbronico (dismissed):

adbronico (approved):
lgtm

PR #2524, opened 2026-06-04
⟡ Gorian Shard's Corsair, Pirate Warship

moyerr:
| ⟡ Gorian Shard's Corsair |
| :---: |
| Pirate Warship |
|  |

### DIFF
--- a/server/game/gameSystems/DamageSystem.ts
+++ b/server/game/gameSystems/DamageSystem.ts
@@ -282,8 +282,13 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext, TPro
             abilityDamageSource.controller = context.event.lastKnownInformation.controller;
         }
 
+        const sourceCard = properties.source ?? context.source;
+        const sourceMakesDamageUnpreventable =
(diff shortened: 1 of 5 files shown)


server/game/gameSystems/DamageSystem.ts
```diff
@@ -283,13 +287,30 @@ export class DamageSystem<TContext extends AbilityContext = AbilityContext, TPro
...
+    private sourcesMakeDamageUnpreventable(event: any): boolean {
+        const damageSource = event.damageSource;
+        if (!damageSource) {
+            return false;
+        }
+
+        const hasUnpreventableEffect = (source: Card | null | undefined) =>
+            source?.hasOngoingEffect?.(EffectName.DamageDealtByThisCardIsUnpreventable) ?? false;
+
+        if (damageSource.type === DamageSourceType.Attack) {
```
AMMayberry1: nit: instead of falling through to the assumption that it's DamageSourceType.Ability, can we use a switch with a default that hits `Contract.fail()` just in case we rework that enum at any point

AMMayberry1:
I haven't thought about this before, but can unpreventable damage be reduced by effects like Vigil or Boba Fett's Armor?

AMMayberry1 (approved):

PR #2525, opened 2026-06-04
Domesticated Loth-Cat

santiagovazquez:
test file
```json
{
  "phase": "action",
  "player1": {
    "hand": [
      "escort-skiff",
      "doctor-pershing#dedicated-to-research",
      "takedown"
    ],
    "groundArena": [
      "domesticated-lothcat",
      "escort-skiff",
      "doctor-pershing#dedicated-to-research"
    ],
    "spaceArena": [
      "patrolling-vwing"
    ]
  },
  "player2": {
    "hand": [
      "escort-skiff",
      "doctor-pershing#dedicated-to-research",
      "takedown"
    ],
    "groundArena": [
      "escort-skiff"
    ],
    "spaceArena": [
      "patrolling-vwing"
    ],
    "resources": 10,
    "hasInitiative": true
  }
}

```


test/server/cards/08_ASH/units/DomesticatedLothCat.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+            });
+
+            const { context } = contextRef;
+            const friendlyEscortSkiff = context.player1.findCardByName('escort-skiff');
+            const friendlyDoctorPershing = context.player1.findCardByName('doctor-pershing#dedicated-to-research');
+            const enemyEscortSkiff = context.player2.findCardByName('escort-skiff');
+            const enemyDoctorPershing = context.player2.findCardByName('doctor-pershing#dedicated-to-research');
+
+            expect(friendlyEscortSkiff.hasSomeKeyword('ambush')).toBeTrue();
+            expect(friendlyDoctorPershing.hasSomeKeyword('support')).toBeTrue();
```
AMMayberry1: Wherever possible we want to test behaviors by using game situations and card effects that would be visible to the player instead of doing internal checks like `hasSomeKeyword`. So in this case we would want to play cards that use Ambush and Support and then set expectations that their effects don't trigger correctly (and we would probably want those to be separate test cases for hygiene).

And like you have tested here, we'd want to make sure they still work for friendly units but again those would be broken into their own unit `it()` cases. For situations that share a lot of common setup you can use a `beforeEach()` for a cluster of `it()` cases, there's a lot of examples of that in the repo for reference.

In addition to checking that the on-play functionality of Ambush and Support goes away, it would also be good to test that the keyword itself goes away even if the units were already on the field before Loth-Cat was played. This could be done using any number of abilities which check for the presence of those keywords - Daimyo Boba leader is a great candidate for a way to check for this using game effects

test/server/cards/08_ASH/units/DomesticatedLothCat.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+                    player1: {
+                        groundArena: ['domesticated-lothcat']
+                    },
+                    player2: {
+                        hand: ['escort-skiff'],
+                        spaceArena: ['patrolling-vwing'],
+                        groundArena: ['wampa'],
+                        resources: 10,
+                        hasInitiative: true
+                    }
```
moyerr, line 14: The conditional ambush setup is not wrong, but it's a bit more convoluted than just testing with an Ambush unit.

AMMayberry1 (approved):

PR #2527, opened 2026-06-05
Barriss Offee, Redeeming Herself

santiagovazquez:
test file

```json
{
  "phase": "action",
  "player1": {
    "hand": [
      "barriss-offee#redeeming-herself"
    ],
    "groundArena": [
      {
        "card": "wampa",
        "damage": 3
      }
    ]
  },
  "player2": {
    "groundArena": [
      {
        "card": "atst",
        "damage": 1
      }
    ]
  }
}

```


server/game/cards/08_ASH/units/BarrissOffeeRedeemingHerself.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addWhenPlayedAbility({
+            title: 'Heal up to 2 damage from a unit. Give an Advantage token to it for each damage healed this way',
+            immediateEffect: abilityHelper.immediateEffects.distributeHealingAmong({
+                amountToDistribute: 2,
+                controller: WildcardRelativePlayer.Any,
+                canChooseNoTargets: true,
+                cardTypeFilter: WildcardCardType.Unit,
+                cardCondition: (card) => card.isUnit() && card.damage > 0,
```
AMMayberry1: We shouldn't have the `card.damage > 0` condition here, it's fine to allow the player to choose a unit with no damage. In general we should assume that the targeting logic will correctly know whether or not a target is legal for an effect and we should not add additional conditions to prevent selecting targets where there will be no effect, this can actually create subtle bugs

server/game/cards/08_ASH/units/BarrissOffeeRedeemingHerself.ts
```diff
@@ -0,0 +1,35 @@
...
+                maxTargets: 1
+            }),
+            ifYouDo: (ifYouDoContext) => ({
+                title: 'Give an Advantage token to that unit for each damage healed this way',
+                ifYouDoCondition: () => ifYouDoContext.events[0].totalDistributed > 0,
+                immediateEffect: abilityHelper.immediateEffects.giveAdvantage({
+                    amount: ifYouDoContext.events[0].totalDistributed,
+                    target: ifYouDoContext.events[0].individualEvents[0].card
```
AMMayberry1, line 29: nit: slightly cleaner to use `ifYouDoContext.events[0].card`

test/server/cards/08_ASH/units/BarrissOffeeRedeemingHerself.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+                context.player1.setDistributeHealingPromptState(new Map([
+                    [context.wampa, 2]
+                ]));
+
+                expect(context.wampa.damage).toBe(1);
+                expect(context.wampa).toHaveExactUpgradeNames(['advantage', 'advantage']);
+            });
+
+            it('gives only one Advantage token when only 1 damage is healed', function() {
```
AMMayberry1, line 32: Let's add a test where 0 is healed from a unit

server/game/cards/08_ASH/units/BarrissOffeeRedeemingHerself.ts
santiagovazquez: I made [this change](https://github.com/SWU-Karabast/forceteki/pull/2527/changes#diff-3823aca3c2b98c493528bdf4bd7a9a81841eb696d897a6a20b7fbf83de37c0bcR199) so the proposed change works, is it correct?

santiagovazquez: I reverted the change since it breaks other tests (also not sure that's what we want). I believe it's not possible to use `ifYouDoContext.events[0].card` but I could be wrong

AMMayberry1 (approved):

PR #2528, opened 2026-06-05
Pit Droid Team

santiagovazquez:
Test file:
```json
{
  "phase": "action",
  "player1": {
    "leader": "chewbacca#walking-carpet",
    "base": "jedha-city",
    "hand": [
      "on-top-of-things",
      "electrostaff",
      "jedi-lightsaber"
    ],
    "groundArena": [
      "pit-droid-team",
      "battlefield-marine",
      "jedi-guardian"
    ]
  },
  "player2": {
    "groundArena": [
      "wampa"
    ],
    "base": "remote-village"
  }
}

```


test/server/cards/08_ASH/units/PitDroidTeam.spec.ts
```diff
@@ -0,0 +1,86 @@
+describe('Pit Droid Team', function() {
+    integration(function(contextRef) {
```
AMMayberry1, line 2: These "first upgrade played" conditions are pretty nuanced and difficult to get correct, so we want to add a _lot_ of tests to cover the various edge cases. See tests for Guardian of the Whills to get a sense of what we typically expect to see.

Here's a few cases off the top of my head that would be good for this one:
- Opponent plays an upgrade on their own unit; opponent plays an upgrade on our unit; we play an upgrade on an opponent's unit (none of these should trigger the discount). After each of these, we play an upgrade onto an eligible friendly unit and the discount works correctly
- Play two upgrades onto separate eligible units (second one does not get discount)
- Pit Droid Crew is played after the first eligible upgrade has already been played, it should not be able to apply its effect at all that phase

AMMayberry1 (approved):

PR #2535, opened 2026-06-05
Flarestar Attack Shuttle

SAEM2710:

### DIFF
--- a/server/game/cards/08_ASH/units/FlarestarAttackShuttle.ts
+++ b/server/game/cards/08_ASH/units/FlarestarAttackShuttle.ts
@@ -13,7 +13,7 @@ export default class FlarestarAttackShuttle extends NonLeaderUnitCard {
 
     public override setupCardAbilities (registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
         registrar.addTriggeredAbility({
-            title: 'Give an Advantage token to a unit.',
+            title: 'Give an Advantage token to a unit',
             optional: true,
             when: {
                 whenPlayed: true,
@@ -21,9 +21,7 @@ export default class FlarestarAttackShuttle extends NonLeaderUnitCard {
             },
             targetResolver: {
                 cardTypeFilter: WildcardCardType.Unit,
-                immediateEffect: AbilityHelper.immediateEffects.giveAdvantage({
-                    amount: 1
-                })
+                immediateEffect: AbilityHelper.immediateEffects.giveAdvantage()
             }
         });
     }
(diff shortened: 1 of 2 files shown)


server/game/cards/08_ASH/units/FlarestarAttackShuttle.ts
```diff
@@ -0,0 +1,30 @@
...
+    protected override getImplementationId () {
+        return {
+            id: '7534790849',
+            internalName: 'flarestar-attack-shuttle',
+        };
+    }
+
+    public override setupCardAbilities (registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addTriggeredAbility({
+            title: 'Give an Advantage token to a unit.',
```
moyerr: Nit: we usually don't add a period to these end of these ability titles

Full disclosure: this probably isn't completely consistent across the codebase, but we try to get more consistent about it over time 😄

server/game/cards/08_ASH/units/FlarestarAttackShuttle.ts
```diff
@@ -0,0 +1,30 @@
...
+            optional: true,
+            when: {
+                whenPlayed: true,
+                whenDefeated: true,
+            },
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                immediateEffect: AbilityHelper.immediateEffects.giveAdvantage({
+                    amount: 1
+                })
```
moyerr: I believe 1 is the default, so you should be able to use `giveAdvantage()` with no props.

test/server/cards/08_ASH/units/FlarestarAttackShuttle.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+                    spaceArena: ['tie-fighter'],
+                },
+            });
+
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.flarestarAttackShuttle);
+
+            expect(context.player1).toHavePassAbilityButton();
+            expect(context.player1).toBeAbleToSelectExactly([context.awing, context.wampa, context.atst, context.tieFighter, context.flarestarAttackShuttle]);
```
moyerr, line 22: For at least one test, I like to have the prompt text under an assertion. Just something like

```ts
expect(context.player1).toHavePrompt('Give an Advantage token to a unit');
```

SAEM2710: Done

moyerr (approved):

PR #2536, opened 2026-06-05
clan wren loyalist, stolen eta, warrior of clan kryze, r2d2 bug fix

sethhelman:
clan wren loyalist
warrior of clan lryze
stolen eta
r2


adbronico (approved):
lgtm

PR #2539, opened 2026-06-05
Improvised Identity

moyerr:
| Improvised Identity |
| :---: |
| |

### DIFF
--- a/test/server/cards/08_ASH/upgrades/ImprovisedIdentity.spec.ts
+++ b/test/server/cards/08_ASH/upgrades/ImprovisedIdentity.spec.ts
@@ -381,6 +381,312 @@ describe('Improvised Identity', function() {
                 // Wampa deals its base 4 damage — no discard means no gained abilities, but the attack still happens
                 expect(context.p2Base.damage).toBe(4);
             });
+
+            describe('with unusual gained abilities', function() {
+                it('should still resolve the attack after gaining Loth-Wolf\'s "can\'t attack" ability (restrictions already passed)', async function() {
+                    // Loth-Wolf has Sentinel + "This unit can't attack."
+                    // The "gainNonKeywordAbilitiesFromUnit" effect grants the "can't attack" constant ability as an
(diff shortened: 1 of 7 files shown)


test/server/cards/08_ASH/upgrades/ImprovisedIdentity.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+                    expect(context.player1).toHavePrompt('Attack with Wampa. It gains Kage Elite\'s abilities for this attack.');
+
+                    // Attack the base — optional attack picker is shown directly (no separate Trigger)
+                    context.player1.clickCard(context.p2Base);
+
+                    // Wampa (4 power) + Raid 2 = 6 damage to base
+                    expect(context.p2Base.damage).toBe(6);
+                });
+
+                it('should grant an On Attack ability from the discarded unit, firing it during the attack', async function() {
```
AMMayberry1, line 190: Do we know if "when defeated" or "when attack ends" abilities are supposed to function, or if they get dropped before they can trigger?

AMMayberry1: Also, if "when defeated" is supposed to work, do we need to worry about JTL L3 😅

test/server/cards/08_ASH/upgrades/ImprovisedIdentity.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+                // The optional attack is still offered even when nothing was discarded; player declines
+                expect(context.player1).toHavePrompt('Attack with Wampa');
+                context.player1.clickPrompt('Pass attack');
+
+                expect(context.p2Base.damage).toBe(0);
+                expect(context.wampa.exhausted).toBeFalse();
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            describe('optional attack with gained abilities', function() {
```
AMMayberry1, line 130: A few oddball cases I can think of:
- Units like Loth-Wolf that have an attack restriction
- Strafing Gunship
- TWI Maul

AMMayberry1 (approved):

PR #2540, opened 2026-06-05
operation cinder, wipe them out, kachirho militia

sethhelman:
wipe them out
kachirho
operation cinder

### DIFF
--- a/server/game/cards/08_ASH/events/WipeThemOut.ts
+++ b/server/game/cards/08_ASH/events/WipeThemOut.ts
@@ -2,6 +2,7 @@ import type { IAbilityHelper } from '../../../AbilityHelper';
 import { EventCard } from '../../../core/card/EventCard';
 import type { IEventAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
 import type { TriggeredAbilityContext } from '../../../core/ability/TriggeredAbilityContext';
+import { Contract } from '../../../core/utils/Contract';
 import { AbilityType, DamageType, RelativePlayer, WildcardCardType, WildcardZoneName } from '../../../core/Constants';
 import { DefeatCardSystem } from '../../../gameSystems/DefeatCardSystem';
 
@@ -37,15 +38,15 @@ export default class WipeThemOut extends EventCard {
                                 }
 
                                 const arena = context.event.lastKnownInformation.arena;
-                                return arena != null && card.zoneName === arena && card !== context.source;
+                                return card.zoneName === arena && card !== context.source;
                             },
                             immediateEffect: AbilityHelper.immediateEffects.excessDamage((context) => {
-                                const triggeredContext = context as TriggeredAbilityContext;
+                                Contract.assertTrue(context.isTriggered());
                                 return {
                                     type: DamageType.Excess,
-                                    sourceEventForExcessDamage: triggeredContext.event.defeatSource.event
+                                    sourceEventForExcessDamage: context.event.defeatSource.event
                                 };
-                            }) as unknown as never
+                            })
                         }
                     })
                 }]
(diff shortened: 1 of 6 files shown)


server/game/cards/08_ASH/events/WipeThemOut.ts
```diff
@@ -0,0 +1,55 @@
...
+                            cardTypeFilter: WildcardCardType.Unit,
+                            controller: RelativePlayer.Opponent,
+                            zoneFilter: WildcardZoneName.AnyArena,
+                            cardCondition: (card, context?: TriggeredAbilityContext) => {
+                                if (!context?.source || !context.event?.lastKnownInformation) {
+                                    return false;
+                                }
+
+                                const arena = context.event.lastKnownInformation.arena;
+                                return arena != null && card.zoneName === arena && card !== context.source;
```
AMMayberry1: Is there ever an expected situation where arena would be null? If not, we can remove that check since it might just silently hide a failure we would want to know about

server/game/cards/08_ASH/events/WipeThemOut.ts
```diff
@@ -0,0 +1,55 @@
...
+                            },
+                            immediateEffect: AbilityHelper.immediateEffects.excessDamage((context) => {
+                                const triggeredContext = context as TriggeredAbilityContext;
+                                return {
+                                    type: DamageType.Excess,
+                                    sourceEventForExcessDamage: triggeredContext.event.defeatSource.event
+                                };
+                            }) as unknown as never
```
AMMayberry1: This is a bit weird, do you know why it's needed? Typically this kind of thing is covering up a type error that we would want to straighten out

server/game/cards/08_ASH/events/WipeThemOut.ts
```diff
@@ -0,0 +1,55 @@
...
+                            cardCondition: (card, context?: TriggeredAbilityContext) => {
+                                if (!context?.source || !context.event?.lastKnownInformation) {
+                                    return false;
+                                }
+
+                                const arena = context.event.lastKnownInformation.arena;
+                                return arena != null && card.zoneName === arena && card !== context.source;
+                            },
+                            immediateEffect: AbilityHelper.immediateEffects.excessDamage((context) => {
+                                const triggeredContext = context as TriggeredAbilityContext;
```
AMMayberry1: To make this a bit safer, we can instead just do `Contract.assertTrue(context.isTriggered())`. That will force `context` into being identified as a TriggeredAbilityContext or throw an error if it's not. The current version will just silently pretend that `triggeredContext` is a TriggeredAbilityContext even if it isn't for some reason, which is much harder to debug

server/game/cards/08_ASH/units/KachirhoMilitia.ts
```diff
@@ -0,0 +1,27 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '6989468664',
+            internalName: 'kachirho-militia'
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addTriggeredAbility({
+            title: 'Ready when an enemy ground unit attacks base',
```
AMMayberry1: nit: I think our standard titling convention here would just be something like `'Ready Kachirho Milita'`

server/game/cards/08_ASH/units/KachirhoMilitia.ts
```diff
@@ -0,0 +1,27 @@
...
+            limit: AbilityHelper.limit.perRound(1),
+            when: {
+                onAttackDeclared: (event, context) =>
+                    event.attack.attacker.controller !== context.player &&
```
AMMayberry1: To avoid issues around controllership of the attacker changing, we should check `event.attackerLastKnownInformation.controller === context.player`. Same for checking the arena on the line below - obviously nothing right now could change arena as part of an on-attack trigger, but might as well future-proof it

AMMayberry1 (approved):

PR #2547, opened 2026-06-06
Ezra Bridger Its Now Or Never

aruizc:
image
image


server/game/cards/08_ASH/leaders/EzraBridgerItsNowOrNever.ts
```diff
@@ -0,0 +1,83 @@
...
+                onAttackEnd: (event, context) =>
+                    event.attack.attackingPlayer === context.player && this.damageDealtThisPhaseWatcher.getDamageDealtToBaseByUnitThisAttack(context.event.attack.attacker, context) >= 3 &&
+                    this.damageDealtThisPhaseWatcher.unitHasDealtCombatDamageToBaseThisAttack(event.attack.attacker, context)
+            },
+            optional: true,
+            immediateEffect: abilityHelper.immediateEffects.exhaust(),
+            ifYouDo: {
+                title: 'Give an Advantage token to a different unit',
+                contextTitle: (context) => this.getContextTitle(context),
+                immediateEffect: abilityHelper.immediateEffects.conditional({
```
VJubert, line 38: why another conditional ? it's already check by trigger non?

server/game/cards/08_ASH/leaders/EzraBridgerItsNowOrNever.ts
```diff
@@ -0,0 +1,83 @@
...
+                    event.attack.attackingPlayer === context.player &&
+                    this.damageDealtThisPhaseWatcher.unitHasDealtCombatDamageToBaseThisAttack(event.attack.attacker, context)
+            },
+            immediateEffect: abilityHelper.immediateEffects.conditional({
+                condition: (context) => {
+                    const damage = this.damageDealtThisPhaseWatcher.getDamageDealtToBaseByUnitThisAttack(context.event.attack.attacker, context);
+                    return damage >= 3;
+                },
+                onTrue: abilityHelper.immediateEffects.selectCard({
+                    activePromptTitle: 'Give an Advantage token to a different unit',
```
VJubert, line 69: why not using context to customize title too

test/server/cards/08_ASH/leaders/EzraBridgerItsNowOrNever.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+                context.player1.clickCard(context.atst);
+                context.player1.clickCard(context.p2Base);
+
+                expect(context.player1).toHavePassAbilityPrompt('Give an Advantage token to a different unit');
+                context.player1.clickPrompt('Pass');
+                expect(context.player2).toBeActivePlayer();
+                expect(context.p2Base.damage).toBe(6);
+                expect(context.ezraBridger.exhausted).toBeFalse();
+            });
+        });
```
VJubert, line 78: can you test with overwhelm

server/game/cards/08_ASH/leaders/EzraBridgerItsNowOrNever.ts
```diff
@@ -0,0 +1,71 @@
...
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar, abilityHelper: IAbilityHelper): void {
+        this.damageDealtThisPhaseWatcher = abilityHelper.stateWatchers.damageDealtThisPhase();
+    }
+
+    protected override setupLeaderSideAbilities(registrar: ILeaderUnitLeaderSideAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addTriggeredAbility({
+            title: 'Give an Advantage token to a different unit',
+            when: {
+                onAttackEnd: (event, context) =>
+                    event.attack.attackingPlayer === context.player && this.damageDealtThisPhaseWatcher.getDamageDealtToBaseByUnitThisAttack(context.event.attack.attacker, context) >= 3 &&
```
AMMayberry1: Sorry, I think we've been giving conflicting advice on this - but the "damage dealt" check should not be here since it's not part of the trigger condition. The trigger condition is just "when a friendly unit's attack ends", and then the "if" check happens in the body of the ability.

So the correct thing to do here is to use a `conditional` check in the `immediateEffect` so that the exhaust can only happen if the unit did the 3 damage to base.

Same comment for the deployed ability

server/game/cards/08_ASH/leaders/EzraBridgerItsNowOrNever.ts
```diff
@@ -0,0 +1,71 @@
...
+                    this.damageDealtThisPhaseWatcher.unitHasDealtCombatDamageToBaseThisAttack(event.attack.attacker, context)
```
AMMayberry1: I believe this check is redundant since we're already checking that the unit dealt 3 damage to base with `getDamageDealtToBaseByUnitThisAttack()`. Same comment for the deployed ability.

AMMayberry1 (approved):

PR #2553, opened 2026-06-07
Arcana Star Map

adbronico:
image

### DIFF
--- a/test/server/cards/08_ASH/upgrades/ArcanaStarMap.spec.ts
+++ b/test/server/cards/08_ASH/upgrades/ArcanaStarMap.spec.ts
@@ -159,6 +159,194 @@ describe('Arcana Star Map', function () {
                 });
                 context.player2.clickPrompt('Take nothing');
             });
+
+            it('', async function () {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        hand: ['search-your-feelings'],
+                        groundArena: [{ card: 'battlefield-marine', upgrades: ['arcana-star-map#path-to-peridea'] }],
+                        deck: ['wampa', 'atst', 'yoda#old-master']
+                    },
+                });
+
+                const buildCardName = (card) => `${card.title}${card.subtitle ? ', ' + card.subtitle : ''}`;
+
+                const { context } = contextRef;
+
+                // TODO REFACTOR THIS WHEN ENABLING FULL DECK SEARCH
+
+                context.player1.clickCard(context.searchYourFeelings);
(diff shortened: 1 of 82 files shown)


test/server/cards/08_ASH/upgrades/ArcanaStarMap.spec.ts
```diff
@@ -0,0 +1,164 @@
+describe('Arcana Star Map', function () {
+    integration(function (contextRef) {
+        describe('Arcana Star Map\'s ability', function () {
+            it('should cause the attached unit\'s controller to search twice as many cards from the top of the deck', async function () {
```
AMMayberry1, line 4: Since we haven't done a replacement like this before I think we need to add pretty broad coverage of different search types to make sure that they all work (this could even potentially go in a dedicated test suite).

A few cases I can think of that merit explicit coverage:
- Search Your Feelings
- Search more than one card (e.g. SOR Tarkin unit)
- Search for aggregate cost (SOR Vader)
- Bounty Hunter's Quarry on both unique and non-unique
- Annihilator where the opponent is searching our deck

test/server/cards/08_ASH/upgrades/ArcanaStarMap.spec.ts
```diff
@@ -0,0 +1,417 @@
...
+            });
+
+            it('should double the search count for effects that deploy units from the deck (Darth Vader)', async function () {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        groundArena: [{ card: 'wampa', upgrades: ['arcana-star-map#path-to-peridea'] }],
+                        hand: ['darth-vader#commanding-the-first-legion'],
+                        deck: [
+                            'battlefield-marine',           // 1  — Rebel unit (not Imperial → invalid)
```
VJubert: Darth vader don't play imperial unit but villainy

Ravarix:
Can we add a test for if star map is attached to a unit with a When Defeated search effect?

adbronico:
> Can we add a test for if star map is attached to a unit with a When Defeated search effect?

Yep, it should hopefully not double that since the constant ability is gone by the time it would resolve

Ravarix:
> > Can we add a test for if star map is attached to a unit with a When Defeated search effect?
> 
> Yep, it should hopefully not double that since the constant ability is gone by the time it would resolve

I think it would apply because of last known information, no?

adbronico:
> > > Can we add a test for if star map is attached to a unit with a When Defeated search effect?
> > 
> > 
> > Yep, it should hopefully not double that since the constant ability is gone by the time it would resolve
> 
> I think it would apply because of last known information, no?

I don't think so. It's similar to Targeting Computer
image

AMMayberry1 (approved):

PR #2554, opened 2026-06-07
Jabba the Hutt

VJubert:

### DIFF
--- a/server/game/cards/08_ASH/units/JabbaTheHuttEminenceOfTatooine.ts
+++ b/server/game/cards/08_ASH/units/JabbaTheHuttEminenceOfTatooine.ts
@@ -1,0 +1,37 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { WildcardCardType } from '../../../core/Constants';
+import { CostAdjustType } from '../../../core/cost/CostAdjuster';
+
+export default class JabbaTheHuttEminenceOfTatooine extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: '7317430733',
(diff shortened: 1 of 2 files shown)


server/game/cards/08_ASH/units/JabbaTheHuttEminenceOfTatooine.ts
```diff
@@ -0,0 +1,37 @@
...
+        registrar.addWhenPlayedAbility({
+            title: 'Return an upgrade to its owner\'s hand. If it\'s returned to your hand, you may play it for free',
+            optional: true,
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Upgrade,
+                immediateEffect: AbilityHelper.immediateEffects.returnToHand()
+            },
+            then: (thenContext) => ({
+                title: `Play ${thenContext.target?.title} for free`,
+                thenCondition: () => thenContext.player === thenContext.target?.controller,
```
AMMayberry1: nit: let's maybe use `owner` instead of `controller` just to avoid any potential weirdness down the road if we end up refactoring how LKI works (I'm kind of thinking of some ideas that might break this as written)

test/server/cards/08_ASH/units/JabbaTheHuttEminenceOfTatooine.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+                const { context } = contextRef;
+                context.player1.clickCard(context.jabbaTheHutt);
+                context.player1.clickCard(context.mastery);
+
+                expect(context.player2).toBeActivePlayer();
+                expect(context.atst).toHaveExactUpgradeNames([]);
+                expect(context.mastery).toBeInZone('hand', context.player2);
+            });
+
+            it('should return an upgrade to its owner\'s hand and play it for free if upgrade return to AP\'s hand (friendly upgrade)', function() {
```
adbronico: Can we make these titles more descriptive of what the test is doing? Having the whole ability listed is kind of confusing as it implies something will be replayed in all of them

adbronico (dismissed):

AMMayberry1 (approved):

PR #2558, opened 2026-06-08
⟡ Baylan Skoll, Fallen Jedi

moyerr:
| ⟡ Baylan Skoll |
| :---: |
| Fallen Jedi |
|  |


server/game/cards/08_ASH/units/BaylanSkollFallenJedi.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    protected override getImplementationId() {
+        return {
+            id: '8242761292',
+            internalName: 'baylan-skoll#fallen-jedi',
+        };
+    }
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar, AbilityHelper: IAbilityHelper): void {
+        this.cardsLeftPlayThisPhaseWatcher = AbilityHelper.stateWatchers.cardsLeftPlayThisPhase();
```
AMMayberry1: I don't think this is the right watcher to be using, since "left play" would count upgrades that were returned to hand. We might actually need to expand the "units defeated" watcher to be a "cards defeated" watcher so it can cover both types

AMMayberry1 (approved):

PR #2563, opened 2026-06-09
Fix Grav Charge bug

santiagovazquez:
I was watching this [video](https://www.youtube.com/watch?v=fB_cfNi0rLY&t=1572s) and it appears to be a bug with "Grav Charge" not triggering "When defeated" abilities on the attached unit.

I added a test covering the exact same scenario and a fix.


adbronico (commented):
Before you did your fix, was the new test failing?

santiagovazquez:
> Before you did your fix, was the new test failing?

Yes but not because of an assertion but because an exception on the GravCharge.ts code

adbronico (dismissed):

AMMayberry1 (approved):

PR #2566, opened 2026-06-09
Fix upgrade conditional targeting

### LINKED ISSUE #1970: Upgrade targeting should not include units if costs cannot be paid to play the upgrade on them
## Description

Currently, there are scenarios where an upgrade can only be played on specific units due to targeted cost reduction (e.g. The Darksaber & Guardian of the Whills). However, the targeting logic still includes units that the player cannot afford to play the upgrade on. 

The result when clicking one of the invalid targets is that the action gets cancelled (same as manually clicking Cancel), but the more ideal solution would be that the targeting logic excludes those units altogether.

## Steps to Reproduce

Use the test setup below, and follow these steps:

1. Click Darksaber
2. Notice that both Bo-Katan and Jedi Guardian are selectable

Clicking Jedi Guardian results in the action being cancelled since costs cannot be paid
Clicking Bo-Katan plays the upgrade, ignoring aspect penalty because she has the Mandalorian trait

```json
{
    "phase": "action",
    "player1": {
        "resources": 4,
        "hand": ["the-darksaber"],
        "groundArena": [
            "jedi-guardian",
            "bokatan-kryze#fighting-for-mandalore"
        ]
    }
}
```

adbronico:
Fixes issues where an upgrade has a conditional cost reduction and targeting shows invalid targets

Should fix #1970

### DIFF
--- a/server/game/core/ability/abilityTargets/CardTargetResolver.ts
+++ b/server/game/core/ability/abilityTargets/CardTargetResolver.ts
@@ -27,6 +27,8 @@ export class CardTargetResolver extends TargetResolver<ICardTargetsResolver<Abil
 
     private readonly onSelectionSetChanged?: (selected: Card | Card[], context: AbilityContext) => void;
 
+    private _inAffordabilityCheck = false;
+
     private static readonly choosingFromHiddenPrompt = '\n(because you are choosing from a hidden zone you may choose nothing)';
 
     public static allZonesAreHidden(zoneFilter: ZoneFilter | ZoneFilter[], controller: RelativePlayer): boolean {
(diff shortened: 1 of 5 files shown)


server/game/core/ability/abilityTargets/CardTargetResolver.ts
```diff
@@ -72,6 +74,25 @@ export class CardTargetResolver extends TargetResolver<ICardTargetsResolver<Abil
...
             }
+
+            // Check whether the resource cost can be paid for this specific target. This ensures that
+            // when cost adjusters are target-dependent (e.g. Mastery only discounts on unique units),
+            // targets that the player cannot afford are excluded from the selectable list rather than
+            // appearing selectable and thencancelling the action after selection.
+            if (context.stage === Stage.PreTarget && !this._inAffordabilityCheck) {
+                // Guard against re-entry: some cost adjusters (e.g. Exploit) call hasLegalTarget on their own CardTargetResolver from within canPay,
+                // which would invoke this same check recursively. Setting __inAffordabilityCheck prevents that inner resolver from entering this block
+                this._inAffordabilityCheck = true;
```
AMMayberry1, line 85: This is a pretty big hack, as evidenced by the fact that this re-entry guard is needed. It's basically trying to shoehorn in the filtering for the upgrade cost evaluation into the general "card targeting" flow in a way that doesn't really make sense, especially since (as the comment notes) that flow isn't tightly coupled with the upgrade cost evaluation and therefore will be called multiple times in the same process.

Realistically we're probably going to have to overhaul the "play upgrade" flow and / or the cost evaluation flow a bit to add an explicit step that considers the cost on a target-by-target basis

PR #2567, opened 2026-06-10
Mandalorian + Multiple damaged units bugfix

adbronico:
This PR solves an error that was happening when multiple units were damaged simultaneously with The Mandalorian out. It also simplifies his UX so that if triggered, a shield is auto-defeated, with priority being given to any Jetpack-style priority removal shields.


test/server/cards/08_ASH/units/TheMandalorianDevotedRescuer.spec.ts
```diff
@@ -260,6 +282,85 @@ describe('The Mandalorian, Devoted Rescuer', function () {
...
+                });
+
+                const { context } = contextRef;
+
+                context.player2.clickCard(context.turbolaserSalvo);
+                context.player2.clickPrompt('Ground');
+                context.player2.clickCard(context.concordDawnInterceptors);
+
+                // Resolve Mando's ability for BFM — auto-defeats first shield
+                expect(context.player1).toHavePrompt('You have multiple triggers to resolve. Choose which to resolve first:');
```
AMMayberry1, line 356: Let's also add an expectation to check the exact prompts that appear, same for the above test

adbronico: I've pushed my fix, but I noticed that the card titles are doubled up in Turbolaser Salvo's tests. This seems to be related to the getOverrideTitle logic in TriggerWindowBase.

I'm wondering if we need a way to not use getOverrideTitle or if it just isn't a big deal

AMMayberry1 (dismissed):

AMMayberry1 (approved):

PR #2573, opened 2026-06-10
Chebato/constant effects

CheBato:
This is draft so we know where we can expand regarding constantEffects look at issue [Issue link] for more details


moyerr:
This will be subsumed by a future PR on [`feature/constant-effects-on-gameboard`](https://github.com/SWU-Karabast/forceteki/tree/feature/constant-effects-on-gameboard)

PR #2575, opened 2026-06-10
[Bugfix] Darksaber update for cards that check friendly leaders

moyerr:
We now read from one of a few methods on `Player` instead of accessing `player.leader` directly:
- `getLeaderCards(filter)`
- `hasSomeLeaderCard(filter)`

The property on player has been renamed to `deckLeader` to explicitly indicate that this is the leader card in the player's deck, and to avoid accidental usage of this property in ability implementations.


server/game/core/gameSteps/phases/RegroupPhase.ts
```diff
@@ -81,8 +81,8 @@ export class RegroupPhase extends Phase {
             cardsToReady.push(...player.getArenaUnits({ condition: (card) => !card.hasRestriction(AbilityRestriction.DoesNotReadyDuringRegroup) }));
             cardsToReady.push(...player.resources);
 
-            if (player.leader.zoneName === ZoneName.Base) {
-                cardsToReady.push(player.leader);
+            if (player.deckLeader.zoneName === ZoneName.Base) {
+                cardsToReady.push(player.deckLeader);
```
AMMayberry1, line 85: Is there any reason to even have `deckLeader`? I feel like this and most other cases would be pretty easy to adapt to just using `getLeaders()`, so it feels like we can maybe save Chaqa some legwork by doing those conversions now

adbronico: It's all just gonna merge conflicts anyway 😎😂

AMMayberry1 (approved):

PR #2578, opened 2026-06-12
Add Deck Validation suite

### LINKED ISSUE #664: Test harness for deck validator
We currently have no unit tests for the deck validator since it requires a unique test setup that is different from how we test cards. We need something set up for this not just to have test coverage of the validator but also so that we can implement the bases that change deck size and be able to test those as well.

adbronico:
Supports https://github.com/SWU-Karabast/forceteki-client/pull/747


server/utils/deck/DeckValidator.ts
```diff
@@ -412,11 +418,33 @@ export class DeckValidator {
...
-                name: cardData.titleAndSubtitle
+                name: cardData.titleAndSubtitle,
+                reason: IllegalInFormatReason.Suspended
             });
         }
     }
 
+    /**
+     * Determines why a card is not legal in the requested format.
+     * - If the card has no recognized sets (unknown set code), it is treated as Unreleased.
```
AMMayberry1: This assumes that the only reason we could ever have an unrecognized set code is that it's from an unreleased set. That's not a safe assumption to make since there could also be typos, or even the "prestige" sets that we don't ever plan to support.

server/utils/deck/SwuSetData.ts
```diff
@@ -77,6 +77,7 @@ export interface IFormatRules {
     rotationBlockCount?: number;
 }
 
+// TODO: Can we clean this up once these rotate out? Are they technically still on the ban list?
```
AMMayberry1: Yeah we definitely can, I'm not opposed to doing that in this PR even

test/server/utils/DeckValidator.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+        it('should reject 4 copies of a card', function () {
+            const filler = helpers.getDeckFiller(48);
+            const quadCard: IInternalCardEntry = { ...filler[0], count: 4 };
+            const deck = buildDeck([quadCard, ...filler.slice(1)]);
+            const failures = validator.validateInternalDeck(deck, eternalProps());
+            expect(failures[DeckValidationFailureReason.TooManyCopiesOfCard]).toBeDefined();
+        });
+
+        // Sideboard
+        it('should allow a sideboard of up to 10 cards', function () {
```
AMMayberry1: We should also add a case with less than 10 sideboard cards

test/server/utils/DeckValidator.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+
+        // Format legality
+        it('should accept cards from all released sets (no rotation lock)', function () {
+            const sorFiller = helpers.getDeckFiller(50, new Set(['SOR']));
+            const deck = buildDeck(sorFiller);
+            const failures = validator.validateInternalDeck(deck, eternalProps());
+            expect(failures[DeckValidationFailureReason.IllegalInFormat]).toBeUndefined();
+        });
+
+        it('should reject a card from an unrecognized set with the Current card pool', function () {
```
AMMayberry1: Why does it say Current card pool specifically? This should be true for any pool, right?

test/server/utils/PremierValidator.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+import { DeckValidationFailureReason, IllegalInFormatReason } from '../../../server/utils/deck/DeckInterfaces';
+import type { IInternalCardEntry } from '../../../server/utils/deck/DeckInterfaces';
+import { DeckValidator } from '../../../server/utils/deck/DeckValidator';
+import { UnitTestCardDataGetter } from '../../../server/utils/cardData/UnitTestCardDataGetter';
+import { makeDeckValidatorHelpers } from './DeckValidatorTestUtils';
+
+const PREMIER_SETS = new Set([...DeckValidator.getLegalSets(SwuGameFormat.Premier, CardPool.Current)].map((s) => s.toUpperCase()));
+const PREMIER_NEXT_SETS = new Set([...DeckValidator.getLegalSets(SwuGameFormat.Premier, CardPool.NextSet)].map((s) => s.toUpperCase()));
+
+describe('Premier validation', function () {
```
AMMayberry1: Why is premier broken out into a separate file from the other three?

test/server/utils/DeckValidator.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    function openProps() {
+        return { format: SwuGameFormat.Open, cardPool: CardPool.Current };
+    }
+
+    function buildDeck(deckCards: IInternalCardEntry[], overrides = {}) {
+        return helpers.buildDeck(LEADER, BASE, deckCards, overrides);
+    }
+
+    describe('Eternal validation', function () {
```
AMMayberry1: As a meta-comment, normally I'm in favor of explicitly writing out test cases but there's a _lot_ of duplication across these so maybe it makes more sense to use a parameterized approach. Unfortunately Jasmine doesn't have native support for that but it's not hard to set up manually, e.g.:

```ts
describe('Calculator Utilities', () => {
  // Define your test datasets
  const testCases = [
    { a: 1, b: 2, expected: 3 },
    { a: -1, b: 1, expected: 0 },
    { a: 0, b: 0, expected: 0 },
    { a: 10, b: -5, expected: 5 }
  ];

  // Dynamically generate an 'it' block for each case
  testCases.forEach(({ a, b, expected }) => {
    it(`should correctly add ${a} and ${b} to equal ${expected}`, () => {
      const result = a + b;
      expect(result).toEqual(expected);
    });
  });
});
```

And then tests that are the same across all formats (or even two or three) could be consolidated a lot more.

test/server/utils/DeckValidator.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+        it('should reject a card from an unrecognized set with the Current card pool', function () {
+            const { validator: extValidator, unreleasedEntry } = helpers.makeValidatorWithUnreleasedCard();
+            const filler = helpers.getDeckFiller(49);
+            const deck = buildDeck([...filler, unreleasedEntry]);
+            const failures = extValidator.validateInternalDeck(deck, eternalProps());
+            expect(failures[DeckValidationFailureReason.IllegalInFormat]).toBeDefined();
+            expect(failures[DeckValidationFailureReason.IllegalInFormat][0].reason).toBe(IllegalInFormatReason.Unreleased);
+        });
+
+        it('should reject a leader card placed in the main deck', function () {
```
AMMayberry1: Can we add test cases for rejecting a deck with two leaders (or no leader at all, or no base at all, or no deck at all)

test/server/utils/DeckValidator.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+            expect(failures[DeckValidationFailureReason.MinDecklistSizeNotMet]).toBeUndefined();
+        });
+
+        it('should reject a deck with fewer than 30 cards', function () {
+            const limitedSets = new Set([...DeckValidator.getLegalSets(SwuGameFormat.Limited, CardPool.Current)].map((s) => s.toUpperCase()));
+            const deck = buildLimitedDeck(helpers.getDeckFiller(29, limitedSets));
+            const failures = validator.validateInternalDeck(deck, limitedProps());
+            expect(failures[DeckValidationFailureReason.MinDecklistSizeNotMet]).toBeDefined();
+            expect(failures[DeckValidationFailureReason.MinDecklistSizeNotMet].actualDecklistSize).toBe(29);
+        });
```
AMMayberry1, line 97: Do limited decks have to be 30 exactly? If so there should be a > 30 check as well. Also, we'll want to test the card pool options

test/server/utils/PremierValidator.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+        it('should reject a sideboard of more than 10 cards', function () {
+            const all = helpers.getDeckFiller(61, PREMIER_SETS);
+            const deck = buildDeck(all.slice(0, 50), { sideboard: all.slice(50, 61) });
+            const failures = validator.validateInternalDeck(deck, premierProps());
+            expect(failures[DeckValidationFailureReason.MaxSideboardSizeExceeded]).toBeDefined();
+            expect(failures[DeckValidationFailureReason.MaxSideboardSizeExceeded].actualSideboardSize).toBe(11);
+        });
+
+        it('should allow a sideboard of more than 10 cards when using the NextSet card pool', function () {
```
AMMayberry1: The test coverage across card pools is very spotty, this seems to be the only one that checks anything in that regard

test/server/utils/DeckValidatorTestUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+import type { ICardDataJson } from '../../../server/utils/cardData/CardDataInterfaces';
+
+// Released set IDs (uppercase to match card.setId.set from JSON)
+export const RELEASED_SETS = new Set<string>([
+    ...rotationBlocks.flatMap((b) => b.sets).filter((s) => s.released)
+        .map((s) => s.id.toUpperCase()),
+    ...nonRotatingSets.filter((s) => s.released).map((s) => s.id.toUpperCase()),
+]);
+
+export function makeDeckValidatorHelpers(cardDataGetter: UnitTestCardDataGetter) {
```
AMMayberry1: We don't use this "helper builder" pattern with any of other test util files, can we switch to just directly exporting functions like we do in the others

server/utils/deck/DeckInterfaces.ts
```diff
@@ -78,6 +78,22 @@ export interface ICardIdAndName {
...
+
+    /** Card's set is not part of the legal rotation for this format (e.g. a SOR card in Premier). */
+    RotatedOut = 'rotatedOut',
+
+    /** Card is from a set that has not yet been officially released (or whose set code is unrecognized). */
+    Unreleased = 'unreleased',
+
+    /** Card is on this format's suspension list. */
+    Suspended = 'suspended',
+}
```
AMMayberry1, line 91: I think you made a comment to this effect already but yeah, there will need to be corresponding changes in the FE for any of these that we add

server/utils/deck/DeckInterfaces.ts
```diff
@@ -78,6 +78,22 @@ export interface ICardIdAndName {
     name: string;
 }
 
+export enum IllegalInFormatReason {
+
+    /** Card's set is not part of the legal rotation for this format (e.g. a SOR card in Premier). */
+    RotatedOut = 'rotatedOut',
+
+    /** Card is from a set that has not yet been officially released (or whose set code is unrecognized). */
+    Unreleased = 'unreleased',
```
AMMayberry1: Bit of a nit but could we rename this to `preview`, I think that makes it a bit clearer

test/server/utils/DeckValidatorTestUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+    /** Returns a set code string like "SOR_005" for a given internal card name. */
+    function sc(internalName: string): string {
```
AMMayberry1: I think the card data getter already has utilities for this and `entry` below

test/server/utils/DeckValidatorTestUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+              !card.types[0].startsWith('token');
+            if (isPlayable) {
+                return { id: setCodeToString(card.setId), count: 1, internalName: card.internalName };
+            }
+        }
+        throw new Error(`No playable card found in set '${set}'`);
+    }
+
+    /** Builds a minimal IDecklistInternal with the provided defaults overrideable via `overrides`. */
+    function buildDeck(defaultLeader: string, defaultBase: string, deckCards: IInternalCardEntry[], overrides: Partial<IDecklistInternal> = {}): IDecklistInternal {
```
AMMayberry1: nit: can we name this something like `buildValidationTestDeck` or something more specific, since we have a few similar use cases for something like this in the test suite

test/server/utils/PremierValidator.spec.ts
adbronico: I actually just never got back to finishing the other ones before, I just realized. I've updated with the other tests

test/server/utils/PremierValidator.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+        });
+    });
+
+    registerCommonDeckRuleTests(() => ({ validator, cardDataGetter, leader, base }), {
+        format: SwuGameFormat.Premier,
+        cardPool: CardPool.Current,
+        legalSets: PREMIER_SETS,
+        minDeckSize: 50,
+        maxCardCopies: 3,
+        sideboardCap: 10,
```
AMMayberry1: These settings need to be tied to something defined in the DeckValidator code so that if we change any of the settings then the changes will be reflected here in the test setup. Could just reuse the same pattern for getting the list of `legalSets`

test/server/utils/CommonDeckRuleTests.ts
```diff
@@ -0,0 +1,206 @@
...
+import { DeckValidationFailureReason, IllegalInFormatReason } from '../../../server/utils/deck/DeckInterfaces';
+import type { IDecklistInternal } from '../../../server/utils/deck/DeckInterfaces';
+import type { DeckValidator } from '../../../server/utils/deck/DeckValidator';
+import type { UnitTestCardDataGetter } from '../../../server/utils/cardData/UnitTestCardDataGetter';
+import { buildValidationTestDeck, getDeckFiller, getFirstBase, getFirstCardInSet, getFirstLeader, getLegalSetCodes } from './DeckValidatorTestUtils';
+
+// Well-known sets in the test card data used to exercise format-legality rules.
+const OLD_RELEASED_SET = 'SOR';   // an early released set; rotated out of Premier and not the current Limited set
+const NON_ROTATING_SET = 'TS26';  // a non-rotating set, legal only in Eternal/Open
+const PREVIEW_SET = 'ASH';        // an unreleased ("preview") set
```
AMMayberry1: Is there a way we can set things up so that we won't need to come update this on a regular basis?

server/utils/deck/DeckValidator.ts
```diff
@@ -54,11 +54,16 @@ export class DeckValidator {
         return new DeckValidator(allCardsData, cardDataGetter.setCodeMap);
     }
 
+    public static createForTesting(allCardsData: ICardDataJson[], setCodeToId: Map<string, string>): DeckValidator {
```
AMMayberry1: I'm generally opposed to having dedicated test constructors like this, it forks the test path from the prod path in ways that can make the tests not useful or hide issues. Can we instead use mocking techniques to override the `cardDataGetter` object for the normal constructor so that it gives the behavior we want

test/server/utils/EternalValidator.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    function eternalProps() {
+        return { format: SwuGameFormat.Eternal, cardPool: CardPool.Current };
+    }
+
+    function buildDeck(deckCards: IInternalCardEntry[], overrides = {}) {
+        return buildValidationTestDeck(cardDataGetter, LEADER, BASE, deckCards, overrides);
+    }
+
+    describe('deck structure', function () {
+        it('should reject a SWUDB deck with a second leader', function () {
```
AMMayberry1, line 35: I'm honestly not 100% sure why we have the two different methods `validateSwuDbDeck` and `validateInternalDeck` in the first place, it looks like the only functional difference is that the former checks for multiple leaders and the latter doesn't.

We probably should consolidate those if we can, since otherwise when we move to the TS tests we may accidentally miss updating one or the other

.vscode/launch.json
```diff
@@ -77,7 +78,8 @@
         "--config=./jasmine.json",
         "--random=true",
         "--parallel=2",
-        "${file}"
+        // Jasmine runs the compiled JS in build/, so translate the open .ts source to its built .js path
+        "${workspaceFolder}/build/${relativeFileDirname}/${fileBasenameNoExtension}.js"
```
AMMayberry1, line 82: What is this for?

server/utils/deck/DeckValidator.ts
```diff
@@ -199,11 +204,11 @@ export class DeckValidator {
...
     public getMaxSideboardSize(format: SwuGameFormat, cardPool: CardPool): number {
-        // Sideboard is only restricted in Premier and Eternal. We relax the restriction in Next Set mode.
-        if (format === SwuGameFormat.Open || format === SwuGameFormat.Limited || cardPool === CardPool.NextSet) {
-            return -1;
+        // Sideboard is only restricted in Premier and Eternal. We relax the restriction in other modes.
+        if ((format === SwuGameFormat.Premier || format === SwuGameFormat.Eternal) && cardPool === CardPool.Current) {
+            return 10;
         }
-        return 10;
+        return -1;
```
AMMayberry1: nit: could we switch to using `null` instead of `-1`, I think it makes the meaning a bit more explicit and will produce clearer errors if there's any kind of mixup around it

server/utils/deck/DeckValidator.ts
```diff
@@ -412,11 +418,32 @@ export class DeckValidator {
...
             });
         }
     }
 
+    /**
+     * Determines why a card is not legal in the requested format.
+     * - If the card has no recognized sets (e.g. a typo or an unsupported set), the set code is UnknownSet.
+     * - If the card would become legal in this format once unreleased sets are included (i.e. it is in a
+     *   preview set that belongs to this format's pool), the card is from a Preview set.
+     * - Otherwise the card's set exists but is outside the format's legal pool entirely: RotatedOut.
```
AMMayberry1: Adding the "Preview" vs "Rotated Out" distinction creates an issue here, since neither one correctly describes a situation like trying to use a TS26 card in Premier. TS26 was never legal in Premier so it's not really correct to say that it was rotated out.

Given that, would it be simpler to just have a single value like "NotLegalInFormat"?

test/server/utils/EternalValidator.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    registerCommonDeckRuleTests(() => ({ validator, cardDataGetter, leader: LEADER, base: BASE }), {
+        format: SwuGameFormat.Eternal,
+        cardPool: CardPool.Current,
+        legalSets: RELEASED_SETS,
+        minDeckSize: 50,
+        maxCardCopies: 3,
+        sideboardCap: 10,
+    });
+
+    describe('ban list', function () {
```
AMMayberry1: This should probably be moved to the common rules, since eventually Premier will presumably have a ban list again (and also TS)

test/server/utils/PremierValidator.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+            const filler = getDeckFiller(cardDataGetter, 50, PREMIER_SETS);
+            const sorCard = getFirstCardInSet(cardDataGetter, 'SOR');
+            const deck = buildDeck(filler, { sideboard: [sorCard] });
+            const failures = validator.validateInternalDeck(deck, premierProps());
+            expect(failures[DeckValidationFailureReason.IllegalInFormat]).toBeDefined();
+            expect(failures[DeckValidationFailureReason.IllegalInFormat][0].reason).toBe(IllegalInFormatReason.RotatedOut);
+        });
+    });
+
+    describe('NextSet card pool', function () {
```
AMMayberry1, line 78: I think this one is also a good candidate to move into the common tests with some kind of flag to enable it in the config

test/server/utils/PremierValidator.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+        format: SwuGameFormat.Premier,
+        cardPool: CardPool.Current,
+        legalSets: PREMIER_SETS,
+        minDeckSize: 50,
+        maxCardCopies: 3,
+        sideboardCap: 10,
+    });
+
+    describe('format legality', function () {
+        it('should reject a rotated-out base in the base slot', function () {
```
AMMayberry1, line 61: Good test, let's also add one for leader

server/utils/deck/DeckValidator.ts
adbronico: Note: this will require updates on the FE too since it uses -1 to mean "no sideboard limits" and will default to 10 otherwise

.vscode/launch.json
adbronico: Without this we can't run the new deck validator tests in VS Code debug because they contain imports

server/utils/deck/DeckValidator.ts
adbronico: We'll need to make a corresponding FE update too.

test/server/utils/LimitedValidator.spec.ts
```diff
@@ -73,13 +70,25 @@ describe('Limited deck validation', function () {
...
 
-    describe('NextSet card pool', function () {
-        it('should reject a card from the latest released set, which rotates out under NextSet', function () {
-            const filler = getDeckFiller(cardDataGetter, 29, LIMITED_NEXT_SETS);
-            const currentSetCard = getDeckFiller(cardDataGetter, 1, LIMITED_SETS)[0];
-            const deck = buildValidationTestDeck(cardDataGetter, nextSetLeader, nextSetBase, [...filler, currentSetCard]);
-            const failures = validator.validateInternalDeck(deck, limitedProps(CardPool.NextSet));
-            expect(failures[DeckValidationFailureReason.IllegalInFormat]).toBeDefined();
+    // Only meaningful while an upcoming mainline set exists; otherwise Limited has no NextSet pool.
+    if (LIMITED_HAS_NEXT_SET) {
```
AMMayberry1: This doesn't seem right, we should be able to test these features regardless of the current state of the SWU release schedule. iirc there's already some stuff for creating mocked cards, maybe we could extend that to sets?

moyerr (approved):

PR #2579, opened 2026-06-12
Re-enable full-deck search effects

### LINKED ISSUE #2500: Re-enable full deck search on cards that use it
Now that the search cards modal can scroll, we can allow cards that search the deck to show the entirety of the deck contents in the search popup. We originally had this for cards like Annihilator but it overflowed the screen so we had to cut back to just the cards that are legal targets.

moyerr:
Resolves #2500

PR #1492 introduced SearchEntireDeckSystem as a workaround for a UI limitation: the card display prompt couldn't scroll, so a deck-wide prompt would overflow the screen. The system hid invalid cards from the prompt so only legal targets were shown, and Search Your Feelings switched to a name-only dropdown. Now that the prompt can scroll, the workaround is no longer needed.

- Switch Bounty Posting, Annihilator, and Great Pit of Carkoon from entireDeckSearch to plain deckSearch — the whole deck is now shown with invalid cards greyed out.
- Restore Search Your Feelings to a plain whole-deck deckSearch with drawSpecificCard.
- Delete SearchEntireDeckSystem and its entireDeckSearch helper.
- Make ISearchDeckProperties.searchCount optional — omit it to search the whole deck.
- Skip the move-to-bottom step for whole-deck searches in SearchDeckSystem so the chat message no longer redundantly says "move N cards to the bottom, and shuffle" — whole-deck searches always shuffle (CR 8.26.2).

Tests: uncomment the deferred "invalid:" assertions in the affected specs, rewrite Search Your Feelings and Great Pit of Carkoon's no-Sarlacc-in-deck cases for the display prompt, fix two Pre Vizsla tests that drove Search Your Feelings via the dropdown UI, and update the search-system scenario spec to expect the full deck contents.


server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -23,12 +23,8 @@ export interface ISearchDeckProperties<TContext extends AbilityContext = Ability
...
     activePromptTitle?: string;
 
-    /**
-     * The number of cards from the top of the deck to search, or a function to determine how many cards to search. Default is -1, which indicates the whole deck.
-     *
-     * This is currently required while SearchEntireDeckSystem exists, but once the prompt UI issue is fixed we can remove that system and make this optional again.
-     */
-    searchCount: number | ((context: TContext) => number);
+    /** The number of cards from the top of the deck to search, or a function to determine it. Omit to search the whole deck. */
+    searchCount?: number | ((context: TContext) => number);
```
AMMayberry1, line 27: I'd prefer we add an explicit bool property for flagging that this search should be "entire deck", I feel like the implicit omission behavior is a bit unclear and folks will do it by accident. If they don't have good enough test expectations it could slip through.

We'd just need to add a validation check in the system that `searchCount` has to be populated if we're not doing an entire deck search. At some point this interface needs some refactoring and cleanup generally, and that point we could redefine things so it'd be a compile-time check (not for this PR though, obviously)

adbronico: Maybe we add an enum? FullDeck, Numeric, Function?

AMMayberry1: Yeah, theoretically I feel like the `TargetMode` should cover this but we are very much not using that consistently right now lol

server/game/gameSystems/SearchDeckSystem.ts
```diff
@@ -95,6 +95,10 @@ export class SearchDeckSystem<TContext extends AbilityContext = AbilityContext,
         const properties = super.generatePropertiesFromContext(context, additionalProperties);
 
         properties.cardCondition = properties.cardCondition || (() => true);
+
+        const searchCount = this.computeSearchCount(properties.searchCount, context);
+        Contract.assertTrue(!!properties.searchWholeDeck === (searchCount == null), 'searchCount must be set if and only if searchWholeDeck is false');
```
adbronico: Maybe it's just personal taste but I'd probably prefer this to be structured a little differently to not do the !! into double comparison thing

adbronico (approved):

PR #2582, opened 2026-06-13
Add related card id when asking for multiple triggers selection

santiagovazquez:
**WARNING:** It must go out with its [FE counterpart](https://github.com/SWU-Karabast/forceteki-client/pull/748). 

The goal is to support card display when selecting multiple action triggers:

Screenshot 2026-06-13 at 12 04 22 PM

### DIFF
--- a/server/game/core/gameSteps/PromptInterfaces.ts
+++ b/server/game/core/gameSteps/PromptInterfaces.ts
@@ -29,7 +29,10 @@ export interface IButton {
     arg: string;
     command?: string;
     disabled?: boolean;
-    relatedCardId?: string;
+}
+
+export interface ITriggerWindowButton extends IButton {
+    sourceCardSetId?: ISetId;
 }
 
 export interface INumberPromptData {
--- a/server/game/core/gameSteps/PromptInterfaces.ts
+++ b/server/game/core/gameSteps/PromptInterfaces.ts
@@ -29,6 +29,7 @@ export interface IButton {
     arg: string;
     command?: string;
     disabled?: boolean;
+    relatedCardId?: string;
 }
 
 export interface INumberPromptData {
--- a/server/game/core/gameSteps/PromptInterfaces.ts
+++ b/server/game/core/gameSteps/PromptInterfaces.ts
@@ -42,6 +42,7 @@ export interface ITriggerWindowSourceCard {
 
 export interface ITriggerWindowButton extends IButton {
     sourceCard?: ITriggerWindowSourceCard;
+    hasLegalEffects: boolean;
 }
 
 export interface INumberPromptData {
--- a/server/game/core/gameSteps/PromptInterfaces.ts
+++ b/server/game/core/gameSteps/PromptInterfaces.ts
@@ -31,8 +31,17 @@ export interface IButton {
     disabled?: boolean;
 }
 
+export interface ITriggerWindowSourceCard {
+    id: string;
+    uuid: string;
+    name: string;
+    setId?: Partial<ISetId>;
+    type: string;
+    printedType?: string;
+}
+
 export interface ITriggerWindowButton extends IButton {
-    sourceCardSetId?: ISetId;
+    sourceCard?: ITriggerWindowSourceCard;
 }
 
 export interface INumberPromptData {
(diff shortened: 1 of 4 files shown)


server/game/core/gameSteps/PromptInterfaces.ts
```diff
@@ -29,6 +29,7 @@ export interface IButton {
     arg: string;
     command?: string;
     disabled?: boolean;
+    relatedCardId?: string;
```
AMMayberry1: nit: can we go with a name that's a bit more specific than "related" so that the intent is clear. In this case maybe "sourceCard" makes sense?

Also I think we've been bad about this but I'd like to start being more consistent about distinguishing "ID" (guid or FFG id) vs "set code", where I believe this is the latter

server/game/core/gameSteps/abilityWindow/TriggerWindowBase.ts
```diff
@@ -257,11 +257,14 @@ export abstract class TriggerWindowBase extends BaseStep {
...
 
         // If its a multi-select, append the card name at the end of the ability name to differentiate them
-        choices = abilitiesToResolve.map((context) => this.getChoiceTitle(context));
+        choices = abilitiesToResolve.map((context) => ({
+            text: this.getChoiceTitle(context),
+            relatedCardId: context.source.uuid
+        }));
         handlers = abilitiesToResolve.map((context) => () => this.resolveAbility(context));
 
         this.game.promptWithHandlerMenu(this.currentlyResolvingPlayer, {
```
AMMayberry1: Since there are now going to be special properties around this prompt type I feel like it's probably worth promoting to its own UiPrompt subclass. Right now there's not a huge distinction and it's only used here, but I think this will start getting more elaborate pretty soon and having a dedicated place to put that logic will help us.

Honestly even here, if we create a dedicated class for it we can maybe just pass in the list of triggered abilities directly and it can pull off the source card set code internally instead of needing to do it in this class

server/game/core/gameSteps/prompts/TriggeredAbilityResolutionPrompt.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export class TriggeredAbilityResolutionPrompt extends UiPrompt {
+    private readonly source = new OngoingEffectSource(this.game, 'Choose Triggered Ability Resolution Order');
+
+    public constructor(
+        game: Game,
+        private readonly player: Player,
+        private readonly triggeredAbilities: TriggeredAbilityContext[],
+        private readonly getChoiceTitle: (context: TriggeredAbilityContext) => string,
+        private readonly resolveAbility: (context: TriggeredAbilityContext) => void
```
AMMayberry1: One nit I just caught - we're trying to move away from having property declarations in the constructor, I find that it decreases readability. Could you change these to be declared at the top of the class and then assigned in the body of ctor?

server/game/core/gameSteps/prompts/TriggeredAbilityResolutionPrompt.ts
```diff
@@ -0,0 +1,95 @@
...
+import type { IPlayerPromptStateProperties } from '../../PlayerPromptState';
+import { UiPrompt } from './UiPrompt';
+
+export class TriggeredAbilityResolutionPrompt extends UiPrompt {
+    private readonly source = new OngoingEffectSource(this.game, 'Choose Triggered Ability Resolution Order');
+
+    public constructor(
+        game: Game,
+        private readonly player: Player,
+        private readonly triggeredAbilities: TriggeredAbilityContext[],
```
AMMayberry1: As discussed in chat, let's manually rearrange these to put any "No Effect" ones at the end

server/game/core/gameSteps/prompts/TriggeredAbilityResolutionPrompt.ts
```diff
@@ -0,0 +1,106 @@
...
+    public override activeCondition(player: Player): boolean {
+        return player === this.player;
+    }
+
+    public override activePromptInternal(): IPlayerPromptStateProperties {
+        const buttons = this.triggeredAbilities
+            .map((context, index) => this.makeChoiceButton(context, index));
+
+        // Sort buttons so that those with legal effects are first
+        buttons.sort((a, b) => {
```
AMMayberry1: Shouldn't we be reordering the contents of `this.triggeredAbilities`, since that's what we apply the index arg to (line 96)?

AMMayberry1 (approved):

PR #2597, opened 2026-06-22
Convert Advantage tokens to a single global trigger listener

### LINKED ISSUE #2587: Convert Advantage tokens to using global trigger
Currently, each Advantage token creates an individual trigger listener. Because large numbers of event listeners cause engine overhead, this could potentially cause issues if one game has a massive amount of Advantage tokens.

Other similar systems use a single global event listener that's registered at game start, we should switch to that pattern to guard against this issue and to generally make things more efficient.

moyerr:
Each Advantage token previously registered its own persistent onAttackEnd listener while attached, which adds engine overhead when many are in play. Register one game-level listener at game start (alongside the existing unit keyword rules listeners) that, when an attack ends, routes the defeat ability of every Advantage token on the involved units into the trigger window.

The token keeps its declarative defeat ability for correct title/source/effect and snapshot safety, but its per-token zone-based registration is suppressed so the single global listener is the only thing that invokes it.

Closes #2587

### DIFF
--- a/server/game/cards/08_ASH/tokens/Advantage.ts
+++ b/server/game/cards/08_ASH/tokens/Advantage.ts
@@ -51,17 +51,9 @@ export default class Advantage extends TokenUpgradeCard {
         this._whenAttackOrDefenseEndsAbility = null;
     }
 
-    /**
-     * Registers the game-level rules listeners for token upgrades. Called once at game start from
-     * {@link Game.registerGlobalRulesListeners}, mirroring {@link UnitPropertiesCard.registerRulesListeners}.
-     */
     public static registerRulesListeners(game: Game) {
-        game.on(EventName.OnAttackEnd + ':preResolve', (event) => {
+        game.on(EventName.OnAttackEnd + ':postResolve', (event) => {
             const attack = event.attack as Attack;
-
-            if (attack == null) {
-                return;
-            }
 
             const involvedAdvantageTokens = [attack.attacker, ...attack.getAllTargets()]
                 .filter((card) => card.isUnit())
--- a/server/game/cards/08_ASH/tokens/Advantage.ts
+++ b/server/game/cards/08_ASH/tokens/Advantage.ts
@@ -1,9 +1,17 @@
-import type { IAbilityHelper } from '../../../AbilityHelper';
-import type { IUpgradeAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { TriggeredAbilityBase } from '../../../core/ability/TriggeredAbility';
+import type { Attack } from '../../../core/attack/Attack';
 import { TokenUpgradeCard } from '../../../core/card/TokenCards';
-import type { ZoneName } from '../../../core/Constants';
+import { EventName } from '../../../core/Constants';
+import type { GameEvent } from '../../../core/event/GameEvent';
+import type { Game } from '../../../core/Game';
+import { registerState, stateRef } from '../../../core/GameObjectUtils';
+import { Contract } from '../../../core/utils/Contract';
+import type { ITriggeredAbilityProps } from '../../../Interfaces';
 
 export default class Advantage extends TokenUpgradeCard {
+    @stateRef()
+    private accessor _whenAttackOrDefenseEndsAbility: TriggeredAbilityBase | null = null;
+
     protected override getImplementationId() {
         return {
             id: '5844562972',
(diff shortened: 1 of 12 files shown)


server/game/cards/08_ASH/tokens/Advantage.ts
```diff
@@ -1,28 +1,92 @@
...
+
+    public override getTriggeredAbilities(): TriggeredAbilityBase[] {
+        const abilities = super.getTriggeredAbilities();
+
+        return this._whenAttackOrDefenseEndsAbility != null
+            ? [...abilities, this._whenAttackOrDefenseEndsAbility]
+            : abilities;
+    }
+
+    public checkRegisterWhenAttackOrDefenseEndsAbility(event: GameEvent) {
```
AMMayberry1: Since this is the same general pattern as used for the other stuff in `UnitProperties`, we might want to do something like setting up the basic pieces there and making things overridable so that this class can just hook into the `checkRegisterWhenAttackOrDefenseEndsAbility` part.

Then `registerRulesListeners` could just be merged into the existing one in `UnitProperties`

server/game/cards/08_ASH/tokens/Advantage.ts
```diff
@@ -1,28 +1,92 @@
...
+        game.on(EventName.OnAttackEnd + ':postResolve', (event) => {
+            const attack = event.attack as Attack;
+
+            const involvedAdvantageTokens = [attack.attacker, ...attack.getAllTargets()]
+                .filter((card) => card.isUnit())
+                .flatMap((unit) => (unit.isInPlay() ? unit.upgrades : []))
+                .filter((upgrade) => upgrade.isAdvantage());
+
+            for (const token of involvedAdvantageTokens) {
+                if (!token.isBlank()) {
```
AMMayberry1: The equivalent checks in `UnitProperties` don't do this check, is there a reason it's specially needed here? My recollection is that the `checkRegister*Abilities` pattern is setting up the triggered abilities like normal so they should still respect blanking without any extra logic

server/game/cards/08_ASH/tokens/Advantage.ts
moyerr: Done. Since `Advantage` is an upgrade, `UnitProperties` (the unit mixin) isn't a common superclass, so I hoisted the generic machinery to `InPlayCard` (the shared ancestor of units and upgrades): the state field, `getTriggeredAbilities` inclusion, register/unregister, and an overridable `buildWhenAttackOrDefenseEndsAbility()` hook that defaults to `null`. `Advantage` now just overrides that hook. The game-level listener is merged into `UnitPropertiesCard.registerRulesListeners`, and the separate `Advantage.registerRulesListeners` call in `Game.ts` is removed.

server/game/cards/08_ASH/tokens/Advantage.ts
moyerr: Good call — removed. Gating the JIT ability behind the blank check in `getTriggeredAbilities()` (mirroring `UnitProperties`) makes the `isBlank()` guard redundant: `TriggeredAbility.eventHandler` only adds itself to the window when `getTriggeredAbilities().includes(this)`, so a blanked token's ability is never collected.

server/game/core/card/baseClasses/InPlayCard.ts
```diff
@@ -404,6 +424,40 @@ export class InPlayCard extends InPlayCardParent implements IInPlayCard {
...
+
+        this._whenAttackOrDefenseEndsAbility.unregisterEvents();
+        this._whenAttackOrDefenseEndsAbility = null;
+    }
+
+    /**
+     * Builds this card's "When Attack/Defense Ends" ability, or returns null if it has none. Overridden by cards
+     * (e.g. the Advantage token) that rely on the just-in-time registration described above.
+     */
+    protected buildWhenAttackOrDefenseEndsAbility(): TriggeredAbilityBase | null {
```
AMMayberry1: Since we could hit cases where there'd be multiple, can we switch this to return an array and call it `buildWhenAttackOrDefenseEndsAbilities`

moyerr: Done — switched to a nullable array (`_whenAttackOrDefenseEndsAbilities`) tracked via `@stateRefArray()`, mirroring the keyword pattern (`_whenPlayedKeywordAbilities`, etc.). `checkRegister`/`unregister` now iterate the array, and the hook is `buildWhenAttackOrDefenseEndsAbilities()` returning `[]` by default; `Advantage` returns a single-element array. Pushed in 311496ad5.

AMMayberry1 (approved):

PR #2599, opened 2026-06-23
Apply text replacement tokens for Keywords

moyerr:
⚠️  Requires Client PR: SWU-Karabast/forceteki-client#757 ⚠️ 

## Description

- Update `TextHelper` to support keyword replacements: `"Ambush"` becomes `"{keyword:ambush}"`
- Update `no-raw-token-text` lint rule to prevent usage of raw Keyword names
- Update all card abilities to use the `TextHelper` keyword methods instead of raw keyword strings
- Update user-facing strings deeper in the engine to also use the `TextHelper` methods (e.g. append `"using ${TextHelper.Smuggle}"` in PlayCardAction.ts)

### DIFF
--- a/server/game/cards/07_LAW/units/UndercityHuntingTeam.ts
+++ b/server/game/cards/07_LAW/units/UndercityHuntingTeam.ts
@@ -13,6 +13,7 @@ export default class UndercityHuntingTeam extends NonLeaderUnitCard {
 
     public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
         registrar.addWhenPlayedAbility({
+            // eslint-disable-next-line forceteki/no-raw-token-text -- "Bounty Hunter" is a trait, not the Bounty keyword
             title: 'Search the top 5 cards of you deck for a Bounty Hunter unit, reveal it, and draw it',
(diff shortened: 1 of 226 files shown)


server/game/cards/07_LAW/units/UndercityHuntingTeam.ts
```diff
@@ -13,6 +13,7 @@ export default class UndercityHuntingTeam extends NonLeaderUnitCard {
 
     public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
         registrar.addWhenPlayedAbility({
+            // eslint-disable-next-line forceteki/no-raw-token-text -- "Bounty Hunter" is a trait, not the Bounty keyword
```
AMMayberry1, line 16: Would it be feasible to update the regex to specifically allow "Bounty Hunter" to pass through? I'm always fuzzy on what's easy or hard to do with regex

moyerr: I think we'll consider this moot for now since I will follow up with replacement tokens for traits. Also, aside from Bounty Hunter, just "Bounty" is both a trait and a keyword.

AMMayberry1 (approved):

Issue #2606, opened 2026-06-29
Moff Jerjerrod can't double tokens created by multiple events


moyerr:
There are many scenarios where tokens are generated by a single effect, but that use multiple `OnTokensCreated` events. Some examples are:

- The multiple experience given by The Crucible
- The shield _and_ experience given by Three Lessons

Because Moff Jerjerrod can only replace one `OnTokensCreated` event, his effect is not behaving as intended in these specific scenarios.

### Relevant Rules Clarification

> Moff Jerjerrod - We Shall Redouble Our Efforts is intended to double an entire creation event from an ability. This means that for Three Lessons, he will give both an additional Shield and an additional Experience. For Executor - Final Destruction of the Alliance, this is also one creation event, so each unit will receive an additional Advantage token. Any restrictions to the tokens also apply to the newly created tokens, so the additional eight TIE Fighters created from Scramble Fighters will come into play and be readied, but will still not be able to attack bases this phase. This replacement effect is tied to the ability, so if you double the creation of Mandalorian tokens, they do not also get two shields each. You get twice the Mandalorian tokens, but their Shielded ability is separate from the creation effect.

PR #2608, opened 2026-06-30
Bugfix: Moff Jerjerrod doubles entire multi-target token creation

moyerr:
Crucible (and other "give a token to each friendly unit" effects) didn't work with Moff Jerjerrod. `GiveTokenUpgradeSystem` emitted one `OnTokensCreated` event per target unit, so Jerjerrod's replacement fired per-unit — and since he can only be defeated once, only one unit got doubled.

Fix: a single give-token-upgrade ability now emits one `OnTokensCreated` event covering every affected unit, so one defeat doubles the whole creation event (matches the publisher ruling). Also generalizes to Executor – Final Destruction.

Note: Three Lessons (XP + Shield via two separate systems) is still a separate cross-system grouping issue and is not fixed here.

Refs #2606 (does not fully resolve)

### DIFF
--- a/server/game/gameSystems/GiveTokenUpgradeSystem.ts
+++ b/server/game/gameSystems/GiveTokenUpgradeSystem.ts
@@ -1,10 +1,12 @@
 import type { AbilityContext } from '../core/ability/AbilityContext';
 import type { Card } from '../core/card/Card';
 import type { CardTypeFilter, TokenUpgradeName } from '../core/Constants';
-import { EventName, WildcardCardType } from '../core/Constants';
+import { EventName, GameStateChangeRequired, WildcardCardType } from '../core/Constants';
 import type { ICardTargetSystemProperties } from '../core/gameSystem/CardTargetSystem';
 import { CardTargetSystem } from '../core/gameSystem/CardTargetSystem';
+import type { GameEvent } from '../core/event/GameEvent';
 import { Contract } from '../core/utils/Contract';
+import { Helpers } from '../core/utils/Helpers';
 import { ChatHelpers } from '../core/chat/ChatHelpers';
 import { AttachUpgradeSystem } from './AttachUpgradeSystem';
 import type { Player } from '../core/Player';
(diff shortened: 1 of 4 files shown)


server/game/gameSystems/GiveTokenUpgradeSystem.ts
```diff
@@ -49,51 +51,102 @@ export abstract class GiveTokenUpgradeSystem<TContext extends AbilityContext = A
...
+    public override queueGenerateEventGameSteps(events: GameEvent[], context: TContext, additionalProperties: Partial<IGiveTokenUpgradeProperties> = {}): void {
+        const { target } = this.generatePropertiesFromContext(context, additionalProperties);
+        const cards = Helpers.asArray(target).filter((card) => this.canAffect(card, context, additionalProperties));
+
+        if (cards.length === 0) {
+            return;
+        }
+
+        const event = this.createEvent(cards, context, additionalProperties);
+        this.updateEventForCards(event, cards, context, additionalProperties);
```
AMMayberry1: This is a bit odd, typically we just override `updateEvent()` since the `generateEvent()` base method does the create followed by update. Is there a reason to fully replace that logic here instead of just calling `generateEvent()` and having a standard override of `updateEvent()` (as opposed to this bespoke `updateEventForCards()`)?

server/game/gameSystems/GiveTokenUpgradeSystem.ts
```diff
@@ -49,51 +51,102 @@ export abstract class GiveTokenUpgradeSystem<TContext extends AbilityContext = A
...
+            Contract.assertTrue(card.isInPlay());
+        }
+
+        // `event.card` is a single-target convention (read by e.g. DistributeAmongTargetsSystem's chat message), so only
+        // populate it when there's genuinely one target. A batched multi-target give exposes its full set via `event.cards`
+        const singleCardTarget = cardsArray.length === 1 ? cardsArray[0] : null;
 
-        super.addPropertiesToEvent(event, card, context, additionalProperties);
+        // GameSystem sets event.player/event.contingentSourceEvent; CardTargetSystem sets event.card to the passed target
+        super.addPropertiesToEvent(event, singleCardTarget, context, additionalProperties);
```
AMMayberry1: This feels a bit hacky, we probably should just update the parent method to allow passing a scalar or an array

server/game/gameSystems/GiveTokenUpgradeSystem.ts
moyerr: Good call — switched to `generateRetargetedEvent` + a standard `updateEvent` override (calling `super.updateEvent`), and dropped the bespoke `updateEventForCards` helper.

server/game/gameSystems/GiveTokenUpgradeSystem.ts
moyerr: I dug into moving this into `CardTargetSystem`, but `event.card` is already overloaded and the consumers conflict: `ViewCardSystem`/`RevealSystem` store an array there and `Grand Admiral Thrawn` / `Han Solo` / `Vermillion` read `event.card[0]`, while single-target consumers read it as a scalar `Card` (DistributeAmongTargetsSystem's chat message and the replacement-effect prompt title `event.card.title`). A uniform base rule breaks one camp or the other — `single-or-null` breaks the array readers, `raw` breaks the scalar readers.

So I kept the single-vs-multi normalization local here (`event.cards` = full target set, `event.card` = the single unit or null for multiple), and only widened the base param type to `Card | Card[]` (it already receives arrays from `ViewCardSystem`, so the old `Card` type was inaccurate) — no behavior change, no cast needed. Happy to do a proper `event.cards`-everywhere migration as a separate follow-up if we want to standardize this pattern.

AMMayberry1 (approved):

PR #2609, opened 2026-06-30
Track username change history in DB

AMMayberry1:
FE PR: https://github.com/SWU-Karabast/forceteki-client/pull/763

As a first step towards improving the username management tools in the mod page, we are adding support for tracking username changes as first-class citizens in the DB.

The following will now be tracked in the DB:

- Initial username state - either for new accounts or accounts that existed before this change which will be migrated to the new system
- Manual username changes (not due to mod action)
- Force renames are gaining context on the before / after for the name change

### How to Test: New Account Scenario
- Enable the docker DB like normal
- Go to the local server and log in to create the account. Grab the user ID from the Preferences page
- Set the username in `seedModAction.ts` and then run it with `ts-node scripts/seedModAction.ts` to make the user a mod
- Restart the dev server to refresh the mod cache
- You should see a "New account" entry in the history for your account. Try applying a force-rename and doing a standard rename on your account to see the entries  

### How to Test: Migrated Account Scenario
Same as above except you need to start with the repos on `main`. After creating the account, switch over to this branch and set `DRY_RUN` to false in `migrateUsernameChangeHistory.ts` and then run it with `ts-node scripts/migrateUsernameChangeHistory.ts` to create the migration entry


scripts/migrateUsernameChangeHistory.ts
```diff
@@ -0,0 +1,112 @@
...
+// and you'll need to set USE_LOCAL_DYNAMODB == false and use ENVIRONMENT for production. Additionally set the DRY_RUN
+// to false if you want the changes to actually happen.
+
+// Usage: ts-node scripts/migrateUsernameChangeHistory.ts
+import { v4 as uuid } from 'uuid';
+import { getDynamoDbServiceAsync } from '../server/services/DynamoDBService';
+import { UsernameChangeSource } from '../server/services/DynamoDBInterfaces';
+import '../server/env';
+
+const DRY_RUN = true; // set to false to actually write
```
moyerr, line 16: Does this need to be flipped before we deploy this?

server/gamenode/GameServer.ts
```diff
@@ -1701,14 +1706,17 @@ export class GameServer {
 
                 // If single match, include mod actions directly
                 let modActions = [];
+                let usernameChanges = [];
                 if (players.length === 1) {
                     modActions = await this.userFactory.getModActionHistoryAsync(players[0].id);
+                    usernameChanges = await this.userFactory.getUsernameChangeHistoryAsync(players[0].id);
```
moyerr: Are these sequential? Is there a way to do them in parallel since they don't depend on each other?

scripts/migrateUsernameChangeHistory.ts
AMMayberry1: This script only needs to be run once for the migration and it's run manually, I keep it in DRY_RUN mode to avoid someone accidentally messing something up when they're testing it out.

It shouldn't really be an issue since it targets the local dynamodb by default anyway, but when it comes to prod data I feel better with multiple safety nets

server/gamenode/GameServer.ts
AMMayberry1: Good call, definitely doable

server/utils/user/UserFactory.ts
```diff
@@ -299,6 +302,15 @@ export class UserFactory {
...
             logger.info(`Username for ${userId} changed to ${newUsername}`);
 
+            // Record the change in the username history
+            await this.recordUsernameChangeAsync(
+                userId,
+                userProfile.username,
+                newUsername,
+                options?.source ?? UsernameChangeSource.UserInitiated,
+                options?.relatedModActionId,
+            );
```
CheBato: i'm not really sure how it should be but as it is now in a scenario where **recordUsernameChangeAsync** fails won't it notify the user that he wasn't able to change his username despite it successfully saving before? And additionally it won't get to **this.modActionService.onRenameCompleted** one level above. Just seems like it might get to a weird state in this scenario. Would it make sense to make this "silently fail"?

server/utils/user/UserFactory.ts
```diff
@@ -405,6 +417,8 @@ export class UserFactory {
             await dbService.saveUserProfileAsync(newUser);
             // create username link
             await dbService.saveUsernameLinkAsync(newUser.username, newUser.id);
+            // record the initial username in the change history
+            await this.recordUsernameChangeAsync(newUser.id, null, newUser.username, UsernameChangeSource.AccountCreation);
```
CheBato: to continue the point of failing (https://github.com/SWU-Karabast/forceteki/pull/2609/changes#r3515090750) could we potentially move this after the 
```
dbService.saveEmailLinkAsync(email, newUser.id);
```
since if it fails here it'll leave the account in a half finished state? Let me know if that would work.

moyerr (approved):

PR #2612, opened 2026-07-01
⟡ Darth Vader, No One to Stop Us

moyerr:
| ⟡ Darth Vader |
| :---: |
| No One to Stop Us |
|  |
|  |

### DIFF
--- a/test/server/cards/09_IC27/leaders/DarthVaderNoOneToStopUs.spec.ts
+++ b/test/server/cards/09_IC27/leaders/DarthVaderNoOneToStopUs.spec.ts
@@ -142,6 +142,33 @@ describe('Darth Vader, No One to Stop Us', function() {
                 expect(context.player1.handSize).toBe(0);
                 expect(context.p1Base.damage).toBe(1);
             });
+
+            it('should defeat the base when it has 2 remaining HP and an empty deck, because the empty-deck draw resolves before the heal', async function() {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        leader: 'darth-vader#no-one-to-stop-us',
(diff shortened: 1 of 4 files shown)


test/server/cards/09_IC27/leaders/DarthVaderNoOneToStopUs.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+                context.player1.clickCard(context.p2Base);
+
+                // Ability skips entirely, it is now player2's turn
+                expect(context.player2).toBeActivePlayer();
+                expect(context.player1.handSize).toBe(initialHandSize);
+                expect(context.p2Base.damage).toBe(5);
+            });
+        });
+    });
+});
```
VJubert, line 267: What happens when your base have only 2 hp left and no cards in deck ? Heal and draw should happens simulteanously ?

moyerr: They describe in the clarification that things which cannot happen simultaneously are done in printed order. Since deal & heal cannot be simultaneous, they have to be sequential, and the deal happens first (draw is printed first).

VJubert: Maybe you can add a test to enforce this behavior ?

moyerr: Added a test (`should defeat the base when it has 2 remaining HP and an empty deck...`) that sets the base to 2 remaining HP with an empty deck: activating the ability defeats the base because the empty-deck draw resolves (and deals its 3 damage) before the heal. Pushed in f073d51.

VJubert (approved):

PR #2617, opened 2026-07-05
Boba Fett

VJubert:

### DIFF
--- a/server/game/cards/09_IC27/units/BobaFettCompensatedIfHeDies.ts
+++ b/server/game/cards/09_IC27/units/BobaFettCompensatedIfHeDies.ts
@@ -1,7 +1,6 @@
 import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
 import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
 import type { IAbilityHelper } from '../../../AbilityHelper';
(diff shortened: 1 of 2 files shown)


server/game/cards/09_IC27/units/BobaFettCompensatedIfHeDies.ts
```diff
@@ -0,0 +1,33 @@
...
+        };
+    }
+
+    protected override setupStateWatchers(registrar: StateWatcherRegistrar, AbilityHelper: IAbilityHelper): void {
+        this.cardsDefeatedThisPhaseWatcher = AbilityHelper.stateWatchers.cardsDefeatedThisPhase();
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addWhenAttackEndsAbility({
+            title: `Ready ${TextHelper.resource(2)}`,
```
moyerr: Did the linter catch this or did you add it yourself? I think this is a scenario where we don't want the icon since the printed version of abilities like this don't use the icon

VJubert: That was me. I saw you add this for resources and I didn't think where to use or not use. Thanks for explanation ^^

moyerr (approved):

PR #2619, opened 2026-07-08
Bugfix: exclude upgrade attach targets when the upgrade can't be afforded on them

### LINKED ISSUE #1970: Upgrade targeting should not include units if costs cannot be paid to play the upgrade on them

### LINKED ISSUE #1971: Multiple upgrade cost adjusters with different targets may cause an upgrade to appear playable when it isn't
## Description

If there are multiple unit-specific cost adjusters for upgrade attach targeting in play, and the targets for the adjustments do not overlap, playing the upgrade may appear as a valid action, even though the cost cannot fully be paid for any single target.

For example, if Guardian of the Whills and any Mandalorian unit are in play. If the user has 3 ready resources, an out-of-aspect Darksaber will appear as playable, even though it would cost 4 on the Mandalorian unit and 5 on Guardian of the Whills.

## Test Setup

Notice how The Darksaber appears selectable, even though costs cannot be paid to play it on either unit. The action will be automatically cancelled if the player attempts to play it on either unit, but the engine should really be able to determine that it is not playable at all.

```json
{
    "phase": "action",
    "player1": {
        "resources": 3,
        "hand": [
            "the-darksaber"
        ],
        "groundArena": [
            "guardian-of-the-whills",
            "follower-of-the-way"
        ]
    }
}
```

moyerr:
Fixes #1970
Fixes #1971

## Problem

Upgrade target legality only checked `attachTargetCondition`, never per-target affordability. Targeted cost reductions (Guardian of the Whills, The Darksaber) apply only when the upgrade is attached to specific units, but the cost engine computes the *global minimum* cost — it applies a reduction if *any* unit qualifies. As a result:

- **#1970:** units the upgrade couldn't actually be afforded on were still offered as attach targets. Clicking one cancelled the action. Worse, via **The Armorer**'s play-from-resources ability, the play failed after committing costs while the "if you do" (resource the top card) still resolved — letting the player **ramp** a resource for free.
- **#1971:** when two different unit-specific reductions each apply to a different unit (e.g. Guardian's −1 and a Mandalorian's ignore-penalty), the global minimum combined both, so the upgrade appeared **playable from hand** even though it couldn't be afforded on any single target.

## Fix

Add a per-target affordability check to the upgrade attach `cardCondition` in `PlayUpgradeAction`. It's skipped at the post-cost `Target` stage, where resources are already spent and re-checking would incorrectly invalidate the chosen target.

This flows up through `meetsRequirements → hasSomeLegalTarget`, so both the selectable target set (#1970) and whether the card is playable at all (#1971) are now correct.

## Tests

- Updated the two tests that documented these bugs with TODOs (Guardian of the Whills, The Darksaber) to assert unaffordable units are no longer selectable.
- Added a regression test for The Armorer proving no ramp occurs.
- Added a regression test for the non-overlapping-adjuster scenario (#1971).
- Full suite green.

### DIFF
--- a/server/game/actions/PlayUpgradeAction.ts
+++ b/server/game/actions/PlayUpgradeAction.ts
@@ -3,7 +3,7 @@ import type { IPlayCardActionProperties, PlayCardContext } from '../core/ability
 import { PlayCardAction } from '../core/ability/PlayCardAction';
 import type { Card } from '../core/card/Card';
 import type { UpgradeCard } from '../core/card/UpgradeCard';
-import { AbilityRestriction, CardType, KeywordName, PlayType, RelativePlayer, ZoneName } from '../core/Constants';
+import { AbilityRestriction, CardType, KeywordName, PlayType, RelativePlayer, Stage, ZoneName } from '../core/Constants';
 import type { Restriction } from '../core/ongoingEffect/effectImpl/Restriction';
 import type { Player } from '../core/Player';
 import type { Game } from '../core/Game';
@@ -24,11 +24,19 @@ export class PlayUpgradeAction extends PlayCardAction {
                 ...properties,
                 targetResolver: {
                     activePromptTitle: `Attach ${card.title} to a unit`,
-                    cardCondition: (card, context) => (
-                        properties.attachTargetCondition
-                            ? properties.attachTargetCondition(card, context)
-                            : true
-                    ),
+                    cardCondition: (card, context) => {
+                        if (properties.attachTargetCondition && !properties.attachTargetCondition(card, context)) {
+                            return false;
+                        }
+
+                        // Exclude units the upgrade can't be afforded on when attached to them (see issue #1970).
+                        // Skipped at the post-cost Target stage, where resources are already spent.
+                        if (context.stage !== Stage.Target && !context.ability.canPayCosts(context)) {
+                            return false;
+                        }
+
+                        return true;
+                    },
                     immediateEffect: attachUpgrade<AbilityContext<UpgradeCard>>((context) => ({ upgrade: context.source }))
                 }
             }
--- a/server/game/actions/PlayUpgradeAction.ts
+++ b/server/game/actions/PlayUpgradeAction.ts
@@ -30,8 +30,8 @@ export class PlayUpgradeAction extends PlayCardAction {
                         }
 
                         // Exclude units the upgrade can't be afforded on when attached to them (see issue #1970).
-                        // Skipped at the post-cost Target stage, where resources are already spent.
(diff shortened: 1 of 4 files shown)


VJubert (dismissed):

server/game/actions/PlayUpgradeAction.ts
```diff
@@ -24,11 +24,19 @@ export class PlayUpgradeAction extends PlayCardAction {
...
-                            : true
-                    ),
+                    cardCondition: (card, context) => {
+                        if (properties.attachTargetCondition && !properties.attachTargetCondition(card, context)) {
+                            return false;
+                        }
+
+                        // Exclude units the upgrade can't be afforded on when attached to them (see issue #1970).
+                        // Skipped at the post-cost Target stage, where resources are already spent.
+                        if (context.stage !== Stage.Target && !context.ability.canPayCosts(context)) {
```
AMMayberry1: Would it work if we do `context.stage === Staget.Cost` instead, just to be as specific as possible

moyerr: Tried that first, but it turns the check into dead code — `cardCondition` only ever runs at the `preTarget` and `target` stages, never `cost` (cost payment goes through `cost.resolve()`, not the target resolver). The `Stage.Cost` you see is inside `canPayCosts`, which does `context.copy({ stage: Stage.Cost })` for its own eval — local copy, so the outer stage here is still preTarget/target.

It's phrased negatively because the `target` pass runs after costs are paid, so re-checking affordability there would invalidate the target we just paid for. Hence `!== Stage.Target`.

AMMayberry1 (approved):

PR #2620, opened 2026-07-08
Add ongoing effects to player summary

moyerr:
⚠️ Client PR: SWU-Karabast/forceteki-client#740 ⚠️ 

## Description

- Adds an ongoing effect summary to the player state for rendering active constant, lasting, and delayed effects on the game board
- Adds a mechanism for describing the ongoing effects (custom title > ability title > ongoing effect description)
- Adds a custom title property to the Lasting Effect props, mirroring the existing functionality we have for delayed effects
- Updates many lasting effect and ability titles for clarity when rendered on the game board
- Provides custom matchers for asserting ongoing effect state in tests


AMMayberry1:
For the DJ effect, I know that technically there isn't a "target" property but I'm wondering if it would be worth somehow marking the stolen resource as the target so that it will be highlighted for clarity. I think that's how most players would expect it to work, as opposed to right now where it just says "no target"

image

AMMayberry1:
For lasting effects, I think it would be cool if we could include the expiration condition in the text somehow. Like if DJ's box said "when defeated" or Han1's said "when readying resources" or something

image

AMMayberry1:
Any idea why Tech's ability doesn't present as having any targets?

image

AMMayberry1:
For "give keyword" and similar, it would be cool if we could actually name the keyword (or trait or whatever)

image

danbastin (approved):

PR #2622, opened 2026-07-09
⟡ Darth Sidious, Move Against the Jedi

moyerr:
| ⟡ Darth Sidious |
| :---: |
| Move Against the Jedi |
|  |

### DIFF
--- a/test/server/cards/09_IC27/units/DarthSidiousMoveAgainstTheJedi.spec.ts
+++ b/test/server/cards/09_IC27/units/DarthSidiousMoveAgainstTheJedi.spec.ts
@@ -1,0 +1,215 @@
+describe('Darth Sidious, Move Against the Jedi', function() {
+    integration(function(contextRef) {
+        describe('Sidious\'s heal-triggered ability', function() {
+            it('deals damage equal to the amount healed when he attacks and heals via Restore', async function () {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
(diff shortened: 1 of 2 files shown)


test/server/cards/09_IC27/units/DarthSidiousMoveAgainstTheJedi.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+                    [context.p1Base, 3]
+                ]));
+
+                expect(context.p1Base.damage).toBe(2);
+                expect(context.wampa.damage).toBe(0);
+                expect(context.player1).toBeActivePlayer();
+            });
+        });
+    });
+});
```
VJubert, line 272: can you add some tests:
- when you can't heal your base (confederate tri-fighter or wolffe ability)
- when you heal your base from event (grassroot resistance ?)

moyerr: Added both tests:
- `does not trigger when the base can't be healed` — uses Confederate Tri-Fighter ("Bases can't be healed.") so Restore heals nothing and Sidious doesn't trigger.
- `triggers off base healing from an event` — uses Grassroots Resistance to heal the base, which triggers Sidious to deal 3 to an enemy unit.

Pushed in b8cd90a.

VJubert (approved):

PR #2626, opened 2026-07-13
Bugfix: Introduce a PlayMultipleCardsFromDiscardSystem to fix Dathomiri Magicks

moyerr:
## Description

Dathomiri Magics plays multiple cards from discard. Because these modified play-a-card actions happen sequentially, they should each happen after the previous one has fully resolved, possibly adding more cards to the discard pile. The initial implementation of this forced the player to choose all 3 cards they wanted to play at the start of the ability.

Because this type of effect is easy to get wrong, and difficult to do correctly (as evidenced by LOF Kylo Ren leader's implementation), this seemed like a good candidate for a dedicated game system. The system guarantees that the plays happen sequentially, and fully resolve the resulting triggers before choosing the next card to play.

Both Dathomiri Magicks and LOF Kylo Ren leader's implementations have been updated to use the new system, and are massively simplified as a result.

### DIFF
--- a/server/game/gameSystems/PlayMultipleCardsFromDiscardSystem.ts
+++ b/server/game/gameSystems/PlayMultipleCardsFromDiscardSystem.ts
@@ -1,0 +1,158 @@
+import type { AbilityContext } from '../core/ability/AbilityContext';
+import type { Card } from '../core/card/Card';
+import type { CardTypeFilter } from '../core/Constants';
+import { GameStateChangeRequired, MetaEventName, PlayType, RelativePlayer, TargetMode, ZoneName } from '../core/Constants';
+import type { GameEvent } from '../core/event/GameEvent';
+import type { IGameSystemProperties } from '../core/gameSystem/GameSystem';
+import { GameSystem } from '../core/gameSystem/GameSystem';
+import type { GameObject } from '../core/GameObject';
+import type { Player } from '../core/Player';
+import { Helpers } from '../core/utils/Helpers';
+import type { IPlayCardProperties } from './PlayCardSystem';
+import { PlayCardSystem } from './PlayCardSystem';
+import type { ISelectCardProperties } from './SelectCardSystem';
+import { SelectCardSystem } from './SelectCardSystem';
+
+/**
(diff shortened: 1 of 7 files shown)


server/game/gameSystems/PlayMultipleCardsFromDiscardSystem.ts
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * The subset of {@link IPlayCardProperties} forwarded to the {@link PlayCardSystem} that plays each individual card.
+ * `playType` and `nested` are always set by this system, so they're intentionally not included here.
+ * - `playAsType`: the type each card is played as (unit, upgrade, etc.)
+ * - `adjustCost`: cost adjustment applied to each card (e.g. play for free); omit to pay the printed cost
+ * - `attachTargetCondition`: restricts the unit a played upgrade / pilot may attach to (e.g. "on this unit")
+ */
+type ForwardedPlayCardProperties = Pick<IPlayCardProperties, 'playAsType' | 'adjustCost' | 'attachTargetCondition'>;
+
+export interface IPlayMultipleCardsFromDiscardProperties<TContext extends AbilityContext = AbilityContext> extends IGameSystemProperties, ForwardedPlayCardProperties {
```
VJubert: I don't know if it's possible but shouldn't it be a pick on union of IPlayCardProperties and ISelectCardProperties ? (maxcards, cardcondition, activeprompttitle and cardtypefilter seems to be very common)

moyerr: Good call — done. I picked `activePromptTitle`, `cardCondition`, and `cardTypeFilter` from `ISelectCardProperties` (mirroring the existing `IPlayCardProperties` pick) so they stay in sync with the selection layer.

I left `maxCards` as a local prop, though: it isn't a select/play passthrough but the bound on this system's own selection loop. The analogous `ISelectCardProperties.numCards` only applies to the `UpTo`/`Exactly` batch-select modes, and the whole point here is that we deliberately don't batch-select — each card is chosen via an independent single-card selection so the discard pile is re-evaluated between plays.

AMMayberry1 (approved):

VJubert (approved):

PR #2648, opened 2026-08-06
Add Plot declare-then-order step matching paper rules (SWU 7.6.2b)

priisig:
## Summary

Star Wars Unlimited's comprehensive rule 7.6.2b for Plot states:

> After deploying a leader, show your opponent each card with Plot you plan to play, then play them one at a time in any order. Any abilities triggered by playing a card with Plot must be resolved before playing the next card with Plot.

The previous implementation didn't match this: when a leader deployed with multiple Plot-eligible resources, each Plot trigger was just added to the normal simultaneous-trigger resolution list, and the player decided whether to play each one (via a per-card Trigger/Pass prompt) only as they got to it in the ordering step. There was no upfront "declare which cards you intend to play" step, and no way to decline Plot entirely in one action when multiple cards were eligible — you had to Pass on each one individually.

This PR adds a one-time **declare step**: after deploying a leader, if 2 or more resource-zone cards with Plot are eligible, the player first selects the whole subset of cards they intend to play via Plot (including "none of them," in a single action). Only afterward do they choose the order to actually play the declared cards, one at a time — with any other simultaneous triggers (Shielded, leader deploy abilities, etc.) still freely interleavable in that ordering step, exactly as before.

For the common case of exactly **one** eligible Plot card, the existing 1-click accept / 1-click decline (Trigger/Pass) flow is unchanged — the new declare step only kicks in when there's an actual subset to choose from.

## How it works

- **`TriggeredAbilityBase`** (`server/game/core/ability/TriggeredAbility.ts`) now exposes a `keyword` field, read from the ability's `properties.keyword`. This lets generic engine code identify "this pending trigger is a Plot ability" without importing the keyword-specific `PlotAbility` class. (Note: keyword-ability classes like `PlotAbility`, `ShieldedAbility`, etc. are constructed via a shared generic `TriggeredAbility` class in production — their own subclass constructors are effectively dead code — so the `keyword` has to be threaded through the shared props object rather than set in a subclass field. `PlotAbility.buildPlotAbilityProperties()` was updated accordingly; the other keyword classes only needed an `override` modifier to compile against the new base field.)
- **`TriggeredAbilityContext`** gained a `preConfirmed` flag (`markPreConfirmed()` / `preConfirmed` getter), mirroring the existing `overrideTitle` pattern.
- **`AbilityResolver`** skips the per-ability Trigger/Pass sub-prompt when the context is pre-confirmed, since that decision was already made during the declare step.
- **`TriggerWindowBase`** runs the new declare step once per resolving player per window (tracked via a `plotDeclarationComplete` set), using the same "select any number of cards" display-prompt style used elsewhere (e.g. Chancellor Palpatine's search-and-reveal ability) so only the actual Plot-eligible cards are shown — not the player's whole resource zone. Declared cards are marked pre-confirmed; undeclared cards are removed from the trigger window entirely so they're never asked about individually. Everything downstream (the multi-trigger ordering prompt, interleaving with Shielded/leader abilities, etc.) is untouched.

## Test plan

- [x] Updated `test/server/core/abilities/keyword/Plot.spec.ts` for the new declare-step flow, plus a new test asserting Plot can be declined entirely in one click even with 2+ eligible cards
- [x] Updated `ChancellorPalpatineHowLibertyDies.spec.ts` and `BailOrganaDoingEverythingHeCan.spec.ts` (both exercise 2+ simultaneous Plot cards)
- [x] Full test suite: 7513 specs, 0 failures (9 pre-existing/unrelated pending specs)
- [x] Manually verified via a local dev game (leader with 2 off-aspect Plot-eligible resources + a leader deploy ability like Support) that the declare step shows only the eligible cards, supports "choose nothing," and that declared cards interleave correctly with other simultaneous triggers in any order

## Screenshots
image
image

### DIFF
--- a/server/game/core/gameSteps/abilityWindow/TriggerWindowBase.ts
+++ b/server/game/core/gameSteps/abilityWindow/TriggerWindowBase.ts
@@ -1,7 +1,7 @@
 import type { Player } from '../../Player';
 import type { GameEvent } from '../../event/GameEvent';
 import type { EventWindow } from '../../event/EventWindow';
-import { AbilityType, KeywordName, SubStepCheck } from '../../Constants';
+import { AbilityType, SubStepCheck } from '../../Constants';
 import { Contract } from '../../utils/Contract';
 import type { TriggeredAbilityContext } from '../../ability/TriggeredAbilityContext';
 import type { TriggeredAbilityBase } from '../../ability/TriggeredAbility';
@@ -12,7 +12,7 @@ import type { Game } from '../../Game';
 import { TriggeredAbilityResolutionPrompt } from '../prompts/TriggeredAbilityResolutionPrompt';
 import { BatchTriggerResolutionPrompt } from '../prompts/BatchTriggerResolutionPrompt';
 import type { IResolutionChoice, ITriggerWindowSourceCard } from '../PromptInterfaces';
-import { TextHelper } from '../../utils/TextHelper';
+import { EnumHelpers } from '../../utils/EnumHelpers';
 
 /** Builds the lightweight card summary attached to trigger-style prompt buttons. */
 export function getTriggerSourceCardSummary(card: Card): ITriggerWindowSourceCard {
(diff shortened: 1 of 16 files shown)


server/game/core/gameSteps/abilityWindow/TriggerWindowBase.ts
```diff
@@ -185,6 +192,19 @@ export abstract class TriggerWindowBase extends BaseStep {
             abilitiesToResolve = this.unresolved.get(this.currentlyResolvingPlayer);
         }
 
+        if (!this.plotDeclarationComplete.has(this.currentlyResolvingPlayer)) {
```
danbastin: Is there any way to generalize this step? Looking at the rest of the file it doesn't seem like we reference any particular key words and it would be nice to try and keep it that way so that we avoid allowing specific systems to leak into the Base.

priisig: Yeah, that makes sense. I tried to decouple it from the keyword and make it more generic, so it would also be usable if, e.g., an ability wants to use the feature. So now the caller passes the display label in, and TriggerWindowBase only handles the declare logic

moyerr:
After discussing in discord, we will close this in favor of a streamlined simultaneous trigger prompt. We can revisit later if we still feel its necessary to add the in-paper reveal step up front

PR #2650, opened 2026-08-07
Beast, Weakness, Clone X Assassin & Ezra Bridger

VJubert:

### DIFF
--- a/eslint-rules/no-raw-token-text.mjs
+++ b/eslint-rules/no-raw-token-text.mjs
@@ -51,6 +51,7 @@ const TOKEN_CATEGORIES = [
             'TIE Fighter',
             // "Mandalorian" is both a trait and a token; skip only the token ("Mandalorian token(s)").
             'Mandalorian\\s+tokens?',
+            'Beast\\s+tokens?',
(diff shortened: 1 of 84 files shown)


eslint-rules/no-raw-token-text.mjs
```diff
@@ -51,6 +51,7 @@ const TOKEN_CATEGORIES = [
             'TIE Fighter',
             // "Mandalorian" is both a trait and a token; skip only the token ("Mandalorian token(s)").
             'Mandalorian\\s+tokens?',
+            'Beast\\s+tokens?',
```
moyerr: I don't think this is needed because Beast is not a trait. The cards/regex patterns you see here are to disambiguate card names that happen to contain Traits within their name (e.g. `TIE Fighter` token name contains the string for the trait `Fighter`)

moyerr (commented):
Can we also ensure that the new tokens can be created via a test setup? There's a `TokenUnits.spec.js` where you can add the Beast. I don't think there is an equivalent for Weakness, so that can just have its own small spec file.

moyerr (approved):

PR #2651, opened 2026-08-07
HMW Chewbacca leader

VJubert:


test/server/cards/09_HMW/leaders/ChewbaccaRelentlessRebel.spec.ts
```diff
@@ -0,0 +1,85 @@
+describe('Chewbacca, Relentless Rebel', function () {
+    integration(function (contextRef) {
+        const promptAbility = 'Attack with a unit, even if it\'s exhausted. It can\'t attack bases for this attack';
+
+        it('Chewbacca\'s leader side ability should initiate an attack with a unit (even if it\'s exhausted). It cannot attack base for this attack', async function () {
```
moyerr: Looks like both test cases attack with a ready unit. Can we choose an exhausted one to attack with?

Also wondering if we can add a test for a unit that cannot attack (Loth Wolf)?

moyerr (approved):

PR #2653, opened 2026-08-07
Fix two flaky shuffle/RNG tests

moyerr:
Fixes two intermittent test failures, both caused by the RNG not being pinned before a randomized step.

## Bounty Posting (`BountyPosting.spec.ts`)
The first spec compared the post-shuffle deck against a pre-shuffle reference — but that reference is a *live view* of the same 5 remaining cards (Top Target having been drawn out of both), differing only in order. The shuffle reproduces the original order roughly 1 in 120 runs (5!), causing an intermittent failure.

**Fix:** seed the RNG (`setRandomSeed('12345')`) and assert the exact resulting deck order — deterministic and strictly stronger than "it changed".

## DelayedEffectSystem (`DelayedEffectSystem.spec.ts`)
In the unique-rule describe block, the seed was set *after* the first DJ was already played, so the first steal was unseeded and nondeterministic — and the "seeded" second steal drew from a pool whose composition depended on it. The second DJ re-stole the same resource ~1/10 runs, failing the `not.toBe` check.

**Fix:** move `setRandomSeed('12345')` above the first steal so the entire sequence is deterministic.

Both follow the seeded-shuffle pattern used in `AnnihilatorTaggesFlagship.spec.ts`.

### DIFF
--- a/test/server/cards/02_SHD/events/BountyPosting.spec.ts
+++ b/test/server/cards/02_SHD/events/BountyPosting.spec.ts
@@ -16,7 +16,8 @@ describe('Bounty Posting', function() {
 
                 const { context } = contextRef;
 
-                const preShuffleDeck = context.player1.deck;
+                // Seeded so the shuffle result is identical on every run.
+                context.game.setRandomSeed('12345');
 
                 context.player1.clickCard(context.bountyPosting);
                 expect(context.player1).toHaveExactDisplayPromptCards({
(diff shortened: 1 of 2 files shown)


jamboriu:
Hello! I am reviewing the intermittent test failures in `BountyPosting.spec.ts` caused by live view reference mutation during shuffle. I plan to fix the RNG pin and isolate the array snapshot before shuffling. May I handle this issue?

/attempt

danbastin (approved):

PR #2658, opened 2026-08-11
Bases' trait, Adamant Ewoks & Wookiee Rangers

VJubert:


eslint-rules/no-raw-token-text.mjs
```diff
@@ -36,6 +36,14 @@ const TOKEN_CATEGORIES = [
...
             'Transport', 'Trick', 'Trooper', 'Tusken', 'Twi\'lek', 'Undead', 'Underworld', 'Vehicle',
             'Walker', 'Weapon', 'Wookiee',
+            // bases traits
+            'Aldhani', 'Atollon', 'Bracca', 'Cantonica', 'Castilon', 'Christophsis', 'CloudCity', 'Concordia',
+            'Coruscant', 'Corellia', 'Dagobah', 'Dathomir', 'Death Star', 'D\'Qar', 'Eadu', 'Endor', 'Ferrix',
+            'Geonosis', 'Hoth', 'Hosnian Prime', 'Ilum', 'Jedha', 'Kamino', 'Kashyyk', 'Kessel', 'Kalevala',
+            'Lothal', 'Lowick', 'Malachor', 'Mandalore', 'Mortis', 'Mustafar', 'Nadiri', 'Narkina 5', 'Nevarro',
+            'ObaDiah', 'Onderon', 'Peridea', 'Pillio', 'Quarzite', 'Ryloth', 'Scarif', 'Seatos', 'Segra Milo',
+            'Serenno', 'Sorgan', 'Starkiller Base', 'Starlight Beacon', 'StygeonPrime', 'Takodana', 'Tatooine',
+            'Utapau', 'Vardos', 'Vassek', 'Wayland', 'Yavin 4', 'Zanbar', 'Zeffo',
```
moyerr, line 46: Need spaces: Cloud City, Oba Diah, Stygeon Prime
Typo: Kashyyyk has 3 y's

scripts/fetchdata.js
```diff
@@ -210,6 +210,332 @@ function populateMissingData(attributes, id) {
...
+        case '7297371836': // Imperial Command Complex: Lothal
+            attributes.traits = {
+                data: [{ attributes: { name: 'Lothal' } }]
+            };
+            break;
+        case '3810584393': // Dagobah Swamp: Dagobah
+        case '9756609647': // Dragonsnake Bog: Dagobah
+            attributes.traits = {
+                data: [{ attributes: { name: 'Dagobah' } }]
+            };
```
moyerr: These are a bit tedious. Would it be better for this to live in a file that contains the mappings?

scripts/mockdata.js
```diff
@@ -324,7 +324,7 @@ const mockCards = [
         hp: 30,
         hasNonKeywordAbility: false,
         aspects: ['vigilance'],
-        traits: [],
+        traits: ['tatooine'],
```
moyerr, line 327: In the fetchdata script, can we emit an error (or warning) for any bases that are missing a trait? That way we don't forget to add these during preview season.

We would have to do it _after_ we add the mocks, so it would catch real-data and mock-data cards.

server/game/cards/04_JTL/upgrades/DeathStarPlans.ts
```diff
@@ -21,6 +21,7 @@ export default class DeathStarPlans extends UpgradeCard {
 
     public override setupCardAbilities(registrar: IUpgradeAbilityRegistrar, AbilityHelper: IAbilityHelper) {
         registrar.addTriggeredAbility({
+            // eslint-disable-next-line forceteki/no-raw-token-text
             title: 'Take control of Death Star Plans and attach it to a unit you control',
```
moyerr, line 25: Not completely necessary since this is the only conflict, but we could add an ignore pattern in the lint rule.  So it doesn't catch Death Star Plans.

server/game/core/Constants.ts
```diff
@@ -477,6 +477,66 @@ export enum Trait {
...
+    Eadu = 'eadu',
+    Endor = 'endor',
+    Ferrix = 'ferrix',
+    Geonosis = 'geonosis',
+    Hoth = 'hoth',
+    HosnianPrime = 'hosnian prime',
+    Ilum = 'ilum',
+    Jedha = 'jedha',
+    Kamino = 'kamino',
+    Kashyyk = 'kashyyk',
```
moyerr: Kashyyyk

server/game/cards/04_JTL/upgrades/DeathStarPlans.ts
VJubert: I would have done it if there were lots of files referring to the Death Star plans, but it's the only one

scripts/fetchdata.js
```diff
@@ -501,6 +507,16 @@ async function main() {
...
 
+    // Validate that all bases have traits
+    const basesWithoutTraits = uniqueCards.filter((card) => card.types.includes('base') && (!card.traits || card.traits.length === 0));
+    if (basesWithoutTraits.length > 0) {
+        console.error('\nERROR: The following bases are missing traits:');
+        for (const base of basesWithoutTraits) {
+            console.error(`  - ${base.title} (${base.setId.set}_${String(base.setId.number).padStart(3, '0')})`);
+        }
+        throw new Error('Bases are missing traits. Please add traits to these bases.');
+    }
```
moyerr, line 518: Can we give instruction for _where_ the developer needs to add them? Probably worth calling out both `mockdata.js` and `baseTraits.js`

Also we can use `makeSetCodeString(base.setId)` here

moyerr (approved):

Issue #2588, opened 2026-06-14
LKI cost value is not captured correctly for Clone


AMMayberry1:
The LKI for the cost value of Clone is not captured correctly when it leaves the field. Due to the timing of the effect falling off, for some reason the LKI grabs the cost of Clone instead of the cost of the card it's cloning. This is an issue for things like Purrgil Ultra.

moyerr:
Fixed in #2659

PR #2659, opened 2026-08-12
Fix attachment and reading of LKI (#2588)

hibou-fiscal:
Currently, we do not calculate the LKI cost value for Clone across several cards (https://github.com/SWU-Karabast/forceteki/issues/2588). This is because of two reasons. 

Firstly, Drengir Spawn and Targeted for Removal do not correctly read LKI, and instead read the current state of the card. These cards are fixed by changing their abilities to read from the LKI instead.

Secondly, LKI was not available for Purgill Ultra and Qui-Gon Jinn because in returning units to hand, they trigger `OnCardMoved`, and LKI is only added for `OnCardDefeated`. This is fixed here by moving `addLastKnownInformationToEvent` to `addLeavesPlayPropertiesToEvent` instead of having it only in `DefeatCardSystem`, which provides the LKI information they can then read from.


moyerr (approved):
Thanks for your contribution!

This fixes #2588

PR #2671, opened 2026-08-14
Refine ongoing effect summaries: suppress low-value effects

moyerr:
## Description

Follow-up refinements to the recently released **Ongoing Effect Summaries** feature, based on player feedback that certain effects are noise on the board. The summary now suppresses effects that carry no useful board information, while keeping genuinely informative ones.

### Suppressed everywhere
- **Self cost adjusters** (e.g. Mastery) — abilities that only change the cost to play the source card itself. Implemented via a new `omitFromOngoingEffectSummary` flag on the constant ability, set centrally in `buildCostAdjusterAbilityProps`, so it applies to every self cost-adjuster / aspect-penalty-ignore helper without per-card edits.
- **"Enters play ready"** constant abilities (e.g. Millennium Falcon) — they only describe how a card already entered play.
- **Setup-only base abilities** (`ModifyStartingHandSize`, `NoMulligan` — e.g. Colossus, Nabat Village) — inert for the entire rest of the game.

### Gated to the zone they matter in
- **Play-time modifiers** (e.g. R2-D2's "can be played on a Vehicle with a Pilot") are surfaced only while the source card is in a zone it can be played from (hand), and suppressed once it's in play or elsewhere. Handled by a small effect-type → relevant-zone map that's easy to extend.

### Still shown (intentionally not over-filtered)
- Temporary cost reductions (e.g. GNK Power Droid's "next unit costs 1 less this phase").
- Constant-ability cost adjusters that discount **other** cards (leaders/units like Piett, Jabba, Wedge, Krennic).

Also includes minor ongoing-effect description tweaks (Colossus, Grand Inquisitor, Marrok) for clearer board rendering.

## Testing
- `test/server/core/ongoingEffects/OngoingEffectSummary.spec.ts` extended with coverage for each exclusion category, the play-time-modifier zone gating (shown in hand / suppressed in play), and positive cases that must still surface. Zone-visibility tests were reworked onto a non-excluded from-any-zone effect (R2-D2). All 22 specs pass; lint clean.

### DIFF
--- a/server/game/core/ability/ConstantAbility.ts
+++ b/server/game/core/ability/ConstantAbility.ts
@@ -61,6 +61,10 @@ export class ConstantAbility extends GameObjectBase implements IConstantAbility 
         return this.properties.gainAbilitySource;
     }
 
+    public get omitFromOngoingEffectSummary(): boolean {
+        return this.properties.omitFromOngoingEffectSummary ?? false;
+    }
+
     public get ongoingEffect(): IOngoingEffectGenerator | IOngoingEffectGenerator[] {
         return this.properties.ongoingEffect;
     }
--- a/server/game/core/ongoingEffect/OngoingEffectEngine.ts
+++ b/server/game/core/ongoingEffect/OngoingEffectEngine.ts
@@ -94,9 +94,9 @@ function findRegisteringConstantAbility(effect: OngoingEffect) {
 }
 
 /**
- * Effect types that never carry useful board information and are hidden from the summary per player
- * feedback: "enters play ready" only describes how a card already entered play (already resolved), and
- * the starting-hand-size / no-mulligan base modifiers are setup-only and inert for the rest of the game.
+ * Effect types that are never included in the summary: they describe setup or how a card enters play
+ * rather than the ongoing board state ("enters play ready", and the starting-hand-size / no-mulligan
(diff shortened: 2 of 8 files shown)


server/game/core/ability/ConstantAbility.ts
```diff
@@ -61,6 +61,10 @@ export class ConstantAbility extends GameObjectBase implements IConstantAbility
         return this.properties.gainAbilitySource;
     }
 
+    public get omitFromOngoingEffectSummary(): boolean {
+        return this.properties.omitFromOngoingEffectSummary ?? false;
```
danbastin, line 65: 👍

server/game/core/ongoingEffect/OngoingEffectEngine.ts
```diff
@@ -93,6 +93,58 @@ function findRegisteringConstantAbility(effect: OngoingEffect) {
         .find((ability) => ability.registeredEffects?.includes(effect));
 }
 
+/**
+ * Effect types that never carry useful board information and are hidden from the summary per player
```
danbastin: Can we clean up some of the comments in this file to be more decriptive of the consts and what type of effects should end up here?

danbastin (approved):

PR #2678, opened 2026-08-16
Darth Vader

VJubert:


test/server/cards/09_HMW/units/DarthVaderAnyMethodsNecessary.spec.ts
```diff
@@ -0,0 +1,74 @@
+describe('Darth Vader Any Methods Necessary', function() {
+    integration(function(contextRef) {
+        describe('Darth Vader Any Methods Necessary\'s ability', function() {
```
moyerr, line 3: Can we get some tests with timing, as it relates to the damage? Because this is a modified play-a-card action, the damage should happen before any triggered abilities. So for example, Shielded units would get the damage before the shield.

Judges said it basically goes:

1. Play 1st unit and deal 2 damage to it
2. Resolve any triggered abilities that result from playing 1st unit
3. Play 2nd unit and deal 2 damage to it
4. Resolve any triggered abilities that result from playing 2nd unit

server/game/cards/09_HMW/units/DarthVaderAnyMethodsNecessary.ts
```diff
@@ -0,0 +1,28 @@
...
+            title: 'Search the top 8 cards of your deck for up to 2 units that each cost 4 or less, play them for free, and deal 2 damage to each of them',
+            immediateEffect: abilityHelper.immediateEffects.playMultipleCardsFromDeck({
+                searchCount: 8,
+                selectCount: 2,
+                canChooseFewer: true,
+                cardCondition: (card) => card.isUnit() && card.cost <= 4,
+                playedCardEnterPlayEffect: abilityHelper.immediateEffects.damage((context) => ({
+                    amount: 2,
+                    target: context.source
+                })),
```
moyerr: These effects always target the the played card, so we can drop the `context.source` here.

server/game/gameSystems/PlayMultipleCardsFromDeckSystem.ts
```diff
@@ -15,6 +16,12 @@ export interface IPlayMultipleCardsFromDeckProperties<TContext extends AbilityCo
       | 'remainingCardsHandler'> {
     multiSelectCondition?: (card: Card, currentlySelectedCards: Card[], context: TContext) => boolean;
     playAsType?: WildcardCardType.Upgrade | WildcardCardType.Unit | CardType.Event;
+
+    /**
+     * Effect(s) resolved for each played units before any trigger and next units
+     * See {@link IPutIntoPlayProperties.enterPlayEffect}.
```
moyerr: Grammar nit: "Effect(s) resolved for each played unit as they enter play, before any triggers and before subsequent units enter play"

test/server/cards/09_HMW/units/DarthVaderAnyMethodsNecessary.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+                context.player1.clickCardInDisplayCardPrompt(context.tradeFederationShuttle);
+                context.player1.clickPrompt('Play cards in selection order');
+
+                expect(context.heroicArc170.damage).toBe(2);
+                expect(context.player1).toBeAbleToSelectExactly([context.battlefieldMarine]);
+                context.player1.clickCard(context.battlefieldMarine);
+
+                expect(context.tradeFederationShuttle.damage).toBe(2);
+                expect(context.player2).toBeActivePlayer();
+            });
```
moyerr: Can we make it clearer what specifically is being tested in these cases? The test case names are the same, and the prompt steps are not clear.

moyerr (approved):

PR #2690, opened 2026-08-18
Carbonite Chamber & Trap Field

VJubert:


server/game/cards/09_HMW/upgrades/CarboniteChamber.ts
```diff
@@ -0,0 +1,29 @@
...
+        registrar.addActionAbility({
+            title: `Choose a non-${TextHelper.Trait.Vehicle} unit. It doesn't ready during the next regroup phase`,
+            cost: [abilityHelper.costs.defeatSelf()],
+            targetResolver: {
+                cardTypeFilter: WildcardCardType.Unit,
+                cardCondition: (card) => !card.hasSomeTrait(Trait.Vehicle),
+                immediateEffect: abilityHelper.immediateEffects.forThisRoundCardEffect({
+                    effect: abilityHelper.ongoingEffects.cardCannot(AbilityRestriction.DoesNotReadyDuringRegroup),
+                    ongoingEffectDescription: 'prevent {0} from readying',
+                })
```
moyerr: The card only says "the next regroup phase", so I think `forThisRound` is too long.

Max Rebo can create an additional regroup phase, and the chosen unit should be able to ready in the second one based on the "next" wording.

test/server/cards/09_HMW/upgrades/TrapField.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+            const { context } = contextRef;
+
+            context.player2.clickCard(context.dedraMeero);
+            context.player2.clickCard(context.p1Base);
+            expect(context.player1).toHavePassAbilityPrompt('Defeat this upgrade to deal 3 damage to Spy');
+            context.player1.clickPrompt('Trigger');
+
+            expect(context.player1).toBeActivePlayer();
+            expect(context.trapField).toBeInZone('discard');
+            expect(() => context.player1.findCardByName('spy')).toThrowError('Could not find any cards matching name spy');
```
moyerr: It's P2's Dedra, so it would be P2's Spy. And i think i will be able to find it in the `outsideTheGame` zone

moyerr (approved):

PR #2694, opened 2026-08-18
⟡ Vernestra Rwoh, We Should Handle This Ourselves

moyerr:
| ⟡ Vernestra Rwoh |
| :---: |
| We Should Handle This Ourselves |
|  |

## Engine tweaks
- Add ability registration support for additional costs (Vernestra) and alternate costs (Bamboozle) to make implementations easier
- Add mulitple-card support for the `CopyStandardTriggeredAbilitiesEffect`


server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -224,6 +252,86 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
...
+                title: alternate.title,
+                additionalCosts: this.buildPlayCost(alternate),
+                costAdjusters: [CostAdjusterFactory.create(this.game, this, { costAdjustType: CostAdjustType.Free })],
+            })
+        );
+    }
+
+    /**
+     * The additional play costs registered on this card (see `addAdditionalPlayCost`). These are card
+     * abilities, so a card blanked out of play (all abilities lost) has none.
```
AMMayberry1, line 347: This doesn't feel like it's generically true - there could definitely be play costs that are imposed on the card by other effects. I see the comment above for Saw which I'm assuming is related to that, but worth calling out and maybe leaving a note on that in the comment here as well

server/game/Interfaces.ts
```diff
@@ -187,6 +187,28 @@ export interface IAbilityPropsWithSystems<TContext extends AbilityContext> exten
     initiateAttack?: IInitiateAttackProperties | ((context: TContext) => IInitiateAttackProperties);
 }
 
+/** Interface definition for addAdditionalPlayCost and addAlternatePlayCost abilities */
+export type IPlayCostProperties<TSource extends Card = Card> = { title?: string } & (
```
AMMayberry1: Could we refactor this definition to use interface inheritance like the other ones in this file? That would help consolidate the shared properties a bit and make it easier to avoid typos if we need to add other variations

server/game/Interfaces.ts
```diff
@@ -187,6 +187,28 @@ export interface IAbilityPropsWithSystems<TContext extends AbilityContext> exten
...
+  | {
+      cost: ICost<AbilityContext<TSource>> | ICost<AbilityContext<TSource>>[];
+      costName?: undefined;
+      immediateEffect?: undefined;
+      targetResolver?: undefined;
```
AMMayberry1: Is there any way to make these not have to be undefined?

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
```diff
@@ -194,6 +212,16 @@ export class PlayableOrDeployableCard extends Card implements IPlayableOrDeploya
 
         const actions: PlayCardAction[] = [defaultPlayAction];
 
+        // Alternate play costs are offered alongside the default for the base play types only (not for
+        // keyword-alternate play types like Smuggle/Piloting), and not when the card is blanked out of play.
+        if (
+            (playType === PlayType.PlayFromHand || playType === PlayType.PlayFromOutOfPlay) &&
```
AMMayberry1, line 229: Don't necessarily need to address it this PR, but if this concept of "base play types" comes up more frequently it may be worth it to massage the enums a bit to add a concept of that

server/game/cards/09_HMW/units/VernestraRwohWeShouldHandleThisOurselves.ts
```diff
@@ -0,0 +1,48 @@
...
+export default class VernestraRwohWeShouldHandleThisOurselves extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: 'vernestra-rwoh#we-should-handle-this-ourselves-id',
+            internalName: 'vernestra-rwoh#we-should-handle-this-ourselves',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        const chosenUnitsCostName = 'unitsPutOnBottomOfDeck';
```
AMMayberry1, line 17: Not sure how feasible this is, but I wonder if we could have issues where costs from two different sources end up having the same name and causing collisions. If there's a way we could check and throw an error if a cost is being added which already exists with the same name, that might be a useful guardrail against an issue that'd be kind of painful to debug

server/game/Interfaces.ts
moyerr: This (as I understand it) is a compiler trick to ensure users of the interface can _only_ supply `exactly` one of: `targetResolver`, `immediateEffect`, or `cost`. Is there a better way to do this?

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
moyerr: Yeah, I think this is just poor naming. This is really "additional cost _abilities_" and should be blanked while out of play just like any other printed ability.

moyerr: Renamed to `additionalPlayCostAbilities` and reworded the doc to call these out as card abilities (blanked out of play), distinct from effect-imposed costs like Saw's.

server/game/Interfaces.ts
moyerr: Done — split into `IPlayCostPropertiesBase` plus three variant interfaces that extend it.

server/game/Interfaces.ts
moyerr: Replaced the `undefined` keys with an `ExclusiveUnion` helper. Still enforces exactly-one-of, without the explicit `undefined`s.

server/game/core/card/baseClasses/PlayableOrDeployableCard.ts
moyerr: Agreed, leaving this for a follow-up if base play types come up again.

server/game/cards/09_HMW/units/VernestraRwohWeShouldHandleThisOurselves.ts
moyerr: Added `assertPlayCostNameUnused`, so now registering a duplicate `costName` on a card throws. It only covers a card's own play costs. Something global would require more changes since there's no guaranteed way to derive a cost name after it is created. We'd probably need a method on `ICost` to get there.

AMMayberry1 (approved):

PR #2713, opened 2026-08-22
⟡ Darth Sidious, There is No Mercy

moyerr:
| ⟡ Darth Sidious |
| :---: |
| There is No Mercy |
|  |
|  |

### DIFF
--- a/test/server/cards/09_HMW/leaders/DarthSidiousThereIsNoMercy.spec.ts
+++ b/test/server/cards/09_HMW/leaders/DarthSidiousThereIsNoMercy.spec.ts
@@ -1,0 +1,463 @@
+describe('Darth Sidious, There is No Mercy', function() {
+    integration(function(contextRef) {
+        describe('Darth Sidious\'s leader side ability', function() {
+            const exhaustAbilityPrompt = 'Exhaust this leader to deal 1 damage to a different unit or base';
+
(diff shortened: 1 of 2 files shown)


test/server/cards/09_HMW/leaders/DarthSidiousThereIsNoMercy.spec.ts
```diff
@@ -0,0 +1,463 @@
...
+                expect(context.player2).toBeActivePlayer();
+                expect(context.p2Base.damage).toBe(1);
+            });
+
+            it('should only exclude the exact object that was dealt 4 or more damage, allowing all other units and bases to be selected', async function() {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        leader: 'darth-sidious#there-is-no-mercy',
+                        groundArena: ['massassi-group-marines', 'wampa']
```
VJubert, line 177: Wampa make me think some interesting tests. 
First, you should try with overwhelm to see if base is selectable (maybe you need atst to have more than 4 power) 
Secondly, what happens if massassi or wampa kill someone ? Does it consider to have been dealt 4 damage even if unit cannot take them all ? There's something similar with Overwelming Barrage, you can put more damage on a unit than he has remaining HP (to be funnier you could also test with SOR Chirrut ou SEC Cassian 😆)

VJubert (approved):

PR #2720, opened 2026-08-24
The Warrior

VJubert:

### DIFF
--- a/server/game/cards/09_HMW/leaders/TheWarriorDeftDuelist.ts
+++ b/server/game/cards/09_HMW/leaders/TheWarriorDeftDuelist.ts
@@ -1,0 +1,39 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import type {
+    ILeaderUnitLeaderSideAbilityRegistrar
+} from '../../../core/card/AbilityRegistrationInterfaces';
+import { LeaderUnitCard } from '../../../core/card/LeaderUnitCard';
+import { KeywordName, RelativePlayer, WildcardCardType, ZoneName } from '../../../core/Constants';
+import { ResolutionMode } from '../../../gameSystems/SimultaneousOrSequentialSystem';
+import { TextHelper } from '../../../core/utils/TextHelper';
+
+export default class TheWarriorDeftDuelist extends LeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: 'the-warrior#deft-duelist-id',
+            internalName: 'the-warrior#deft-duelist',
+        };
+    }
+
+    protected override setupLeaderSideAbilities(registrar: ILeaderUnitLeaderSideAbilityRegistrar, abilityHelper: IAbilityHelper): void {
+        registrar.addActionAbility({
+            title: `Play a unit with 3 or less power from your hand. Give it ${TextHelper.Ambush} for this phase`,
+            cost: [abilityHelper.costs.abilityActivationResourceCost(1), abilityHelper.costs.exhaustSelf()],
+            cannotTargetFirst: true,
+            targetResolver: {
+                cardCondition: (card) => card.isUnit() && card.getPower() <= 3,
+                controller: RelativePlayer.Self,
+                zoneFilter: ZoneName.Hand,
+                immediateEffect: abilityHelper.immediateEffects.simultaneous({
+                    gameSystems: [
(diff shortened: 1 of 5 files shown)


server/game/cards/09_HMW/leaders/TheWarriorDeftDuelist.ts
```diff
@@ -0,0 +1,39 @@
...
+            id: 'the-warrior#deft-duelist-id',
+            internalName: 'the-warrior#deft-duelist',
+        };
+    }
+
+    protected override setupLeaderSideAbilities(registrar: ILeaderUnitLeaderSideAbilityRegistrar, abilityHelper: IAbilityHelper): void {
+        registrar.addActionAbility({
+            title: `Play a unit with 3 or less power from your hand. Give it ${TextHelper.Ambush} for this phase`,
+            cost: [abilityHelper.costs.abilityActivationResourceCost(1), abilityHelper.costs.exhaustSelf()],
+            cannotTargetFirst: true,
```
moyerr, line 22: Why was this needed? I think we usually use this for cases where the cost or the play action could affect potential targets (like Restock needing to be in discard for targeting step, or Bamboozle not being able to discard itself to pay its own cost)

server/game/core/card/propertyMixins/UnitProperties.ts
```diff
@@ -1027,7 +1027,7 @@ export function WithUnitProperties<TBaseClass extends InPlayCardConstructor>(Bas
             const modifierEffects: OngoingCardEffect[] = rawEffects.filter((effect) => effect.type === EffectName.ModifyStats);
             const wrappedStatsModifiers = modifierEffects.map((modifierEffect) => StatsModifierWrapper.fromEffect(modifierEffect, this));
 
-            if (!this.isAttached()) {
+            if (!this.isAttached() && this.isInPlay()) {
```
moyerr: What's the story here?

server/game/cards/09_HMW/leaders/TheWarriorDeftDuelist.ts
```diff
@@ -0,0 +1,39 @@
...
+            targetResolver: {
+                cardCondition: (card) => card.isUnit() && card.getPower() <= 3,
```
moyerr: These effects always use `printedPower` since modifiers don't apply in hand (see Neel)

server/game/core/card/propertyMixins/UnitProperties.ts
VJubert: getPower() check upgrade even if card is in hand. But upgrades cannot be accessed if unit is not in play

server/game/cards/09_HMW/leaders/TheWarriorDeftDuelist.ts
VJubert: 🤔 It exists some ability which change attributes in hand (HMW Zam Wessel, LAW Malakili). There is not for power ATM but it could be in future

server/game/cards/09_HMW/leaders/TheWarriorDeftDuelist.ts
VJubert: Hum I'll look more for this attribute, it's a copy-paste from SHD Fennec

server/game/cards/09_HMW/leaders/TheWarriorDeftDuelist.ts
moyerr: The devs have clarified that these abilities _always_ check printed power

server/game/core/card/propertyMixins/UnitProperties.ts
moyerr: Okay, that should be covered by the `printedPower` conversation below

server/game/cards/09_HMW/leaders/TheWarriorDeftDuelist.ts
VJubert: Ok I found another usage for this attribute. It is used when we play cards (paying their cost) and ability have cost

VJubert: Maybe we should do a check on recent cards (from 3 or 4 last sets, maybe it's an forgotten attribute 🤣)

moyerr (approved):

PR #2725, opened 2026-08-25
Update deck validation logic Next Set mode

moyerr:
- Next Set mode should include the next set to be released (HMW), but still exclude any known future sets (IC27)
- Ban lists now have an optional expiration set. For example, the current Eternal suspensions expire with the release of Homeworlds
- Add Cad Bane leader to the premier ban list

### DIFF
--- a/server/utils/deck/SwuSetData.ts
+++ b/server/utils/deck/SwuSetData.ts
@@ -79,8 +79,6 @@ export const rotationBlocks: IRotationBlock[] = [
         sets: [
             { id: SwuSetId.LAW, releaseStage: ReleaseStage.Released, mainline: true },
             { id: SwuSetId.ASH, releaseStage: ReleaseStage.Released, mainline: true },
-            // HMW is the next set to release; IC27 previews overlap HMW's preview season but release later, so
-            // it stays Future and is excluded from HMW's NextSet meta.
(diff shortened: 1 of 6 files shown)


server/utils/deck/SwuSetData.ts
```diff
@@ -69,23 +89,36 @@ export const nonRotatingSets: INonRotatingSet[] = [
...
     minDeckSize: number;
     maxCardCopies?: number;
-    bannedCards: Map<string, string>;
+    bannedCards: Map<string, IBannedCard>;
     rotationBlockCount?: number;
 }
 
-const bannedPremierCards = new Map<string, string>();
+const bannedPremierCards = new Map<string, IBannedCard>([
+    ['5648009238', { name: 'cad-bane#still-faster-than-you' }]
```
danbastin, line 116: Is it worth the hassle to add a `Next` prop to these so that if bans come through while we have NSP up we can ban there but retain the pending ban in the current pool? I think with this the easiest thing to do is to just wait until Monday to push this and open up NSP at the same time.

moyerr: I don't think it's really needed in this case, since we'll ban and open NS modes at the same time. But if it comes up again, we can definitely revisit

danbastin (approved):

PR #2734, opened 2026-08-27
Fix Latts Razzi interactions with Krayt Dragon

VJubert:


server/game/core/card/baseClasses/InPlayCard.ts
VJubert, line 110: I have an issue with this. LOF Qui-Gon return the unit to deck, but the property is not readable even if I need it for watcher check

VJubert: @moyerr

server/game/core/card/baseClasses/InPlayCard.ts
```diff
@@ -102,16 +102,10 @@ export class InPlayCard extends InPlayCardParent implements IInPlayCard {
...
-     * If the card is in a non-hidden, non-arena zone, this property is the most recent value of {@link inPlayId} for the card.
+     * The most recent value of {@link inPlayId} for this card, or `-1` if it has never been played.
      * This is used to determine e.g. if a card in the discard pile was defeated this phase.
      */
     public get mostRecentInPlayId() {
-        // TODO PR COMMENT
-        this.assertPropertyEnabledForZoneBoolean(
-            !this.isInPlay() && this.zone.hiddenForPlayers == null,
-            'mostRecentInPlayId'
-        );
```
moyerr, line 113: I think we still want the `!this.isInPlay()` part of the assertion, but drop the hidden zone check. Right?

moyerr (approved):

PR #2742, opened 2026-08-30
Exploit Confidence, Crosshair & Sun Fac

VJubert:


server/game/cards/09_HMW/units/CrosshairIveChanged.ts
```diff
@@ -0,0 +1,39 @@
...
+            when: {
+                onCardsDrawn: (event, context) =>
+                    event.player === context.player.opponent && context.game.currentPhase === PhaseName.Action,
+            },
+            immediateEffect: abilityHelper.immediateEffects.damage((context) => ({
+                amount: 2,
+                target: context.player.opponent.base
+            }))
+        });
+    }
```
moyerr, line 39: I think this needs a `collectiveTrigger` right? So it doesn't double-trigger for an effect that draws multiple cards? Similar to `SeasonedFleetAdmiral`

moyerr (approved):

PR #2749, opened 2026-08-31
Fix return-to-hand effects against Fortify (base) upgrades

moyerr:
## Bug

Return-to-hand effects (e.g. **Jabba the Hutt, Eminence of Tatooine**) couldn't target Fortify upgrades attached to a base. The upgrade wasn't selectable, and forcing the move hit a contract assertion.

**Cause:** `MoveCardSystem` used `isArena(card.zoneName)` as a stand-in for "in play." Fortify upgrades attach to the base, so they live in the base zone — an in-play zone that isn't an arena — and were excluded from both the target-validity check (`canAffectInternal`) and the leaves-play/unattach path (`eventHandler`/`updateEvent`).

## Fix

Use the existing `canBeInPlay() && isInPlay()` predicate (which already accounts for base-attached upgrades) in place of the arena-only checks, centralized in an `isLeavingPlay` helper. It's a strict superset of the old check, so arena behavior is unchanged — only base upgrades are now additionally handled. Generic fix for any return-to-hand effect, not just Jabba.

## Testing

- New regression test for Jabba returning a Fortify upgrade to hand
- Full suite green: 7983 specs, 0 failures

### DIFF
--- a/server/game/gameSystems/MoveCardSystem.ts
+++ b/server/game/gameSystems/MoveCardSystem.ts
@@ -1,6 +1,6 @@
 import type { AbilityContext } from '../core/ability/AbilityContext';
 import type { Card } from '../core/card/Card';
-import type { MoveZoneDestination } from '../core/Constants';
+import type { MoveZoneDestination, ZoneFilter } from '../core/Constants';
 import { AbilityRestriction, EffectName, RelativePlayer } from '../core/Constants';
 import {
     CardType,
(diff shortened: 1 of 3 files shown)


server/game/gameSystems/MoveCardSystem.ts
```diff
@@ -132,11 +132,20 @@ export class MoveCardSystem<TContext extends AbilityContext = AbilityContext> ex
...
         }
     }
 
+    /**
+     * A card is leaving play when it is currently in play and is being moved to a non-arena zone. `isInPlay()` accounts
+     * for base-attached (Fortify) upgrades, which live in the base zone rather than an arena (`SWU 4.9.1`).
+     */
+    private isLeavingPlay(card: Card, destination: MoveZoneDestination): boolean {
+        // Deck-position destinations (top/bottom) aren't arenas, so casting to ZoneFilter is safe for the arena check.
+        return card.canBeInPlay() && card.isInPlay() && !EnumHelpers.isArena(destination as ZoneFilter);
```
VJubert: It could be moving from a base to another. It do not leave play but it is not an arena

moyerr: Fixed, but there are no scenarios where this is possible, so no tests added

VJubert (approved):

PR #2764, opened 2026-09-03
add mos eisley and shield generator complex reprint info

VJubert:


scripts/fetchdata.js
```diff
@@ -212,6 +212,30 @@ function populateMissingData(attributes, id) {
...
+                data: [{
+                    attributes: {
+                        expansion: { data: { attributes: { code: 'HMW' } } },
+                        cardNumber: 32
+                    }
+                }]
+            };
+            break;
+    }
+
```
VJubert: Will it break builder which still use JTL numbers ?

VJubert: @AMMayberry1 @moyerr

moyerr (approved):

PR #2766, opened 2026-09-03
Implement 5 HMW cards: Blockade Ship, Dragonboat Freighter, Kelnacca Solitary Master, Noxious Refinery, Surveillance Cruiser

priisig:
## Summary

Implements ability logic and integration tests for five upcoming *Homeworlds* (HMW) cards.

---

### Blockade Ship — HMW 251

**Sentinel.** Enemy ground units get -1/-0 while attacking.

---

### Dragonboat Freighter — HMW 231

**When Played:** Give a Weakness token to a unit. If it's a unique unit, exhaust it.

---

### Kelnacca, Solitary Master — HMW 36

**When Played:** You may pay any number of resources. For every 3 resources paid this way, deal damage equal to this unit's power to an enemy unit (each instance of damage is dealt separately and can be re-targeted).

---

### Noxious Refinery — HMW 160

**Fortify.** Attached base gains: "When the regroup phase starts: Reveal the top card of your deck. If it's Aggression, deal 1 damage to an enemy unit."

---

### Surveillance Cruiser — HMW 247

**When Played:** If an opponent controls an Endor, Kashyyyk, Naboo, or Tatooine base, draw a card.

### DIFF
--- a/server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
+++ b/server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
@@ -1,0 +1,50 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { RelativePlayer, TargetMode, WildcardCardType } from '../../../core/Constants';
+
+export default class KelnaccaSolitaryMaster extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: 'kelnacca#solitary-master-id',
+            internalName: 'kelnacca#solitary-master',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addWhenPlayedAbility({
+            title: 'You may pay any number of resources. For every 3 resources paid this way, deal damage equal to this unit\'s power to an enemy unit',
+            targetResolver: {
+                mode: TargetMode.ChooseNumber,
+                min: 0,
+                max: (context) => context.player.readyResourceCount,
+            },
+            then: (thenContext) => ({
+                title: 'Pay the chosen number of resources',
+                immediateEffect: AbilityHelper.immediateEffects.payResourcesWithoutAdjustment({
+                    amount: parseInt(thenContext.select),
+                    target: thenContext.player
+                }),
+                then: {
+                    title: 'Deal damage equal to this unit\'s power to an enemy unit',
+                    thenCondition: (context) =>
+                        Math.floor(parseInt(thenContext.select) / 3) >= 1 &&
+                        context.player.opponent.hasSomeArenaUnit(),
+                    immediateEffect: AbilityHelper.immediateEffects.sequential(
+                        Array.from({ length: Math.floor(parseInt(thenContext.select) / 3) }, () =>
+                            AbilityHelper.immediateEffects.selectCard({
+                                activePromptTitle: `Deal ${thenContext.source.getPower()} damage to an enemy unit`,
+                                cardTypeFilter: WildcardCardType.Unit,
+                                controller: RelativePlayer.Opponent,
+                                immediateEffect: AbilityHelper.immediateEffects.damage({
+                                    amount: thenContext.source.getPower(),
+                                    source: thenContext.source,
+                                })
+                            })
+                        )
+                    )
+                }
+            })
+        });
+    }
+}
--- a/server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
+++ b/server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
@@ -13,37 +13,31 @@ export default class KelnaccaSolitaryMaster extends NonLeaderUnitCard {
 
     public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
         registrar.addWhenPlayedAbility({
-            title: 'You may pay any number of resources. For every 3 resources paid this way, deal damage equal to this unit\'s power to an enemy unit',
+            title: 'Pay any number of resources. For every 3 resources paid this way, deal damage equal to this unit\'s power to an enemy unit',
+            contextTitle: (context) => `Pay any number of resources. For every 3 resources paid this way, deal ${context.source.getPower()} damage to an enemy unit`,
             targetResolver: {
                 mode: TargetMode.ChooseNumber,
                 min: 0,
                 max: (context) => context.player.readyResourceCount,
+                immediateEffect: AbilityHelper.immediateEffects.payResourcesWithoutAdjustment((context) => ({
+                    amount: parseInt(context.select),
+                    target: context.player
+                })),
             },
             then: (thenContext) => ({
-                title: 'Pay the chosen number of resources',
-                immediateEffect: AbilityHelper.immediateEffects.payResourcesWithoutAdjustment({
-                    amount: parseInt(thenContext.select),
-                    target: thenContext.player
-                }),
-                then: {
-                    title: 'Deal damage equal to this unit\'s power to an enemy unit',
-                    thenCondition: (context) =>
-                        Math.floor(parseInt(thenContext.select) / 3) >= 1 &&
-                        context.player.opponent.hasSomeArenaUnit(),
-                    immediateEffect: AbilityHelper.immediateEffects.sequential(
-                        Array.from({ length: Math.floor(parseInt(thenContext.select) / 3) }, () =>
-                            AbilityHelper.immediateEffects.selectCard({
-                                activePromptTitle: `Deal ${thenContext.source.getPower()} damage to an enemy unit`,
-                                cardTypeFilter: WildcardCardType.Unit,
-                                controller: RelativePlayer.Opponent,
-                                immediateEffect: AbilityHelper.immediateEffects.damage({
-                                    amount: thenContext.source.getPower(),
-                                    source: thenContext.source,
-                                })
+                title: `Deal ${thenContext.source.getPower()} damage to an enemy unit`,
+                immediateEffect: AbilityHelper.immediateEffects.simultaneous(
+                    Array.from({ length: Math.floor(parseInt(thenContext.select) / 3) }, () =>
+                        AbilityHelper.immediateEffects.selectCard({
+                            activePromptTitle: `Deal ${thenContext.source.getPower()} damage to an enemy unit`,
+                            cardTypeFilter: WildcardCardType.Unit,
+                            controller: RelativePlayer.Opponent,
+                            immediateEffect: AbilityHelper.immediateEffects.damage({
+                                amount: thenContext.source.getPower(),
                             })
-                        )
+                        })
                     )
-                }
+                )
             })
         });
     }
--- a/server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
+++ b/server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
@@ -24,21 +24,24 @@ export default class KelnaccaSolitaryMaster extends NonLeaderUnitCard {
                     target: context.player
                 })),
             },
-            then: (thenContext) => ({
-                title: `Deal ${thenContext.source.getPower()} damage to an enemy unit`,
-                immediateEffect: AbilityHelper.immediateEffects.simultaneous(
-                    Array.from({ length: Math.floor(parseInt(thenContext.select) / 3) }, () =>
-                        AbilityHelper.immediateEffects.selectCard({
-                            activePromptTitle: `Deal ${thenContext.source.getPower()} damage to an enemy unit`,
-                            cardTypeFilter: WildcardCardType.Unit,
-                            controller: RelativePlayer.Opponent,
-                            immediateEffect: AbilityHelper.immediateEffects.damage({
-                                amount: thenContext.source.getPower(),
+            then: (thenContext) => {
+                const damageInstances = Math.floor(parseInt(thenContext.select) / 3);
+                return {
+                    title: `Deal ${thenContext.source.getPower()} damage to an enemy unit`,
+                    immediateEffect: AbilityHelper.immediateEffects.simultaneous(
+                        Array.from({ length: damageInstances }, (_, i) =>
+                            AbilityHelper.immediateEffects.selectCard({
+                                activePromptTitle: `Deal ${thenContext.source.getPower()} damage to an enemy unit (${i + 1}/${damageInstances})`,
+                                cardTypeFilter: WildcardCardType.Unit,
+                                controller: RelativePlayer.Opponent,
+                                immediateEffect: AbilityHelper.immediateEffects.damage({
+                                    amount: thenContext.source.getPower(),
+                                })
                             })
-                        })
+                        )
                     )
-                )
-            })
+                };
+            }
         });
     }
 }
--- a/server/game/cards/09_HMW/units/SurveillanceCruiser.ts
+++ b/server/game/cards/09_HMW/units/SurveillanceCruiser.ts
@@ -1,0 +1,24 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { Trait } from '../../../core/Constants';
+import { TextHelper } from '../../../core/utils/TextHelper';
+
+export default class SurveillanceCruiser extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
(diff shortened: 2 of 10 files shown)


test/server/cards/09_HMW/units/SurveillanceCruiser.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    integration(function(contextRef) {
+        it('Surveillance Cruiser\'s ability should draw a card if the opponent controls an Endor, Kashyyyk, Naboo, or Tatooine base', async function() {
+            await contextRef.setupTestAsync({
+                phase: 'action',
+                player1: {
+                    hand: ['surveillance-cruiser'],
+                    deck: ['porg', 'rey#skywalker']
+                },
+                player2: {
+                    base: 'jabbas-palace'
```
VJubert, line 77: you should have 5 test, 4 for each Trait + 1 without (but a Naboo unit could be good)

server/game/cards/09_HMW/units/SurveillanceCruiser.ts
```diff
@@ -0,0 +1,24 @@
...
+    protected override getImplementationId() {
+        return {
+            id: 'surveillance-cruiser-id',
+            internalName: 'surveillance-cruiser',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addWhenPlayedAbility({
+            title: `If an opponent controls an ${TextHelper.Trait.Endor}, ${TextHelper.Trait.Kashyyyk}, ${TextHelper.Trait.Naboo}, or ${TextHelper.Trait.Tatooine} base, draw a card`,
```
VJubert: Does the if part is needed ? Maybe when there's multiple trigger, but in this case, No Effect prompt should appear ?

test/server/cards/09_HMW/upgrades/NoxiousRefinery.spec.ts
```diff
@@ -0,0 +1,68 @@
+describe('Noxious Refinery', function() {
+    integration(function(contextRef) {
+        it('should deal 1 damage to an enemy unit when the revealed card has the Aggression aspect', async function() {
+            await contextRef.setupTestAsync({
+                phase: 'action',
+                player1: {
+                    deck: ['wampa'],
```
VJubert: can you set more than one card ? To see if ability is not limited by deck size

test/server/cards/09_HMW/units/BlockadeShip.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+            context.player1.clickCard(context.atteVanguard);
+            context.player1.clickCard(context.battlefieldMarine);
+
+            expect(context.battlefieldMarine).toBeInZone('discard');
+            expect(context.atteVanguard.damage).toBe(3);
+            expect(context.player2).toBeActivePlayer();
+        });
+    });
+});
```
VJubert, line 114: a good test to add is with hera syndulla, should heal 1 less

test/server/cards/09_HMW/units/KelnaccaSolitaryMaster.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+            context.player1.chooseListOption('6');
+
+            expect(context.player1).toHavePrompt('Deal 4 damage to an enemy unit');
+            expect(context.player1).toBeAbleToSelectExactly([context.wampa, context.battlefieldMarine]);
+            // uses the non-checking click variant since this prompt looks identical (same title, same two
+            // selectable units) both before and after — the underlying game state (damage dealt) still changes,
+            // it's just not reflected in the prompt shape the "expect change" heuristic compares
+            context.player1.clickCardNonChecking(context.wampa);
+
+            // second damage instance is a fresh target selection, not part of a single lump-sum hit
```
VJubert, line 131: does damage are dealt before each prompt ? if not player will never understand what's happening

test/server/cards/09_HMW/units/KelnaccaSolitaryMaster.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+                player2: {
+                    groundArena: ['wampa'],
+                    leader: { card: 'cad-bane#still-faster-than-you', deployed: true }
+                }
+            });
+
+            const { context } = contextRef;
+
+            context.player1.clickCard(context.kelnacca);
+            context.player1.chooseListOption('3');
```
VJubert, line 223: can you try to exhaust 4 or 5 resources ? And see if we do not deal 4.5 damage 🤣

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
```diff
@@ -0,0 +1,50 @@
...
+                min: 0,
+                max: (context) => context.player.readyResourceCount,
+            },
+            then: (thenContext) => ({
+                title: 'Pay the chosen number of resources',
+                immediateEffect: AbilityHelper.immediateEffects.payResourcesWithoutAdjustment({
+                    amount: parseInt(thenContext.select),
+                    target: thenContext.player
+                }),
+                then: {
```
moyerr: The `ChooseNumber` target mode also supports an `immediateEffect` (see Force Lightning), so we can collapse this down to a single `then` by moving this effect up into the `targetResolver`

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
```diff
@@ -0,0 +1,50 @@
...
+    protected override getImplementationId() {
+        return {
+            id: 'kelnacca#solitary-master-id',
+            internalName: 'kelnacca#solitary-master',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addWhenPlayedAbility({
+            title: 'You may pay any number of resources. For every 3 resources paid this way, deal damage equal to this unit\'s power to an enemy unit',
```
moyerr: We don't usually put "You may" in these ability titles, since it's implied by the ability to pass (or choose 0 in this case). It can just say:

> Pay any number of resources. For every 3 resources paid this way, deal damage equal to this unit\'s power to an enemy unit.

We can also add a `contextTitle` that uses the context to call out Kelnacca's actual power

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
```diff
@@ -0,0 +1,50 @@
...
+                max: (context) => context.player.readyResourceCount,
+            },
+            then: (thenContext) => ({
+                title: 'Pay the chosen number of resources',
+                immediateEffect: AbilityHelper.immediateEffects.payResourcesWithoutAdjustment({
+                    amount: parseInt(thenContext.select),
+                    target: thenContext.player
+                }),
+                then: {
+                    title: 'Deal damage equal to this unit\'s power to an enemy unit',
```
moyerr: Since we have a `thenContext` here, we can call out the actual damage amount

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
```diff
@@ -0,0 +1,50 @@
...
+                    thenCondition: (context) =>
+                        Math.floor(parseInt(thenContext.select) / 3) >= 1 &&
+                        context.player.opponent.hasSomeArenaUnit(),
```
moyerr: I don't think either of these conditions are necessary since the ability would fizzle anyway

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
```diff
@@ -0,0 +1,50 @@
...
+                    immediateEffect: AbilityHelper.immediateEffects.sequential(
```
moyerr: This should be simultaneous

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
```diff
@@ -0,0 +1,50 @@
...
+                        Array.from({ length: Math.floor(parseInt(thenContext.select) / 3) }, () =>
+                            AbilityHelper.immediateEffects.selectCard({
+                                activePromptTitle: `Deal ${thenContext.source.getPower()} damage to an enemy unit`,
+                                cardTypeFilter: WildcardCardType.Unit,
+                                controller: RelativePlayer.Opponent,
+                                immediateEffect: AbilityHelper.immediateEffects.damage({
+                                    amount: thenContext.source.getPower(),
+                                    source: thenContext.source,
```
moyerr: Is the source necessary here? I would think the ability's source is the default.

server/game/cards/09_HMW/units/SurveillanceCruiser.ts
moyerr: Agreed, this would just be "Draw a card"

test/server/cards/09_HMW/units/BlockadeShip.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+                    spaceArena: ['blockade-ship']
+                },
+                player2: {
+                    groundArena: ['wampa']
+                }
+            });
+
+            const { context } = contextRef;
+
+            context.player1.passAction();
```
moyerr: FYI you can set `hasInitiative: true` for player 2 if you want that player to start with an action

test/server/cards/09_HMW/units/KelnaccaSolitaryMaster.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+            expect(context.player1).toBeAbleToSelectExactly([context.wampa, context.battlefieldMarine]);
+            context.player1.clickCardNonChecking(context.battlefieldMarine);
+
+            expect(context.wampa.damage).toBe(4);
+            expect(context.battlefieldMarine).toBeInZone('discard');
+            expect(context.player1.exhaustedResourceCount).toBe(10);
+            expect(context.player2).toBeActivePlayer();
+        });
+
+        it('should allow choosing the same unit for each damage instance, applied as separate hits rather than a multiplied lump sum', async function () {
```
moyerr: This should be simultaneous, so all would be prevented by a single shield:

test/server/cards/09_HMW/units/SurveillanceCruiser.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+            await contextRef.setupTestAsync({
+                phase: 'action',
+                player1: {
+                    hand: ['surveillance-cruiser'],
+                    deck: ['porg', 'rey#skywalker']
+                },
+                player2: {
+                    base: 'command-center'
+                }
+            });
```
moyerr, line 105: Should give P1 one of the Endor, Kashyyyk, Naboo, or Tatooine bases in this test, just to make sure it's only checking opponent

test/server/cards/09_HMW/upgrades/NoxiousRefinery.spec.ts
moyerr: Might also ensure timing since the player will also draw 2 for regroup

test/server/cards/09_HMW/units/SurveillanceCruiser.spec.ts
priisig: Added 4 trait-specific tests (Endor/Kashyyyk/Naboo/Tatooine) plus a no-match test that also gives the opponent a Naboo-*trait unit* to confirm unit traits don't count, and gives player1 a matching base too, per moyerr's comment below.

server/game/cards/09_HMW/units/SurveillanceCruiser.ts
priisig: Done — title simplified to `'Draw a card'`.

test/server/cards/09_HMW/upgrades/NoxiousRefinery.spec.ts
priisig: Deck now has 3 cards (Aggression card on top + 2 fillers), and the test also asserts the revealed card ends up in hand from the phase's automatic draw-2 afterward — confirming the reveal/damage trigger resolves before that draw step.

test/server/cards/09_HMW/units/BlockadeShip.spec.ts
priisig: Added a test with Hera Syndulla, Renegade General attacking into Blockade Ship — her attack-end heal (equal to damage dealt) comes out 1 lower than her printed power since Blockade Ship reduces her power while attacking.

test/server/cards/09_HMW/units/KelnaccaSolitaryMaster.spec.ts
priisig: Each instance is still its own target-selection prompt happening one after another, but now that the damage is `simultaneous`, the actual damage events all resolve together in one window after every target is chosen, rather than being applied between prompts. The prompt title/range already show the numbers involved, so I think it's clear enough — but flag it if you'd rather see it differently.

test/server/cards/09_HMW/units/KelnaccaSolitaryMaster.spec.ts
priisig: Added a test paying 5 resources that asserts exactly one 4-damage instance (`floor(5/3) = 1`), never a fractional or rounded-up amount.

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
priisig: Done — payment now happens directly in the `ChooseNumber` `targetResolver`'s `immediateEffect`, leaving a single `then` for the damage step.

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
priisig: Done — dropped "You may" and added a `contextTitle` that shows the actual power via `context.source.getPower()`.

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
priisig: Done — the `then` title now shows the actual damage number via `thenContext.source.getPower()`.

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
priisig: Removed both — confirmed the ability fizzles cleanly without them (empty array / no legal target).

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
priisig: Done — switched to `simultaneous`.

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
priisig: Removed — defaults to the ability's own source.

test/server/cards/09_HMW/units/BlockadeShip.spec.ts
priisig: Switched both affected tests to `hasInitiative: true` for player2 and dropped the `passAction()` calls.

test/server/cards/09_HMW/units/KelnaccaSolitaryMaster.spec.ts
priisig: Updated this test to assert both instances get prevented by the single Shield, and added a further test with 3 simultaneous instances against a 3-HP unit to confirm no damage leaks through even partially.

test/server/cards/09_HMW/units/SurveillanceCruiser.spec.ts
priisig: Folded into the no-match test above — player1 now has `jabbas-palace` as well, confirming only the opponent's base is checked.

server/game/cards/09_HMW/units/KelnaccaSolitaryMaster.ts
```diff
@@ -24,21 +24,24 @@ export default class KelnaccaSolitaryMaster extends NonLeaderUnitCard {
...
-                            immediateEffect: AbilityHelper.immediateEffects.damage({
-                                amount: thenContext.source.getPower(),
+            then: (thenContext) => {
+                const damageInstances = Math.floor(parseInt(thenContext.select) / 3);
+                return {
+                    title: `Deal ${thenContext.source.getPower()} damage to an enemy unit`,
+                    immediateEffect: AbilityHelper.immediateEffects.simultaneous(
+                        Array.from({ length: damageInstances }, (_, i) =>
+                            AbilityHelper.immediateEffects.selectCard({
+                                activePromptTitle: `Deal ${thenContext.source.getPower()} damage to an enemy unit (${i + 1}/${damageInstances})`,
```
VJubert, line 34: Should we hide indicators if there's only 1 damage instance ? @moyerr

priisig: I was wondering about that, but I think it makes sense to keep it in case you pay e.g. 5 resources, you immediately see that this only gives you one damage instance.

VJubert (approved):

PR #2773, opened 2026-09-04
Phee Genoa, Liberator of Ancient Wonders

AlejandroRigau:
[Card image — Phee Genoa, Liberator of Ancient Wonders (HMW 214)](https://swudb.com/cdn-cgi/image/width=300/images/cards/HMW/214.png)

Implements Phee Genoa, Liberator of Ancient Wonders (HMW 214).

### Implementation

A triggered ability on `onLeaderDeployed` for an opponent's leader, using a `SelectUnless` target resolver so the leader's controller chooses between paying 2 resources and having the leader exhausted.

Pilot deploys are handled as a "no effect" case rather than as a non-trigger: a leader deployed as a pilot is an upgrade, so it can't be exhausted, but its controller is still offered the choice. `showUnresolvable: true` keeps the Exhaust button on the prompt in that situation — picking it simply does nothing. If the controller can't afford 2 resources, the prompt is skipped entirely and the exhaust resolves directly.

### Tests

`test/server/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.spec.ts` — 6 specs covering: exhaust chosen, pay chosen, controller unable to pay (no prompt), friendly leader deploy (no trigger), and both branches of the prompt after a pilot deploy.

All 6 pass in both normal and whole-suite undo mode; `tsc` (server + test) and `eslint --quiet` are clean.

https://claude.ai/code/session_013PpWW7rka2C8WstDh6LFqB

### DIFF
--- a/server/game/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.ts
+++ b/server/game/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.ts
@@ -1,0 +1,47 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import type { INonLeaderUnitAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { NonLeaderUnitCard } from '../../../core/card/NonLeaderUnitCard';
+import { NamedAction, RelativePlayer, TargetMode } from '../../../core/Constants';
+import { TextHelper } from '../../../core/utils/TextHelper';
+
+export default class PheeGenoaLiberatorOfAncientWonders extends NonLeaderUnitCard {
+    protected override getImplementationId() {
+        return {
+            id: 'phee-genoa#liberator-of-ancient-wonders-id',
+            internalName: 'phee-genoa#liberator-of-ancient-wonders',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addTriggeredAbility({
+            title: `Exhaust the deployed leader unless its controller pays ${TextHelper.resource(2)}`,
+            contextTitle: (context) => `Exhaust ${context.event.card.title} unless its controller pays ${TextHelper.resource(2)}`,
+            when: {
+                onLeaderDeployed: (event, context) => event.card.controller === context.player.opponent
+            },
+            targetResolver: {
+                mode: TargetMode.SelectUnless,
+                choosingPlayer: RelativePlayer.Opponent,
+                activePromptTitle: (context) => `[Exhaust] ${context.event.card.title} or [Pay] ${TextHelper.resource(2)}`,
+
+                // A leader deployed as a pilot is an upgrade and cannot be exhausted, but the choice is still offered
+                // to its controller. Choosing to exhaust it simply has no effect.
+                showUnresolvable: true,
+                unlessEffect: {
+                    effect: (context) => abilityHelper.immediateEffects.payResources({
+                        target: context.event.card.controller,
+                        amount: 2
+                    }),
+                    promptButtonText: NamedAction.Pay
+                },
+                defaultEffect: {
+                    effect: (context) => abilityHelper.immediateEffects.exhaust({
+                        target: context.event.card
+                    }),
+                    promptButtonText: NamedAction.Exhaust
+                },
+                highlightCards: (context) => context.event.card,
+            }
+        });
+    }
+}
--- a/test/server/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.spec.ts
+++ b/test/server/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.spec.ts
@@ -1,0 +1,126 @@
+describe('Phee Genoa, Liberator of Ancient Wonders', function() {
+    integration(function(contextRef) {
+        describe('Phee Genoa\'s triggered ability', function() {
+            beforeEach(async function() {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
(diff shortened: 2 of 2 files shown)


test/server/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+                expect(context.player2).toHaveEnabledPromptButtons(['Pay', 'Exhaust']);
+                context.player2.clickPrompt('Pay');
+
+                expect(context.wedgeAntilles).toBeAttachedTo(context.cartelSpacer);
+                expect(context.player2.exhaustedResourceCount).toBe(2);
+                expect(context.player1).toBeActivePlayer();
+            });
+        });
+    });
+});
```
VJubert, line 180: Can you add a test asserting there is no phase limit, you'll need grogu, thraya and 2 4-drop unique unit

server/game/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addTriggeredAbility({
+            title: `Exhaust the deployed leader unless its controller pays ${TextHelper.resource(2)}`,
+            contextTitle: (context) => `Exhaust ${context.event.card.title} unless its controller pays ${TextHelper.resource(2)}`,
+            when: {
+                onLeaderDeployed: (event, context) => event.card.controller === context.player.opponent
+            },
+            targetResolver: {
+                mode: TargetMode.SelectUnless,
```
VJubert, line 23: I'm a bit worried about using selectunless instead ifyoudonot like card is worded, but prompt are very clear 👍 (no change required)

test/server/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.spec.ts
AlejandroRigau: Happy to add this, but I need a bit more from you first — I can't see the card's real text to work from. In this repo `phee-genoa#liberator-of-ancient-wonders` is still a mock (`hasNonKeywordAbility: true`, text renders as `mock ability text`), and the card isn't publicly revealed yet, so I have nothing to check the wording against.

Two things I'd need:

1. **Is there a once-per-phase limit on the ability?** As implemented there's no `limit`, so a test asserting one would fail today — if the card says "Once per phase" then this is an implementation gap, not just a missing test, and I'll add `limit: abilityHelper.limit.perPhase(1)` alongside the test.
2. **What's the interaction that gets a second trigger in one phase?** I'm not following the Grogu / Thraya / two 4-drop uniques setup — a leader deploys once, so I can't see how `onLeaderDeployed` fires twice. If the real trigger is broader than leader deploys (e.g. it also covers playing a unique unit), that's a bigger gap and I'd rather fix the trigger than paper over it with a test.

If you paste the card text I'll push the limit and the test in one go.

---

server/game/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.ts
AlejandroRigau: Thanks — for what it's worth, `SelectUnless` is the existing idiom in this repo for the "if you do not pay, then X" pattern (see `IAmYourFather`, `LookTheOtherWay`, `InDebtToCrimsonDawn`, `DedraMeeroNotWastingTime`), and it gives the same semantics: the choice goes to the affected card's controller, and the default effect resolves automatically when the payment isn't affordable. Leaving it as-is per your note.

---

test/server/cards/09_HMW/units/PheeGenoaLiberatorOfAncientWonders.spec.ts
AlejandroRigau: Saw the edit — "no phase limit" rather than "there is". Good news on half of it: the ability has no `limit`, so nothing to change in the implementation.

But I can't write that test against the trigger as implemented, and I think that points at a real gap rather than a test I'm just failing to construct. Deploying a leader is an **epic action — once per game, and explicitly not reset on defeat**:

```ts
// LeaderUnitCard constructor
this.deployEpicActions.push(registrar.addActionAbility({
    limit: this.deployEpicActionLimit,   // EpicActionLimit
    title: `Deploy ${this.title}`,
    ...
}));
```
```ts
// AbilityLimit.ts:250
export class EpicActionLimit extends PerPlayerPerGameAbilityLimitBase {
    public constructor(game: Game) { super(game, 1); }
    // this prevents the limit from being reset when a leader is defeated
    public override reset() {}
}
```

So `onLeaderDeployed` can fire at most **once per game** for a given opponent — a second trigger in the same phase isn't reachable, with or without Grogu and a defeat/undeploy line. Which makes me think the real trigger is broader than leader deploys (playing a unique unit too?), and that's what the two 4-drop uniques in your setup are for.

I also can't find "Thraya" anywhere — not in `test/json/` after `npm run get-cards`, not in `scripts/mockdata.js` on current `main` (`fe5842c`), and the card isn't publicly revealed, so I can't read its text either.

Could you paste the card's actual wording? If the trigger genuinely is leader-deploy-only then I'd argue the no-phase-limit test isn't expressible and we drop it; if it's broader, I'll fix the trigger and add the test in the same push. Either way I'd rather get the trigger right than write a test around a guess.

---

AlejandroRigau: Added the test

moyerr (approved):

PR #2775, opened 2026-09-04
Krrsantan, Grand Army Marine & Fambaa Shield Team

VJubert:


AlejandroRigau (dismissed):
lgtm

test/server/cards/09_HMW/units/KrrsantanSanto.spec.ts
```diff
@@ -0,0 +1,62 @@
+describe('Krrsantan, Santo', function() {
+    integration(function(contextRef) {
+        describe('Krrsantan\'s When Played ability', function() {
```
moyerr, line 3: What about a case where he's played for 1 less and the damage total is 0? Han Solo, Worth The Risk can be used for that case

moyerr (approved):

PR #2781, opened 2026-09-04
Fix Overgrowth and Command crashing when no friendly unit can deal damage

AlejandroRigau:
## The bug

Play **Overgrowth** with a Kashyyyk base while every friendly unit has 0 power (Moisture Farmer, Coruscanti Spy, …) and the opponent controls at least one unit. The friendly unit prompt never appears; instead the game asks to `Deal undefined damage to an enemy unit`, and clicking a target throws:

```
Contract assertion failure: Null-like object value: undefined
    at DamageSystem.addAbilityDamagePropertiesToEvent
```

The ability aborts there, so Overgrowth is never resourced — it sits in the discard pile.

**Command** hits the identical crash: choose `A friendly unit deals damage equal to its power to a non-unique enemy unit` with only 0-power units on board.

## Why it happens

`CardTargetResolver` applies two different bars to a selection:

1. **Legality** (`checkCardCondition`) at `GameStateChangeRequired.None` — a 0-power unit passes, because 0 damage is still a legal damage effect.
2. **Effectiveness** (`CardTargetResolver.ts:151-168`) at `MustFullyOrPartiallyResolve` — when *no* legal target would change game state, the resolver deliberately skips the prompt and sets the target to `null`.

So with every friendly unit at 0 power, the selection resolves to no unit at all. `SelectCardSystem` then ran its wrapped effect anyway — it only checked `targetResults.cancelled`, never whether a card was actually chosen.

For most cards that is invisible: the wrapped effect is a card target system, and no target means no events. But when the wrapped effect is *another selection that reads the outer target* — the shape both Overgrowth and Command use — the enemy unit prompt opens with `friendlyUnit` still `null`, the damage amount evaluates to `undefined`, and `DamageSystem` asserts on it.

## Before / after

| Scenario | Before | After |
| --- | --- | --- |
| Overgrowth, at least one friendly unit with power | works | unchanged |
| Overgrowth, 0-power unit picked alongside a real attacker | selectable, deals no damage | unchanged |
| Overgrowth, **every** friendly unit at 0 power | `Deal undefined damage` prompt → crash → event **not** resourced | no damage prompt, no damage, event resourced |
| Command damage mode, every friendly unit at 0 power | same crash | mode fizzles, remaining modes resolve normally |
| "Up to" selections where the player chooses nothing | effect resolves against the empty selection | unchanged |

## The fix

`SelectCardSystem` skips the wrapped effect when a **mandatory** selection resolved without a card:

```ts
if (selectedCards.length === 0 && !this.selectionAllowsChoosingNoCards(context, additionalProperties)) {
    return;
}
```

The `selectionAllowsChoosingNoCards` half is load-bearing, not defensive. Selections like `TargetMode.UpTo` treat choosing nothing as a real choice and still need their effect: The Eye Of Aldhani exhausts every *un*selected unit, so an unconditional skip breaks it. The full suite caught that.

Two smaller changes ride along:

- The emptiness check reads `context.targets[properties.name]` rather than `context.target`. `context.target` is only written when a selection is unnamed (`TargetResolver.setTargetResult`), so the existing `cancelIfNoTargets` check silently never fired for a named selection. Behavioural no-op for its only caller (`DiscloseAspectsSystem`, unnamed), and the flag now works if anyone reaches for it.
- Overgrowth drops `?.` to `.` on its two `friendlyUnit` reads, since a selection is now guaranteed before that effect runs. A regression should fail loudly rather than quietly become `undefined`.

## Tests

- `Overgrowth.spec.ts` — 0-power unit is offered alongside a real attacker and deals nothing when picked; neither prompt offers a pass or choose-nothing button; the event is resourced in all four dead ends (no friendly units, no enemy units, no unit that can deal damage, non-Kashyyyk base).
- `Command.spec.ts` — new fizzle case for "no friendly unit has any power", matching the two fizzle tests already there.

Both new crash tests were confirmed to fail with the guard reverted and pass with it, so they genuinely cover the bug.

Full suite green: 8099 specs, 0 failures (8 pre-existing `xit`s). Lint clean. This touches core targeting that all 106 `selectCard` cards route through, so the suite is the main evidence of safety — the `selectionAllowsChoosingNoCards` condition is the line worth a close read.

https://claude.ai/code/session_01PBpcPepAx7Xh2J8HTBpBxj

### DIFF
--- a/server/game/gameSystems/SelectCardSystem.ts
+++ b/server/game/gameSystems/SelectCardSystem.ts
@@ -104,11 +104,8 @@ export class SelectCardSystem<TContext extends AbilityContext = AbilityContext> 
                 return;
             }
 
-            // a mandatory selection can still resolve without a card, e.g. when the resolver finds no target that the
-            // wrapped effect could do anything to and skips the prompt entirely. there is no selection to apply the
-            // effect to in that case. a selection that lets the player choose no cards (e.g. TargetMode.UpTo) is
-            // different: choosing nothing is a real choice there, and the effect still resolves against the empty
-            // selection
+            // a mandatory selection resolves without a card when the resolver finds no effective target and skips the
+            // prompt, leaving nothing to apply the effect to. choosing nothing in an "up to" selection is a real choice
             if (selectedCards.length === 0 && !this.selectionAllowsChoosingNoCards(context, additionalProperties)) {
                 return;
             }
--- a/server/game/gameSystems/SelectCardSystem.ts
+++ b/server/game/gameSystems/SelectCardSystem.ts
@@ -99,22 +99,25 @@ export class SelectCardSystem<TContext extends AbilityContext = AbilityContext> 
         context.game.queueSimpleStep(() => {
             const selectedCards = Helpers.asArray(context.targets[properties.name]);
 
-            // a mandatory selection can still resolve without a card, e.g. when the resolver finds no target that the
-            // wrapped effect could do anything to and skips the prompt. there is nothing to apply the effect to in that
-            // case, unlike a selection that lets the player choose no cards (e.g. TargetMode.UpTo), where the effect
-            // still resolves against the empty selection
-            const resolvedWithoutSelection = selectedCards.length === 0 &&
-              !this.selectionAllowsChoosingNoCards(context, additionalProperties);
-
             if (targetResults.cancelled || (properties.cancelIfNoTargets && selectedCards.length === 0)) {
                 properties.cancelHandler?.();
-            } else if (!resolvedWithoutSelection) {
-                if (!properties.isCost && Helpers.asArray(context.target).length > 0) {
-                    this.addOnSelectEffectMessage(context, properties);
-                }
-                properties.onSelectHandler?.(context.targets[properties.name] ?? context.target);
-                properties.immediateEffect.queueGenerateEventGameSteps(events, context, additionalProperties);
+                return;
             }
+
+            // a mandatory selection can still resolve without a card, e.g. when the resolver finds no target that the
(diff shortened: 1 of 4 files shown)


server/game/gameSystems/SelectCardSystem.ts
```diff
@@ -97,15 +97,24 @@ export class SelectCardSystem<TContext extends AbilityContext = AbilityContext>
...
+
+            // a mandatory selection resolves without a card when the resolver finds no effective target and skips the
+            // prompt, leaving nothing to apply the effect to. choosing nothing in an "up to" selection is a real choice
+            if (selectedCards.length === 0 && !this.selectionAllowsChoosingNoCards(context, additionalProperties)) {
+                return;
+            }
+
+            if (!properties.isCost && Helpers.asArray(context.target).length > 0) {
+                this.addOnSelectEffectMessage(context, properties);
             }
```
moyerr, line 115: should this use the `selectedCards` from above instead of `context.target`?

AlejandroRigau: This line is unchanged from main — I only added `selectedCards` a few lines above it.

They're not equivalent though: `context.target` is only set for unnamed selections, so this message never fires for named ones. Not sure that's wrong as such, but switching it does change game logs. Grievous starts logging `uses General Grievous to play Imperial Dark Trooper from their hand` directly above the existing `plays Imperial Dark Trooper`, and its spec asserts the log verbatim, so it fails. Overgrowth gains a useful damage line plus a meaningless `choose a target for Gungi` one.

Happy to switch it if you think that's better.

---

moyerr (approved):

PR #2782, opened 2026-09-04
Tireless Magnaguard

AlejandroRigau:
[](https://swudb.com/cdn-cgi/image/width=300/images/cards/HMW/109.png)

**Tireless Magnaguard** — When Defeated: If this unit had 5 or more power, for this phase you may play this unit from your discard pile for free and give 2 Weakness tokens to it.

## Implementation
- The 5 power check reads the defeat event's last known information, so buffs and Weakness tokens at the moment of defeat count.
- The permission is only granted when the player resolving the ability owns the unit. A unit defeated under an opponent's control goes to its owner's discard pile, so it is never in "your discard pile" for the player who resolved the ability.
- Play permission and the free cost follow the Stolen AT-Hauler pattern: a phase-long play-from-discard effect on the card, plus a free cost adjuster on the player that only matches while the card is in the discard pile.
- The Weakness tokens are a phase-long delayed effect on the player, fired when that player plays this card out of the discard pile. It cannot live on the card itself because lasting effects are stripped when a card moves from discard into an arena.

## Test cases
- replayed for free and comes back as a 3/1 with 2 Weakness tokens
- once replayed at 3 power, a second defeat grants nothing
- replayed three times in a single action phase, pumped back to 5 power before each defeat (Commence the Festivities, then Tactical Advantage) — there is no per-phase limit
- no permission when it had 4 power at defeat
- power modifiers count: Weakness plus Experience is 5 power and triggers
- the permission expires with the phase
- returned from the discard pile to hand with Renewed Friendship and played from hand: costs full price, no Weakness tokens
- defeated under an opponent's control by No Glory, Only Results: neither player can play it from the owner's discard pile

https://claude.ai/code/session_01LJnXH2vY5dpcQuEdU6qzDP

### DIFF
--- a/test/server/cards/09_HMW/units/TirelessMagnaguard.spec.ts
+++ b/test/server/cards/09_HMW/units/TirelessMagnaguard.spec.ts
@@ -1,0 +1,294 @@
+describe('Tireless Magnaguard', function() {
+    integration(function(contextRef) {
+        describe('Tireless Magnaguard\'s When Defeated ability', function() {
+            it('should allow its owner to play it from their discard pile for free this phase and give it 2 Weakness tokens if it had 5 or more power', async function() {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
+                        groundArena: ['tireless-magnaguard']
+                    },
+                    player2: {
+                        hand: ['vanquish'],
+                        hasInitiative: true
+                    }
+                });
+
+                const { context } = contextRef;
+
+                // Player 2 defeats Tireless Magnaguard while it has 5 power
+                context.player2.clickCard(context.vanquish);
+                context.player2.clickCard(context.tirelessMagnaguard);
+
+                expect(context.tirelessMagnaguard).toBeInZone('discard', context.player1);
+                expect(context.player1).toBeActivePlayer();
+
+                // Player 1 plays it from their discard pile for free
+                const readyResourceCount = context.player1.readyResourceCount;
+                context.player1.clickCard(context.tirelessMagnaguard);
+
+                expect(context.tirelessMagnaguard).toBeInZone('groundArena', context.player1);
+                expect(context.player1.readyResourceCount).toBe(readyResourceCount);
+                expect(context.tirelessMagnaguard).toHaveExactUpgradeNames(['weakness', 'weakness']);
+                expect(context.tirelessMagnaguard.getPower()).toBe(3);
+                expect(context.tirelessMagnaguard.getHp()).toBe(1);
+                expect(context.player2).toBeActivePlayer();
+            });
+
+            it('should not allow it to be played again if it is defeated again with less than 5 power', async function() {
+                await contextRef.setupTestAsync({
+                    phase: 'action',
+                    player1: {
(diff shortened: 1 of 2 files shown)


test/server/cards/09_HMW/units/TirelessMagnaguard.spec.ts
```diff
@@ -0,0 +1,259 @@
...
+                context.player1.clickCard(context.tirelessMagnaguard);
+
+                expect(context.tirelessMagnaguard).toBeInZone('groundArena', context.player1);
+                expect(context.player1.readyResourceCount).toBe(readyResourceCount);
+                expect(context.tirelessMagnaguard).toHaveExactUpgradeNames(['weakness', 'weakness']);
+                expect(context.tirelessMagnaguard.getPower()).toBe(3);
+                expect(context.tirelessMagnaguard.getHp()).toBe(1);
+
+                // ROUND 3: same again against a different defender
+                context.moveToNextActionPhase();
```
VJubert: why changing phase between action ? There is not limit per phase to play it from discard pile

test/server/cards/09_HMW/units/TirelessMagnaguard.spec.ts
```diff
@@ -0,0 +1,259 @@
...
+                // The ability resolved for player 1, so the owner does not get to play it from their discard pile
+                expect(context.tirelessMagnaguard).not.toHaveAvailableActionWhenClickedBy(context.player2);
+                context.player2.passAction();
+
+                // Player 1 cannot play it from player 2's discard pile
+                expect(context.tirelessMagnaguard).not.toHaveAvailableActionWhenClickedBy(context.player1);
+            });
+        });
+    });
+});
```
VJubert, line 294: can you add a test when you defeat him while having 5 or more power, bouncing it back to the hand (SEC Renewed Friendship) and play it from hand. It should not have free discount

test/server/cards/09_HMW/units/TirelessMagnaguard.spec.ts
AlejandroRigau: Good catch, there is no per-phase limit and the test read like there was. Rewritten so all three replays happen in a single action phase.

The phase changes were only there because a unit played from the discard pile enters exhausted, so Commence the Festivities could not attack with it again in the same phase. It now opens with Commence the Festivities while the unit is still ready, then uses Tactical Advantage to pump it back to 5 power before each of the next two defeats. The separate test for the permission expiring in a later phase still covers the "for this phase" duration.

---

test/server/cards/09_HMW/units/TirelessMagnaguard.spec.ts
AlejandroRigau: Added. The new test defeats it at 5 power, plays Renewed Friendship to return it from the discard pile to hand, then plays it from hand. It costs the full 4 resources and gets no Weakness tokens.

The free-cost adjuster only matches while the card is in the discard pile, and the Weakness tokens are only given on a play whose original zone was the discard pile.

---

VJubert (approved):

PR #2785, opened 2026-09-07
Fix Ravager and add Luminara tests

bwohldmann:
Fixed interaction when Maul9 Leader plays a unit while Ravager is on board.  No triggers were resolving due to an exception caused by the Ravager ability prompt trying to read the power from the card that was played (which was at that point in the discard).

Added LKI to the CardLeftPlayThisPhaseWatcher, also added a convenience function to that class to make sure we are getting the most recent instance of that card.

Also added a couple tests for HMW Luminara with Maul9

### DIFF
--- a/server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
+++ b/server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
@@ -1,5 +1,5 @@
 import type { IInPlayCard } from '../core/card/baseClasses/InPlayCard';
-import type { CardType } from '../core/Constants';
+import type { CardType, ZoneName } from '../core/Constants';
 import { StateWatcherName } from '../core/Constants';
 import type { Game } from '../core/Game';
 import type { UnwrapRef } from '../core/GameObjectBase';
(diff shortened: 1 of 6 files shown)


server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -14,6 +14,12 @@ export interface CardLeftPlayEntry {
     card: GameObjectId<IInPlayCard>;
     controlledBy: GameObjectId<Player>;
     cardType: CardType;
+
+    /** The card's power when it left play */
+    power?: number;
+
+    /** Arena card was in when it left play */
+    arena?: ZoneName;
```
VJubert: I did something similar on #2734 but it's not reviewed, maybe you can take watcher part to add partial LKI to this interface

bwohldmann: Thanks.  Pivoted to partial LKI approach bases on your draft PR.

server/game/stateWatchers/CardsLeftPlayThisPhaseWatcher.ts
```diff
@@ -46,6 +49,11 @@ export class CardsLeftPlayThisPhaseWatcher extends StateWatcher<CardLeftPlayEntr
             .map((entry) => entry.card);
     }
 
+    public getMostRecentLeftPlayEntry(card: IInPlayCard): UnwrapRef<CardLeftPlayEntry> | undefined {
+        const entries = this.getCurrentValue().filter((entry) => entry.card === card);
+        return entries[entries.length - 1];
```
VJubert: I don't know if we should do this and trust order of list and filter or if we should use mostRecentPlayId

VJubert (approved):

PR #2793, opened 2026-09-09
Seismic Detonation

VJubert:

### DIFF
--- a/server/game/cards/09_HMW/events/SeismicDetonation.ts
+++ b/server/game/cards/09_HMW/events/SeismicDetonation.ts
@@ -1,0 +1,42 @@
+import type { IAbilityHelper } from '../../../AbilityHelper';
+import type { IEventAbilityRegistrar } from '../../../core/card/AbilityRegistrationInterfaces';
+import { EventCard } from '../../../core/card/EventCard';
+import type { Arena } from '../../../core/Constants';
+import { PhaseName, TargetMode, ZoneName } from '../../../core/Constants';
+
+export default class SeismicDetonation extends EventCard {
+    protected override getImplementationId() {
+        return {
+            id: 'seismic-detonation-id',
+            internalName: 'seismic-detonation',
+        };
+    }
+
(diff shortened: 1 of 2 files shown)


server/game/cards/09_HMW/events/SeismicDetonation.ts
```diff
@@ -0,0 +1,42 @@
...
+                    ['Space']: this.eventEffect(ZoneName.SpaceArena, abilityHelper),
+                    ['Ground']: this.eventEffect(ZoneName.GroundArena, abilityHelper),
+                }
+            }
+        });
+    }
+
+    private eventEffect(arena: Arena, abilityHelper: IAbilityHelper) {
+        return abilityHelper.immediateEffects.delayedPlayerEffect({
+            title: 'Deal 3 damage to each enemy unit in that arena',
```
moyerr: We can name the arena here. The user will see this effect in the ongoing effect tiles. I think EnumHelpers has a helper for arena name

server/game/cards/09_HMW/events/SeismicDetonation.ts
```diff
@@ -0,0 +1,42 @@
...
+            effectDescription: `deal 3 damage to each enemy unit in the ${arena === ZoneName.SpaceArena ? 'Space' : 'Ground'} arena`,
```
moyerr: yeah, we can use the EnumHelper here too. I think we've stopped capitalizing arena names since they don't do so on the cards

test/server/cards/09_HMW/events/SeismicDetonation.spec.ts
```diff
@@ -0,0 +1,60 @@
+describe('Seismic Detonation', function() {
+    integration(function(contextRef) {
+        it('deals 3 damage to each enemy unit in the chosen arena at the start of the next regroup phase, and only triggers once even with an additional regroup phase', async function() {
```
moyerr: Should probably add a case for the other arena too, just to be sure. We can also assert the ongoing effect text with `expect(context.player1).toHaveOngoingEffect(text)`

moyerr (approved):

PR #2804, opened 2026-09-11
Garnac

VJubert:


test/server/cards/09_HMW/units/GarnacLetTheHuntBegin.spec.ts
```diff
@@ -0,0 +1,85 @@
+describe('Garnac, Let The Hunt Begin', function() {
+    integration(function(contextRef) {
+        describe('Garnac\'s constant ability', function () {
+            it('should have Hidden while an opponent controls a unique unit', async function() {
```
moyerr, line 4: Can we add a case that it _gains_ hidden after being played when the opponent plays a unique unit? So basically:

- Player 1 plays Garnac, opponent has no unique units in play -- not hidden
- Player 2 plays a unique unit
- Garnac is now Hidden (because its the same phase he was played)

moyerr (approved):

PR #2814, opened 2026-09-14
do not initalize upgrade when moving (pilot are failing)

VJubert:


test/server/cards/09_HMW/events/LowAltitudeCombat.spec.ts
```diff
@@ -41,6 +41,43 @@ describe('Low Altitude Combat', function () {
             expect(context.p2Base.damage).toBe(4);
         });
 
+        it('Low Altitude Combat\'s ability should moves a space unit to the ground arena (friendly), then may attack with a ground unit that gets +2/+0', async function () {
```
moyerr: Test case title should mention pilot upgrade case. Do we also have cases for other upgrades? I'm worried that normal upgrades are also broken because the `UpgradeCard` override of `moveTo` ignores the `InitializeCardStateOption` parameter

VJubert: Added, no issue with classic upgrade

moyerr (approved):

PR #2840, opened 2026-09-17
Landing Pad

VJubert:


server/game/cards/09_HMW/upgrades/LandingPad.ts
```diff
@@ -0,0 +1,23 @@
...
+    protected override getImplementationId() {
+        return {
+            id: '0920001727',
+            internalName: 'landing-pad',
+        };
+    }
+
+    public override setupCardAbilities(registrar: IUpgradeAbilityRegistrar, abilityHelper: IAbilityHelper) {
+        registrar.addGainConstantAbilityTargetingAttached({
+            title: 'Friendly space units gain +1/+0',
```
moyerr: Nit: We use "gain" with abilities. We use "get" with stat modifications.

moyerr (approved):

PR #2846, opened 2026-09-17
child of dathomir, saw gerrera, soaring can-cell

sethhelman:
soaring
child of dathomir
saw


server/game/cards/09_HMW/units/SawGerreraShadowlandsInsurgent.ts
```diff
@@ -0,0 +1,23 @@
...
+            id: '3947921548',
+            internalName: 'saw-gerrera#shadowlands-insurgent',
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addTriggeredAbility({
+            title: 'Resource the top card of your deck',
+            when: {
+                onCardPlayed: (event, context) => event.player !== context.player && event.cardTypeWhenInPlay === CardType.Event,
```
moyerr: `cardTypeWhenInPlay` seems odd for a card that is never in play. I think we can just use `event.card.isEvent()` similar to TS16 Ahsoka leader

server/game/cards/09_HMW/units/SoaringCanCell.ts
```diff
@@ -0,0 +1,25 @@
...
+            id: '2209630427',
+            internalName: 'soaring-cancell'
+        };
+    }
+
+    public override setupCardAbilities(registrar: INonLeaderUnitAbilityRegistrar, AbilityHelper: IAbilityHelper) {
+        registrar.addConstantAbility({
+            title: `While you control an ${TextHelper.Trait.Kashyyyk} base, this unit gains ${TextHelper.Ambush}`,
+            condition: (c) => c.player.base.hasSomeTrait(Trait.Kashyyyk),
+            matchTarget: (card, context) => card === context.source,
```
moyerr: We don't need to match self for constant abilities

server/game/cards/09_HMW/units/SoaringCanCell.ts
```diff
@@ -0,0 +1,25 @@
...
+            ongoingEffect: [
+                AbilityHelper.ongoingEffects.gainKeyword(KeywordName.Ambush),
+            ]
```
moyerr: This doesn't need to be wrapped in an array

test/server/cards/09_HMW/units/SawGerreraShadowlandsInsurgent.spec.ts
```diff
@@ -0,0 +1,115 @@
+describe('Saw Gerrera, Shadowlands Insurgent', function() {
+    integration(function(contextRef) {
+        it('should resource the top card of the deck when the opponent plays an event', async function () {
```
moyerr, line 3: Can we add a case that checks for double-ramp when the opponent uses HMW L3-37 to play an event twice?

test/server/cards/09_HMW/units/SoaringCanCell.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+                player2: {
+                    groundArena: ['porg'],
+                }
+            });
+
+            const { context } = contextRef;
+
+            expect(context.soaringCancell.hasSomeKeyword('ambush')).toBeFalse();
+
+            expect(context.player1).toBeActivePlayer();
```
moyerr: can we play the unit for this case, so we can verify with game mechanics?

server/game/cards/09_HMW/units/SawGerreraShadowlandsInsurgent.ts
sethhelman: Making this fix but just a heads up that this was a direct copy past from Adi Gallia

moyerr (approved):

PR #2850, opened 2026-09-17
migrate, howler pack, local support

sethhelman:
local support
howler pack
migrate


server/game/cards/09_HMW/upgrades/LocalSupport.ts
```diff
@@ -0,0 +1,34 @@
...
+                useDisplayPrompt: true,
+                promptedPlayer: RelativePlayer.Self,
+                target: context.player.getTopCardOfDeck()
+            })),
+            then: (thenContext) => ({
+                title: 'Draw the top card of your deck',
+                thenCondition: () => {
+                    const revealedCard = thenContext.events[0]?.cards?.[0];
+                    return revealedCard != null && thenContext.player.isTraitInPlay([...revealedCard.traits]);
+                },
```
moyerr: We can avoid this whole `thenCondition` dance by using `ifYouDo` instead. And the revealed card should be `ifYouDoContext.target`

Also since we have the card here, the title can name the card

test/server/cards/09_HMW/events/Migrate.spec.ts
```diff
@@ -0,0 +1,27 @@
+describe('Migrate', function() {
+    integration(function(contextRef) {
+        it('Migrates\'s ability should create a Beast token for every 3 resources they control', async function() {
```
moyerr, line 3: Add a case with < 3 resources (can use credits to pay for the event). Player should see warning that event will do nothing.

moyerr (approved):

PR #2877, opened 2026-09-18
⟡ Zam Wesell, Not What She Seems

moyerr:
| ⟡ Zam Wesell |
| :---: |
| Not What She Seems |
|  |

Grants the unit each friendly leader's traits except Force, active in every zone ("even while she's not in play") via sourceZoneFilter: Any.

Engine support:
- Add dynamic gainTraits() ongoing-effect helper (parallel to gainKeywords)
- getTraits() now flattens array-valued GainTrait effects; static gainTrait path is unchanged
- Fix Card.hasEvery() returning false after its loop instead of true (affected hasEveryTrait/hasEveryKeyword/hasEveryAspect; previously unused)

Tests cover in-play/out-of-play (hand, discard) trait gain, Force exclusion, dedup, dynamic recalc on deploy, and leader-by-designation cases (The Darksaber, pilot leader deployed as an upgrade + host vehicle).


test/server/cards/09_HMW/units/ZamWesellNotWhatSheSeems.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+                context.player1.clickPrompt('Trigger');
+                context.player1.clickCard(context.battlefieldMarine);
+
+                // Combat damage was dealt on both sides, proving the attack actually resolved
+                expect(context.battlefieldMarine.damage).toBe(2);
+                expect(context.zamWesell.damage).toBe(3);
+            });
+        });
+    });
+});
```
VJubert, line 345: Can you add a test while she's on deck ? LAW R2 could be useful for this one

moyerr: R2 looks for aspects, not traits. But added test with This Is The Way

VJubert (approved):

PR #2889, opened 2026-09-18
Catch the Scent

sethhelman:
catch the scent


test/server/cards/09_HMW/events/CatchTheScent.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+            expect(p1Beasts.filter((beast) => beast.exhausted).length).toBe(0);
+            expect(p1Beasts.filter((beast) => !beast.exhausted).length).toBe(2);
+
+            const p2Beast = context.player2.findCardsByName('beast');
+            expect(p2Beast.length).toBe(0);
+
+            expect(context.player2).toBeActivePlayer();
+        });
+    });
+});
```
VJubert, line 78: Nice shot with TWI Palpa 👍

Can you add a test with Jerjerrod ? Should create 4 but still ready only 1

VJubert (approved):

PR #2894, opened 2026-09-18
⟡ Rex, Outserved His Purpose

moyerr:
| ⟡ Rex |
| :---: |
| Outserved His Purpose |
|  |


test/server/cards/09_HMW/units/RexOutservedHisPurpose.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+                context.player1.clickCard(context.targetedForRemoval);
+                context.player1.clickCard(context.battlefieldMarine);
+
+                // Battlefield Marine now has an ability, so Rex's bonus no longer applies
+                expect(context.battlefieldMarine.getPower()).toBe(3);
+                expect(context.battlefieldMarine.getHp()).toBe(3);
+            });
+        });
+    });
+});
```
VJubert, line 270: Few more tests to add :
- an upgrade which do not give ability (upgrade with When Played like Durasteel), maybe a Piloting upgrade could be interesting too (for example: JTL Luke unit)
- I see your comment on experience token, but even if shield have ability, it's not an ability on unit, so buff should apply
- even if I don't see why it wouldn't working, but a test with Beast token could be interesting, it's one of mechanics of this set

VJubert (approved):

PR #2904, opened 2026-09-20
Fix blanked triggered abilities firing on the event that un-blanks them

### LINKED ISSUE #2903: ASH Chimaera improperly healing off of defeat of units that stop healing
ASH Chimaera in play, SEC Galen naming chimaera, play anything/attack into Galen. This should not heal as Chimaera's when an enemy unit is defeated trigger should not miss its window

rshutton1:
Triggering events are emitted twice: once before resolution, and again after resolveGameState() so that cards which entered play or gained keywords mid-resolution can trigger. An ability that was blanked at the first emit and came back online because of the event itself (e.g. the card blanking it was defeated) passed the second emit and triggered. Chimaera (ASH) healed off the defeat of the Galen Erso naming it.

Record abilities that were registered but inactive when an event was first emitted and skip them on the post-resolution emit. Cards entering play mid-resolution are unaffected: their listeners were not registered at the first emit, so they are never marked.

Fixes #2903


moyerr (dismissed):
Curious if there are any more test cases we should add before merging? But specs aside, this LGTM!

moyerr (approved):

